PR #336, opened 2014-03-19
Create src-greenplum.r

mwillumz:
Sorry for the separate pull request.  I'm having to enter the code via the website due to company firewalls.  I made the src-greenplum.r changes suggested.  Regarding has_lahman() are you suggesting edits to the data-lahman.r file?  I'm not sure I follow otherwise.  Unfortunately I don't have write access to the greenplum db so I'll be unable to really test some of the lahman stuff.

### DIFF
--- a/R/src-greenplum.r
+++ b/R/src-greenplum.r
@@ -196,7 +196,7 @@ translate_env.src_greenplum <- function(x) {
       min   = win_recycled_greenplum("min"),
       max   = win_recycled_greenplum("max"),
       n     = function() {
-        over_greenplum(sql("COUNT(*)"), partition_group(), frame = c(-Inf, Inf))
+        over_greenplum(sql("COUNT(*)"), partition_group(), frame = NULL)
         },
                    
       # Cumulative function are like recycled aggregates except that R names


will0955:
I've created a backend in a separate package. This works up to a point. In order for more complex window functions to work properly I need to modify the internal `over()` function. Specifically I need to change from:

``` r
  if (!is.null(partition)) {
    partition <- build_sql("PARTITION BY ", 
                           sql_vector(partition, collapse = ", "))
  }
  if (!is.null(order)) {
    order <- build_sql("ORDER BY ", sql_vector(order, collapse = ", "))
  }
```

to

``` r
  if (!is.null(partition)) {
    partition <- build_sql("PARTITION BY ", 
                           sql_vector(partition, collapse = ", ", parens = FALSE))
  }
  if (!is.null(order)) {
    order <- build_sql("ORDER BY ", sql_vector(order, collapse = ", ", parens = FALSE))
  }
```

so that my query appears as

```
LAST_VALUE("x") OVER (PARTITION BY "a", "b" ORDER BY "y") AS "x_result"
```

rather than

```
LAST_VALUE("x") OVER (PARTITION BY ("a", "b") ORDER BY "y") AS "x_result"
```

I've verified that this solves the problem by using `fixInNamespace(over,'dplyr')` but I'm not sure how to go about implementing changes on the internal `function(over)`...


The following results in an error on greenplum (can't test postgres... is including parens correct syntax in postgres?)
```
mtcars2 <- copy_to(src_postgres(), mtcars)

Test <- group_by(mtcars2,cyl,gear) %>%
  arrange(mpg) %>%
  mutate(low=first(mpg))

Test$query

Test
```
But if I add `parens=FALSE` via `fixInNamespace(over,'dplyr')` the following runs as expected.
```
mtcars2 <- copy_to(src_postgres(), mtcars)

Test <- group_by(mtcars2,cyl,gear) %>%
  arrange(mpg) %>%
  mutate(low=first(mpg))

Test$query
```

PR #844, opened 2014-12-12
added function: db_create_database

sumprain:
Added function for adding new database.  I am on way to enhance db_create_table. Bye.


hadley:
Thanks, but I currently don't think this belongs in dplyr.

PR #850, opened 2014-12-16
Properly remove cluster in stop_cluster

saurfang:
I believe you need to remove the cluster in `cluster_env` rather than setting it to `NULL` because you use `exists` to check if the clusters have been initialized or not in `has_cluster`.


hadley:
Thanks!

PR #874, opened 2015-01-08
translation of plyr::ddply

paul4forest:
This statement "most of dplyr is equivalent to ddply()[...]" was slightly misleading for me. I took me a while to realise that ddply can also be replaced by do() in some cases. After a couple of months of using dplyr, I realised this today while converting a function that returns a data frame, calculated on a window for each country (GDP deflator by country).

This iris example might be best placed in a another page on translating plyr statement into dplyr.  But this introduction vignette is probably the first place most people will look at.

### DIFF
--- a/vignettes/introduction.Rmd
+++ b/vignettes/introduction.Rmd
@@ -408,7 +408,7 @@ Compared to plyr:
 
 * it provides a better thought out set of joins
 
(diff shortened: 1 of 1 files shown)


jimhester:
You might also want to checkout http://jimhester.github.io/plyrToDplyr/ for some more examples of converting plyr to dplyr code.

hadley:
I rather tackle this problem more broadly. Thanks for the PR though!

PR #905, opened 2015-01-19
Handle drop=FALSE correctly for empty factors in id, id_var

alexbbrown:
Addresses 

https://github.com/hadley/tidyr/issues/56
and
https://github.com/hadley/dplyr/issues/904

not otherwise tested.


alexbbrown:
fixed and tested

hadley:
Would you mind adding a bullet point to NEWS and a few unit tests?

hadley:
See #808

PR #908, opened 2015-01-19
Hive does not accept table names that begin with a "_".

rbdixon:
See https://issues.apache.org/jira/browse/HIVE-6431
Switch to using "x" as the prefix for unique_name().

I've started work on a Hive backend for dplyr that is working well so far. I'll be staging that for pull soon. This is a simple change that seems innocuous and is required by Hive due to its restrictions on tables and variables starting with an underscore.

### DIFF
--- a/R/utils.r
+++ b/R/utils.r
@@ -114,7 +114,7 @@ unique_name <- local({
 
   function() {
     i <<- i + 1
-    paste0("_W", i)
+    paste0("xW", i)
   }
 })
 


rbdixon:
Hmmm... one of the builds failed during setup attempting to install: "./travis-tool.sh install_github jimhester/covr" with a HTTP 403 error. That doesn't seem to be an issue with this one character change. Any ideas how to kick the Travis CI build off again to see if this will pass with a repeated run?

rbdixon:
Docs say to close PR and re-open. Standby.

Issue #891, opened 2015-01-14
Add alias summarize_each


avsmith:
For people from countries that don't don't spell things correctly, please add an alias of `summarise_each` as `summarize_each`. Similar exists for `summarise`, creates confusion that same doesn't exist for "`_each`" variants.

hadley:
Fixed in dev version

PR #923, opened 2015-01-25
Added aliases summarize_each and summarize_each_ per request in issue #891.

rbdixon:


rbdixon:
I just noticed the [earlier PR](https://github.com/hadley/dplyr/pull/787) for this same matter created by matthieugomez. I apologize for the duplication. Use whichever you prefer but credit is due to matthieugomez as he submitted first.

PR #932, opened 2015-01-29
add sql escape functions for POSIXlt and POSIXt

pschmied:
Encountered a data.frame with posix dates and times, that I couldn't mash into a postgres database. Added some methods for escaping these, just the same as escape.Date.


hadley:
It's not clear that this is the correct approach when the time zone is different on the client and server.

PR #970, opened 2015-02-15
Add read from .sql file functionality

jenswirf:
For those times when explicitly writing out a full SQL query is unavoidable, this is is a nice way of at least avoiding long quoted SQL _query strings_ in the chain. Keep a separate file/tab open where all the SQL goes (and is syntax highlighted). Then source it into your R script, like so:

`results <- tbl(my_conn, sql(file="my_query.sql")) %>% collect()`


hadley:
To me, this doesn't seem like a big win over `sql(readLines("my_query.sql"))`

PR #974, opened 2015-02-17
Add mapping for ifelse <-> if in MySQL

cosinequanon:
Lets you do the following:

    my_db <- src_mysql('test')
    copy_to(my_db, iris)

    myiris <- tbl(my_db, 'iris')
    myiris %>%
        mutate(is_setosa = ifelse(Species == 'setosa', 'setosa', 'nope')) %>%
        group_by(is_setosa) %>%
        tally


hadley:
Does the existing translation to `CASE WHEN` not work for you?

cosinequanon:
I didn't realize that existed, yeah, that works fine for me.

PR #980, opened 2015-02-17
MonetDB is supported via MonetDB.R

xiaodaigh:
I had to search the news to find out that MonetDB support is now via MonetDB.R. So the README was updated to reflect this fact.


hadley:
Thanks!

PR #981, opened 2015-02-18
x -> con to be consistent with rest of the example

xiaodaigh:
The argument x in the function was not referenced and the variable con was used in the body of the function but not an argument. The intention was for the argument to be called con.


hadley:
Thanks!

PR #985, opened 2015-02-19
Sweep over several rows or columns simultaneously

kasiajanocha:
As mentioned in https://github.com/hadley/dplyr/issues/944, no intuitive way to sweep over several rows or columns simultaneously was available.


hadley:
I'd rather discuss this more before reviewing an implementation.

PR #987, opened 2015-02-19
summarise / summarize alias thing for summarise_each

jennybc:


jennybc:
Oh now I see a similar PR here: #787 but build failed, I think for reasons having nothing to do with the proposed change. My PR also includes updated `Rd` files and `NAMESPACE`. I'll add the same text as @matthieugomez did to `NEWS.md` to get best of both worlds in one place.

hadley:
Sorry pulled this from someone else

PR #988, opened 2015-02-19
Print the number of variables in glimpse.

ilarischeinin:
Adds the number of variables to the output of glimpse. Also, wherever thousand separators are used, use "," only if it isn't already being used for the decimal point, in which case use ".".


hadley:
Good idea overall, but can you please pull out the comma/period-adding code into its own function to avoid the duplication?

ilarischeinin:
Good point. I made it a function called big_mark(), which I put at the end of file utils-format.r (not sure if that's the best file to include it in though).

R/utils-format.r
```diff
@@ -181,3 +181,9 @@ obj_type.data.frame <- function(x) {
 obj_type.data_frame <- function(x) {
   paste0("<data_frame [", paste0(dim(x), collapse = ","), "]", ">")
 }
+
+# function for the thousand separator,
+# returns "," unless it's used for the decimal point, in which case returns "."
+big_mark <- function() {
```
hadley: I think you can reduce the duplication even further if `big_mark()` looked like this:

```R
big_mark <- function(x, ...) { 
  mark <- if (identical(getOption("OutDec"), ",") "." else ","
  format(d, big.mark = mark, ...)
}
```

Note that here it's better to use non-vectorised equals (i.e. `identical()`) and `ifelse()` (i.e. `if`)

ilarischeinin:
True. Earlier I made it separate so that `big_mark()` could be used with `format()`, `formatC()`, and `prettyNum()`. But the three of them are a bit redundant I guess.

R/utils-format.r
```diff
@@ -25,7 +25,7 @@ NULL
 #' @rdname dplyr-formatting
 dim_desc <- function(x) {
   d <- dim(x)
-  d2 <- format(d, big.mark = ",", justify = "none", trim = TRUE)
+  d2 <- big_mark(d)
```
hadley: Do you no longer need the other args?

ilarischeinin: No, as I used `formatC()` instead of `format()`. (Because it seems to be slightly more efficient.)

But actually, while it makes no difference here, for `glimpse()` there is a small cosmetic choice to be made. With the code above (that uses `formatC()`), its output currently looks like this:

```r
> glimpse(flights)
Observations: 336,776
Variables: 16
$ year      (int) 2013, 2013, 2013, 2013, 2013, 2013, 2013, 2013, 2013, 201...
$ month     (int) 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, ...
```

But with `format()` it would be possible to align the two numbers like this:

```r
> glimpse(flights)
Observations: 336,776
Variables:         16
$ year      (int) 2013, 2013, 2013, 2013, 2013, 2013, 2013, 2013, 2013, 201...
$ month     (int) 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, ...
```

The reason I didn't align them is that I'd expect the number of variables to practically always be way way smaller than the number of observations (and to very very rarely actually use the thousands separator).

But if you'd prefer the aligned version, I'll change `big_mark()` from `formatC()` to `format()`, which will by default produce right-aligned output. Then the function call inside `dim_desc()` would also get back its `trim=TRUE` argument so that it'll continue to print
```
Source: local data frame [336,776 x 16]
```
instead of
```
Source: local data frame [336,776 x       16]
```

hadley: Looks good as is - thanks for the explanation!

hadley:
One last thing - can you please add a bullet point to NEWS? It should look like ` * description (@yourusername, #thisprnumber)`

ilarischeinin:
There you go, I added it as two separate bullet point entries.

hadley:
Thanks!

PR #990, opened 2015-02-20
Consistent behavior on distinct() when key is set to data.table

pimentel:
Really small change when calling `distinct()` on a data.table with a key.

**Background**

When using a `data.table`, if a key is set, a call to `unique` returns only unique keys, not unique rows. This causes unexpected behavior with `distinct()`:

```{r}
> dt_no_key <- data.table(x = rep("a", 4), y = rep(c("f", "g"), 2))
> dt_key <- copy(dt_no_key)
> setkey(dt_key, x)
> dt_no_key %>%
+     distinct()
   x y
1: a f
2: a g
> dt_key %>%
+     distinct()
   x y
1: a f
```

**New behavior**

`distinct()` on a data.table with a key returns unique rows, not just the unique key

```{r}
> dt_key %>%
+     distinct()
   x y
1: a f
2: a g
```


hadley:
Could you please also add a bullet point to NEWS and a unit test?

hadley:
Bump.

pimentel:
will deal with it soon (by end of week)

pimentel:
Okay, I think we are good. @hadley

hadley:
Thanks!

PR #992, opened 2015-02-20
Allow any number of data frames and lists of dataframes in binding functions

lionel-:
Currently `bind_rows()`, `bind_cols()` and `combine()` have complicated rules for their arguments. They also ignore silently some data frame arguments when the first arg is a list:
```{r}
bind_cols(list(mtcars, mtcars), diamonds)  # diamonds is silently ignored here
```

This PR simplifies the rules by allowing indistinctively any number of data frames or lists of data frames, so that it's  ok to do:
```{r}
bind_rows(mtcars, list(mtcars, mtcars), mtcars)
```

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,7 +1,10 @@
 # dplyr 0.4.1.9000
 
-* `lag` and `lead` for grouped data were confused about indices and therefore 
+* `lag` and `lead` for grouped data were confused about indices and therefore
   produced wrong results (#925, #937)
+
+* `bind_rows()`, `bind_cols()` and `combine()` now accept indistinctively
+  any number of data frames and list of data frames as arguments (@lionel-).
 
 # dplyr 0.4.1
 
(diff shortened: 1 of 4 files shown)


hadley:
I think the reason for the original design was to avoid copying data frames. Maybe it's not so bad >3.1.1 where they only references might get copied, but I'd appreciate some before and after benchmarks to confirm.

lionel-:
Hi Hadley,

There is a small cost to calling `enlist()`. It does not increase with the size of the data frame, so your conjecture about >3.1.1 behavior is correct. If anything, it decreases.

```{r}
small <- mtcars
big <- rerun(1000, mtcars) %>% bind_rows()
```
```
neval: 100
                                         expr median         unit
                                             expr median         unit
1 rbind_all(list(mtcars, mtcars, mtcars, mtcars))     86 microseconds
2 bind_rows(mtcars, list(mtcars, mtcars), mtcars)    513 microseconds
3         rbind_all(list(data, data, data, data))  13233 microseconds
4         bind_rows(data, list(data, data), data)  13319 microseconds

```

hadley:
I think the use case is more like a list of 50 or 100 data frames. Could you benchmark that too please?

lionel-:
hmm something like this?

```{r}
data <- rerun(100, mtcars) %>% bind_rows()
data_list <- rerun(100, data)
```
```
neval: 100
                                 expr median         unit
1                rbind_all(data_list)     74 milliseconds
2 rbind_all(c(data_list, list(data)))     75 milliseconds
3  rbind_all(enlist(data_list, data))     75 milliseconds
```

hadley:
I think I can live with a 1ms slowdown ;)

lionel-:
Not relevant anymore, see #1104

PR #994, opened 2015-02-24
updated the do() example so that it works now

xiaodaigh:


hadley:
Sorry for taking to long to review this PR. I fixed it in the meantime.

Issue #996, opened 2015-02-25
Improve duplicated column name error message


hammer:
The error message should include the duplicated column name in the error message; currently it just reads `Error: found duplicated column name`.

PR #1001, opened 2015-03-03
Displaying names of duplicated columns in error message

kasiajanocha:
As mentioned in #996, error message for duplicated column names only said
<code> Error: found duplicated column name </code>.
Added printing out a list of duplicated column names.


hadley:
Sorry for taking to long to review this PR. It was fixed in the meantime.

Issue #804, opened 2014-11-25
Row-wise data frame creation


hadley:
Often when creating example data you want to see a row of data at a time. It would be useful to have an easy way to create this with dplyr. Maybe something like?

```R
rdf(ncol = 4, 
  "a", 1, 3, "b",
  "c", 2, 5, "e"
)
```

?

Have you thought about this at all @kevinushey ?

jennybc:
Hopefully dplyr will eventually have a function like `read_table()` where

dplyr::read_table is to dplyr::data_frame as read.table is to data.frame

in which case it could offer a `text =` argument like `read.table()` does. I use that a lot to do this row-wise creation of data for examples, e.g.:

```r
read.table(text = c("'a' 1 3 'b'",
                    "'c' 2 5 'e'"))
```

hadley:
(That function will be called `fastread::read_table()`)

hadley:
Another option would be

```R
rdf(
  "a" | 1 | 3 | "b",
  "c" | 2 | 5 | "e"
)
```

kevinushey:
I like making tiny DSLs within function calls; I can take a stab at that.

hadley:
@kevinushey that would be great!

PR #1010, opened 2015-03-10
add 'rdf' (#804)

kevinushey:
Adds `rdf` as a little function helper for creating `tbl_df`s inline, e.g.

```R
rdf(
    a | b | 1 | 2,
    c | d | 3 | 4
)
```

Comments appreciated (probably most pertinent one -- what about column names?)

Other things to think about:

- is `rdf` the best name?
- should we provide a column-wise analog (ie, like `rdf` but transposes the created frame)? Such a function could accept named arguments (for named columns)
- allow for leading or trailing commas?


hadley:
Maybe the first row should always be column names?

I definitely don't like `rdf()` as a function name.

Another approach that would be a bit easier to parse would be:

```R
rdf(
    hdr(a, b, c)
    row(a, b, 1, 2)
    row(c, d, 3, 4)
)
```

Another approach would require a sentinel value at the end of each row:

```R
rdf(
  a, b, c, d, END
  a, b, 1, e, END,
  c, d, 3, 4
)
```

(But I don't think you could make that as elegant as the other two approaches)

kevinushey:
I definitely prefer not having to add 'extra' syntax to make the table; we should keep it as 'markdown-like' as possible.

Perhaps a dummy 'delimiter', or 'delimiter-like' symbol, could be used, e.g. a series of dots:

```R
rdf(
    these | are | names,
    ...................,
    these | are | cells,
)

)
```

I would muc

hadley:
I like the series of dots idea.

kevinushey:
Just added some code to support the dots delimiter, and also we drop missing arguments. So we can now write e.g.

```R
df <- rdf(
   , these | are | names
   , ...................
   , these | are | cells
 )
```

(note that symbols are implicitly converted to character)

lionel-:
`doodle()` ?

lionel-:
another idea for specifying column headers:

```r
data <- doodle(
  ha : ho : hi,
   1 |  x | NA,
   3 | NA | NA,
)
```

hadley:
@jennybc I know this is something that you're interested in. Any thoughts on name, syntax, etc?

jennybc:
Here's a mash-up of the above ideas

```
tibble(
  these, are, names,
  ---,
  these, are, values,
  these, are, values
)
```

Who doesn't like `tibble()`?!?

I am much faster at typing commas than `|` (tho the nod to markdown is not lost on me).

You *could* use the markdown-inspired "3 dashes" to signal the presence of variable names. A series of dots might lead to confusion with `...`?

Maybe instead of a sentinel value to signal the end of one observation, just require variable names and get length there? That might be less of a drag than repeatedly typing the sentinel.

lionel-:
`tibble()` is an awesome name.

I like your `---` idea. It could be 3 characters or more, so that a fully formatted table looks like:

```r
tibble(
  these, are,  names,
  ------------------,
  these, are, values,
  these, are, values
)
```

hadley:
I love the name tibble, but I don't think `-` will work as a separator because it's not a valid R expression. Unless some else has a really bright idea, I think we're stuck with `......`.

I like the idea of forcing a header row so that we can automatically determine how many columns there are, so can use the implicit newline as a row separator.

lionel-:
ah yes... and with the dots there's another issue, the function won't work with only three dots.

the following is maybe a bit annoying to type
```r
tibble(
  these, are,  names,
  `----------------`,
  these, are, values,
  these, are, values
)
```

This would work but looks like a hack:
```r
tibble(
  these, are,  names,
  -----------------.,
  these, are, values,
  these, are, values
)
```

jennybc:
I definitely don't understand exactly what will work as the thing that separates variable names from the data, but I get the general drift.

That said, can the twiddle be pressed into service? It's easy to make a valid expression with it.

lionel-:
maybe the simplest thing:

```r
tibble(
  'these', 'are', 'names',
   these,   are,   values,
   these,   are,   values
)
```

hadley:
The problem with using an infix function (like `~`) is that you need to have something on the right hand side of it (i.e. `~~~~~~~~~~x` would work but, `~~~~~~~` would not). I think we need to use something that's a valid R variable name (i.e. can start with letter or `.`, and contain letters, digits, `.` and `_`. That doesn't leave many options, but we could do something like:

```R
tibble(
  "col1", "col2", 
  HEADER,
  1, 3
  2, 6
)
```

I also don't think we want to drop the quotes for strings altogether because that will confuse people who want to put spaces in their values. So maybe we could flip @llionel-'s idea around and do:

```R
tibble(
  col1, col2,
  1, 2,
  3, 4
)
```

But then the header doesn't stand out much. Another idea would be to eliminate the NSE altogether and require formula quoting for column headers:

```R
tibble(
  ~col1, ~col2,
  1, "a",
  3, "b"
)
```
 
I quite like that.

(I also just thought of a another name: `frame_data()` because it's a transposed `data_frame()`.)

jennybc:
@hadley's last proposal looks really good to me:

```
tibble(
  ~col1, ~col2,
  1, "a",
  3, "b"
)
```

lionel-:
and once again `~` saves the day

hadley:
@jennybc how do you feel about `frame_data()` as a name?

@kevinushey do you want to take a shot at implementing this? Or should I? I think it should be fairly simple now.

jennybc:
I like `frame_data()` and think it's clever. I still like `tibble()` too: brings back fond memories of the epic [tibble diff](https://twitter.com/hadleywickham/status/524200537861128192) discussion. You have a great track record with naming things so I'm not bothered either way.

kevinushey:
@hadley I'll give it a shot and update this PR.

kevinushey:
New PR at https://github.com/hadley/dplyr/pull/1358.

PR #1057, opened 2015-04-04
Fixed typo in new-sql-backend.Rmd

teramonagi:
`mysql` -> `postgres`


hadley:
Ooops. Thanks for the fix!

PR #1058, opened 2015-04-04
Make select() work with non-English names

hkjinlee:
An add-hoc solution to #866. 

https://github.com/hadley/lazyeval/pull/31 deals with mutate(), summarise(), etc...
This deals with select().


hadley:
Should be fixed in master now.

PR #1070, opened 2015-04-11
Fix broken link

krlmlr:
I wonder if this is the right paper.

Also, [LINQ](https://msdn.microsoft.com/en-us/library/bb397926.aspx) seems to have some similarity.


hadley:
Fixed with DOI link

Issue #659, opened 2014-10-06
Options to not copy data.tables during a join


matthieugomez:
At least for data.table, it seems that join operations copy x and y when setting their key. Would it be possible to avoid this behavior either automatically when data.tables are already keyed by the relevant variable or through an option like `setkey = TRUE` ?

PR #1071, opened 2015-04-12
Test for join over data tables

krlmlr:
Tests that join operations don't change the source data tables. Fails if data.table::copy is omitted in `join_dt()`.

Related to #659.


hadley:
Thanks!

PR #1072, opened 2015-04-12
Fix filter_.data.table

### LINKED ISSUE #906: filter_ with a data.table throws 'argument ".dots" is missing, with no default' error
To reproduce:

    library(data.table)
    library(dplyr)
    DT <- data.table(x = rnorm(10))
    DT %>% filter_(~ x > 0)
    ## Error in lazyeval::common_env(.dots) : 
    ##   argument ".dots" is missing, with no default

The error persists if you replace the formula with a string or call:

    DT %>% filter_("x > 0")       # throws same error
    DT %>% filter_(quote(x > 0))  # throws same error

...and if you don't use piping:

    filter_(DT, ~ x > 0)          # throws same error

The interactive version is OK:

    DT %>% filter(x > 0)          # OK

krlmlr:
Minor glitch in call to lazyeval::common_env.

Fixes #906.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -3,6 +3,8 @@
 * `lag` and `lead` for grouped data were confused about indices and therefore 
   produced wrong results (#925, #937)
 
+* Fix `filter_` for data tables (#906)
+
 # dplyr 0.4.1
 
 * Don't assume that RPostgreSQL is available.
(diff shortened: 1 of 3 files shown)


hadley:
Thanks!

PR #1079, opened 2015-04-14
show example with multiple criteria for filter

almartin82:
current documentation shows how to filter for a specific value, and with an inequality.  might also be nice to show an example that uses multiple criteria (a common use case).


sfirke:
If adding an example with multiple criteria, could you show it w/ the arguments separated by commas?  I love that I don't have to use the `&` operator.  E.g., `filter(mtcars, cyl == 8, mpg > 16)`

hadley:
Both great ideas - thanks!

PR #1082, opened 2015-04-15
fix typo in funs() documentation

almartin82:
comment references `funs_q` where it appears to mean `funs_`


hadley:
I think someone else got this, but thanks for the PR!

PR #1091, opened 2015-04-21
```{cpp} means ```{r, engine='cpp'} in the next version of knitr

yihui:
And because `cpp` is not an existing engine name in knitr, this will lead to errors.


hadley:
Thanks!

PR #1116, opened 2015-04-30
Fix typo

JasonAizkalns:


hadley:
Would you mind fixing in the original R file? (See comment at top of file)

JasonAizkalns:
Done.

Living and learning...mostly still learning...

PR #1128, opened 2015-05-06
Add escape.POSIXt

### LINKED ISSUE #857: no applicable method for 'escape' applied to an object of class "c('POSIXct', 'POSIXt')"
At first, I want to thank your hard work of developing `dplyr` package to the R community. This is one of the best package to do data manipulation in R and it does a lot of helps in my work.

Recently, I am trying the database API and filter a datetime column in PostgreSQL. I have the following error message:

```
> filter(hourly_rps_mode, time == time_value)
Error in UseMethod("escape") : 
  no applicable method for 'escape' applied to an object of class "c('POSIXct', 'POSIXt')"
```

It could be fixed easily by:

```
escape.POSIXt <- dplyr:::escape.Date
```

Could you add this to the package?

eibanez:
To close #857


hadley:
Could you please also add a bullet point to NEWS?

ReportMort:
@eibanez Your function escape.POSIXt drops the timezone of the data it's escaping https://github.com/eibanez/dplyr/blob/master/R/sql-escape.r#L132. 

Would you consider switching to:
 `x <- format(x, "%Y-%m-%d %H:%M:%S %Z")`

I think the expectation should be that the timezone is preserved while escaping. Data without a timezone that is inserted or updated to a database column with time zone would adopt the default timezone of the database. So if your function escapes and drops the timezone of some data while it's being sent to the database, then it will always be reassigned to the default database timezone, potentially making it different from the values the user had in their data.frame.

eibanez:
This could pose a problem (at least) with SQLite, as times are compared as characters only (see [here](http://stackoverflow.com/questions/2930768/how-to-compare-sqlite-timestamp-values)). I don't know what the implications are for MySQL.

hadley:
Thanks!

hadley:
Ooops, didn't read the full discussion. I'm going to revert

Issue #1132, opened 2015-05-12
Typo in `percent_rank()`?


jdthorpe:
In trying to understand what 'percent_rank' does, I took a peak at the code and found the expression `length(!is.na(x))`.  I can't think of any situation in which `length(x) != length(!x)` and I suspect that `length()` should be replaced with `sum()`

PR #1133, opened 2015-05-12
Update rank.R to fix percent_rank() and cume_dist()

sfirke:
From http://stackoverflow.com/questions/30202658/typo-in-dplyrpercent-rank.

`percent_rank()` and `cume_dist()` currently divide by `length(!is.na(x))`, which is always `length(x)` and returns erroneous values for inputs with `NA` values.  Should be `sum(!is.na(x))`.


hadley:
Ooops!   Would you mind also including a bullet point in NEWS.md, and a test?

sfirke:
Done - I've added the bullet point to NEWS.md and created a test, and added those commits to this request.  I haven't created a test before so please review to see if I've done it correctly.

tests/testthat/test-percent_rank-and-cume_dist.R
```diff
@@ -0,0 +1,20 @@
+context("percent_rank and cume_dist")
+
+
+test_that("#1132 is fixed - percent_rank", {
+  testdata <- c(1:4, rep(NA, 4))
+  res <- percent_rank(testdata)
+  manually_calculated_res <- (min_rank(testdata) - 1) / (sum(!is.na(testdata)) - 1)
```
hadley: Rather than doing the calculation in the code, I'd just do it once and included it inline, e.g. `expect_equal(c(1, 2, NA), c(0, 1, NA)` - I think that's a bit simpler and still tests the desired behaviour

tests/testthat/test-percent_rank-and-cume_dist.R
```diff
@@ -0,0 +1,20 @@
+context("percent_rank and cume_dist")
```
hadley: I'd include these tests in `test-window.R`

tests/testthat/test-percent_rank-and-cume_dist.R
```diff
@@ -0,0 +1,20 @@
...
+
+
+test_that("#1132 is fixed - percent_rank", {
```
hadley: I'd make the name more descriptive, e.g. "percent_rank ignores NAs [#1132]"

sfirke:
Thanks for the feedback on the tests, I've adjusted them and moved to test-window.R

tests/testthat/test-window.R
```diff
@@ -25,3 +25,17 @@ test_that("cumany handles NA (#408)", {
   expect_true( all( cumany(x) ) )
   
 })
+
+test_that("percent_rank ignores NAs (#1132)", {
+  testdata <- c(1:4, rep(NA, 4))
+  expect_equal( percent_rank(testdata), (min_rank(testdata) - 1) / (sum(!is.na(testdata)) - 1) )
```
hadley: Could you please just inline the results of the expected value computation? It's important to be able to see the correct results when reading the tests.

sfirke:
Now the test vectors are created as part of the `expect_equal` calls, so there's a single line for each test - is that what you meant by making it in-line?  And I omitted the redundant 2nd test for each of the two functions.

hadley:
@sfirke I meant actually include the correct numbers in the test so you can read it without having to do any computation

sfirke:
Ah okay - I have added the correct numbers for the test comparison results

sfirke:
if you've had a chance to look at the updated tests: all set or do you see any other needed tweaks?

PR #1137, opened 2015-05-13
Update documentation for %>%

jimhester:
The previous documentation was not entirely updated for %>%, and did not
mention that %>% used in dplyr actually comes from the magrittr package.

In addition this uses R's built-in .Defunct and .Deprecated functions to
mark `chain()` and `%.%` and defunct and deprecated respectively.

I included only the code changes and chain man page in this pull request, because I am using a newer version of Roxygen there were a number of inconsequential version changes in the rest of the man pages I omitted.

The code comment mentions that `%.%` was supposed to be `Defunct` as of `0.4`.  I left it as `Deprecated` for now, but if you do want it to be `Defunct` let me know and I can do that.


hadley:
Thanks!

Issue #1127, opened 2015-05-06
Installation problem - Reducer.h:6:115: error: wrong number of template arguments


kykuo:
```
> devtools::install_github("hadley/dplyr")
Downloading github repo hadley/dplyr@master
Installing dplyr
"C:/PROGRA~1/R/R-32~1.0/bin/x64/R" --vanilla CMD  \
  INSTALL  \
  "C:/Users/kevinkuo/AppData/Local/Temp/Rtmp8CRTju/devtools2798a4078ec/hadley-dplyr-5b3cc9b"  \
  --library="C:/Users/kevinkuo/Documents/R/win-library/3.2"  \
  --install-tests 

* installing *source* package 'dplyr' ...
** libs

*** arch - i386
g++ -m32 -I"C:/PROGRA~1/R/R-32~1.0/include" -DNDEBUG -I../inst/include -DCOMPILING_DPLYR   -I"C:/Users/kevinkuo/Documents/R/win-library/3.2/Rcpp/include" -I"C:/Users/kevinkuo/Documents/R/win-library/3.2/BH/include" -I"d:/RCompile/r-compiling/local/local320/include"     -O2 -Wall  -mtune=core2 -c RcppExports.cpp -o RcppExports.o
In file included from ../inst/include/dplyr/Result/all.h:23:0,
                 from ../inst/include/dplyr.h:140,
                 from RcppExports.cpp:4:
../inst/include/dplyr/Result/Reducer.h:6:115: error: wrong number of template arguments (0, should be 1)
C:/Users/kevinkuo/Documents/R/win-library/3.2/Rcpp/include/Rcpp/traits/r_sexptype_traits.h:33:30: error: provided for 'template<class T> struct Rcpp::traits::r_sexptype_traits'
In file included from ../inst/include/dplyr/Result/all.h:23:0,
                 from ../inst/include/dplyr.h:140,
                 from RcppExports.cpp:4:
../inst/include/dplyr/Result/Reducer.h:6:148: error: template argument 1 is invalid
../inst/include/dplyr/Result/Reducer.h:6:150: error: template argument 1 is invalid
../inst/include/dplyr/Result/Reducer.h:6:150: error: template argument 2 is invalid
../inst/include/dplyr/Result/Reducer.h:8:70: error: wrong number of template arguments (0, should be 1)
[...]

PR #1152, opened 2015-05-20
avoid using name OUT

kevinushey:
`OUT` is `#define`'d in such a way on Windows that it breaks `Reducer.h`. This PR just renames that template parameter to avoid the problem.


hadley:
Thanks!  This fixes #1127, right?

kevinushey:
Yes indeed!

PR #1153, opened 2015-05-20
ensure 'data_frame' returns 'tbl_df' (closes #1151)

### LINKED ISSUE #1151: data_frame() returns a data.frame
This is a minor issue but I discovered it in testing something else. If you use `data_frame()` to make a zero row, zero column `tbl_df` … instead, you get a `data.frame` back.

```
> class(data_frame(a = 1:5, b = a * 2))
[1] "tbl_df"     "tbl"        "data.frame"

> class(data_frame())
[1] "data.frame"
```

kevinushey:


R/dataframe.R
```diff
@@ -86,11 +87,7 @@ data_frame_ <- function(columns) {
     output[short] <- lapply(output[short], rep, max)
   }
 
-  # Set attributes
-  attr(output, "row.names") <- c(NA_integer_, max)
-  attr(output, "class") <- c("tbl_df", "tbl", "data.frame")
-
-  output
+  tbl_df(output)
```
hadley: I'd rather keep the old code, and have `tbl_df()` use `as_data_frame()`

hadley:
Would you mind also adding a bullet to news?

PR #1160, opened 2015-05-21
print.tbl_df for duplicated column names. Fixes #1159.

### LINKED ISSUE #1159: print.tbl_df mungles duplicate column names
Duplicated column names are made unique (call to `make.unique`) before printing. This is related to [Issue #820](https://github.com/hadley/dplyr/issues/820) and [Issue 896](https://github.com/hadley/dplyr/issues/896), where the duplicated column names cause additional problems. Part of the problems to these issues is that they are harder to notice because of the printing issue. 

    # create data frame with duplicate column names
    set.seed(1)
    df <- data.frame(x = 1, x = 2, check.names = FALSE)
    df
    ##   x x
    ## 1 1 2
    as_data_frame(df)
    ## Source: local data frame [1 x 2]
    
    ##   x x.1
    ## 1 1   2

blasern:


hadley:
Would you mind also adding a bullet to NEWS?

hadley:
Thanks!

PR #1195, opened 2015-06-03
mutate_each with one call now creates new variables if given suffix inside funs()

msaidf:
this change is for fixing https://github.com/hadley/dplyr/issues/712 , where mutate_each replace the original variable when calling only one function even when the suffix name is already given inside funs().


hadley:
I don't really follow your implementation here. It would be better to start with more discussion in the original issue, including some test cases.

PR #1218, opened 2015-06-14
Fixed issue 926 (problems with escaping table names for SQLite databases)

NikNakk:
Fixed issues described with SQLite at https://github.com/hadley/dply/issues/926 by creating new function for db_insert_into that supports escaping (unlike the current DBI::dbWriteTable that was used) and using build_sql rather than paste0 within db_query_fields.

This was also raised at http://stackoverflow.com/questions/28132697/bracket-escaped-table-names-with-dplyr

### DIFF
--- a/R/src-sqlite.r
+++ b/R/src-sqlite.r
@@ -149,5 +149,6 @@ db_insert_into.SQLiteConnection <- function(con, table, values, ...) {
   assert_that(is.string(table), is.data.frame(values))
   valStr <- paste(rep("?", ncol(values)), collapse = ",")
   sql <- build_sql("insert into ", ident(table), " values (", sql(valStr), ")")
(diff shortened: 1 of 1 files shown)


NikNakk:
Realised I'd not included the code to clear up resources (dbClearResult) from DBI::dbWriteTable; included now. It should be noted also that the db_insert_into.SQLiteConnection differs from the previous version in that it will never create a new table. At present, it's only called from copy_to which creates the table first anyway, and to my mind a function called `db_insert_to` generally should be appending into an existing table, with a separate call to create the table. If the previous behaviour is preferred, then additional code to call `db_create_table` as needed could be added.

NikNakk:
Closed because was based on my master branch which made making any other changes problematic until this PR was successful

PR #1220, opened 2015-06-16
Implemented join for Google BigQuery

akhmed1:
Hello everyone! This pull request features an implementation of `sql_join` for "bigquery".

This is my first attempt to contribute to `dplyr` project so I apologize in advance if I am missing some key details (or may be even being overly verbose in describing it)

### Motivation
The major motivation for this pull request is that Google BigQuery has powerful SQL join facilities but requires a particular dialect of SQL when doing joins such that regular `sql_join` fails to produce a runnable query.
* Example: SQL keyword "USING" is not supported and the complete column name specification is required. (The default `sql_join.DBIConnection` solution when joining two tables A and B on the common column id is to say  "A JOIN B USING id" while Google BigQuery requires to say "A JOIN B on A.id = B.id")
* Example: Subsequently, Google Query is very sensitive to ambigous names inside SELECT statement and would require one to specify SELECT A.id or SELECT B.id instead of SELECT id which required some change in the logic of `sql_join`.
* Example: When joining large tables "JOIN EACH" keyword is required instead of "JOIN"

### Implementation Details
Some points that are probably worth clarifying for this pull request:
* *Pull request is submitted to dplyr instead of bigrquery*. I thought it is better to submit the pull request to dplyr since the new function heavily mimicks `sql_join.DBIConnection` and relies on pretty much the same non-exported dependencies: `auto_names, unique_names, common_by` etc. 

* *`sql_subquery.bigquery` is modified*. The problem with the current version `sql_subquery.bigquery` that is presented inside `bigrquery` package is that `build_sql` erases "vars" attribute when creating a subquery. However, "vars" are needed for correct subsequent processing of the join. I decided to go with the smallest and safest change possible and just added back "vars" to SQL object.

* only `collect()` works and not `compute()` would work with this pull request. (`compute()` fails due to missing `db_save_query.bigquery` which I submitted as a separate pull request here [hadley/bigrquery#51](https://github.com/hadley/bigrquery/pull/51))

**Please note** that now there are two different definitions `sql_subquery.bigquery` (one inside `dplyr` and another one inside `bigrquery`). It feels that a better approach is to move `sql_subquery.bigquery` to dplyr as well so it could be closer to `sql_join.bigquery`.

I can submit a separate pull request for moving `sql_subquery.bigquery` out of `bigrquery` to `dplyr` if that solution is found preferable.

### Tests
The code belows demonstrates a test of BigQuery JOIN via dplyr using publicly available data on Google Cloud. Please note that running the test does require to have a Google Cloud billing project set up already. 

The test should run for several seconds and billed for ~9GB of data (~$0.05 USD)
``` r
library(dplyr)
library(bigrquery)

shakespeare <- 
   src_bigquery("publicdata", "samples", billing="<BILLING_PROJECT>") %>% 
   tbl("shakespeare")
wikipedia <- 
   src_bigquery("publicdata", "samples", billing="<BILLING_PROJECT>") %>% 
   tbl("wikipedia")

result <-
  shakespeare %>% 
  select(word, word_count) %>%
  inner_join(wikipedia %>% select(title, num_characters),
             by=c("word"="title")) %>% 
  summarise( n = n(),
             word_count = sum(word_count),
             num_characters = max(num_characters)) %>%
  collect

result
```


akhmed1:
Moving this pull request to a separate branch. Please see new pull request.

PR #1249, opened 2015-07-05
[proposal] ORDER BY in window functions should not be automatically assumed to be also the order of the full final result

akhmed1:
### Introduction
When `build_query` generates window functions with `OVER (PARTITION BY x ORDER BY y)`, the query builder then also automatically assumes that not just the partition but also the full final result needs to be ordered by `x` and `y` and thus, inserts an extra `ORDER BY x,y` at the end even though it was never requested. 

Problems with current behavior:
* This creates a significant performance hit for larger queries for no reason. (to the extent that otherwise doable queries become impossible to run). 
* it appears that it is not possible for the user to disable that behavior and remove the offending `ORDER BY x,y` (without modifying the library code).
* this does not appear to be an SQL requirement

### Details
Consider the following test code:
``` r
library(dplyr)

if (!has_lahman("postgres")) {
  stop("You need to have Lahman data installed")
}

salaries <- lahman_postgres() %>% tbl("Salaries")

salaries %>%
  group_by( teamID ) %>%
  arrange( salary ) %>%
  mutate( lower_salary = lag(salary)) %>%
  explain
```
This generates the following query:
```
SELECT "yearID", "teamID", "lgID", "playerID", "salary", "lower_salary"
FROM (SELECT "yearID", "teamID", "lgID", "playerID", "salary", LAG("salary", 1, NULL) OVER (PARTITION BY "teamID" ORDER BY "salary") AS "lower_salary"
FROM "Salaries"
ORDER BY "teamID", "salary") AS "_W1"
```

As can be seen `build_query` adds `ORDER BY "teamID", "salary"` at the end even though it was never specifically requested. 

For large datasets, that could be a large performance hit. As of now, there seems to be no way to disable that behavior.

### Proposed fix

This pull request:
* Disables inheriting the `order by` from inside window functions. (It can always be added by the user explicitly if needed).

For example, with this pull request, the above code produces:
```
SELECT "yearID", "teamID", "lgID", "playerID", "salary", "lower_salary"
FROM (SELECT "yearID", "teamID", "lgID", "playerID", "salary", LAG("salary", 1, NULL) OVER (PARTITION BY "teamID" ORDER BY "salary") AS "lower_salary"
FROM "Salaries") AS "_W1"
```
which is valid SQL and removes the final ORDER BY performance hit

If it is desired to have a particular full final result order, then the users always have the option to specify it explicitly.
``` r
salaries %>%
  group_by( teamID ) %>%
  arrange( salary ) %>%
  mutate( lower_salary = lag(salary)) %>%
  arrange( teamID, salary ) %>%
  explain
```
which generates
``` sql
SELECT "yearID", "teamID", "lgID", "playerID", "salary", "lower_salary"
FROM (SELECT "yearID", "teamID", "lgID", "playerID", "salary", LAG("salary", 1, NULL) OVER (PARTITION BY "teamID" ORDER BY "salary") AS "lower_salary"
FROM "Salaries") AS "_W1"
ORDER BY "teamID", "salary"
```

### DIFF
--- a/R/tbl-sql.r
+++ b/R/tbl-sql.r
@@ -207,10 +207,10 @@ build_query <- function(x, limit = NULL) {
     # Don't use group_by - grouping affects window functions only
     group_by_sql <- NULL
 
-    # If the user requested ordering, ensuring group_by is included
+    # If the user does not use window functions, then order full results as requested
     # Otherwise don't, because that may make queries substantially slower
-    if (!is.null(x$order_by) && !is.null(x$group_by)) {
-      order_by_sql <- translate(c(x$group_by, x$order_by))
+    if (uses_window_fun(x$select, x)) {
+      order_by_sql <- NULL
     } else {
       order_by_sql <- translate(x$order_by)
     }


hadley:
I feel like I'm missing something here - didn't you explicitly request a sorting with `arrange( salary )`?

akhmed1:
@hadley I may misunderstand something but isn't `arrange(salary)` meant to apply only within the group (window function) in this example and not for the full result ? (There is a `group_by` right before it).

By the way,  `arrange(teamid)` was never requested yet `build_query` attempts to sort the final result by `teamid` anyway. 

My understanding is that there are two different (and independent) sortings here:
(1) within window function (this determines how `lag` is to be computed)
(2) full final result of the query (this determines how to sort the final-final results once everything is done)

*My concern is about (2) and not (1)*. As of right now, sorting order from step (1) is also enforced as sorting order at step (2) but I don't see why it is so. Unfortunately, the drawback is clear: there is a huge performance hit for very large datasets.

Another way to phrase it: there seems to be no way to generate this minimal and efficient SQL query:
```
SELECT  "teamID",  "salary", 
LAG("salary", 1, NULL) OVER
   (PARTITION BY "teamID" ORDER BY "salary") AS "lower_salary"
FROM "Salaries"
```
this is a valid SQL query but `build_query` will *always* insert the (unsolicited) `ORDER BY "teamID", "salary"` at the very end of the query and make it:
```
SELECT  "teamID",  "salary", 
LAG("salary", 1, NULL) OVER
   (PARTITION BY "teamID" ORDER BY "salary") AS "lower_salary"
FROM "Salaries"
ORDER BY "teamID", "salary"
```

Basically, the latter query fails for me due to a huge performance hit because of the sorting that was never requested. I can provide an example dataset (this requires a large dataset though).

hadley:
Arrange currently acts group wise, which is where the teamID comes from. 

What chain did you use to create and create the SQL?

IDSC3101:
@hadley I see but isn't the resulting SQL then somewhat inefficient with window functions? Sorting by `teamID` could be an expensive procedure on some backends and may not be desired. The chain that results in the "undesired" sorting by `teamID` is this (if that's your question?): 
``` r
salaries %>%
  group_by( teamID ) %>%
  arrange( salary ) %>%
  mutate( lower_salary = lag(salary)) %>%
  explain
```
As an example, this `teamID` sorting kills otherwise normal queries on BigQuery and unfortunately, there is no other way to disable it.

akhmed1:
@hadley I see but isn't the resulting SQL then somewhat inefficient with window functions? Sorting by `teamID` could be an expensive procedure on some backends and may not be desired. The chain that results in the "undesired" sorting by `teamID` is this (if that's your question?): 
``` r
salaries %>%
  group_by( teamID ) %>%
  arrange( salary ) %>%
  mutate( lower_salary = lag(salary)) %>%
  explain
```
As an example, this `teamID` sorting kills otherwise normal queries on BigQuery and unfortunately, there is no other way to disable it.

hadley:
If you want that behaviour, why wouldn't you do this?

```R
salaries %>%
  group_by( teamID ) %>%
  mutate( lower_salary = lag(salary, order_by = salary)) %>%
  explain
```
?

akhmed1:
@hadley  `mutate( lower_salary = lag(salary, order_by = salary))` does not work with `tbl_sql`. But your point is well-taken: I sifted through the code and got your example to work by replacing  `order_by` with `order`: `lag(salary, order_by = salary)` -- > `lag(salary, order = salary)`.
(Which is kind of strange... is it intentional?)

Regarding this pull request -- I see the point of these lines now. Even if sorting within group is requested, one must sort by `teamID` as well because there is no other way in SQL. One can't sort within the groups while leaving the order of groups untouched.

I am closing this pull request as it will violate this principle above.
Thanks!

PR #1251, opened 2015-07-06
Update join.r

aadler:
Adding descriptions of right and full join. Putting x, y, and NA consistently in \code{} wrapper.


hadley:
Thanks!

Issue #926, opened 2015-01-27
Escaping odd table names in SQLite not working


pschmied:
This report is reproduced from this Stack Overflow question:
http://stackoverflow.com/questions/28132697/bracket-escaped-table-names-with-dplyr?noredirect=1#comment44719663_28132697

I'm programmatically fetching a bunch of datasets, many of them having silly names that begin with numbers and have special characters like minus signs in them. Because none of the datasets are particularly large, and I wanted the benefit R making its best guess about data types, I'm (ab)using dplyr to dump these tables into SQLite.

I am using square brackets to escape the horrible table names, but this doesn't seem to work. For example:

    data(iris)
    foo.db <- src_sqlite("foo.sqlite3", create = TRUE)
    copy_to(foo.db, df=iris, name="[14m3-n4m3]")

This results in the error message:

   `Error in sqliteSendQuery(conn, statement, bind.data) : error in statement: no such table: 14m3-n4m3`

This works if I choose a sensible name. However, due to a variety of reasons, I'd really like to keep the cumbersome names. I am also able to create such a badly-named table directly from sqlite:

    sqlite> create table [14m3-n4m3](foo,bar,baz);
    sqlite> .tables
    14m3-n4m3

Without cracking into things too deeply, this looks like dplyr is handling the square brackets in some way that I cannot figure out. My suspicion is that this is a bug, but I wanted to check here first to make sure I wasn't missing something.

EDIT: I forgot to mention the case where I just pass the janky name directly to dplyr. This errors out as follows:

    library(dplyr)
    
    data(iris)
    foo.db <- src_sqlite("foo.sqlite3", create = TRUE)
    copy_to(foo.db, df=iris, name="14M3-N4M3")

    Error in sqliteSendQuery(conn, statement, bind.data) : 
      error in statement: unrecognized token: "14M3"

pschmied:
I've started trying to dig into this myself a bit. Just for fun, I decided to install / test against a Postgres database. This appears to work correctly without escaping:

    library(dplyr)
    library("RPostgreSQL")
    
    foo.db <- src_postgres("foo")
    
    foo <- copy_to(foo.db, df=iris, name="14M3-N4M3")

FvD:
Apparently the issue is with the first character being a number in "14M3". It tried to save a sequence of databases with a UUID and `copy_to` will fail as soon as there is a table name with a number as the first character (this is copying to a sqlite database).

Reports are added nicely as long as there is a letter in front:
```
[...]

PR #1258, opened 2015-07-10
Fixed issue #926 (problems with escaping table names for SQLite databases)

NikNakk:
Fixed issues described with SQLite at https://github.com/hadley/dplyr/issues/926 by creating new function for db_insert_into that supports escaping (unlike the current DBI::dbWriteTable that was used) and using build_sql rather than paste0 within db_query_fields.

This was also raised at http://stackoverflow.com/questions/28132697/bracket-escaped-table-names-with-dplyr

### DIFF
--- a/R/src-sqlite.r
+++ b/R/src-sqlite.r
@@ -127,7 +127,7 @@ src_translate_env.src_sqlite <- function(x) {
 
 #' @export
 db_query_fields.SQLiteConnection <- function(con, sql, ...) {
-  rs <- DBI::dbSendQuery(con, paste0("SELECT * FROM ", sql))
+  rs <- DBI::dbSendQuery(con, build_sql("SELECT * FROM ", sql))
   on.exit(DBI::dbClearResult(rs))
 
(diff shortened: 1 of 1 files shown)


hadley:
I'd rather fix this in RSQLite than in dplyr

NikNakk:
Fair enough, but I can't find the relevant code to fix in RQLite; it all seems to be in dplyr at present. Are you planning to migrate the src-sqlite code across?

Nick

hadley:
The `dbWriteTable()` method needs a fix?

NikNakk:
The `dbWriteTable` method seems to have been fixed upstream in the current `RSQLite` on github. There's still an outstanding issue with `dplyr` code (just a single line!), so I've opened a new pull request with that. It's a pretty trivial fix - changing `paste0` to `build_sql`. Part of the issue for me before was following the chain of S3 and S4 methods to find out which code was where (`copy_to` is an S3 generic, while `DBI::dbWriteTable` is an S4 one.)

PR #1260, opened 2015-07-12
makes small change to documentation `_q` -> `_`

ijlyttle:
also roxygenizes and catches something from `n_distinct.Rd`


hadley:
Thanks!

PR #1264, opened 2015-07-13
Progress bar now respects dplyr.show_progress

### LINKED ISSUE #1226: Switch for showing progress bar
This is kind of in the same vein as #428. The progress bar shown when using do() is often unwanted if you use dplyr inside a function. If the dplyr operation is only part of the computation it signals a wrong message about the running time to users, as there is no indication that the progress bar is only related to part of the calculation. I would like to either have the possibility of switching it off as an argument in do, or only have it shown if it is used in the global environment.

jimhester:
This allows the user to disable the progress bar by setting
dplyr.show_progress to FALSE.

Closes #1226

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,8 @@
 # dplyr 0.4.2.9000
 
+* `Progress` respects global option `dplyr.show_progress` (default is TRUE)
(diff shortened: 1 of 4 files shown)


hadley:
Thanks!

PR #1276, opened 2015-07-17
added option `show_classes` for `print.tbl_df`

w9:
Adding option `show_classes` for `print.tbl_df`
-----------------

I think being able to see classes directly would drastically increase my productivity and lower the chance of mistakes. Especially sometimes I need to distinguish between `character` and `factor`, or between `double` and `integer` etc. Plus, we have class names for omitted columns then why not the columns that are actually shown?

image

Adding option `m` for `print.tbl_df`
------------------

Instead of printing the top 10 by default, I changed it to printing 5 top and 5 bottom. Users can set the added `m` option to change how many bottom rows to show. Now the option `dplyr.show_min` will dictate the default values of both `n` and `m`, and is set to `5`.

I made so that setting one of `n` or `m` without the other implies the other to be `0`. So the behavior of previous code `print(n=...)` will not change.

image

image

image

image

### DIFF
--- a/R/tbl-df.r
+++ b/R/tbl-df.r
@@ -104,10 +104,10 @@ as.data.frame.tbl_df <- function(x, row.names = NULL, optional = FALSE, ...) {
 
 #' @rdname dplyr-formatting
 #' @export
-print.tbl_df <- function(x, ..., n = NULL, width = NULL, show_classes = NULL) {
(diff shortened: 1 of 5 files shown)


hadley:
I don't think this needs to be an option - it should always happen.  Can you please:

* fix the mangled code
* use `()` instead of `<>`  (to be consistent)
* add a bullet point to NEWS

cosinequanon:
This is cool, would love to see it actually merged

PR #1310, opened 2015-08-10
Respect print(n = Inf) for remote sources

### LINKED ISSUE #1308: print(n = Inf) does not work for some remote sources
Using `print(n = Inf)` for remote sources results in a query that ends with `LIMIT NULL`.This is not supported in several languages and causes the query to fail.

```
> db <- src_mysql('information_schema')
> t <- tbl(db, 'CHARACTER_SETS')
> t %>% print(n = Inf)
Source: mysql 5.6.25 [root@localhost:/information_schema]
From: CHARACTER_SETS [40 x 4]

Error in .local(conn, statement, ...) :
  could not run statement: You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near 'NULL' at line 3
In addition: Warning message:
In eval(expr, envir, enclos) : NAs introduced by coercion
> options(error = recover)
> t %>% print(n = Inf)
Source: mysql 5.6.25 [root@localhost:/information_schema]
From: CHARACTER_SETS [40 x 4]

Error in .local(conn, statement, ...) :
  could not run statement: You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near 'NULL' at line 3
In addition: Warning message:
In eval(expr, envir, enclos) : NAs introduced by coercion

Enter a frame number, or 0 to exit

 1: t %>% print(n = Inf)
 2: withVisible(eval(quote(`_fseq`(`_lhs`)), env, env))
 3: eval(quote(`_fseq`(`_lhs`)), env, env)
 4: eval(expr, envir, enclos)
 5: `_fseq`(`_lhs`)
 6: freduce(value, `_function_list`)
 7: withVisible(function_list[[k]](value))
 8: function_list[[k]](value)
 9: print(., n = Inf)
10: print.tbl_sql(., n = Inf)
11: print(trunc_mat(x, n = n, width = width))
12: trunc_mat(x, n = n, width = width)
13: as.data.frame(head(x, n))
14: head(x, n)
15: head.tbl_sql(x, n)
16: build_query(x, limit = as.integer(n))$fetch()
17: dbSendQuery(self$con, self$sql)
18: .valueClassTest(standardGeneric("dbSendQuery"), "DBIResult", "dbSendQuery")
19: is(object, Cl)
20: is(object, Cl)
21: .local(conn, statement, ...)

Selection: 16
Called from: top level
Browse[1]> self$sql
<SQL> SELECT `CHARACTER_SET_NAME`, `DEFAULT_COLLATE_NAME`, `DESCRIPTION`, `MAXLEN`
FROM `CHARACTER_SETS`
LIMIT NULL
```

Some languages allow for `LIMIT ALL` so that is one option but I think it makes the most sense to remove the limit part altogether when using `print(n = Inf)`

cosinequanon:
Fixes #1308

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -65,6 +65,8 @@ Until now, dplyr's support for non-UTF8 encodings has been rather shaky. This re
 
 * Up Rcpp dependency to 0.12.0, and remove the obsolete SHALLOW_COPY workaround
 
+* `print(n = Inf)` now works for remote sources (#1310).
+
 ## Hybrid evaluation
 
 * Hybrid evaluation does not take place for objects with a class (#1237).
--- a/R/tbl-sql.r
+++ b/R/tbl-sql.r
@@ -152,7 +152,12 @@ dim.tbl_sql <- function(x) {
 head.tbl_sql <- function(x, n = 6L, ...) {
   assert_that(length(n) == 1, n > 0L)
 
-  build_query(x, limit = as.integer(n))$fetch()
+  if (is.infinite(n)) {
+    limit <- NULL
+  } else {
+    limit <- as.integer(limit)
+  }
+  build_query(x, limit)$fetch()
 }
 
 #' @export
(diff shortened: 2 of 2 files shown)


hadley:
Could you please add a bullet point to NEWS?

hadley:
Thanks!

PR #1354, opened 2015-08-27
Add test for join not changing source data tables

krlmlr:
Same as #1071, which doesn't seem to have made it into master.


hadley:
Weird. I definitely remember merging it.

krlmlr:
Perhaps somebody force-pushed afterwards.

PR #1356, opened 2015-08-27
Update README.md

lindbrook:
Could you clarify Line 113? "They all work as similarly as possible across the range of data sources. The main difference is performance:"

### DIFF
--- a/README.md
+++ b/README.md
@@ -104,11 +104,11 @@ Single table verbs
 
 `dplyr` implements the following data manipulation verbs :
 
--   `select()`: selects a subset of columns or variables
--   `filter()`: selects a subset of rows or observations
--   `mutate()`: adds new columns or variables
+-   `select()`: selects a subset of variables
+-   `filter()`: selects a subset of observations
+-   `mutate()`: adds new variables
 -   `summarise()`: reduces a group(s) to a smaller number of values (e.g., summary statistics)
--   `arrange()`: re-orders rows or observations
+-   `arrange()`: re-orders observations
 
 They all work as similarly as possible across the range of data sources. The main difference is performance:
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -102,13 +102,13 @@ carriers_db2 <- flights_db2 %>% group_by(carrier)
...
--   `mutate()`: add new columns
--   `summarise()`: reduce each group to a smaller number of summary statistics
--   `arrange()`: re-order the rows
+-   `select()`: selects a subset of columns or variables
+-   `filter()`: selects a subset of rows or observations
+-   `mutate()`: adds new columns or variables
+-   `summarise()`: reduces a group(s) to a smaller number of values (e.g., summary statistics)
+-   `arrange()`: re-orders rows or observations
 
 They all work as similarly as possible across the range of data sources. The main difference is performance:
```
hadley: How about: "dplyr verbs are designed to work in the same way across multiple backends. This means that you can use the same code to operate on a local data frame or a remote database. The main difference is performance:"

README.md
```diff
@@ -102,13 +102,13 @@ carriers_db2 <- flights_db2 %>% group_by(carrier)
 Single table verbs
 ------------------
 
-`dplyr` implements the following verbs useful for data manipulation:
+`dplyr` implements the following data manipulation verbs :
 
--   `select()`: focus on a subset of variables
--   `filter()`: focus on a subset of rows
```
hadley: I'd prefer to stick consistently to variables and observations rather than columns and rows.

README.md
lindbrook: The difference in performance essentially boils down to the local v. remote
issue, right?

hadley: Right. Databases store the data on disk so are slower, but can handle more data

lindbrook: How about: "Because you use the same verb regardless of whether you are
working with local data frames or remote databases, your coding task is
simplified. The only real difference is that performance is affected by the
data's location."

hadley:
Thanks!

PR #1374, opened 2015-09-01
Swap order of dim_names and met_name in as.tbl_cube

krlmlr:
- to make it consistent with the `tbl_cube` constructor
- to allow adding a `met_name` argument to `as.tbl_cube.data.frame` later

Also, the `met_name` argument to `as.tbl_cube.table` now defaults to `"Freq"` for consistency with `as.data.frame.table`.

This is a breaking change, but perhaps justified since
> `tbl_cube` support is currently experimental

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -31,6 +31,12 @@
 
 * When printing a grouped data frame the number of groups is now printed with thousands separators. (#1398)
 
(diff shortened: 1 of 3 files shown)


hadley:
Can you please move NEWS up?

krlmlr:
Rebased and moved news.

hadley:
Thanks!

PR #1377, opened 2015-09-01
fix error in column building of 'frame-data'

kevinushey:
Fixes an unfortunate error in building the columns of `frame_data()`.


hadley:
Bummer :(  Can you please add a bullet to NEWS?

kevinushey:
Updated!

hadley:
Thanks!

PR #1394, opened 2015-09-08
Specify behavior of do() for empty input

### LINKED ISSUE #1393: Grouped do() with empty data frame
```
data_frame(a=character()) %>% group_by(a) %>% do(stop("Hey!"))
```

throws an error. I wouldn't expect `do()` to be called at all in this case (there are no groups!), but even if this is desired behavior, I think it should be documented.

Slightly related: #341.

krlmlr:
Fixes #1393.


hadley:
Thanks!

PR #1475, opened 2015-10-25
Update join.r to fix doc

potterzot:
Fix documentation error for inner join (I think). Should be all rows in x that match rows in *y*, not x.


hadley:
Thanks!

PR #1493, opened 2015-10-31
Support complex values in matrixToDataFrame

krlmlr:
Probably we also want a meaningful error message if the type is not supported (`default` branch).

Related: hadley/tidyr#134.


hadley:
Thanks! This would be another target for tibble.

krlmlr:
Sure. If you're planning any more changes to the tibble-related code in the near future, I'd better wait for the dplyr release before working again on tibble. Otherwise, it would be great to freeze that code during the transition period (probably a few weeks).

PR #1536, opened 2015-11-16
added CRAN-related badges to README

vsimko:
I just added 2 badges to README.Rmd source file and README.md file.
These badges show CRAN version of dplyr and number of downloads.


hadley:
I'd rather not include the download count, but the CRAN badge is a good idea. I'll add it myself.

PR #1548, opened 2015-11-24
Don't execute query for determining variable names if they are known

krlmlr:
Solved by forcing the `vars` argument in `tbl_sql()`. The default for this argument is an attribute of the `from` argument, but that argument gets overwritten before this attribute is extracted.

Currently, `update.tbl_sql()` always needs to run `db_query_fields()` e.g. for a `left_join()`.

Closes #1549.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,1 @@
+/NEWS.md merge=union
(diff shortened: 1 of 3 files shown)


kevinushey:
Nice catch!

hadley:
Thanks!

PR #1550, opened 2015-11-25
New arguments "unique_indexes" and "indexes" to compute()

### LINKED ISSUE #1499: Add index option to compute()
It would be great if `collect()` could be accompanied by the creation of indexes for performance.

krlmlr:
- New argument `unique` to db_create_indexes() and db_create_index()
- New argument `unique_indexes` to `copy_to()`
- Index creation (also the uniqueness check) currently works only for SQL sources and is a no-op for data.frames and data.tables
- With tests

Fixes #1499.


hadley:
What's the advantage of unique indices in the dplyr context? People are not going to be adding data to the database with dplyr, so why worry about the uniqueness constraint?

krlmlr:
I'm still wondering how you're going to do #1619 -- a compute() with adding unique indexes could just do the trick. It's a cheap way to ensure that something is unique.

hadley:
But then you have to catch the error raised by the database, which is a bit icky. I don't think 1619 will necessary work with databases.

krlmlr:
Let's forget about #1619 for now. To your original question: I think the creation of intermediate tables is very well suited for the scope of dplyr (think data pipelines); here, you may want to double-check uniqueness in the process. My approach makes it cheap and potentially improves performance of further queries, while being non-intrusive (one extra arg to compute()).

Perhaps you prefer adding a new verify_key() verb that calls compute() behind the scenes for SQL sources?

hadley:
Ok, that makes sense.  And I like the idea of `verify_key()` function.

PR #1568, opened 2015-12-02
Improve construction and coercion of tbl_cube

krlmlr:
- new argument met_name = guess_met() in as.tbl_cube.data.frame
    - all coercion methods generate identical cubes for the same data
- properly handle incomplete cubes and duplication of dimensions

With tests.


hadley:
Thanks!

PR #1582, opened 2015-12-11
Update JoinVisitorImpl.h

### LINKED ISSUE #1578: Problem joining with POSIXct
I noticed when joining on POSIXct the class of what should be the returned data frame's "time" column is missing the virtual class "POSIXt". This created an issue for me when I passed the data to another function to send to a database (MySQL). I added a step to explicitly add the virtual class, but this seems like it should be added in the join?

Example:
```R
> d1 <- data_frame(t=seq(as.POSIXct("2015-12-01", tz="UTC"), length.out=2, by="days"), x=1:2)
> d1
Source: local data frame [2 x 2]

           t     x
      (time) (int)
1 2015-12-01     1
2 2015-12-02     2
> class(d1$t)
[1] "POSIXct" "POSIXt" 
> d2 <- d1 %>% inner_join(d1, by = "t")
> d2
Source: local data frame [2 x 3]

           t   x.x   x.y
      (POSI) (int) (int)
1 2015-12-01     1     1
2 2015-12-02     2     2
> class(d2$t)
[1] "POSIXct"
```

joel23888:
Add virtual class POSIXt to class of join result to fix #1578.

### DIFF
--- a/inst/include/dplyr/JoinVisitorImpl.h
+++ b/inst/include/dplyr/JoinVisitorImpl.h
@@ -359,7 +359,7 @@ namespace dplyr{
         RObject tzone ;
 
         inline SEXP promote( NumericVector x){
-          x.attr("class") = "POSIXct" ;
+          x.attr("class") = Rcpp::CharacterVector::create("POSIXct", "POSIXt") ;
           if( !tzone.isNULL() ){
             x.attr("tzone") = tzone ;
           }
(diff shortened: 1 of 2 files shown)


hadley:
Would you mind adding a test for this?

joel23888:
I added a test but the Travis build returned an error, which I am not sure about.

hadley:
Thanks!

PR #1583, opened 2015-12-11
Use 'select' generic

jimhester:
This is mainly a proof of concept for using a 'select' generic from the generics package as discussed with @mtmorgan and @hadley yesterday.


jimhester:
The main drawback from this implementation compared to previous is that previously the first formal was called `.data` to prevent clashes with variables but when using the generic the first argument needs to be named `x` (or whatever the first argument in generics ends up being, it can be changed).

I don't really know a good way to fix this problem. You need to explicitly specify an object to dispatch on because if you just use `...` the S4 generic does multiple dispatch on _all_ of the arguments rather than just the first.

jimhester:
We discussed using `.data` as the first formal argument for `select()` in generics and decided it was pragmatic to do so.

Also Michael Lawrence suggested simply defining a `select.default` method in dplyr which would just call `select_(.data, .dots = lazyeval::lazy_dots(...))`, which would then avoid having to define all of the select methods. Which seems like a good idea to me and would drastically reduce the diff.

hadley:
But is it "ok" for someone other than the creator of the generic to define a default method?

jimhester:
You are right it is probably dicey. I am not really sure what would happen if two different packages both defined a default method on the same generic (defined in another package), seems like it would be confusing to at the very least.

I somehow overlooked the fact that there is a 'tbl' superclass for all of the tbl types. https://github.com/hadley/dplyr/commit/595bab568ceeeda77d75ae1b6b3b1bff3be9cea5 defines a method for just `select.tbl` and goes back to using `.data`, which makes this much nicer all around.

jimhester:
https://github.com/jimhester/dplyr/commit/148bf56e7a5c66a1b21b222a07e5698e4eb8990c is now passing, the RCpp release build failed for an unrelated GitHub 403.

However it is generating a WARNING from R CMD check (https://travis-ci.org/hadley/dplyr/jobs/96756094#L1450-L1451) because of two `\description{}` sections in the select man page (https://github.com/jimhester/dplyr/blob/148bf56e7a5c66a1b21b222a07e5698e4eb8990c/man/select.Rd#L40-L50).

The second `\description{}` is automatically added when re-exporting a function. Not sure if there is a way to turn this off, or have it append to the existing description if already specified.

hadley:
I'm going to close for now, but lets reconsider in the future.

PR #1593, opened 2015-12-22
Update sets.r

helix123:
I feel like "compare" is not the right term here (makes me think about "=="), especially so for union()


hadley:
Thanks!

PR #1664, opened 2016-02-15
Make current behavior of create argument in src_sqlite explicit

ethanwhite:
The current behavior of src_sqlite when create = TRUE is to connect to
the existing database if it is present and to create a new database if
there is no file at the specified path. The docs are not explicit about
using the existing database if it is present, which has caused some
confusion in at least one instance [1].

This change makes the current behavior explicit in the documentation.

[1] https://github.com/weecology/bbs-forecasting/pull/16#discussion_r52901968


hadley:
Thanks!

PR #1687, opened 2016-03-05
add blank lines so md lists render correctly.

jayhesselberth:
trivial fix for correctly formatted MD lists in HTML output.


hadley:
Thanks!

Issue #1703, opened 2016-03-11
count does not use current group_by (unintuitive)


edwindj:
Hi, 

Currently `count` does not 'respect' `group_by` statements, which is not intuitive (I observed this multiple times when training novice R/dplyr users).

Example:
```R
data(diamonds, package="ggplot2")
diamonds %>% count(cut)

diamonds %>% group_by(cut) %>% count # gives total count, not group_by

# Source: local data frame [1 x 1]
#
# n
# (int)
# 1 53940

diamonds %>% group_by(cut) %>% count(clarity) # gives count by clarity (but not cut)

# Source: local data frame [8 x 2]
# 
# clarity     n
# (fctr) (int)
# 1      I1   741
# 2     SI2  9194
# 3     SI1 13065
# 4     VS2 12258
# ...
```
I know that one way to 'fix' this is to use `tally` and to expand the group_by statement, however I will submit a PR in which `count` expands the current grouping.

PR #1704, opened 2016-03-11
count now uses group_by  (issue #1703)

edwindj:
This is a fix for issue #1703.

Best regards,

Edwin

### DIFF
--- a/R/tally.R
+++ b/R/tally.R
@@ -74,6 +74,7 @@ count <- function(x, ..., wt = NULL, sort = FALSE) {
 #' @export
 #' @rdname tally
 count_ <- function(x, vars, wt = NULL, sort = FALSE) {
-  grouped <- group_by_(x, .dots = vars)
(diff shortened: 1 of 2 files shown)


R/tally.R
```diff
@@ -74,6 +74,7 @@ count <- function(x, ..., wt = NULL, sort = FALSE) {
 #' @export
 #' @rdname tally
 count_ <- function(x, vars, wt = NULL, sort = FALSE) {
-  grouped <- group_by_(x, .dots = vars)
+  old_vars <- lapply(groups(x), lazyeval::as.lazy, env=parent.frame())
```
hadley: I think all you need is to set `add = TRUE` in the call to `group_by_`

edwindj: Thanks! I didn't notice the `add` argument of `group_by_`: 
I've updated the PR with your suggestion

Best regards!

hadley:
Thanks - did this by hand since I wanted to add a test and the code was simple.

Issue #631, opened 2014-09-26
Equivalent of SQL CASE


edwindj:
In data cleaning data often contains values that one would like to impute/replace.

For example: 

In a survey for smokers, one would not ask a person his or her cigarette consumption when he/she has indicated not to be a smoker. So technically the number of cigarretes is not measured, but it can be deduced to be zero.

```S
smoke <- read.csv(textConnection(
"name,is_smoker,cigarrettes,cigares
 John,TRUE,15, 0
 Adam,FALSE,NA,NA
 Phil,TRUE,0,1
"
))

#standard R way
smoke$cigarrettes[!smoke$is_smoker & is.na(smoke$cigarrettes)] <- 0
```
How does one do this elegantly in dplyr?

My 2 cents:
```S
#dplyr option 1: split the data set, fix one part and combine the two
smoker_true <- smoke %>% filter(!is_smoker, is.na(smoke$cigarettes))
# complement
smoker_false <- smoke %>% filter(is_smoker | !is.na(smoke$cigarettes))

smoke2 <- rbind( smoker_true %>% mutate(cigarrettes=0)
               , smoker_false
)
```
which I don't find elegant dplyr code.

So try a little harder
```S
# dplyr option 2: replace the whole variable and do the filtering in the assignment
smoke2 <- 
  smoke %>%
  mutate(cigarrettes=ifelse(!is_smoker & is.na(cigarrettes), 0, cigarrettes))
```
It works, but this is not efficient/workable for the database backends and I find it unelegant to
do the filtering in the assignment.

I suggest a new method:
`mutate_where` that allows for changing values of _existing_ variables of a subset of a `tbl`.
So something like
```S
smoke2 <-
   smoke %>% mutate_where(.where=!is_smoker&is.na(cigarrettes), cigarettes=0)
```  
Or another option (in analogy with `group` and `grouped_<tbl>`) to change `filter` or introduce `filtered_<tbl>`/add_filter: setting a filter lets one work on a subset of the data, but one can `remove_filter`.

```S
smoke2 <-
   smoke %>% 
   add_filter(is_smoker, is.na(cigarrettes) %>%
   mutate(cigarrettes=0) %>%
   remove_filter()
```

Any better ideas or suggestions?

mattwg:
I currently solve this problem as follows:
```
df <- data.frame(duration=c(seq(1,28),NA))

df %>%
    mutate(my_duration=
                       ifelse(is.na(duration),0,
[...]

PR #1718, opened 2016-03-16
More verbose error messages in case_when()

krlmlr:
by showing the expression that fails.  Useful for `transmute()` fests where many variables are recoded.

~~Implementation: New eval_with_expr() and expr() that store the original expression along with the result.~~

Also fixes glitch in the first LHS error message ([ vs. [[).

With test.

Related to #631.

### DIFF
--- a/R/case_when.R
+++ b/R/case_when.R
@@ -32,7 +32,7 @@
 #'   x %% 35 == 0 ~ "fizz buzz"
 #' )
 case_when <- function(...) {
-  formula_args <- lapply(as.list(substitute(list(...))[-1]), deparse)
+  formula_args <- lapply(as.list(substitute(list(...))[-1]), deparse_trunc)
   formulas <- list(...)
   n <- length(formulas)
 
@@ -54,7 +54,7 @@ case_when <- function(...) {
 
     query[[i]] <- eval_with_expr(f[[2]], envir = env)
     if (!is.logical(query[[i]])) {
-      stop("LHS of case ", i, " (", expr(query[[i]]),
+      stop("LHS of case ", i, " (", deparse_trunc(expr(query[[i]])),
            ") is ", typeof(query[[i]]), ", not logical",
         call. = FALSE)
     }
@@ -69,11 +69,11 @@ case_when <- function(...) {
   for (i in seq_len(n)) {
     check_length(
       query[[i]], out,
-      paste0("LHS of case ", i, " (", deparse(expr(query[[i]])), ")"))
+      paste0("LHS of case ", i, " (", deparse_trunc(expr(query[[i]])), ")"))
 
     out <- replace_with(
       out, query[[i]] & !replaced, value[[i]],
-      paste0("RHS of case ", i, " (", deparse(expr(value[[i]])), ")"))
+      paste0("RHS of case ", i, " (", deparse_trunc(expr(value[[i]])), ")"))
     replaced <- replaced | query[[i]]
   }
 
--- a/R/case_when.R
+++ b/R/case_when.R
@@ -32,7 +32,6 @@
 #'   x %% 35 == 0 ~ "fizz buzz"
 #' )
 case_when <- function(...) {
-  formula_args <- lapply(as.list(substitute(list(...))[-1]), deparse_trunc)
   formulas <- list(...)
   n <- length(formulas)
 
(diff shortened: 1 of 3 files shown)


R/case_when.R
```diff
@@ -45,18 +46,20 @@ case_when <- function(...) {
   for (i in seq_len(n)) {
     f <- formulas[[i]]
     if (!inherits(f, "formula") || length(f) != 3) {
-      stop("Case ", i , " is not a two-sided formula", call. = FALSE)
+      stop("Case ", i , " (", formula_args[[i]],
```
hadley: I don't understand why you need `formula_args` here. Why not `deparse_trunc(formula[[i]])` work?

R/case_when.R
```diff
@@ -32,6 +32,7 @@
 #'   x %% 35 == 0 ~ "fizz buzz"
 #' )
 case_when <- function(...) {
+  formula_args <- lapply(as.list(substitute(list(...))[-1]), deparse)
```
hadley: I'm always suspicious of bare `deparse()` because for long expressions it will return a character vector of length > 1

R/case_when.R
```diff
@@ -45,18 +46,20 @@ case_when <- function(...) {
...
+           ") is not a two-sided formula", call. = FALSE)
     }
 
     env <- environment(f)
 
-    query[[i]] <- eval(f[[2]], envir = env)
+    query[[i]] <- eval_with_expr(f[[2]], envir = env)
     if (!is.logical(query[[i]])) {
-      stop("LHS of case ", i, " is ", typeof(query[i]), ", not logical",
+      stop("LHS of case ", i, " (", expr(query[[i]]),
```
hadley: Again, why not use `f[[2]]` directly?

R/case_when.R
krlmlr: Would like to use `deparse_trunc()` as suggested above, but to me it looks like the [current implementation](https://github.com/hadley/dplyr/blob/4920c8b5fe9267bdd9d6f1270980f643fea5657e/R/utils.r#L25-L30) also might return a vector in some cases. Is this desired?

R/case_when.R
krlmlr: For consistency: the pattern expr(query[[i]]) is repeated below, too, where f isn't available anymore.

R/case_when.R
krlmlr: Missed the [1] in text; everything looks well with deparse_trunc().

R/case_when.R
hadley: But you can easily recreate it with `formula[[i]]`?

R/case_when.R
krlmlr: I'm getting `"object of type 'closure' is not subsettable"` if I try this. We need the *representation* of the formula here (as stored in formula_args); below we already work with expressions captured by the formulas.

krlmlr:
PTAL.

R/case_when.R
hadley: Did you notice my typo? The representation of the formula is stored inside the formula.

krlmlr: `deparse_trunc(formulas[[i]])` doesn't work either -- it yields the *value* of `formulas[[i]]`, while we're looking for the *expression* here. (formulas[[i]] isn't a formula in this case.)

krlmlr:
Next try, much simpler now.

R/case_when.R
```diff
@@ -32,6 +32,7 @@
 #'   x %% 35 == 0 ~ "fizz buzz"
 #' )
 case_when <- function(...) {
+  formula_args <- lapply(as.list(substitute(list(...))[-1]), deparse_trunc)
```
hadley: I am certain that you don't need this line because the whole point of using formulas is that they're quoting operators

krlmlr: This is just for the first test -- are all arguments formulas? Catch-22...

krlmlr:
Took me a while to realize you can substitute(list(...)) even after evaluating ... . Now calling deparse_trunc() for the arg itself only in case of error.

tests/testthat/test-case-when.R
```diff
@@ -4,13 +4,33 @@ test_that("zero inputs throws an error", {
   expect_error(case_when(), "No cases provided")
 })
 
+test_that("error messages", {
+  withr::with_options(list(width = 40),
```
hadley: I wouldn't worry about this test - it looks fragile, it's effectively just testing the behaviour of `deparse_trunc`

krlmlr:
Done, PTAL.

hadley:
Thanks!

Issue #1717, opened 2016-03-16
.default param does not work in recode.factor


tklebel:
``` r
x <- factor(c("f", "m", "t"))

# works as expected
recode(x, "f" = "female", "m" = "male", "t" = "trans")
#> [1] female male   trans 
#> Levels: female male trans

# does not work
recode(x, "f" = "female", "m" = "male", .default = x)
#> Error: `.default` has type 'integer' not 'character'
```

hadley:
Yes, because `default.factor()` operates on the levels, not the individual values. So you need `recode(x, "f" = "female", "m" = "male", .default = levels(x))`

PR #1723, opened 2016-03-17
document .default behaviour for factor.recode #1717

tklebel:
For me it was not obvious how to use `.default` with factors, therefore I think it's sensible to document this with an example.


hadley:
Thanks!

PR #1729, opened 2016-03-22
Fix warning message for collect()

krlmlr:
if incomplete results were retrieved: n = -1 raises an error with collect(), only n = Inf works.


hadley:
Thanks!

PR #1595, opened 2015-12-23
Use tibble package

krlmlr:
Reexports all functions exported by tibble.

Fixes #1488.

A few tests (of functions now private to tibble) had to be removed. One test had to be adapted, because `names(data_frame())` is now `character(0L)` instead of `NULL`.

PR #1737, opened 2016-03-24
Use tibble package, but keep tbl_df() here

krlmlr:
Same as #1595, but tbl_df() remains in dplyr. Documentation of tbl_df() still needs to be adapted.

Closes #1595.


krlmlr:
Should work with https://github.com/hadley/tibble/pull/57.

Compare #1595 and #1737 locally:

```
git remote add krlmlr https://github.com/krlmlr/dplyr.git
git fetch krlmlr
git diff krlmlr/tibble-2 krlmlr/tibble-3
```

GitHub compare currently doesn't work here, branches are unrelated.

See 211c10c for documentation update.

hadley:
This should work with release tibble too, right?

krlmlr:
Yes; only if both are attached, there'll be warnings.

hadley:
Ah, that's true.  I think that's ok - unless you have any objections, I'll merge this PR, and by the time dplyr is on CRAN, we'll have had another tibble release.

krlmlr:
Fine with me.

krlmlr:
I'll just add another commit that removes a check warning.

hadley:
Thanks!

PR #1740, opened 2016-03-24
check for zero-length vars in distinct

zkamvar:
I noticed that when distinct is run without any variables selected, it returns an error on the conditional that checks for zero-length vars. This would prevent that.


hadley:
I think this is what you mean, but you didn't supply an example so I'm not sure.

PR #1757, opened 2016-04-05
c.sql() gains con argument

krlmlr:
- purpose: con is passed to escape() to fix inconsistent quoting
- c.sql() is now always called explicitly, consider renaming it
- with test

In general, I think "ident" and "sql" should be different classes, independent of each other, managed by a new "query" class. The "query" class could be modeled as a list that contains objects of classes "ident" and "sql" (and perhaps others), c.query() simply concatenates lists, and sql_render.query() would dispatch to the members of the list and quote the identifiers (when the connection is known for sure).


hadley:
Thanks!

PR #1766, opened 2016-04-09
Update rank.R

thelostscientist:
The function mentioned in the msdn blog does a sort before the split.


hadley:
Thst doesn't seem likely to be correct to me.

Issue #1767, opened 2016-04-12
recode() improvements


lionel-:
##### Create factors

`recode_factor()` would be handy to create factors with levels ordered as they appear in the recode call.

```{r}
recode_factor <- function (.x, ..., .default = NULL, .missing = NULL) {
  levels <- c(...)
  recoded <- recode(.x, ..., .default = .default, .missing = .missing)
  factor(recoded, levels)
}
```

##### Pronoun

Do we need a pronoun for `.x`? Would be useful for default values when mapping over the columns of a data frame.

```{r}
df %>% dmap(recode, `99` = NA_real_, .default = .x)
```

##### `labelled` method

Alias for `recode.numeric()`? Or would it make sense to have a default method that dispatches on vector type?

hadley:
Want to do a PR for `recode_factor()`?

Instead of a pronoun, maybe the default value for `.default` should be an NA, and then `NULL` would mean use instead? Or is there another sentinel value that would make sense?

I think it's reasonable for haven to define `recode.numeric()`. I'm not a huge fan of implementing special dispatch rules.

lionel-:
> Want to do a PR for recode_factor()?

yup I'll do it tomorrow.

> maybe the default value for .default should be an NA, and then NULL would mean use instead?

I like this as well. Which one do you prefer for a PR?

hadley:
I think it's reasonable for `NULL` to mean use `.x` as default

hadley:
And maybe that should be the default behaviour? It seems reasonable that recode leaves unspecified values the same by default.

lionel-:
I think so too. Will change that as well in the forthcoming PR

GarciaRios:
`recode`  is great! it nicely recodes from character to numeric...So this works:

```
x <- c("a", "b", "c")
y <- recode(x, a = 1, b = 2, c= 3)
y 
```
But  it won't work with factors even using `recode_factor`

```
xf <- factor(c("a", "b", "c"))
yf <- recode(xf, a = 1, b = 2, c= 3)
Error: `a` has type 'double' not 'character'

yf <- recode_factor(xf, a = 1, b = 2, c= 3)
Error: `a` has type 'double' not 'character'
```


You have to treat it as character and use recode_factor so that it goes back to factor. Seems pretty verbose

```
recode_factor(as.character(xf), a = 1, b = 2, c= 3)
[1] 1 2 3
Levels: 1 2 3
```

PR #1781, opened 2016-04-21
Recode improvements

lionel-:
As discussed in #1767 

Simply reusing `.x` for unmatched values was producing type mismatch errors that were somewhat obscure. To avoid this, this sets the default values to NA when the types are not compatible. Maybe this needs a `message()` or `warning()`?


R/recode.R
```diff
@@ -7,35 +7,57 @@
...
 #'
 #'   All replacements must be the same type, and must have either
 #'   length one or the same length as x.
-#' @param .default If supplied, all values not otherwise matched will be
-#'   given this value instead of \code{NA}. Must be either length 1 or the same
-#'   length as \code{x}.
-#' @param .missing If supplied, any missing values in \code{x} will be
+#' @param .default If supplied, all values not otherwise matched will
+#'   be given this value. If not supplied and if the replacements are
+#'   type-compatible with the original values in \code{.x}, unmatched
```
hadley: type-compatible => "the same type as"

R/recode.R
```diff
@@ -128,3 +152,35 @@ find_template <- function(...) {
 
   x[[1]]
 }
+
+recode_default <- function(default, x, out) {
```
hadley: This might be slightly too strict for integer vs. numeric (but I see I use the same code elsewhere, so you don't need to change it here). I'm starting to think it might be worthwhile to pull out all this sensible type conversion stuff into one place.

R/recode.R
lionel-: Now changed

R/recode.R
lionel-: Also I think it might be better to convert `NA` values automatically. The typed NAs are not very user-friendly.

hadley:
Thanks!

PR #1784, opened 2016-04-23
Changes to work nicely with Redshift (RPostgres and optional BEGIN on copy_to)

sicarul:
Hello, i tried to use dplyr for my Redshift database with enforced SSL, and the only R library able to connect for me is RPostgres, RPostgreSQL doesn't work for me. I understood from issues i saw that a migration to RPostgres was in order, sorry if i misinterpreted that.

I changed src_postgres to use RPostgres instead of RPostgreSQL and added an additional parameter to copy_to so the BEGIN at the start of the transaction is optional. I added this because for some reason Redshift fails miserably when BEGIN is called, but works flawless otherwise.

I have tried many transformations and this change seems to work fine, though i understand it may break people's code if they are using this assuming it will use the RPostgreSQL driver. If there's a better way to implement this i'll be happy to modify my code per the team's instructions, i'd like to get this upstream for because it'll probably be useful for other redshift users like myself.

Thanks! This R package is magical


sicarul:
Ok, i seem to have some docs issues, sorry, i will fix those

sicarul:
It seems tests keep failing because RPostgres isn't installed in the Vagrant, is there somewhere i should specify that so it builds ok?
Thanks

krlmlr:
Currently, RPostgres and RPostgreSQL are two different packages, consolidation seems unlikely. Do we want to support both?

sicarul:
@krlmlr Hey, sorry about the last state, the PR was messed up. I just consolidated the changes into one commit. My proposal is replacing the driver, but we could just create a new driver too, i just thought mantaining two different drivers for the same database was too much. RPostgres does support Redshift with SSL, while RPostgreSQL does not.

I'm already using this in production and it rocks, i'm able to create very complex ETLs in R which would normally be giant queries or slow R code but with this package i can mix the best of both worlds, i think other people will probably find this useful too

hadley:
Given the maintenance status of RPostgreSQL, I think we'll move to RPostgres once it's on CRAN.

sicarul:
Cool! Looking forward to it 👍  Meanwhile i'l keep using my fork then, thanks!

hadley:
I'll close for now, but will definitely incorporate once RPostgres is on CRAN. In the mean time, would you mind filing issues on RPostgres - ideally you shouldn't need to override any of the methods in dplyr, instead relying on correct DBI implementation.

sicarul:
Ok, thanks @hadley, will look into this when i have some spare time! 👍

PR #1785, opened 2016-04-25
add support for using limit on sql select

javierluraschi:
@hadley I'm implementing the use of `LIMIT` while fetching rows in `collect`. This is my first attempt, planning to run `testthat` and iterate as needed tomorrow. However, I think I could use some feedback before I get further:

1. Is there any particular reason why `dplyr` currently does not use `LIMIT` in `collect` queries?
2. Any pointers on how to test `LIMIT` with queries or other paths? From the code, it seems to me I'm not generating `LIMIT` through all entries, would be good to know which paths I should track.

### DIFF
--- a/R/sql-generic.R
+++ b/R/sql-generic.R
@@ -13,13 +13,15 @@ NULL
 #' @rdname backend_sql
 #' @export
 sql_select <- function(con, select, from, where = NULL, group_by = NULL,
-                       having = NULL, order_by = NULL, distinct = FALSE, ...) {
+                       having = NULL, order_by = NULL, distinct = FALSE,
+                       limit = NULL, ...) {
   UseMethod("sql_select")
 }
 #' @export
 sql_select.default <- function(con, select, from, where = NULL,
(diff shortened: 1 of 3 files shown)


hadley:
I explicitly removed this because `LIMIT` is an SQL extension which is only supported by a handful of databases. I think it's better to rely on `dbFetch()` with restricted `n`.

javierluraschi:
@hadley Sound good. How would you change about this change instead? It adds a `LIMIT` parameter to `sql_select` but without using it by default. It can only be used from backends that decide to use `sql_select` and explicitly pass this parameter. Thoughts?

hadley:
I'm hesitant to add non-compliant SQL 92 to the default method.   Maybe dplyr needs some helper functions so that writing your own `sql_select` method is easier/simpler?

javierluraschi:
@hadley Sounds good. I'll work in my branch and then refactor back helper functions at some point, if needed. Closing this PR for now.

PR #1787, opened 2016-04-26
allow tables to be subclassed

javierluraschi:
@hadley: Am I subclassing a `tbl_sql` incorrectly or are we missing to subclass this in dplyr?

This is what I'm doing:

```
#' @export
tbl.src_s <- function(src, from, ...) {
  tbl_sql("s", src = src, from = from, ...)
}
```

However, the generated table is not of type "s". The change in the PR solves this issue.


hadley:
Oops, thanks

PR #1801, opened 2016-05-06
add COMMIT for postgres

leoncamel:
It seems PostgreSQL backend did not implement "COMMIT" method. I add the missing method for postgresql.


leoncamel:
The travis-ci shows "missing: dtplyr". I am not sure if it is my bad.

@hadley any comments about this patch?

hadley:
Could you please add a comment indicating why you only need this and not begin and rollback as well? (Presumably due to partial support from RPostgreSQL)

And can you please re-document to update the NAMESPACE?

hadley:
@leoncamel do you have time to update this?

PR #1808, opened 2016-05-09
Add dtplyr as remote dependency

krlmlr:
so that Travis tests can find it.


krlmlr:
Installation of dtplyr seems to loop: https://travis-ci.org/krlmlr/dplyr/builds/128871717. I'm working on it, but it's tricky even with dev-devtools...

PR #1815, opened 2016-05-10
WIP: Avoid segfaults for unsupported types

### LINKED ISSUE #1803: segfault from filter/slice with raw vector column
R crashes when I dplyr::filter a tbl_df with a raw column. 

I get this on Windows Rgui and in Ubuntu console - see session_info outputs below. 

```R
library(dplyr)
devtools::session_info()  ## see below
d <-  as.raw(c(0x01, 0x01, 0x00, 0x00))
b <- data_frame(x = d)
## ok
b[1,]

## boom!
filter(as.data.frame(b), row_number() == 1)

## or with slice (boom!)
# slice(b, 1)
```

krlmlr:
Fixes #1803.

The error should be raised in a central location, and textual type descriptions should be printed (with Rf_type2char). This is largely obsolete by #1817, but should be added nevertheless to avoid surprises.

### DIFF
--- a/inst/include/dplyr/Collecter.h
+++ b/inst/include/dplyr/Collecter.h
@@ -382,6 +382,8 @@ namespace dplyr {
         case VECSXP: return new Collecter_Impl<VECSXP>(n) ;
         default: break ;
         }
+
+        stop( "unimplemented vector type: %d", TYPEOF(model) ) ;
         return 0 ;
     }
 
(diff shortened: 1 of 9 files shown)


hadley:
I think these would be more natural in the `default` statements rather than after the switch block. Would you mind changing?

hadley:
Actually, I can do this myself.

krlmlr:
Thanks, sorry I was slow on that.

PR #1816, opened 2016-05-11
escape partition fields for window functions

javierluraschi:
Need to escape partition fields, for instance, this sql:

```
... rank() OVER (PARTITION BY "playerID" ORDER BY "H" DESC) AS `zzz4` ...
```

Now generates as:

```
... rank() OVER (PARTITION BY `playerID` ORDER BY `H` DESC) AS `zzz4` ...
```

### DIFF
--- a/R/translate-sql.r
+++ b/R/translate-sql.r
@@ -143,10 +143,10 @@ translate_sql_ <- function(dots,
   variant <- sql_translate_env(con)
 
   if (window) {
-    old_group <- set_partition_group(vars_group)
+    old_group <- set_partition_group(escape(ident(vars_group), con = con))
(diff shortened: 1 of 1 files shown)


hadley:
It feels like the escaping should happen in `over()`, but I think I've been excessively clever about how `con` gets implicitly passed around so it has to be here.

hadley:
I had to revert this because it was making unit tests fail, because it was double-escaping. Could you please take another look?

javierluraschi:
@hadley yes, taking a look.

javierluraschi:
@hadley Sent PR https://github.com/hadley/dplyr/pull/1861

PR #1826, opened 2016-05-13
Improve default handling in recode.factor() and recode_factor()

lionel-:
This fixes a couple of issues for recoding factors:

- `recode.factor()` now uses `.x` as default when `.default` is NULL, just like the other methods

- `recode_factor()` handles vector default better

### DIFF
--- a/R/recode.R
+++ b/R/recode.R
@@ -86,7 +86,7 @@ recode.numeric <- function(.x, ..., .default = NULL, .missing = NULL) {
     replaced[.x == vals[i]] <- TRUE
   }
 
-  .default <- recode_default(.default, .x, out)
+  .default <- recode_default(.default, .x, out, replaced, recode_default_vector)
   out <- replace_with(out, !replaced & !is.na(.x), .default, "`.default`")
   out <- replace_with(out, is.na(.x), .missing, "`.missing`")
   out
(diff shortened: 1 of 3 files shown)


lionel-:
Also I think we should issue an informative message when:

- not all values are specified in `...` and `.default` is NULL
- the type of `.x` and the replacement values is not compatible.

Otherwise it might be a bit puzzling that sometimes `.x` is used for default values and sometimes they are set to `NA`. The message could be:

```
Cannot fill in defaults as .x is not compatible
Please specify replacements exhaustively or supply .default
```

lionel-:
> Also I think we should issue an informative message

The last commit now issues a warning.

R/recode.R
```diff
@@ -153,34 +154,43 @@ find_template <- function(...) {
...
 }
 
-recode_default <- function(default, x, out) {
-  same_type <- identical(typeof(x), typeof(out))
+recode_default <- function(default, x, out, replaced, default_fun) {
+  default <- default_fun(default, x, out)
+
+  if (is.null(default) && sum(replaced & !is.na(x)) < length(out[!is.na(x)])) {
+    warning("Unreplaced values treated as NA as .x is not compatible
+  Please specify replacements exhaustively or supply .default")
```
hadley: Can you please break this up into two strings? (`warning()` pastes together it's inputs)

hadley:
Maybe also needs a new example?

lionel-:
Here you go

hadley:
Thanks!

PR #1847, opened 2016-05-22
put function to assign at the end of a chain.

jepusto:
`put()` is an alias for `base::assign()`, but can be used in the middle of a chain of manipulations. This lets you finish a chain with `%>% put(new_df)` so that the code reads more like a story. Longer explanation of the motivation is here: http://jepusto.github.io//assigning-after-dplyr


abresler:
Very cool creation, I hope this gets integrated!

hadley:
This would be more appropriate in magrittr, rather than dplyr, since it's not dplyr specific.

Issue #1569, opened 2015-12-03
Request: select by column type


cfhammill:
Being able to quickly separate list columns from standard columns of a data.frame would be handy. Possible to implement generically as a select by column class.

Issue #1229, opened 2015-06-19
Pass custom functions to select()


msjgriffiths:
Right now, you're limited in column selection when using select() and magrittr.

Consider the situation where you have a table, and you want to select certain columns at an intermediate point in a transformation pipeline, but don't want to (i) use fixed column indexes and (ii) don't want to save out an intermediary variable and identify column names separately.

It would make sense to be able to define a custom function in the same format as `everything()` or `one_of()`, and have `select()` and `gather()` and other functions use a custom function for selection variables.

In principle the function you take in a list of column names (e.g. `names(data)`) and then output a list of integer positions (e.g. `c(1, 2, 7)`. 

Example usage:

``` r

# Pass in column that expects the first parameter to be a list of column names.
data %>% gather(key, value, function(vars) {
    vars[-grep("Drop", vars, fixed = TRUE)]
})

# Define a function that expect the first parameter to be column names, second to be matches.
not_contains <- function(vars, match){
    vars[-grep(match, vars, fixed = TRUE)]
}
data %>% select(not_contains("avoid"))
```

Less ideally, allow custom functions to be registered as [part of `select_funs`](https://github.com/hadley/dplyr/blob/master/R/select-vars.R#L67), e.g. something like:

``` r
not_contains <- function(vars, match){
    vars[-grep(match, vars, fixed = TRUE)]
}

dplyr::register_select_funs(not_contains)

data %>% select(not_contains("avoid"))
```

hadley:
You can now use `select_if()` or implement your helper in the same way as `starts_with()` etc

Issue #497, opened 2014-07-14
Selecting all numeric columns


steromano:
What's the best way of doing it?
The following
```
iris %>% select(which(sapply(., is.numeric)))
```
works with the latest version of `magrittr`. However,
```
iris %>% mutate_each(funs(log), which(sapply(., is.numeric)))
```
throws `Error in lapply(X = X, FUN = FUN, ...) : object '.' not found`, even with the updated chain operator. Is this the expected behaviour?

In general, what is the best way to filter columns by some boolean condition in `dplyr`? Would it be worth having another selection function (in addition to `starts_with`, `ends_with`, etc.) acting directly on columns instead of column names? E.g. something along the lines of
```
iris %>% select(satisfies(is.numeric))
```

jennybc:
The [`purrr`](https://github.com/hadley/purrr) package now offers nice solutions to @steromano's original question, at least for data frames.

To keep all numeric columns:

``` r
iris %>% 
  purrr::keep(is.numeric) %>% 
  head(2)
#>   Sepal.Length Sepal.Width Petal.Length Petal.Width
#> 1          5.1         3.5          1.4         0.2
#> 2          4.9         3.0          1.4         0.2
```

To mutate all numeric columns with a single function:

``` r
iris %>%
  purrr::map_if(is.numeric, log) %>%
  head(2)
#>   Sepal.Length Sepal.Width Petal.Length Petal.Width Species
#> 1     1.629241    1.252763    0.3364722   -1.609438  setosa
#> 2     1.589235    1.098612    0.3364722   -1.609438  setosa
```

Fredo-XVII:
Nice! select_if() worked perfectly. Thank you.

PR #1853, opened 2016-05-24
Colwise update

lionel-:
Here you go @hadley

I'm not sure how to handle databases with the `_if()` variants. It seems only a very restricted set of predicates will be compatible and in any case they will generate a full table scan. This doesn't seem very useful, so these variants fail with an error message for now when applied to a lazy source.

Also I think the easiest way to solve the purrr problem is to restore the DataframeSubsetVisitors class as header only, and keep the `dmap()` functions as is (with a deprecation message pointing to dplyr). The semantics are slightly different so there is no easy way to provide a 100% compatible translation to the new dplyr colwise functions.


NEWS.md
```diff
@@ -44,6 +44,9 @@
 * Code related to starting and signalling clusters has been moved out to
   [multidplyr](http://github.com/hadley/multidplyr).
 
+* `summarise_each()` and `mutate_each()` are deprecated in favour of a
```
hadley: Can we not deprecate for now? (And instead deprecate in the next version)

R/colwise.R
```diff
@@ -3,104 +3,175 @@
 #' Apply one or more functions to one or more columns. Grouping variables
 #' are always excluded from modification.
 #'
```
hadley: I think this should say something like:

`summarise_all()` and `mutate_all()` apply the functions to all (non-grouping) columns. `summarise_at()` and `mutate_at()` allow you to select columns in the same way as `select()`. `summarise_each()` and `mutate_each()` are older variants that will be deprecated in the future

R/colwise.R
```diff
@@ -3,104 +3,175 @@
...
-#'   You can use same specifications as in \code{\link{select}}. If missing,
-#'   defaults to all non-grouping variables.
-#'
-#'   For standard evaluation versions (ending in \code{_}) these can
-#'   be either a list of expressions or a character vector.
+#' @param .tbl a tbl
+#' @param .funs List of function calls generated by
+#'   \code{\link{funs}()}, or a character vector of function names, or
+#'   simply a function (only for local sources).
+#' @param .cols A list of columns generated by \code{\link{columns}()},
```
hadley: What if this was `vars()` instead of `columns()`? `funs()` and `vars()` are more symmetric.

R/colwise.R
```diff
@@ -3,104 +3,175 @@
...
-#' by_species %>% summarise_each(funs(length))
-#' by_species %>% summarise_each(funs(mean))
-#' by_species %>% summarise_each(funs(mean), Petal.Width)
-#' by_species %>% summarise_each(funs(mean), matches("Width"))
+#' by_species %>% summarise_all(length)
+#' by_species %>% summarise_all(mean)
+#'
+#' # Use the _at and _if variants for conditional mapping.
+#' # summarise_at() can use select() helpers with the columns() function:
+#' by_species %>% summarise_at(columns(Petal.Width), mean)
```
hadley: Might be worth example with `na.rm = TRUE` now

hadley: Oh don't worry, I see you did that below

R/colwise.R
```diff
@@ -3,104 +3,175 @@
...
+#' by_species %>% summarise_at(columns(matches("Width")), mean)
+#' by_species %>% summarise_if(is.numeric, mean)
+#'
+#' # You can also specify columns with column names or column positions:
+#' by_species %>% summarise_at(c("Sepal.Width", "Petal.Width"), mean)
+#' by_species %>% summarise_at(c(1, 3), mean)
+#'
+#' # Or with logical vectors:
+#' by_species %>% summarise_if(c(TRUE, TRUE, FALSE, TRUE, FALSE), mean)
```
hadley: I'd rather not mention this - it seems like it will lead to confusing data analysis code

R/colwise.R
hadley: And mention the `_if` variants too

R/funs.R
```diff
@@ -93,3 +126,18 @@ make_name <- function(x) {
...
+
+merge_args <- function(call, args) {
+  if (!length(args)) {
+    return(call)
+  }
+  if (is.null(names(args))) {
+    stop("Additional arguments should be named", call. = FALSE)
+  }
+
+  Map(function(arg, value) {
```
hadley: It doesn't really matter, but I think a for loop would be better here

R/colwise.R
lionel-: hmm vars makes me think of programming variables rather than data frame columns, but that may just be me.

R/colwise.R
```diff
@@ -3,104 +3,175 @@
...
 #'
-#' # Alternative variable specification
-#' summarise_each_(iris, funs(max), names(iris)[-5])
-#' summarise_each_(iris, funs(max), list(quote(-Species)))
 #' @aliases summarise_each_q mutate_each_q
 #' @export
-summarise_each <- function(tbl, funs, ...) {
-  summarise_each_(tbl, funs, lazyeval::lazy_dots(...))
+summarise_all <- function(.tbl, .funs, ...) {
+  funs <- as.fun_list(.funs, .env = parent.frame(), ...)
```
lionel-: I'm having doubts with this `parent.frame()` thing now, it doesn't make a lot of sense. I think I better use `lazyeval::lazy(.funs)$env`, would this be reasonable @hadley?

hadley: It's tricky because potentially each function could have a different parent environment. But I think `parent.frame()` will work 90% of the time and then we can fix when we switch to the new lazyeval

lionel-:
All done

NEWS.md
```diff
@@ -13,6 +13,7 @@
   column names (#1513). 
   
 * `one_of()` tolerates unknown variables in `vars`, but warns (#1848, @jennybc).
+  column names (#1513).
```
hadley: Oops

hadley:
@lionel- looks good (apart from the one small merge problem)

@jennybc I'd love you thoughts before I pull the trigger on this

jennybc:
Yes it would be a pleasure to test drive. I had already noticed the lack of American versions but thankfully @lionel- has fixed that :)

jennybc:
These new functions will be very useful! Everything works well. I made a  PR to @lionel- with some [modest changes to the help](https://github.com/lionel-/dplyr/pull/1). I put some other notes made while running things in a [gist](https://gist.github.com/jennybc/4ed67a1c1af22f92191daab571c7cc62).

This is definitely related to https://github.com/hadley/purrr/issues/152 and presumably closes several of the `dplyr` issues linked there. Now that I reread those, I'm still not sure if there's a `dplyr` solution to select based on a predicate, i.e. `select_if()`? Is `purrr::keep()` the best bet https://github.com/hadley/dplyr/issues/497?

lionel-:
A `select_if()` variant would be consistent with the new colwise verbs (with all `_if` variants only compatible with local sources). Then should we also have a `select_at()` that would accept numeric/character vectors and `vars()` specifications? Could be useful for programming but will probably be confusing and redundant wrt both `select()` and `select_()`.

jennybc:
`select_if()` seems to be part or all of what people are asking for in #1569, #1501, #1229, #497 = all the issues I alluded to above. So that would be handy.

The fact that `select_at()` has such potential to create confusion with `select()` itself gets at something I noted in the gist: Why do we have to use the usual `select_helpers` inside `vars()` with the new `_at` functions but not with `select()`?

hadley:
Because there's three things you might want to supply to `summarise_at()`: the variables, the functions, and fixed arguments to the functions. There's only only thing to supply to `select()`: the variables.

hadley:
Thanks!

PR #1854, opened 2016-05-25
WIP: Change file name for in-memory SQLite database

krlmlr:
Use `"file::memory:"` instead of `":memory:"`, the latter doesn't seem to work well on Windows for unknown reasons.

This needs to wait for RSQLite > 1.0.0 to hit CRAN, because the CRAN version simply creates a file named `"file::memory:"`.


hadley:
I think it would be better to fix this upstream in RSQLite. (Otherwise we're going to have to add a lot of `packageVersion()` checks)

PR #1855, opened 2016-05-25
revert 04704bf to enable installing purrr again

### LINKED ISSUE #1814: Header-only API for DataFrameSubsetVisitors
Would you consider keeping the `DataFrameSubsetVisitors` class as header-only? This would make it easier to reuse in external packages.

Purrr has stopped loading a while ago (issue hadley/purrr#196) because dplyr is moving definitions inside unshipped cpp files, cf https://github.com/hadley/dplyr/commit/04704bff3fd2969607fd648e362b289483041a3d.

yutannihilation:
(Disclaimer: I'm not familiar with C++)

This temporarily fixes #1814.

As I wrote in #1814, this problem is about template class. Since the compiler compiles templates only with the types that are actually used, if the type purrr uses is not used in dplyr, it is not compiled and occurs `undefined reference to ...` error. In this case, exporting the class whole in header file is right way in my understanding.

So, could you consider reverting 04704bf for now? I know this may not be an ultimate solution, but not a few people must suffer from failing to install purrr with dev version of dplyr.

### DIFF
--- a/inst/include/dplyr/DataFrameSubsetVisitors.h
+++ b/inst/include/dplyr/DataFrameSubsetVisitors.h
@@ -14,9 +14,43 @@ namespace dplyr {
         public:
             typedef SubsetVectorVisitor visitor_type ;
 
-            DataFrameSubsetVisitors( const Rcpp::DataFrame& data_)  ;
+            DataFrameSubsetVisitors( const Rcpp::DataFrame& data_) :
+                data(data_),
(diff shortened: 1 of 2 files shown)


hadley:
Yeah, I think this is the right fix for now, and I was going to do it anyway. Thanks!

yutannihilation:
Thanks! :+1:

PR #1813, opened 2016-05-10
WIP: Travis testing

krlmlr:
Need to manually install dev version of devtools, then the dplyr package (with essential deps), then dtplyr, then the suggests. I hope devtools can handle this one day.

Also, enables testing of r-devel and r-oldrel. The build fails on r-release because of #1806.

Currently, devtools is installed from a pull request to achieve the functionality; I'll update it once hadley/devtools#1183 is merged.

Closes #1812 (=includes it).

PR #1859, opened 2016-05-26
Improve Travis checks

krlmlr:
- check three R versions in a matrix
- build dependencies in parallel

Trimmed-down version of #1813.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,10 +1,18 @@
+# R for travis: see documentation at https://docs.travis-ci.com/user/languages/r
+
 language: R
 sudo: false
 cache: packages
+r:
+  - oldrel
+  - release
+  - devel
 
 warnings_are_errors: false
 env:
-- _R_CHECK_FORCE_SUGGESTS_=false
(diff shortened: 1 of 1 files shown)


hadley:
Thanks! 

Is there a reason that we can't make ` MAKEFLAGS="-j 2"` the default for R-travis? @jimhester

jimhester:
Kirill and I discussed this in https://github.com/krlmlr/dplyr/commit/63bc1c115b013d7d5e9cb6c0e660b441f628cb59#commitcomment-17423727

Basically parallel make can break a package if the dependencies are not written properly in the makefile, so I felt it would be unwise to turn on by default.

But maybe the short term pain is worth the long term gain, I could be persuaded...

krlmlr:
I have MAKEFLAGS = -j4 as a standard setting on my system, so far I haven't noticed any misbehaving packages except for git2r, which is now fixed on CRAN, too. I agree that opt-in is safer here, it really only makes a difference if the dependencies are very heavy.

PR #1861, opened 2016-05-26
fix to escape partition fields for window functions

javierluraschi:
Follow up to: https://github.com/hadley/dplyr/pull/1816


tests/testthat/test-sql-quote.R
```diff
@@ -0,0 +1,48 @@
...
+  out <- select(testTable, field1)
+  expect_match(sql_render(out), "^SELECT `field1` AS `field1`\nFROM `table1`$")
+})
+
+test_that("Window function quotes work", {
+  dbTest <- src_sql("test", con = dbiTest)
+  testTable <- tbl_sql("test", src = dbTest, from = "table1")
+
+  out <- filter(group_by(testTable, field1), min_rank(desc(field1)) < 2)
+  sqlText <- sql_render(out)
```
hadley: Could you use `sql_build()` here? I think it would make the output easier to test

tests/testthat/test-sql-quote.R
```diff
@@ -0,0 +1,48 @@
+context("SQL translation")
+
+dbiTest <- list()
+attr(dbiTest, "class") <- "DBITestConnection"
+
+assign("db_query_fields.DBITestConnection", function(con, x) {
```
hadley: Do you really need those assign statements?

tests/testthat/test-sql-quote.R
```diff
@@ -0,0 +1,48 @@
+context("SQL translation")
+
+dbiTest <- list()
```
hadley: You can do this in one line with `dbiTest <- structure(list(), class = "DBITestConnection"))`

tests/testthat/test-sql-quote.R
hadley: Actually, you might - but I think it would be little better to pull out into a separate file (in `R/`) that can be re-used elsewhere if needed.

javierluraschi: I tried setting the assignments on the function (as in, declaring an S3 object within a function, but the dispatch didn't work) and I also tried a normal assignment in the test file outside the methods, but the test would still not able to find the references. We could export this object as `DBIMockConnection` from the package to help other users with a baseline for unit tests, just a thought.

tests/testthat/test-sql-quote.R
```diff
@@ -0,0 +1,48 @@
...
+
+assign("sql_translate_env.DBITestConnection", function(con) {
+  dplyr::sql_variant(
+    scalar = dplyr::sql_translator(.parent = dplyr::base_scalar),
+    aggregate = dplyr::sql_translator(.parent = dplyr::base_agg),
+    window = dplyr::sql_translator(.parent = dplyr::base_win)
+  )
+}, env = globalenv())
+
+test_that("Simple quotes work", {
```
hadley: I think these could probably go in `test-sql-translations.R` and `test-window.R`.

And better names for the tests would be: "connection affects quoting character"

tests/testthat/test-sql-quote.R
hadley: If you put it in a regular file, you'll need to `@export` the methods, but not the object itself

R/translate-sql.r
```diff
@@ -188,7 +191,10 @@ sql_env <- function(expr, variant, con, window = FALSE,
 
   # Known sql expressions
   symbol_env <- copy_env(base_symbols, parent = name_env)
-  symbol_env
+
+  # Evaluation context.
+  eval_env <- copy_env(list(".dplyr.con" = "con"), parent = symbol_env)
```
hadley: How does that actually work?

javierluraschi: That was attempt #1! Needs to be removed! Thanks for the catch. I was trying to set a connection on the environment for evaluation but realized that is not the actual environment in which the function runs, which made me think that I need a global way to set this context which got me to the partition_* methods I'm reusing.

tests/testthat/test-sql-quote.R
javierluraschi: @hadley tried... However, not everything gets escaped during `sql_build`. For instance, `sql_render` calls `sql_render.select_query`, which calls `sql_select.default`, which at some point calls `escape` with the `con` set as `DBITestConnection`. I could scope this check to use `sql_build` and only verify the partition fields to be escaped; however, my prefference would be to check all of them even if it involves comparing strings... but up to you.

tests/testthat/test-sql-quote.R
javierluraschi: changed.

tests/testthat/test-sql-quote.R
javierluraschi: changed.

tests/testthat/test-sql-quote.R
javierluraschi: changed.

javierluraschi:
@hadley updated.

javierluraschi:
squashed commits

hadley:
Thanks!

PR #1864, opened 2016-05-27
export additional utility functions

javierluraschi:
This PR exposes a couple additional methods required to subclass `sample_n` and `mutate`, among other methods, for packages subclassing dplyr.


R/utils.r
```diff
@@ -1,3 +1,7 @@
+#' Utility functions
+#' @name utils
```
hadley: I wouldn't worry about making a separate utils doc file, instead just `@keywords internal` the `named_commas` doc so they don't appear in the overall index

javierluraschi: Sounds good, updated.

hadley:
Thanks!

PR #1873, opened 2016-05-31
Remove dead code

krlmlr:
from POSIXctCollecter class, introduced in 2743b605a496b734e8

@romainfrancois: Looks to me like this was intended, but please double-check.


hadley:
This looks right to me - thanks!

PR #1875, opened 2016-06-01
bind_rows() rejects POSIXlt columns

### LINKED ISSUE #1789: POSIXlt-related segfault
Last night I was working with some dates/times and I (incorrectly) thought I needed them in POSIXlt format because of some time zone issues.  Along the way, I stumbled on a non-deterministic segfault and managed to create a reprex (below).

The short version is that I can insert a POSIXlt into an existing tbl_df, and that some functions (like `bind_rows`) will try to work with the invalid result without checking its column classes. The segfault's probability of occurring seems to increase with the size of the tibble and also seems more likely to happen when the corrupted tibble is printed than when it's merely computed or stored.

I think that there are three mutually-compatible ways to defend against this:

1) tibble could be more aggressive about checking for the creation of invalid columns (e.g. with checks inside assignment functions, where possible).  I don't know how much can be done about `$<-` though, since it's primitive.

2) increase the number of dplyr functions that check `white_list.h` for invalid column classes

3) It might be possible to do some kind of bounds check to determine when dplyr will try to access unmapped memory and stop it at that level, but I don't know enough C++ to know.

reprex:

```
library(dplyr)
library(lubridate)

# Unfortunately, this was the smallest tibble I was able to make
# that still caused the segfaults.
x <-
structure(list(time_zone = c("America/New_York", "America/Chicago",
"America/New_York", "America/Chicago", "America/Chicago", "America/Denver",
"America/Chicago", "America/Chicago", "America/Chicago", "America/Chicago",
"America/New_York", "America/Chicago", "America/Chicago", "America/Denver",
"America/New_York", "America/New_York", "America/New_York", "America/Chicago",
"America/Chicago", "America/Chicago"),
date_time = structure(c(834556020,
834123360, 897456720, 1055307420, 456210600, 739604520, 1244696520,
993616680, 1307767800, 1087879560, 1307855400, 1214714040, 393483000,
802675920, 644735580, 898232520, 1341548760, 708498480, 456212280,
771396960), class = c("POSIXct", "POSIXt"), tzone = "UTC")), class = c("tbl_df",
"tbl", "data.frame"), .Names = c("time_zone", "date_time"), row.names = c(NA,
-20L))

# no check for invalid classes in `$<-`
x$date_time = as.POSIXlt(x$date_time)

# Printing x throws the following error:
# Error in `[<-.data.frame`(`*tmp*`, is_list, value = list(date_time = c("<dbl [20]>",  : 
#    replacement element 1 has 9 rows, need 20

for(i in 1:100){
  # binding rows and then printing usually triggers a segfault within the first few iterations
  print(bind_rows(x, x))
}
```

Here's the segfault:

```
Source: local data frame [40 x 2]

 *** caught segfault ***
address 0x0, cause 'unknown'

Traceback:
 1: lapply(x, `[`, i)
 2: `[.tbl_df`(x, seq_len(n), , drop = FALSE)
 3: x[seq_len(n), , drop = FALSE]
 4: head.data.frame(x, n)
 5: head(x, n)
 6: as.data.frame(head(x, n))
 7: trunc_mat(x, n = n, width = width)
 8: print(trunc_mat(x, n = n, width = width))
 9: print.tbl_df(bind_rows(x, x))
10: print(bind_rows(x, x))

Possible actions:
1: abort (with core dump, if enabled)
2: normal R exit
3: exit R without saving workspace
4: exit R saving workspace
```

And here's my session info:

```
R version 3.2.4 (2016-03-10)
Platform: x86_64-apple-darwin13.4.0 (64-bit)
Running under: OS X 10.11.4 (El Capitan)

locale:
[1] en_US.UTF-8/en_US.UTF-8/en_US.UTF-8/C/en_US.UTF-8/en_US.UTF-8

attached base packages:

other attached packages:
[1] lubridate_1.5.6.9000 dplyr_0.4.3.9001    

[1] magrittr_1.5   R6_2.1.2       assertthat_0.1 DBI_0.3.1.9008 tools_3.2.4    tibble_1.0-1  
[7] Rcpp_0.12.4.5  stringi_1.0-1  stringr_1.0.0 
```

See also #1080 from last year

**Edited to add:** I just noticed that I happened to copy-paste a less common segfault message above.  More often, I get one like this:

```
*** caught segfault ***
address 0x7fa200000001, cause 'memory not mapped'
```

krlmlr:
instead of segfaulting.

Fixes #1789.

Also fixes another instance with the same pattern, but I'm not sure if this is strictly necessary.

NEWS line:

```
* `bind_rows()` cleanly aborts if one of the data frames contains a `POSIXlt` column
  (#1789, #1875, @krlmlr).
```

### DIFF
--- a/inst/include/dplyr/Collecter.h
+++ b/inst/include/dplyr/Collecter.h
@@ -379,7 +379,11 @@ namespace dplyr {
             return new Collecter_Impl<CPLXSXP>(n) ;
         case LGLSXP: return new Collecter_Impl<LGLSXP>(n) ;
         case STRSXP: return new Collecter_Impl<STRSXP>(n) ;
-        case VECSXP: return new Collecter_Impl<VECSXP>(n) ;
+        case VECSXP:
+            if( Rf_inherits( model, "POSIXlt" )) {
+                stop( "POSIXlt not supported" ) ;
+            }
+            return new Collecter_Impl<VECSXP>(n) ;
         default: stop("Unsupported vector type %s", Rf_type2char(TYPEOF(model))) ;
         }
         return 0 ;
--- a/inst/include/dplyr/Result/GroupedSubset.h
+++ b/inst/include/dplyr/Result/GroupedSubset.h
@@ -64,7 +64,11 @@ namespace dplyr {
             case LGLSXP: return new GroupedSubsetTemplate<LGLSXP>(x, max_size) ;
             case STRSXP: return new GroupedSubsetTemplate<STRSXP>(x, max_size) ;
             case VECSXP:
-                if( Rf_inherits(x, "data.frame") ) return new DataFrameGroupedSubset(x) ;
+                if( Rf_inherits( x, "data.frame" ) )
+                    return new DataFrameGroupedSubset(x) ;
+                if( Rf_inherits( x, "POSIXlt" ) ) {
+                    stop( "POSIXlt not supported" ) ;
+                }
                 return new GroupedSubsetTemplate<VECSXP>(x, max_size) ;
             case CPLXSXP: return new GroupedSubsetTemplate<CPLXSXP>(x, max_size) ;
             default: stop("Unsupported vector type %s", Rf_type2char(TYPEOF(x)));
--- a/tests/testthat/test-binds.R
+++ b/tests/testthat/test-binds.R
@@ -386,3 +386,9 @@ test_that("bind_cols infers classes from first result (#1692)", {
   expect_equal( class(bind_rows(d5,d1)), c("tbl_df", "tbl", "data.frame") )
 
 })
+
+test_that("bind_rows rejects POSIXlt columns (#1789)", {
+  df <- data_frame(x = Sys.time() + 1:12)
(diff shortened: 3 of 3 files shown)


hadley:
Does the logic in `order_visitor()` look right to you? 

```cpp
                    case VECSXP:
                    {
                        if( Rf_inherits( vec, "data.frame" ) ){
                            return new OrderVisitorDataFrame<true>( vec ) ;
                        }
                        break ;
                    }
```

I think that will fall through to the `return 0`.  I think the break should be removed, since we don't have a general list orderer (and POSIXlt doesn't need to be a special case)

krlmlr:
No, the logic looks odd. I'll submit a pull.

Fall-through in `switch()` blocks (removing the `break`) is an anti-pattern and may even induce a compiler warning.

hadley:
@krlmlr good to know - better to be explicit anyway

krlmlr:
In clang: `-Wimplicit-fallthrough` (http://stackoverflow.com/a/27965827/946850), work in progress for gcc.

hadley:
Thanks!

PR #1876, opened 2016-06-01
Valid R syntax for message printed when performing a natural join

krlmlr:
Now:

```
Joining, by = "a"
Joining, by = c("b", "c")
```

The user can copy the text starting at the comma until the end and just paste it into the join call.

Original issue: https://github.com/hadley/dplyr/issues/1783#issuecomment-221874823


hadley:
Thanks!

PR #1878, opened 2016-06-01
Call stop() outside of switch() statement

krlmlr:
so that break works inside switch() without surprises

https://github.com/hadley/dplyr/pull/1875#issuecomment-223000740

Working on further cleanup based on this commit. Also looking for a way to generalize the omnipresent switch(): http://stackoverflow.com/q/37576790/946850.

### DIFF
--- a/inst/include/dplyr/subset_visitor.h
+++ b/inst/include/dplyr/subset_visitor.h
@@ -13,8 +13,10 @@ namespace dplyr {
             case LGLSXP:  return new MatrixColumnSubsetVisitor<LGLSXP>( vec ) ;
             case STRSXP:  return new MatrixColumnSubsetVisitor<STRSXP>( vec ) ;
             case VECSXP:  return new MatrixColumnSubsetVisitor<VECSXP>( vec ) ;
-            default: stop("Unsupported vector type %s", Rf_type2char(TYPEOF(vec)));
+            default: break ;
             }
+
+            stop("Unsupported matrix type %s", Rf_type2char(TYPEOF(vec))) ;
         }
 
         if( Rf_inherits(vec, "Date") ){
@@ -41,10 +43,11 @@ namespace dplyr {
                     }
                     return new SubsetVectorVisitorImpl<VECSXP>( vec ) ;
             }
-            default: stop("Unsupported vector type %s", Rf_type2char(TYPEOF(vec)));
+            default: break ;
         }
 
-        // should not happen
+        // should not happen, safeguard against segfaults anyway
+        stop("Unsupported vector type %s", Rf_type2char(TYPEOF(vec))) ;
         return 0 ;
     }
 
--- a/inst/include/dplyr/subset_visitor.h
+++ b/inst/include/dplyr/subset_visitor.h
@@ -3,53 +3,65 @@
 
 namespace dplyr {
 
-    inline SubsetVectorVisitor* subset_visitor( SEXP vec ){
+inline SubsetVectorVisitor* subset_visitor_matrix( SEXP vec );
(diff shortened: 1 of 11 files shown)


hadley:
Can you explain why? I think it makes more sense to clearly indicate that the default is to throw an error.

krlmlr:
This way, the `stop()` is reached whenever the `switch()` doesn't handle the condition -- opt-in in the sense that we need to explicitly use a `return` statement to indicate a valid result. If we don't do this, and later add a `break` statement in one of the cases, the code reaches an unprotected `return 0` and possibly segfaults.

hadley:
In that case maybe we should also switch from `return 0` to throwing some sort of exception?

krlmlr:
stop() is already [implemented by throwing an exception]( https://github.com/RcppCore/Rcpp/blob/82d8a3a922fc158f2403b233b2ac6861c538d514/inst/include/Rcpp/exceptions.h#L250I). I wonder why the compiler doesn't pick it up. For instance, if I remove a "return 0":

```
../inst/include/dplyr/subset_visitor.h:51:5: warning: control reaches end of non-void function [-Wreturn-type]
```

hadley:
What if we had a wrapper around `TYPEOF()` that returned an enum so at least we'd get compiler warnings about missing cases?

hadley:
Let's hold off on this until we can at least eliminate the `return 0`s

src/nth.cpp
```diff
@@ -172,9 +174,12 @@ Result* nth_prototype( SEXP call, const LazySubsets& subsets, int nargs){
                     case INTSXP:  return nth_with<INTSXP>( data, idx, order_by ) ;
                     case REALSXP: return nth_with<REALSXP>( data, idx, order_by ) ;
                     case STRSXP:  return nth_with<STRSXP>( data, idx, order_by ) ;
-                    default: stop("Unsupported vector type %s", Rf_type2char(TYPEOF(data)));
+                    default: break ;
                 }
             }
+            else {
+                return 0;
```
hadley: Why are we returning 0 here?

krlmlr: Because the default action in this function is now to call stop() in line 214, which seems to be wrong (according to tests) if the condition in line 169 is false. This PR is only an equivalent transformation of the original code.

hadley: Ah got it

krlmlr:
I've added another commit that extracts functions to simplify some of the more complicated switches. Could you please review and merge if ok? I'll build a PR that eliminates the `return 0` for Rcpp > 0.12.5.2 on top of that.

Also, I've looked into the enum idea: d0b0cc70a3, f40863e300b9b95, not part of this PR yet. Missing cases are detected by GCC, `default` (or code that handles the impossible other cases) is still necessary. Should we pursue this further? Please review.

inst/include/dplyr/subset_visitor.h
```diff
@@ -3,50 +3,65 @@
...
-        return 0 ;
+inline SubsetVectorVisitor* subset_visitor_matrix( SEXP vec );
+inline SubsetVectorVisitor* subset_visitor_data_frame( SEXP vec );
+
+inline SubsetVectorVisitor* subset_visitor( SEXP vec ){
+  if( Rf_isMatrix( vec ) ){
+    return subset_visitor_matrix(vec) ;
+  }
+  else {
+    return subset_visitor_data_frame(vec) ;
```
hadley: I think this should be `_vector` not `_data_frame`

hadley:
Apart from that small naming problem, LGTM.

hadley:
Thanks!

PR #1880, opened 2016-06-02
Add select_if() to select based on predicate

### LINKED ISSUE #497: Selecting all numeric columns

### LINKED ISSUE #1569: Request: select by column type

lionel-:
This function would be analogous to `summarise_if()` and `mutate_if()`.

One way to distinguish responsibilities between `select()` and `select_if()` is that the former operates on metadata such as the column names, while the latter operates on the columns themselves, and thus requires local sources.

So the custom functions suggested in #1229 would belong to `select()` because they operate on the column names. And more general predicates based on column types (also metadata) compatible with DB sources could be added to `select()`.

Closes #497, closes #1569


hadley:
I think it maybe should mimic `mutate_if()` and `summarise_if()` and always return grouped values. (It'll do that now but you'll get a message because of the default `select()` behaviour)

lionel-:
Here you go

hadley:
Thanks!

PR #1889, opened 2016-06-04
New as.table.tbl_cube()

### LINKED ISSUE #1775: Convert tbl_cube to table
I'm using tbl_cube as intermediate step in the conversion of a three-dimensional tidy data frame to an ftable. (The value is of type character, xtabs() and table() don't seem to work.) I'm missing an obvious way to convert a measure of a cube to a table.

I know I can access the measures via `cube$mets`, but I still need to set the dimnames. The code below works for me, how about adding it to dplyr:

```
as_table <- function(x, measure = 1L) {
  ret <- x$mets[[measure]]
  dimnames(ret) <- x$dims
  ret
}
```

krlmlr:
With documentation and test. Fixes #1775.


krlmlr:
Strange: Local coverage testing shows that the new code is covered.

hadley:
Thanks! 

The code-cov pull request coverage reports seem to be rather unreliable

PR #1916, opened 2016-06-13
Make db_create_indexes generic; export

### LINKED ISSUE #1911: Export db_create_indexes
Backends that require implementations of copy_to need to be able to access db_create_indexes. As this is not currently exported, backends will need to take a copy of this into their own code base. This seems unnecessary as there is nothing backend specific about this (simple for loop wrapper of db_create_index)

imanuelcostigan:
Closes #1911 and **this PR builds on PR #1915** (updating DB backend function docs).

Copied text from #1911 (for convenience): 

> Backends that require implementations of copy_to need to be able to access db_create_indexes. As this is not currently exported, backends will need to take a copy of this into their own code base. This seems unnecessary as there is nothing backend specific about this (simple for loop wrapper of db_create_index)

### DIFF
--- a/R/dbi-s3.r
+++ b/R/dbi-s3.r
@@ -30,20 +30,45 @@ sql_translate_env.NULL <- function(con) {
 
 #' Database generics.
 #'
-#' These generics execute actions on the database. All generics have a method
+#' These generics execute actions on the database. Most generics have a method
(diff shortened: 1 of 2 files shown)


hadley:
Thanks!

PR #1938, opened 2016-06-18
Test data tables as part of the generic tests

krlmlr:
Had to disable two tests for data.table-s, and to install the .datatable.aware symbol in the namespace.


krlmlr:
Not sure about this -- perhaps we should call test_register_src() from dtplyr.

PR #1942, opened 2016-06-18
Fix distinct() for SQL sources

### LINKED ISSUE #1937: distinct() problems for SQL sources
1. Doesn't seem to respect columns for SQL sources:

    ```r
    > memdb_frame(a=1, b=2) %>% distinct(a, .keep_all = TRUE) %>% sql_render
    <SQL> SELECT DISTINCT *
    FROM `iksbzxasdl`
    ```
    Expected: `SELECT DISTINCT(a)`
2. Error with `.keep_all = FALSE` for SQL sources seems unnecessary, could be implemented with a second `SELECT`:

    ```r
    > memdb_frame(a=1, b=2) %>% distinct(a) %>% sql_render
    <SQL> SELECT a FROM (SELECT DISTINCT(a)
    FROM `iksbzxasdl`)
    ```

krlmlr:
Without columns:

- `.keep_all` is ignored if no columns are selected

With columns:

- `.keep_all = FALSE` is implemented using a grouped select
- `.keep_all = TRUE` still raises an error, because DISTINCT syntax seems to vary between SQL dialects

Adapted tests.

Fixes #1937.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -8,7 +8,9 @@
 
 * `distinct()` now only keeps the distinct variables. If you want to return
   all variables (using the first row for non-distinct values) use
-  `.keep_all = TRUE` (#1110). (The default behaviour of using all variables
+  `.keep_all = TRUE` (#1110). For SQL sources, `.keep_all = FALSE` is
+  implemented using `GROUP BY`, and `.keep_all = TRUE` raises an error
+  (#1937, #1942, @krlmlr). (The default behaviour of using all variables
   when none are specified remains - this note only applies if you select
   some variables).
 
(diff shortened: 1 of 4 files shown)


hadley:
Would you mind adding a test for the SQL generation too?

krlmlr:
What do you have in mind -- something like the tests in test-sql-render.R?

hadley:
Yeah exactly - when reviewing this PR it would help if I had a concrete example of the SQL that is generated.

krlmlr:
Done. On second thought, `SELECT DISTINCT y` might be a simpler query for the second case; not sure how portable it is though.

hadley:
Thanks - that seems like a reasonable approach to me.

PR #1944, opened 2016-06-20
Hand-format limit when rendering SQL

krlmlr:
to support limits > 2³¹ and to ensure that no decimal point is added to the SQL.

Follow-up to #1929.

See https://github.com/hannesmuehleisen/MonetDBLite/issues/56#issuecomment-227065094.


hadley:
Thanks!

PR #1952, opened 2016-06-21
Update RStudio project settings to install tests

krlmlr:
for https://github.com/hadley/dtplyr/pull/24.

### DIFF
--- a/dplyr.Rproj
+++ b/dplyr.Rproj
@@ -17,6 +17,6 @@ StripTrailingWhitespace: Yes
 
 BuildType: Package
 PackageUseDevtools: Yes
-PackageInstallArgs: --no-multiarch --with-keep.source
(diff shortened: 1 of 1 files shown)


krlmlr:
I haven't found it. Does your RStudio append --install-tests when you press Cmd + Shift + B?

hadley:
An R option, not an RStudio option

krlmlr:
There's the `devtools.install.args` option, but RStudio Preview 0.99.1230 doesn't use it for build + install. It's used for `devtools::install()`, but that function installs tests by default. R itself seems to look exclusively for the `--install-tests` switch: https://github.com/wch/r-source/blob/5cfd0944cb9a7554c597342c8274e556bbd5df91/src/library/tools/R/install.R#L1375.

Could RStudio just call `devtools::install()` for build + install?

hadley:
LGTM

PR #1959, opened 2016-06-23
Support zero-column corner case in vector visitors

### LINKED ISSUE #1957: crash in empty `n_distinct()`
The following crashes RStudio (after a minute of 100% CPU):

```R
library(dplyr)
data.frame(a = c("a","b","a"), b=1:3) %>% group_by(a) %>%  summarise(user_count = n_distinct())
```

It also happens with a simple `n_distinct()`.

Win7, R3.3, compiled from master (e87f6dd). Not sure if this is related: https://github.com/hadley/dplyr/issues/1871

krlmlr:
- hash(): return index
- nrows(): stop() instead of segfault

Fixes #1957.

With tests, which pass locally. Travis fails because of tibble update (#1949).

Currently I have no idea how to make the forward of n_distinct() to n() explicit in the code for hybrid evaluation.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -346,8 +346,9 @@ There were two other tweaks to the exported API, but these are less likely to af
 
 * Hybrid `min()` and `max()` handle empty sets (#1481).
 
-* `n_distinct()` uses multiple arguments (#1084), falls back to R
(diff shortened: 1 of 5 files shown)


hadley:
I think I would prefer to make this an error, rather than switching to `n()` behaviour

hadley:
Thanks!

PR #1975, opened 2016-06-28
Fix copy_to behavior for MySQL

### LINKED ISSUE #2256: error when writing character fields containing NAs to mysql table
I get the following error when trying to write a table to a mysql connection with copy_to():

```
> sanbs_db <- src_mysql("sanbs", host = "localhost", user = "root")
> results_diagnostic.db <- copy_to(sanbs_db, results_diagnostic, temporary = FALSE, indexes = list(c("result_id","donation_id")))
Error in if (n <= 65535) { : missing value where TRUE/FALSE needed
```

I believe the error occurs because I have true NAs in character fields, and the following check in src-mysql.r produces n=NA instead of the expected maximum length:

```
db_data_type.MySQLConnection <- function(con, fields, ...) {
  char_type <- function(x) {
    n <- max(nchar(as.character(x), "bytes"))
    if (n <= 65535) {
      paste0("varchar(", n, ")")
    } else {
      "mediumtext"
    }
}
```

the problem can be demonstrated with:

```
> stringvec <- c("a",NA,"bb")
> max(nchar(stringvec))
[1] NA
> max(na.omit(nchar(stringvec)))
[1] 2
```

I'm not sure what this means for writing NULL values into the db table.

demorenoc:
copy_to() gives an error if a character column has NAs.

Fixes #2256.

### DIFF
--- a/R/src-mysql.r
+++ b/R/src-mysql.r
@@ -145,7 +145,7 @@ db_has_table.MySQLConnection <- function(con, table, ...) {
 #' @export
 db_data_type.MySQLConnection <- function(con, fields, ...) {
   char_type <- function(x) {
-    n <- max(nchar(as.character(x), "bytes"))
+    n <- max(nchar(as.character(x), "bytes"), 0L, na.rm = TRUE)
     if (n <= 65535) {
       paste0("varchar(", n, ")")
     } else {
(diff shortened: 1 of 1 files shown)


krlmlr:
Could you please add a test? I'm not sure if MySQL testing is enabled in the current code base, but we still could execute that test manually.

krlmlr:
@hadley: PTAL. @eduardgrebe has tests for this particular issue in #2263, which we can use here.

hadley:
This also needs to go upstream, but it's fine in dplyr for now.

krlmlr:
Thanks!

PR #1992, opened 2016-07-04
Fixed the problem of function db_has_table returning NA

sailingwave:
The generic function db_has_table is missing the second parameter
needed for dispatching the correct method (db_has_table.DBIConnection).

### DIFF
--- a/R/dbi-s3.r
+++ b/R/dbi-s3.r
@@ -83,7 +83,7 @@ db_list_tables.DBIConnection <- function(con) dbListTables(con)
 #' @name backend_db
 #' @export
 #' @param table A string, the table name.
-db_has_table <- function(con, table) UseMethod("db_has_table")
+db_has_table <- function(con, table) UseMethod("db_has_table",con)
 #' @export
 db_has_table.DBIConnection <- function(con, table) dbExistsTable(con, table)
 


hadley:
That's not how UseMethod works - you don't need to supply and argument unless your dispatching on an argument other than the first.

kevinushey:
@sailingwave can you go into more detail in what problem you were seeing?

I noticed a similar issue where S3 dispatch on S4 objects can fail if that object was `load()`ed from a saved workspace, or similarly after an RStudio session restart.

It's also possible that you maybe just need to reinstall DBI, but I'm curious to know what problem you were seeing.

sailingwave:
I think this should be a problem. I ran these codes and got

```
library(dplyr)
library(DBI)

my_db = src_mysql('MY_DBNAME', 'MY_HOST',user='MY_USER_NAME',pass='***')
db_has_table(my_db$con,'SOME_EXISTING_TABLE')
> [1] NA
dbExistsTable(my_db$con,'SOME_EXISTING_TABLE')
> [1] TRUE
```

After I defined my own db_has_table function as what I committed, it returned 'TRUE' instead of NA.

sailingwave:
Never mind. I've found the explanation of this behavior in the definition of db_has_table.MySQLConnection.

Issue #1999, opened 2016-07-06
case_when not handling is.na statements


tjmahr:
`case_when` doesn't seem to work on `is.na` statements unless they are the very first formula. I'm submitting a pull request.

``` r
library(dplyr)

x <- c(1:6, NA)

case_when(
  is.na(x) ~ "missing",
  TRUE ~ as.character(x)
)
#> [1] "1"       "2"       "3"       "4"       "5"       "6"       "missing"

case_when(
  is.na(x) ~ "missing",
  x %% 2 == 0 ~ "even",
  TRUE ~ as.character(x)
)
#> [1] "1"       "even"    "3"       "even"    "5"       "even"    "missing"

# NA is not recoded
case_when(
  x %% 2 == 0 ~ "even",
  is.na(x) ~ "missing",
  TRUE ~ as.character(x)
)
#> [1] "1"    "even" "3"    "even" "5"    "even" NA
```

<details><summary>

Session info</summary>

``` r
devtools::session_info()
#> Session info --------------------------------------------------------------
#>  setting  value                       
#>  version  R version 3.3.1 (2016-06-21)
#>  system   x86_64, mingw32             
#>  ui       RTerm                       
#>  language (EN)                        
#>  collate  English_United States.1252  
#>  tz       America/Chicago             
#>  date     2016-07-06
#> Packages ------------------------------------------------------------------
#>  package    * version date       source        
#>  assertthat   0.1     2013-12-06 CRAN (R 3.0.2)
#>  DBI          0.4-1   2016-05-08 CRAN (R 3.2.5)
#>  devtools     1.12.0  2016-06-24 CRAN (R 3.3.1)
#>  digest       0.6.9   2016-01-08 CRAN (R 3.2.2)
#>  dplyr      * 0.5.0   2016-06-24 CRAN (R 3.3.1)
#>  evaluate     0.9     2016-04-29 CRAN (R 3.2.5)
#>  formatR      1.4     2016-05-09 CRAN (R 3.2.3)
#>  htmltools    0.3.5   2016-03-21 CRAN (R 3.2.4)
#>  knitr        1.13    2016-05-09 CRAN (R 3.2.3)
#>  magrittr     1.5     2014-11-22 CRAN (R 3.1.2)
#>  memoise      1.0.0   2016-01-29 CRAN (R 3.2.3)
#>  R6           2.1.2   2016-01-26 CRAN (R 3.2.3)
[...]

PR #2000, opened 2016-07-06
case_when support of NA values

tjmahr:
#1999. `case_when` stores results of formula evaluations as vectors in `query`. These values are used to determine which elements in the input vector need to be overwritten. But an NA in an earlier query result would infect TRUE values in other queries. Therefore, I've updated the function to treat NAs as FALSE (when it comes to determining which values need to be replaced).

Also, added a unit test.

### DIFF
--- a/R/case_when.R
+++ b/R/case_when.R
@@ -53,6 +53,7 @@ case_when <- function(...) {
     env <- environment(f)
 
     query[[i]] <- eval(f[[2]], envir = env)
+    query[[i]][is.na(query[[i]])] <- FALSE
     if (!is.logical(query[[i]])) {
       stop("LHS of case ", i, " (", deparse_trunc(f_lhs(f)),
            ") is ", typeof(query[[i]]), ", not logical",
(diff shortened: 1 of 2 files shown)


R/case_when.R
```diff
@@ -53,6 +53,7 @@ case_when <- function(...) {
     env <- environment(f)
 
     query[[i]] <- eval(f[[2]], envir = env)
+    query[[i]][is.na(query[[i]])] <- FALSE
     if (!is.logical(query[[i]])) {
       stop("LHS of case ", i, " (", deparse_trunc(f_lhs(f)),
            ") is ", typeof(query[[i]]), ", not logical",
```
hadley: I think it would make more sense to handle in the `|` below: `replaced | (query[[i]] & !is.na(query[[i]]))`

tjmahr: Thanks. Done

hadley:
Can you please give me a bullet point to put in NEWS?

tjmahr:
* Improved handling of NA values in `case_when` (#1999). Previously, NA-recoding formulas like `is.na(x) ~ "missing"` only worked if they were the first formula given to `case_when`.

PR #2011, opened 2016-07-11
Fix colwise issue when predicate returns a FALSE vector

### LINKED ISSUE #1989: *_if operates on all columns if a predicate returns just character() (= no match)
I don't know if this is an intended behavior (confused me): 

```
library(dplyr)
df <- data_frame( x = as.factor("a"), y = 1L )
str(df)
 # $ x: Factor w/ 1 level "a": 1
 # $ y: int 1

dplyr:::probe_colwise_names(df, is.character)
# character(0)

df <- mutate_if(df, is.character, as.factor)
str(df)
# Classes ‘tbl_df’, ‘tbl’ and 'data.frame':	1 obs. of  2 variables:
#  $ x: Factor w/ 1 level "a": 1
#  $ y: Factor w/ 1 level "1": 1 #
```
`mutate_if` turned  the integer column `y` into a factor on an is.character check. 

My `packageVersion("dplyr")` is `‘0.5.0.9000’`

lionel-:
Closes #1989

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,8 @@
 # dplyr 0.5.0.9000
 
+* Fix issue with `mutate_if()` and `summarise_if()` when a predicate
+  function returns a vector of `FALSE` (#1989, #2009).
+
 # dplyr 0.5.0
 
 ## Breaking changes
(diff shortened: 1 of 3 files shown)


hadley:
Can you add a bullet to NEWS too please?

lionel-:
here you go

PR #2013, opened 2016-07-11
Add indirection for src$con

jcheng5:
This commit adds `con_acquire` and `con_release` S3 interfaces which dplyr can use when it needs to get a connection from a `src` (instead of just `src$con`). This makes it possible to support connection pools properly; without this change, there is no way for dplyr to signal to a connection pool that it is done with a connection.

Feel free to suggest alternate names for `con_acquire`/`con_release` and their parameters.

I previously had code in `con_acquire.src_sql` and `con_release.src_sql` that detected reentrancy (i.e. `con_acquire` called twice on the same `src_sql` without an intervening `con_release`) and this helped me fix some unit tests. But the reentrancy-detection code was quite messy, so I stripped it before committing. If you prefer to have it in (and perhaps enabled only at unit-test time?) I can certainly add it back.

cc @bborgesr


jcheng5:
One other note. I wanted to fail fast (or at least, faster) if anyone out there is using `src$con` directly. So I [changed it to `src$obj`](https://github.com/hadley/dplyr/pull/2013/files#diff-709866ab27135e388b07fd623e7dd38e) (also because pool uses this same var to store the connection pool, not a connection).

bborgesr:
LGTM. When Hadley is ready to merge this here, I think we're also good to go with rstudio/pool#10.  As for the names, I'm fine with `con_acquire` and `con_release`. The only thing maybe worth considering is the variety of synonyms we're generating in between the two packages. In `pool`, we have `pool$fetch()`, `poolCheckout(pool)` and now here we have `con_acquire` -- so three different verbs for essentially the same thing. Similarly, but not quite so bad, we have `pool$release(id, object)`, `poolReturn(object)` and now `con_release`. That said, it doesn't really bug me that much.

hadley:
LGTM. I'll merge when I'm next working on dplyr

hadley:
Can you please add a bullet to NEWS (or paste one in here), and I'll merge

jcheng5:
## New functions

* `con_acquire` and `con_release` are S3 generics that will be called internally by dplyr on `src` objects to get and return connections. (Previously, dplyr called `src$con` to obtain connections from sources.) Overriding these methods allows more flexibility in the types of `src` objects that can be created, such as database connection pools.

jcheng5:
I wasn't sure if `src$con` being renamed to `src$obj` was worth mentioning as a breaking change--I don't think that's a documented field anyway.

krlmlr:
@jcheng5: Good point: This has been suggested in many places as a workaround to access the DBI connection for a dplyr connection. Why can't it just remain `con`?

jcheng5:
@krlmlr Keeping it `con` is an option. It'll mean those workarounds will continue to work, but break when they encounter a pool source (which is fine for scripts, not so fine for packages).

By changing it to `obj` we are essentially failing fast, we'd need to encourage people to use `con_acquire` and `con_release` instead.

bborgesr:
@hadley, is there anything else you need before merging this? Without this in `dplyr`, the integration with `pool` is not great and leaks connections constantly. It's being noticed by pool users, e.g. rstudio/pool#19. No worries if you need more time to review this, just wanted to make sure you don't need anything else from our side

hadley:
Nope, it's good from my perspective. Will merge when I'm next working on dplyr.

hadley:
Can you please add a bullet point to NEWS and we can merge.

@krlmlr can you finish this off please?

krlmlr:
Thanks! I'll add NEWS.

krlmlr:
@jcheng5: 818c6fd9 has the NEWS, could you please double-check?

hannesmuehleisen:
Erm, why did you rename `src$con` to `src$obj`?  I often have the case that people want a raw DBI connection out of a dplyr src (which opens it). The rename breaks those snippets, e.g. `dbDisconnect(src$con)`. I like the new way of getting/releasing connections, but the slot rename seems unnecessary.

hadley:
@hannesmuehleisen I think the reasoning above is sound, as it's better to break code that relies on internal data structures cleanly, rather than hoping they'll continue to work.

hannesmuehleisen:
@hadley any idea when you will release this on CRAN?

hadley:
I'm thinking I'll aim for a March release, which means kicking off the release process in February.

hannesmuehleisen:
Ok. Another question. Should `con_acquire.src_sql` and `con_release.src_sql` not be exported? I am using `src_sql` to construct the source object from a DBI connection, and would have to reach into the `src$obj` now to implement `con_acquire`.

hadley:
Oh yeah, those methods should be exported. Do you mind filing an issue so I don't forget?

Issue #2005, opened 2016-07-07
Problems in handling encodings on Windows


r-suzuki:
Hi, this would be related to #866 and #1555, but seems to be somewhat different.

### Example
Basically encodings are shown as "unknown" on Windows with Japanese locale. For example:

    > x <- transform(iris, test = Sepal.Length + 10, テスト = Petal.Width * 10)
    > Encoding(names(x)) # All "unknown"
    [1] "unknown" "unknown" "unknown" "unknown" "unknown" "unknown" "unknown"

However if we use `mutate` to create new variables, non-English column names are in UTF-8:

    > x.dp <- iris %>% mutate(test = Sepal.Length + 10, テスト = Petal.Width * 10)
    > Encoding(names(x.dp)) # Non-english column is now in UTF
    [1] "unknown" "unknown" "unknown" "unknown" "unknown" "unknown" "UTF-8"  

It does not happen if we make the same variable by `data_frame`:

    > data_frame(test = 1:3, テスト = 1:3) %>% names %>% Encoding
    [1] "unknown" "unknown"

`select` can handle UTF-8 variables:

    > select(x.dp, テスト) %>% head(3)
      テスト
    1      2
    2      2
    3      2

However `group_by` fails:

    > x.dp %>% group_by(テスト) %>% head(3)
    Error: unknown column 'テスト' 

The data made by `transform` has no problem on this:

    > x    %>% group_by(テスト) %>% head(3)
    Source: local data frame [3 x 7]
    Groups: テスト [1]
    
      Sepal.Length Sepal.Width Petal.Length Petal.Width Species  test テスト
             <dbl>       <dbl>        <dbl>       <dbl>  <fctr> <dbl>  <dbl>
    1          5.1         3.5          1.4         0.2  setosa  15.1      2
    2          4.9         3.0          1.4         0.2  setosa  14.9      2
    3          4.7         3.2          1.3         0.2  setosa  14.7      2

### Problem
At first I thought that this is a problem in `group_by` function, but the fundamental problem seems to be that the column names created by `mutate`  are in UTF-8 encoding.

### Environment

    > sessionInfo()
    R version 3.3.1 (2016-06-21)
    Platform: x86_64-w64-mingw32/x64 (64-bit)
    Running under: Windows 7 x64 (build 7601) Service Pack 1
    
    locale:
    [1] LC_COLLATE=Japanese_Japan.932 
    [2] LC_CTYPE=Japanese_Japan.932   
[...]

Issue #1555, opened 2015-11-26
bug: group_by() cannot handle a column name with an international character created by separate()


bbTomas:
The source file encoding is UTF-8. Column "jídlo" created by separate() leads to Error: unknown column 'jídlo' in group_by() function, however the same column "jídlo" created manually (see tab4) works fine.

Thanks,
Tomas

``` r
library(tidyr)
library(dplyr)

tab1 <- data.frame(polevka_1 = 1:4,  hlavni_1 = 2:5, polevka_2 = 3:6,  hlavni_2 = 7:8)
tab2 <- gather(tab1, jidlo_menza, hodnoceni) %>% separate(jidlo_menza, c("jídlo", "menza"))
tab2 %>% group_by(jídlo) %>% summarize(mean(hodnoceni))   # Error: unknown column 'jídlo'

tab3 <- gather(tab1, jidlo_menza, hodnoceni) %>% separate(jidlo_menza, c("jidlo", "menza"))  
tab3 %>% group_by(jidlo) %>% summarize(mean(hodnoceni))  # it works

tab4 <- data.frame(jídlo = rep(c(rep("polevka", 4), rep("hlavni", 4)), 2), menza = c(rep(1, 8), rep(2, 8)), hodnoceni = c(1:4, 2:5, 3:8, 7, 8))
tab4 %>% group_by(jídlo) %>% summarize(mean(hodnoceni))  # it works

# It looks like column 'jídlo' created by separete() function in tab2 has different (hidden?)
# properties than 'jídlo' created by me in tab4.

tab2$jídlo   # ok
tab3$jidlo   # ok
tab4$jídlo   # ok
```

hadley:
Can you confirm that this still fails for you with the dev version of dplyr?

bbTomas:
It still fails, Error: unknown column 'jídlo'

I have installed the following packages
```{r}
devtools::install_github("hadley/dplyr")
devtools::install_github("hadley/tidyr")
```

willmoura:
I was getting the same message error, but when I changed the column name, replacing the character with accent worked fine.

hadley:
@bbTomas could you please give the output of `Sys.getlocale()` and tell me what OS you're using? (And this is Czech, right?)

bbTomas:
It was Windows 7. Now, it is Windows 10. With the same result.

```
sessionInfo()
R version 3.3.0 (2016-05-03)
Platform: x86_64-w64-mingw32/x64 (64-bit)
[...]

Issue #1950, opened 2016-06-21
Encoding problems on Windows caused by character -> symbol -> character roundtrip


krlmlr:
Some of the encoding problems (e.g., grouping) seem to be caused by converting `character` to `symbol` and back. On Linux (UTF-8 locale):

```
> "ä" %>% Encoding
[1] "UTF-8"
> "ä" %>% as.name %>% as.character %>% Encoding
[1] "UTF-8"
> "ä" %>% iconv(to = "latin1") %>% as.name %>% as.character %>% Encoding
[1] "unknown"
> "ä" %>% iconv(to = "latin1") %>% as.name %>% as.character %>% enc2utf8
[1] "<e4>"
```

On Windows (latin-1 locale):

```
> "ä" %>% Encoding
[1] "latin1"
> "ä" %>% as.name %>% as.character %>% Encoding
[1] "latin1"
> "ä" %>% iconv(to = "UTF-8") %>% as.name %>% as.character %>% Encoding
[1] "unknown"
> "ä" %>% iconv(to = "UTF-8") %>% as.name %>% as.character %>% enc2utf8
[1] "Ã¤"
```

On Windows, [one test fails](https://github.com/hadley/dplyr/blob/e87f6dd7181b143dba2ad26d13ec1fb8bbd902d0/tests/testthat/test-group-by.r#L240-L245) because of that.

So, currently we should suggest using ASCII (or at least native-encoded) column names, and _not_ fiddle with the encoding, in particular not set it to UTF-8 on Windows.

A lot of internal dplyr logic seems to be based on the `symbol` type, I don't see a quick solution here.

kevinushey:
IIUC, the over-arching issue is the the `as.name` => `as.character` roundtrip loses the encoding, and so we no longer know how to properly translate from the original encoding to the desired encoding?

hadley:
Maybe dplyr could have it's own symbol coercion methods that always assumed UTF-8?

krlmlr:
We could do that, but I think the safest option is not to use symbols to represent column names -- a simple S3 class should do the same job.

hadley:
I think symbols are the correct way to represent column names for a number of reasons.

krlmlr:
We could settle for a combination:

```
> structure(as.name("col_name"), class = "encoded_symbol")
col_name
attr(,"class")
[1] "encoded_symbol"
```

With a suitable `as.character.encoded_symbol()` function.

krlmlr:
[...]

PR #2058, opened 2016-08-09
Fix the problem that group_by() cannot handle UTF-8 column names on Windows

r-suzuki:
Hi, this is just a workaround to the issue #2005 (and  possibly related to #1555). With this fix the following example works properly:

```
library(dplyr)
data(iris)
iris %>%
    mutate(品種 = Species) %>%
    group_by(品種) %>%
    summarise(avg = mean(Sepal.Length))
```

It would be preferable to fix build_index_cpp to handle multiple encodings, however this workaround is simple and works as expected.

### DIFF
--- a/R/dataframe.R
+++ b/R/dataframe.R
@@ -46,7 +46,14 @@ add_rownames <- function(df, var = "rowname") {
 #' @export
 group_by_.data.frame <- function(.data, ..., .dots, add = FALSE) {
   groups <- group_by_prepare(.data, ..., .dots = .dots, add = add)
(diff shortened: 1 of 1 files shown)


krlmlr:
Thanks. My plan is to fix this at a more general scope as part of #1950.

r-suzuki:
Thanks @krlmlr , general fix is more appreciated. This problem seems to need a fix in base-R so I ended up with just a workaround here.

This encoding problem is the main reason that I cannot use dplyr in our projects, so any fix would be far better than nothing. I hope there is any solution.

PR #2060, opened 2016-08-09
add dplyr.tibble.print to override tibble printing behavior

javierluraschi:
This PR adds a `dplyr.tibble.print` option to allow users to override tibble printing behavior.

screen shot 2016-08-09 at 4 03 43 pm

### DIFF
--- a/R/grouped-df.r
+++ b/R/grouped-df.r
@@ -29,7 +29,7 @@ print.grouped_df <- function(x, ..., n = NULL, width = NULL) {
   grps <- if (is.null(attr(x, "indices"))) "?" else length(attr(x, "indices"))
   cat("Groups: ", commas(deparse_all(groups(x))), " [", big_mark(grps), "]\n", sep = "")
   cat("\n")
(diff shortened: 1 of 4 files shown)


krlmlr:
Have you seen knitr::knit_print()? What other use cases does the new option have?

javierluraschi:
@krlmlr I'm closing this one, but here is some background...

You are right, `knitr::knit_print` could be used here to overwrite output; however, at the time I opened this PR I wasn't interested in overwriting the whole print function, but just the data frame part. I think an alternative fix (or perhaps functionality already exists that I'm not aware of), would be for `dplyr` to provide a helper function to get the summary information that other packages can use to integrate in their UI.

Regardless, outputting the header as console + the data frame is no longer supported in the IDE and instead, a gallery is being created of each output; therefore, there it becomes confusing if we split output in two gallery entries and therefore, if we want to display back this information, it would have to be within the gallery entry itself.

Thanks for taking a look at this.

Issue #2044, opened 2016-07-30
select_if doesn't work on columns that contain spaces


double-y:
Hi

select_if doesn't seem to work on columns which contain spaces

To this dataset,
http://datascienceplus.com/wp-content/uploads/2016/07/dataset_HD_euro2016.csv

``` r
select_if(is.numeric)
> Error in parse(text = x) : <text>:1:7: unexpected symbol 1: Total goals ^
```

Doesn't work but

``` r
rename(Totalgoals = `Total goals`)
select_if(is.numeric)
```

Works

izahn:
http://stackoverflow.com/questions/39134602/how-to-pass-column-names-in-dplyr-select-without-evaluation might be the same problem. Here is a very simple example thanks to http://stackoverflow.com/users/496803/thelatemail

`dat <- data.frame(`A B`=1:3,check.names=FALSE); select_(dat, "A B")`

krlmlr:
Is this a duplicate of #1978?

krlmlr:
@hadley: Column names with spaces make a fairly large cluster, too. Okay to link+close?

hadley:
Yes, please.

krlmlr:
Duplicate of #1978.

krlmlr:
#2243 groups all related issues.

Issue #1978, opened 2016-06-28
*_if and  *_at functions do not handle nonstandard names


rongmu:
I was trying out the new `*_if` and `*_at` functions in dplyr 0.5.0 and found that they don't work well with nonstandard names.

If your data frame contains a column called `Standard Error`, for example, you will get an error when applying any of these functions.

```
Error in parse(text = x) : <text>:1:10: unexpected symbol
1: Standard Error
             ^
```

Example code:

``` r
library(dplyr)

df <- data.frame(
    Estimate = c(
      6.7253,
      3.8621,
      4.9862
      ),
    `Standard Error` = c(
      0.1301,
      0.2656,
      0.2338
      ),
    check.names = FALSE
  )

# none of below works :-(
df %>% mutate_if(is.numeric, funs(round(., 2)))
df %>% summarise_if(is.numeric, mean)

df %>% mutate_at('Standard Error', funs(round(., 2)))
df %>% summarise_at('Standard Error', mean)

df %>% select_if(is.numeric)
```

heavywatal:
Try `vars()` and backticks for `*_at`:
```r
> df %>% summarise_at(vars(`Standard Error`), mean)
  Standard Error
1      0.2098333
```

Sorry I have no idea on `*_if`. They may need to be fixed.

emailmarketingtips:
Yes, I also stumbled across `*_if`: 

```
library(dplyr)
df <- data_frame(`foo bar`=1L)
mutate_if(df, is.integer, `+`, e2=1)
# Error in parse(text = x) : <text>:1:5: unexpected symbol
# 1: foo bar
#         ^

df <- data_frame(`foo`=1L)
mutate_if(df, is.integer, `+`, e2=1)
# Source: local data frame [1 x 1]
# 
#     foo
#   <dbl>
# 1     2

sessionInfo()
# R version 3.3.0 (2016-05-03)
# Platform: x86_64-w64-mingw32/x64 (64-bit)
# Running under: Windows 7 x64 (build 7601) Service Pack 1
# 
# locale:
# [1] LC_COLLATE=German_Germany.1252  LC_CTYPE=German_Germany.1252    LC_MONETARY=German_Germany.1252 LC_NUMERIC=C                    LC_TIME=German_Germany.1252    
# 
# attached base packages:
# [1] stats     graphics  grDevices utils     datasets  methods   base     
# 
# other attached packages:
# [1] dplyr_0.5.0.9000
# 
[...]

PR #2071, opened 2016-08-15
fixes select_if with non standard names, #2044 and #1978

jtrecenti:
Small hack. Maybe it would be better to fix `select_`.


krlmlr:
Could you please add a test that shows where the current implementation fails?

hadley:
I'd prefer a lower-level fix.

PR #2078, opened 2016-08-18
Proposing add_count and add_tally for adding an n column within groups

dgrtwo:
I [very often find myself](https://twitter.com/drob/status/766294758696357888) doing the following three steps (for example, before filtering out all groups below a certain size):

``` r
data %>%
  group_by(...) %>%
  mutate(n = n()) %>%
  ungroup()
```

I've contributed two verbs that shorten this, with working title `add_count` and `add_tally`. They add an `n` column (or an `nn` column if one is already present) to the data frame with counts within groups. Unlike `count` and `tally` the data is otherwise unchanged (same number of observations).

Also included basic unit tests and examples but I can flesh those out more if necessary.


dgrtwo:
Two main questions:

* **why the name, and is there an alternative**: I'm working off the possible precedent of `add_predictions`, `add_predictors`, and `add_residuals` from the modelr package. These are functions that  .

    The alternative we've used in the tidytext package is [bind_tf_idf](http://juliasilge.com/blog/Term-Frequency-tf-idf/), when we add three columns (`tf`, `idf`, and `tf_idf`) to a table without otherwise changing it. If there's another prefix that means "add a column" I'm open to it, though I don't think `mutate_` or `append_` would be right.

    We could try to come up with clever verbs that don't refer to count or tally specifically (like... `enumerate` and `tick` or something) but I think that's making it more obscure

* **why the grouping behavior**: I think it is important to return the tbl to its original grouping, because most of the time when this verb is useful, it is immediately followed by filtering or mutating, which are faster when ungrouped. (filtering is a particularly common pitfall). I also think that if the grouping behavior were preserved, one might as well just have used `group_by` so that it's clearer

mgacc0:
It would be a nice feature to have.
Alternatively I would propose the name `add_tally_by` (or `add_count_by`).

krlmlr:
LGTM.

hadley:
LGTM

krlmlr:
Thanks!

PR #2084, opened 2016-08-23
FIX: column reference "col_name" is ambiguous when joining in Postgresql

alberthkcheng:
When both different and identical names existed in 'by', ambiguity arise.

i.e. `left_join(table_a, table_b , by = c("col_name"="col_name","col_a"="col_b")`

Suggested to force rename all.


krlmlr:
Thanks. Would you mind adding a test?

krlmlr:
I think this has been handled in #1997, now merged. Could you please double-check?

PR #2087, opened 2016-08-24
Creation of the setattr and getattr functions

paulponcet:
Proposal to add to dplyr functions (called setattr, setattr_, getattr, getattr_) to add, remove or get attributes from a tbl


krlmlr:
This looks interesting, but I'm not sure dplyr is the best fit for this functionality.

PR #2097, opened 2016-08-29
Allow unnamed additional arguments in colwise verbs

lionel-:
So that we can do:

```r
df %>% mutate_all(round, 2)
```

Found a weird R inconsistency while writing this:

```r
c <- call("fun", quote(arg))
names(c)[2] <- "n"
names(c)
#> [1] ""  "n"

l <- list(1, 2)
names(l)[2] <- "n"
names(l)
#> [1] NA  "n"
```

also

```r
names(c) <- c(NA, NA)
names(c)
#> NULL

names(l) <- c(NA, NA)
names(l)
#> [1] NA NA
```

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -25,6 +25,8 @@
 * Fix issue with `mutate_if()` and `summarise_if()` when a predicate
   function returns a vector of `FALSE` (#1989, #2009).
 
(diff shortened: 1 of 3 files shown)


hadley:
Nice spotting on the names inconsistency!

PR #2103, opened 2016-08-31
Also format header files in inst/

krlmlr:
basically with the same settings. Ignoring Boost file.

The formatting may not be perfect, but I think it's a good compromise. Okay to merge?

The following changes are more than spaces-only:

6c4b09b: astyle logic
df3d8c1: turn off indentation (astyle does weird things here)
a186af2: modifier alignment
3d80d6e: long expressions
fd859c0: multiline
f3fb181: remove useless if


hadley:
LGTM

PR #2110, opened 2016-09-06
Add .dots argument to dplyr::recode to support passing replacements as list

jlegewie:
… that they can be saved and reused

`dplyr::recode` and `dplyr::recode_factor` should support a list of replacements with an additional argument such as `.dots`. I know that there are workarounds but supporting this natively would be very useful! Here is an example:

```
x <- c(rep("White", 4), rep("African-American", 2), rep("Hispanic", 2))

dplyr::recode(x, "White" = 1, "African-American" = 2, "Hispanic" = 3)
recode_race <- list("White" = 1, "African-American" = 2, "Hispanic" = 3)
dplyr::recode(x, .dots = recode_race)
```

ps: I think the same would be useful in other "tidyverse" functions (e.g. `dplyr::case_when`, `forcats::fct_recode`). Happy to send a pull request for them as well but first want to see how this goes.


R/recode.R
```diff
@@ -63,14 +66,17 @@
...
   UseMethod("recode")
 }
 
 #' @export
-recode.numeric <- function(.x, ..., .default = NULL, .missing = NULL) {
+recode.numeric <- function(.x, ..., .default = NULL, .missing = NULL, .dots = NULL) {
   values <- list(...)
-
+  if (!is.null(.dots)) {
+    values <- .dots
```
krlmlr: What if the caller supplies both ... and .dots? IIRC, lazyeval combines these.

R/recode.R
```diff
@@ -82,7 +88,7 @@ recode.numeric <- function(.x, ..., .default = NULL, .missing = NULL) {
   }
 
   n <- length(.x)
-  template <- find_template(..., .default, .missing)
+  template <- find_template(..., .default, .missing, .dots)
```
krlmlr: I think .dots must be passed as named argument here.

R/recode.R
```diff
@@ -149,8 +161,11 @@ recode.factor <- function(.x, ..., .default = NULL, .missing = NULL) {
   .x
 }
 
-find_template <- function(...) {
+find_template <- function(..., .dots = NULL) {
   x <- compact(list(...))
+  if (!is.null(.dots)) {
+    values <- compact(.dots)
```
krlmlr: Same as above.

R/recode.R
```diff
@@ -63,14 +65,17 @@
...
+recode <- function(.x, ..., .default = NULL, .missing = NULL, .dots) {
   UseMethod("recode")
 }
 
 #' @export
-recode.numeric <- function(.x, ..., .default = NULL, .missing = NULL) {
+recode.numeric <- function(.x, ..., .default = NULL, .missing = NULL, .dots) {
   values <- list(...)
-
+  if (!missing(.dots)) {
```
krlmlr: Can this be made simpler if the default value of .dots is NULL?

R/recode.R
```diff
@@ -121,8 +129,11 @@ recode.character <- function(.x, ..., .default = NULL, .missing = NULL) {
 }
 
 #' @export
-recode.factor <- function(.x, ..., .default = NULL, .missing = NULL) {
+recode.factor <- function(.x, ..., .default = NULL, .missing = NULL, .dots) {
   values <- list(...)
+  if (!missing(.dots)) {
```
krlmlr: Same.

R/recode.R
```diff
@@ -98,14 +103,17 @@ recode.numeric <- function(.x, ..., .default = NULL, .missing = NULL) {
 }
 
 #' @export
-recode.character <- function(.x, ..., .default = NULL, .missing = NULL) {
+recode.character <- function(.x, ..., .default = NULL, .missing = NULL, .dots) {
   values <- list(...)
+  if (!missing(.dots)) {
```
krlmlr: Same.

jlegewie:
Why `forcats`? `recode` is not just for factors...


A couple of changes:
- `...` and `.dots` are now combined
- I added tests

Two remaining issues:
- The tests show a warning `Codoc mismatches from documentation object` for the `.dot` argument even though I added documentation for it. Not sure why
- Continuous integration fails but because of tests in `test-arrange.r` and `test-binds`.

Note that `find_template` just takes one argument now, which is the combined `dots`. The function also took `.missing` and `.default` before but never used them.

krlmlr:
Thanks. The failing tests are strange. Could you please:

- rebase against current master, and force-push
- declare `.dots` with a default value `.dots = NULL`
- assign `values` as `values = c(list(...), .dots)`, without checking for `missing(.dots)`
- double-check the `find_template()` calls in the case of an empty recode list (that's when `.default` and/or `.missing` could have come to play; we might need tests here too)
- optionally: run `devtools::document()` with the current development version of roxygen2 and the current CRAN version of Rcpp (this should fix the codoc warnings)

jlegewie:
Sorry for the delay. I made all five changes. Just two things:

- I am not sure whether I rebased correctly.
- I am not sure what the desired behavior is with empty recode list AND a `.default` and/or `.missing` argument. Currently, it results in an error message `Error: No replacements provided` but I think that was the same before.

R/recode.R
```diff
@@ -98,14 +100,14 @@ recode.numeric <- function(.x, ..., .default = NULL, .missing = NULL) {
 }
 
 #' @export
-recode.character <- function(.x, ..., .default = NULL, .missing = NULL) {
-  values <- list(...)
+recode.character <- function(.x, ..., .default = NULL, .missing = NULL, .dots = NULL) {
+  values = c(list(...), .dots)
```
krlmlr: Please use `<-` for assignment.

R/recode.R
```diff
@@ -121,8 +123,8 @@ recode.character <- function(.x, ..., .default = NULL, .missing = NULL) {
 }
 
 #' @export
-recode.factor <- function(.x, ..., .default = NULL, .missing = NULL) {
-  values <- list(...)
+recode.factor <- function(.x, ..., .default = NULL, .missing = NULL, .dots = NULL) {
+  values = c(list(...), .dots)
```
krlmlr: Here, too.

R/recode.R
```diff
@@ -98,14 +100,14 @@ recode.numeric <- function(.x, ..., .default = NULL, .missing = NULL) {
...
-  values <- list(...)
+recode.character <- function(.x, ..., .default = NULL, .missing = NULL, .dots = NULL) {
+  values = c(list(...), .dots)
   if (!all(has_names(values))) {
     stop("All replacements must be named", call. = FALSE)
   }
 
   n <- length(.x)
-  template <- find_template(..., .default, .missing)
+  template <- find_template(values)
```
krlmlr: The original code probably was be responsible for the following behavior:

```r
> recode(letters[1:3], .default = "A")
[1] "A" "A" "A"
> recode(c(letters[1:3], NA), .missing = "A")
[1] "a" "b" "c" "A"
```

Could you please double-check (and add a test)?

tests/testthat/test-recode.R
```diff
@@ -102,3 +102,25 @@ test_that("conversion of unreplaced values to NA gives warning", {
   expect_warning(recode(1:3, `1` = "a"), "treated as NA")
   expect_warning(recode_factor(letters[1:3], b = 1, c = 2))
 })
+
+test_that(".dot argument works correctly", {
```
krlmlr: Please refer to the PR number in the test title.

jlegewie:
Okay, all fixed. I added the tests for `.missing` and `. default` to the existing test series for the respective argument and not the tests for this PR

krlmlr:
Awesome, thanks!

PR #2148, opened 2016-09-26
allow `n_distinct()` to accept `...` arguments

kevinushey:
This allows `n_distinct()` to be more permissive, for SQL engines which accept one or more columns to `count(DISTINCT ....)`.

### DIFF
--- a/R/translate-sql-base.r
+++ b/R/translate-sql-base.r
@@ -141,9 +141,7 @@ base_agg <- sql_translator(
   sum        = sql_prefix("sum", 1),
   min        = sql_prefix("min", 1),
   max        = sql_prefix("max", 1),
-  n_distinct = function(x) {
-    build_sql("COUNT(DISTINCT ", x, ")")
-  }
+  n_distinct = function(...) build_sql("COUNT(DISTINCT", list(...), ")")
 )
 
 #' @export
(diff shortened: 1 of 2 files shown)


R/translate-sql-base.r
```diff
@@ -141,9 +141,7 @@ base_agg <- sql_translator(
   sum        = sql_prefix("sum", 1),
   min        = sql_prefix("min", 1),
   max        = sql_prefix("max", 1),
-  n_distinct = function(x) {
-    build_sql("COUNT(DISTINCT ", x, ")")
-  }
+  n_distinct = function(...) build_sql("COUNT(DISTINCT", list(...), ")")
```
krlmlr: This seems to add parentheses: `COUNT(DISTINCT(x, y, z))`. Which SQL engines support this syntax?

kevinushey: Interesting; I didn't realize that (that was unintentional). The SQL engine used by Spark does support this syntax (which is where I tested this), although it's documented as:

    count(DISTINCT expr[, expr...])

That is, without parentheses. Ultimately, the goal would be to construct the expression as `COUNT(DISTINCT foo, bar, baz))`.

krlmlr:
Would you mind updating?

hadley:
Bump

PR #2157, opened 2016-09-30
New RowwiseSlicingIndex, NaturalSlicingIndex and OffsetSlicingIndex

krlmlr:
- allows removing redundant code
- preparation for hybrid evaluation refactoring
- reduces allocations
- a template solution (instead of virtual functions) may give better performance, can be refactored later

@hadley: PTAL.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,7 @@
 # dplyr 0.5.0.9000
 
+* `SlicingIndex` is now a virtual class with specialized implementations `GroupedSlicingIndex`, `RowwiseSlicingIndex`, `NaturalSlicingIndex` and `OffsetSlicingIndex` (#2187).
+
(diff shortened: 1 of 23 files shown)


hadley:
Seems reasonable to me. Maybe worth adding a couple of lines of comments to each class to briefly describe their role?

krlmlr:
@hadley: Added comments (e60172b). Okay to merge?

hadley:
LGTM

krlmlr:
For the record: Performance did improve with this PR.

bind_rows

krlmlr:
@lionel-: The `SlicingIndex` is an interface that allows selection of a subset of rows in the hybrid evaluator. The `NaturalSlicingIndex` selects the entire data frame; see #2311 for discussion on making the ungrouped case a special case of the grouped case for all operations.

PR #2162, opened 2016-10-05
enforce integer 'n' for lag()

kevinushey:
This resolves an issue in stricter SQL engines where `double` values are not accepted.


hadley:
LGTM. News bullet please?

krlmlr:
Thanks! I'll update NEWS.

PR #2168, opened 2016-10-10
goodpractice to use TRUE not T

jsta:


krlmlr:
I find are a few more instances of `T` instead of `TRUE` when I search for `T)` and `T,` .

hadley:
It's not super important to me, because I don't think it's that bad a practice, but a fix would be nice.

krlmlr:
@jsta: Would you mind enhancing your PR?

PR #2179, opened 2016-10-16
fix: Add an equal sign to make function select best two years not one…

YiyingW:
… year


krlmlr:
I find this part of the vignette misleading at best. The text around that code refers to percent_rank() but not min_rank().

hadley:
I think that first example could just be deleted (or moved earlier in the doc)

krlmlr:
@YiyingW: Would you mind enhancing your PR?

PR #2232, opened 2016-11-06
fix duplicate name bug https://github.com/edzer/sfr/issues/49#issueco…

### LINKED ISSUE #2248: bind_cols() allows duplicate column names
As @hadley [noted](https://github.com/edzer/sfr/issues/49#issuecomment-258683896), this is likely a bug.

```
df1 <- mtcars %>% select(mpg,cyl,disp) %>% slice(1:5)
df2 <- mtcars %>% select(mpg,hp,drat) %>% slice(6:10)

names(one);names(two)
## [1] "mpg"  "cyl"  "disp"
## [1] "mpg"  "hp"   "drat"

df_combined <- bind_cols(df1,df2)

df_combined
## mpg cyl disp  mpg  hp drat
## 1 21.0   6  160 18.1 105 2.76
## 2 21.0   6  160 14.3 245 3.21
## 3 22.8   4  108 24.4  62 3.69
## 4 21.4   6  258 22.8  95 3.92
## 5 18.7   8  360 19.2 123 3.92

df_combined %>% select(mpg)
## Error: found duplicated column name: mpg
```

Seems like a good opportunity to add a `suffix` argument, as seen in the `dplyr::*_join()` verbs.

edzer:
…mment-258683896

Signed-off-by: Edzer Pebesma <edzer.pebesma@uni-muenster.de>

Fixes #2248.

### DIFF
--- a/R/bind.r
+++ b/R/bind.r
@@ -86,7 +86,10 @@ bind_rows <- function(..., .id = NULL) {
 #' @rdname bind
 bind_cols <- function(...) {
   x <- list_or_dots(...)
-  cbind_all(x)
+  ret <- cbind_all(x)
+  if (anyDuplicated(names(ret)) != 0)
+    names(ret) <- make.unique(names(ret))
+  ret
 }
 
 #' @export


krlmlr:
I'd rather throw an error here. @hadley?

hadley:
I think it would be best to follow the readr principle and automatically fix the names with a warning:

```R
> readr::read_csv("a,a,\n1,1,1")
# A tibble: 1 × 3
      a   a_1    X3
  <int> <int> <int>
1     1     1     1
Warning messages:
1: Missing column names filled in: 'X3' [3] 
2: Duplicated column names deduplicated: 'a' => 'a_1' [2] 
```

We may want to pull this behaviour into a tibble function.

I think it's reasonable for the behaviour to be different to `tibble()`. `tibble()` can be stricter because it's used in contexts where you have total control over the names.

krlmlr:
There's tibble::make_unique(), we could export that and add a warning option.

krlmlr:
Yeah. There's also tibble::repair_names(), which is already exported. @edzer: Would you mind switching to this function?

@hadley: There seems to be some duplication between tibble and readr, how to resolve?

hadley:
Maybe combine into https://github.com/tidyverse/readr/issues/538 ? It's likely that we'll need some compromise on the behaviour

PR #2233, opened 2016-11-06
Avoid coercing pairlist to list

krlmlr:
to fix performance regression in #2220.

### DIFF
--- a/src/utils.cpp
+++ b/src/utils.cpp
@@ -62,10 +62,10 @@ SEXP pairlist_shallow_copy(SEXP p) {
 }
 
 void copy_only_attributes(SEXP out, SEXP data) {
-  List att = ATTRIB(data);
-  const bool has_attributes = (att.length() > 0);
+  SEXP att = ATTRIB(data);
+  const bool has_attributes = !Rf_isNull(att);
   if (has_attributes) {
-    LOG_VERBOSE << "copying attributes: " << CharacterVector(att.names());
+    LOG_VERBOSE << "copying attributes: " << CharacterVector(List(att).names());
 
     SET_ATTRIB(out, pairlist_shallow_copy(ATTRIB(data)));
   }


krlmlr:
filter

hadley:
Yes please!  

BTW graphs look great - what did you end up using for changing detection?

krlmlr:
Thanks. I'm using [`tsoutlier::locate_outliers(types = "LS")`](https://github.com/krlmlr/dplyr.benchmark/blob/f6a3f02b9f3b417c4392ed3713d19b077ec51ff5/R/plot.R#L47), but I'm not entirely satisfied with the result. The package seems to provide a mechanism for removing false positives, but I couldn't get it to work with my data, and nobody whom I reached on [Twitter](https://twitter.com/krlmlr/status/793098859933294593) seems to know how to do that either.

hadley:
It might also be worth trying https://github.com/twitter/AnomalyDetection

And possibly https://github.com/hadley/l1tf

Ideally you want a model that's piecewise constant with anomalies, but I don't know of anything that does that off the shelf

jayhesselberth:
Those graphs are great. Perhaps the approach should be cleaned up and put into a generic (non-dplyr-specific) package? Would be nice to plot any commit series or time frame.

We have used the [changepoint](https://cran.r-project.org/web/packages/changepoint/index.html) package for similar types of data, might be useful here.

krlmlr:
Thanks for the hint, will try as soon as my current method starts giving unsatisfactory results.

@analyticalmonk has a related GSOC project "[Rperform](https://github.com/analyticalmonk/Rperform)", the needs for dplyr seemed to be very specific but the experience gained might be useful for that project, too.

analyticalmonk:
Thanks for the tip @krlmlr 
Will look into this.

Issue #2224, opened 2016-11-02
group_by executes column names as code


davharris:
*Previously posted as  hadley/purrr#257; @lionel- said it was a dplyr::group_by issue, so reposting it here*

Column names can be executed as code when `group_by` & related operations are used. I initially discovered this problem because I had a column name with a space in it:

```
x = mtcars
colnames(x) = paste(colnames(x), 1:ncol(x))
slice_rows(x, 1)

# Error in parse(text = x) : <text>:1:5: unexpected numeric constant
#1: mpg 1
```

After I looked at the traceback information, I found that purrr (via dplyr) was actually using lazyeval to parse and run the column names as code. That seems unsafe to me (analogous to SQL injection, but possibly worse because R can call system).

krlmlr:
Confirmed, using:

```r
x = mtcars
colnames(x) = paste(colnames(x), 1:ncol(x))
group_by_prepare(x, .dots = names(x))
```

We should coerce to names first.

@davharris: Would you like to contribute a testthat test?

davharris:
Sure.  To clarify, did you mean that `names` should be coerced to character? Or is there a `names` class?

krlmlr:
dplyr should take care of the coercion, our code should work as is.

davharris:
Submitted a failing test in #2235.

hadley:
But this is actually correct behaviour so you can do:

```R
mtcars %>% group_by(cut(mpg, 3))
```

`group_by()` has `mutate()` semantics, not `select()` semantics.

krlmlr:
There are so many issues about columns with spaces. Should we give a nicer error message if it turns out that there is actually a column by that name?

davharris:
Also, it's probably a bad idea for column names to be executed as code if they aren't explicitly typed in by the user.  This is basically SQL injection, but for R.

davharris:
It looks like `group_by` can also produce other unexpected behavior when executing column names as code.

```
df <- data_frame(`warning("column name executed as code")` = 1:3, b = 1:3)
group_by(df, `warning("column name executed as code")`) # groups correctly; no warning
group_by(df, warning("column name executed as code"))    # mutates the first column; warning
```

Again, this behavior can be triggered without the user being aware that the column name is anything unusual: `slice_rows(df, 1)` returns instead of 

```
Source: local data frame [3 x 2]
Groups: warning("column name executed as code") [1]

  `warning("column name executed as code")`     b
                                      <chr> <int>
[...]

PR #2235, opened 2016-11-07
Test for #2224

davharris:
This test currently fails because dplyr tries to run the column names as code (#2224); it will succeed once the column names are coerced to text.


krlmlr:
Thanks! If you add skip() calls, we can merge right away. @hadley: OK?

tests/testthat/test-group-by.r
```diff
@@ -256,3 +256,15 @@ test_that("rowwise fails gracefully on raw columns (#1803)", {
   df <- data_frame(a = 1:3, b = as.raw(1:3))
   expect_error( rowwise(df), "unsupported type" )
 })
+
+    
+test_that("group_by will not execute column names as code #2224", {
+  df <- mtcars
```
hadley: I prefer to create a very simple df in the test

Issue #2222, opened 2016-10-31
length(unique(x)) and n_distinct(x) return different answers for some lists


jgabry:
The documentation for `n_distinct` says that it's an "equivalent of length(unique(x))"
but it gives a different answer than `length(unique(x))` if `x` is a list with multiple (identical) elements.

```r
library(dplyr)  # I checked versions 0.5.0 and 0.5.0.9000
x <- list(1,1)
length(unique(x)) # result: 1
n_distinct(x) # result: 2
```

Is this a bug or is there just a missing caveat in the doc?

krlmlr:
This is because values in list columns are compared by reference, so n_distinct() treats them as different unless they really point to the same object:

```r
a <- 1
n_distinct(list(a,a))
## [1] 1
```

Would you like to contribute documentation?

jgabry:
Thanks for the explanation. That makes sense. I'd be happy to submit a PR
that clarifies this in the doc and adds an example demonstrating. Can
hopefully do this today.

hadley:
I think this is more of a bug than something we need to document

jgabry:
When I originally opened the issue it was because it seemed like a bug.
Kirill's explanation of the behavior makes sense but it does seem
preferable to have it behave like the base R equivalent than to have to
explain the unexpected deviation.

On Tue, Jan 31, 2017 at 6:45 PM Hadley Wickham <notifications@github.com>
wrote:

I think this is more of a bug than something we need to document

—
You are receiving this because you authored the thread.
Reply to this email directly, view it on GitHub
<https://github.com/hadley/dplyr/issues/2222#issuecomment-276530534>, or mute
the thread

hadley:
Now part of #2355

Issue #1670, opened 2016-02-20
distinct function and lists in column bug


zhilongjia:
For a column containing lists in data.frame, distinct will fail. 
Example:

```
df <- data.frame(
    x = sample(10, 100, rep = TRUE),
    y = sample(10, 100, rep = TRUE)
)
df$z <- list(c("a", "b", "c"))
df$z[3] <- list(c("a", "b", "c"))
dplyr::distinct(df, z)
# result:
# x y       z
#8 3 a, b, c
#5 5 a, b, c

# shoud be
#8 3 a, b, c
identical(df$z[1], df$z[3])
# TRUE
```

hadley:
Simpler reprex:

```R
df <- dplyr::data_frame(
  x = list(list(1, 2), list(1, 2))
)
dplyr::distinct(df, x)
```

@romainfrancois can you please take a look?

krlmlr:
This is because currently the binary representation is hashed, which is just a pointer to the value, and different in this scenario. The following returns just one row:

```r
a <- list(1, 2)

df <- dplyr::data_frame(
  x = list(a, a)
)
dplyr::distinct(df, x)
```

We should probably just give an error here.

soniampub:
@krlmlr I would like to work on this issue. I am a beginner in open source contribution and also in dplyr. Please let me know what should be fixed here. Thanks!!

krlmlr:
@soniampub: I'd start writing a test that defines the expected behavior, it should fail with the current codebase.

soniampub:
Sure let me know when you had checked in that test. Also in my local git is there a script to run all tests and see which one is failing, is there a documentation or mailing list that I should use to post beginner questions. I appreciate your help.

krlmlr:
The tests can be run with `devtools::test()`. StackOverflow is a good place to ask questions of any kind.

From my gut feeling, this issue requires a change to the C++ code. A pull request that contains a failing test and a fix, or even just a failing test, will be greatly appreciated.

MarcAragones:
It happens the same with vectors:

```r
df <- dplyr::data_frame(
  x = list(c(1, 2), c(1, 2))
)
dplyr::distinct(df, x)

# Result:
# A tibble: 2 × 1
          x
     <list>
1 <dbl [2]>
2 <dbl [2]>
```

krlmlr:
I think we can't do much about it except documenting it, see also #2222.

hadley:
This seems like a bug to me.

[...]

PR #2237, opened 2016-11-07
document case where n_distinct(x) differs from length(unique(x)) (#2222)

jgabry:
This PR resolves issue #2222 by adding a Note section and an additional example to the doc for `n_distinct`.


krlmlr:
@hadley: LGTM. We can update Rd files later.

krlmlr:
This also applies to distinct(), see #1670, and probably a few other places. Do we need to extend the documentation?

krlmlr:
Thanks. I'd like to wait here, because we might actually be able to fix the underlying issue -- see #2355.

Issue #2200, opened 2016-10-26
summarize on a grouped data frame drops order from ordered factors


ateucher:
It seems that summarize on a grouped data frame doesn't respect ordered factors. Possibly similar to #1112

``` r
library(dplyr)

data <- tibble(id = rep(letters[23:26], 3), 
               val = runif(12, 0, 12))

# Function that returns an ordered factor
make_levels <- function(x) {
  cut(x, breaks = c(2,4,6,8,10), labels = LETTERS[1:4], ordered_result = TRUE)
}

is.ordered(make_levels(3))
```

```
## [1] TRUE
```

``` r
ungrouped_summary <- data %>%
  summarize(mean_val = mean(val), 
            level = make_levels(mean_val))

## Keeps ordered
is.ordered(ungrouped_summary$level)
```

```
## [1] TRUE
```

``` r
grouped_summary <- data %>% group_by(id) %>% 
  summarize(mean_val = mean(val), 
            level = make_levels(mean_val))

## Keeps as factor, but not ordered any more:
is.ordered(grouped_summary$level)
```

```
## [1] FALSE
```

``` r
is.factor(grouped_summary$level)
```

```
## [1] TRUE
```

krlmlr:
Confirmed. Would you like to contribute a testthat test?

ateucher:
Great, thanks! Yes I'd be happy to

ateucher:
@krlmlr I added two tests, as I also found that `min` and `max` on an existing ordered factor also drops order. Hope they do the trick.

krlmlr:
Awesome, thanks!

PR #2238, opened 2016-11-07
Add failing tests for summarize preserving ordered factors (#2200)

ateucher:
Two tests added for #2200 :
- One where an ordered factor is created inside `summarize` (order gets dropped)
- One where `min` and `max` are calculated on an already existing ordered factor column (order gets dropped)

### DIFF
--- a/tests/testthat/test-summarise.r
+++ b/tests/testthat/test-summarise.r
@@ -743,30 +743,25 @@ test_that("typing and NAs for rowwise summarise (#1839)", {
 })
 
 test_that("calculating an ordered factor preserves order (#2200)", {
-  test_df <- data_frame(id = rep(letters[24:26], 2),
-                 val = runif(6, 0, 4))
-
-  make_levels <- function(x) {
-    cut(x, breaks = 0:4, labels = LETTERS[1:4], ordered_result = TRUE)
-  }
+  test_df <- tibble(id = c("a", "b"), val = 1:2)
 
   ret <- group_by(test_df, id) %>%
-    summarize(mean_val = mean(val),
-              level = make_levels(mean_val))
-  expect_is(ret$level, "ordered")
-  expect_equal(levels(ret$level), levels(make_levels(1)))
+    summarize(level = ordered(val))
+
+  expect_s3_class(ret$level, "ordered")
+  expect_equal(levels(ret$level), c("1","2"))
 })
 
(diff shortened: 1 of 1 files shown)


ateucher:
#2200

krlmlr:
Thanks! If you add skip() calls, we can merge right away. @hadley: OK?

hadley:
I'd like the tests to be a little simpler. For example, the first one could be:

```R
df <- tibble(id = c("a", "b"), x = 1:2) 

out <- df %>% 
  group_by(id) %>% 
  summarise(y = ordered(x))

expect_s3_class(out$y, "ordered")
expect_equal(levels(out$y), c("1", "2"))
```

(also note the use of `expect_s3_class()` which is now preferred to `expect_is()`)

ateucher:
Sure, will do

ateucher:
Ok, done. What kind of `skip()` message do you want?

krlmlr:
I'm fine with "Currently failing", the issue number is given in the testthat() call.

krlmlr:
LGTM.

krlmlr:
@hadley: Do you want to review once more?

tests/testthat/test-summarise.r
```diff
@@ -741,3 +741,29 @@ test_that("typing and NAs for rowwise summarise (#1839)", {
...
+  ret <- group_by(test_df, id) %>%
+    summarize(level = ordered(val))
+
+  expect_s3_class(ret$level, "ordered")
+  expect_equal(levels(ret$level), c("1","2"))
+})
+
+test_that("min, max preserves ordered factor data  (#2200)", {
+  skip("Currently failing")
+  test_df <- tibble(id = rep(c("a","b"), 2),
```
hadley: Could you please use my standard indenting rules?

ateucher: I think this is what you mean?

hadley: It should look like:

```R
tibble(
  id = ..,
  ord = ...
)
```

so each new variable starts on it's own line, indented by two spaces (similarly for the `summarize()` call below)

ateucher:
@hadley @krlmlr hopefully that should do it

hadley:
LGTM - thanks!

krlmlr:
Thanks!

PR #2239, opened 2016-11-08
added a test case addressing issue 2228 (join hangs on empty suffix a…

simon-anders:
…rgument)

See https://github.com/hadley/dplyr/issues/2228

At the moment, the test case added here cause R to hang in and end-less loop, then crashing due to stack overflow. So, better pull this request only after you fixed the issue. ;)

### DIFF
--- a/tests/testthat/test-joins.r
+++ b/tests/testthat/test-joins.r
@@ -614,6 +614,8 @@ test_that( "left_join handles mix of encodings in column names (#1571)", {
 })
 
 test_that("can handle empty string in suffix argument, left side (#2228, #2182, #2007)", {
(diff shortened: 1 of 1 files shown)


krlmlr:
Thanks. We can merge right away if you add a skip(). Does it also hang with an anti_join() and semi_join()?

simon-anders:
I added the skip.

anti_join and semi_join should be fine: they don;t have a 'suffix' argument

krlmlr:
@hadley: PTAL.

hadley:
LGTM

krlmlr:
Thanks!

Issue #1892, opened 2016-06-07
Integer not automatically coerced to double


drknexus:
This might be a won't fix, but I thought I'd float the balloon anyway.

Consider:

```
library(dplyr)
data_frame(
  id = c(1,2,3,4,5,6),
  value = c(1L,2L,3L,NA,NA,NA),
  group = c("A","A","A","B","B","B")
) %>% 
  group_by(group) %>%
  mutate(value = ifelse(is.na(value),0,value))
```

Which fails.  

Versus:

```
data_frame(
  id = c(1,2,3,4,5,6),
  value = c(1L,2L,3L,4,NA,NA),
  group = c("A","A","A","B","B","B")
) %>% 
  mutate(value = ifelse(is.na(value),0,value)) %>%
  group_by(group)
```

Which does not fail.

As a matter of linguistic parsing one might not expect different outcomes based on ordering.  Is there a reason dplyr can't coerce the returns from the group_by operation to share in the most general common data type (as done elsewhere in R) so that both functional orders yield successful results?

hadley:
Both work in the dev version.

drknexus:
That doesn't seem to be entirely true.

```
ubuntu@ip-10-0-0-75:~$ R --vanilla

R version 3.3.0 (2016-05-03) -- "Supposedly Educational"
Copyright (C) 2016 The R Foundation for Statistical Computing
Platform: x86_64-pc-linux-gnu (64-bit)

R is free software and comes with ABSOLUTELY NO WARRANTY.
You are welcome to redistribute it under certain conditions.
Type 'license()' or 'licence()' for distribution details.

  Natural language support but running in an English locale

R is a collaborative project with many contributors.
Type 'contributors()' for more information and
'citation()' on how to cite R or R packages in publications.

Type 'demo()' for some demos, 'help()' for on-line help, or
'help.start()' for an HTML browser interface to help.
Type 'q()' to quit R.

> library(dplyr)

Attaching package: ‘dplyr’

The following objects are masked from ‘package:stats’:

    filter, lag

The following objects are masked from ‘package:base’:

    intersect, setdiff, setequal, union

> library(dplyr)
> data_frame(
+   id = c(1,2,3,4,5,6),
+   value = c(1L,2L,3L,NA,NA,NA),
+   group = c("A","A","A","B","B","B")
+ ) %>%
+   group_by(group) %>%
+   mutate(value = ifelse(is.na(value),0,value))
Error: incompatible types, expecting a integer vector
[...]

PR #2249, opened 2016-11-11
Add Failing Test for #1892 and a Related Failing Test

drknexus:
This PR adds the failing test associated with https://github.com/hadley/dplyr/issues/1892 (`group_by coerces integer + double -> double (#1892)`) and another (apparently related) failing test (`group_by coerces factor + character -> character (WARN)`).  Please confirm that the second failing test is indeed an issue and not a misunderstanding on my part before merging.

### DIFF
--- a/tests/testthat/test-group-by.r
+++ b/tests/testthat/test-group-by.r
@@ -276,11 +276,11 @@ test_that("grouped mutate coerces factor + character -> character (WARN)", {
     id = c(1, 4),
     value = factor(c("blue", NA)),
     group = c("A", "B")
-  )
+  ) %>%
+    group_by(group)
 
   expect_warning(
     df <- df %>%
-      group_by(group) %>%
       mutate(value = ifelse(id > 3, as.character("foo"), value))
   )
 
--- a/tests/testthat/test-group-by.r
+++ b/tests/testthat/test-group-by.r
@@ -260,9 +260,9 @@ test_that("rowwise fails gracefully on raw columns (#1803)", {
 test_that("group_by coerces integer + double -> double (#1892)", {
   skip("Currently failing")
   data_frame(
-    id = c(1,2,3,4,5,6),
-    value = c(1L,2L,3L,NA,NA,NA),
-    group = c("A","A","A","B","B","B")
+    id = c(1, 2, 3, 4, 5, 6),
+    value = c(1L, 2L, 3L, NA, NA, NA),
+    group = c("A", "A", "A", "B", "B", "B")
   ) %>%
   group_by(group) %>%
   mutate(value = ifelse(is.na(value),double(0),value))
--- a/tests/testthat/test-group-by.r
+++ b/tests/testthat/test-group-by.r
@@ -265,5 +265,6 @@ test_that("group_by coerces integer + double -> double (#1892)", {
     group = c("A", "A", "A", "B", "B", "B")
   ) %>%
   group_by(group) %>%
-  mutate(value = ifelse(is.na(value),double(0),value))
+  mutate(value = ifelse(is.na(value), as.double(0), value))
+})
 })
--- a/tests/testthat/test-group-by.r
+++ b/tests/testthat/test-group-by.r
@@ -257,7 +257,7 @@ test_that("rowwise fails gracefully on raw columns (#1803)", {
   expect_error( rowwise(df), "unsupported type" )
 })
 
-test_that("group_by coherses types to most general common data type (#1892)", {
+test_that("group_by coherses integer + double -> double (#1892)", {
   data_frame(
     id = c(1,2,3,4,5,6),
     value = c(1L,2L,3L,NA,NA,NA),
--- a/tests/testthat/test-group-by.r
+++ b/tests/testthat/test-group-by.r
@@ -256,33 +256,3 @@ test_that("rowwise fails gracefully on raw columns (#1803)", {
   df <- data_frame(a = 1:3, b = as.raw(1:3))
   expect_error( rowwise(df), "unsupported type" )
 })
-
-test_that("grouped mutate coerces integer + double -> double (#1892)", {
-  skip("Currently failing")
-  df <- data_frame(
-    id = c(1, 4),
-    value = c(1L, NA),
-    group = c("A", "B")
-  ) %>%
-    group_by(group) %>%
-    mutate(value = ifelse(is.na(value), as.double(0), value))
-  expect_type(df$value, "double")
-})
-
-test_that("grouped mutate coerces factor + character -> character (WARN) (#1892)", {
-  skip("Currently failing")
-
-  df <- data_frame(
-    id = c(1, 4),
-    value = factor(c("blue", NA)),
-    group = c("A", "B")
-  ) %>%
-    group_by(group)
-
-  expect_warning(
-    df <- df %>%
-      mutate(value = ifelse(id > 3, as.character("foo"), value))
-  )
-
-  expect_type(df$value, "character")
-})
--- a/tests/testthat/test-group-by.r
+++ b/tests/testthat/test-group-by.r
@@ -266,7 +266,7 @@ test_that("group_by coerces integer + double -> double (#1892)", {
   ) %>%
     group_by(group) %>%
     mutate(value = ifelse(is.na(value), as.double(0), value))
-  expect_equal(class(df$value), "double")
+  expect_equal(typeof(df$value), "double")
 })
 
 test_that("group_by coerces factor + character -> character (WARN)", {
@@ -280,5 +280,5 @@ test_that("group_by coerces factor + character -> character (WARN)", {
     group_by(group) %>%
     mutate(value = ifelse(id > 3, as.character("foo"), value))
   )
-  expect_equal(class(df$value), "character")
+  expect_equal(typeof(df$value), "character")
 })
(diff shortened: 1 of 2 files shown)


krlmlr:
@hadley: PTAL.

tests/testthat/test-group-by.r
```diff
@@ -256,3 +256,29 @@ test_that("rowwise fails gracefully on raw columns (#1803)", {
   df <- data_frame(a = 1:3, b = as.raw(1:3))
   expect_error( rowwise(df), "unsupported type" )
 })
+
+test_that("group_by coerces integer + double -> double (#1892)", {
```
hadley: `group_by` doesn't coerce anything. Do you mean `mutate()`?

tests/testthat/test-group-by.r
```diff
@@ -256,3 +256,29 @@ test_that("rowwise fails gracefully on raw columns (#1803)", {
...
+  skip("Currently failing")
+  df <- data_frame(
+    id = c(1, 2, 3, 4, 5, 6),
+    value = c(1L, 2L, 3L, NA, NA, NA),
+    group = c("A", "A", "A", "B", "B", "B")
+  ) %>%
+    group_by(group) %>%
+    mutate(value = ifelse(is.na(value), as.double(0), value))
```
hadley: Could you illustrate the same problem with just a single observation per group?

tests/testthat/test-group-by.r
```diff
@@ -256,3 +256,29 @@ test_that("rowwise fails gracefully on raw columns (#1803)", {
...
+  expect_equal(typeof(df$value), "double")
```
hadley: `expect_type()` would be safer

tests/testthat/test-group-by.r
```diff
@@ -256,3 +256,29 @@ test_that("rowwise fails gracefully on raw columns (#1803)", {
...
+})
+
+test_that("group_by coerces factor + character -> character (WARN)", {
+  skip("Currently failing")
+  expect_warning(
```
hadley: The smaller the block of code that `expect_warning()` surrounds the better.

tests/testthat/test-group-by.r
drknexus: Well a grouped mutate.  If the mutate weren't grouped we wouldn't have the problem.

tests/testthat/test-group-by.r
drknexus: Sure. Can that be done without breaking up the chain?  Which function do you expect to throw the warning?

tests/testthat/test-group-by.r
drknexus: Terminology updated in d9bb142

tests/testthat/test-group-by.r
drknexus: Sure.  Updated in d9bb142.

tests/testthat/test-group-by.r
drknexus: I now use `expect_type()` in d9bb142.  Why is it 'safer'?

tests/testthat/test-group-by.r
drknexus: Minor break in chain and reduced area covered by the expect_warning() test (ff38827).  This should leave you free to decide if group_by or mutate is responsible for throwing the anticipated warning.

tests/testthat/test-group-by.r
```diff
@@ -256,3 +256,33 @@ test_that("rowwise fails gracefully on raw columns (#1803)", {
...
+    id = c(1, 4),
+    value = c(1L, NA),
+    group = c("A", "B")
+  ) %>%
+    group_by(group) %>%
+    mutate(value = ifelse(is.na(value), as.double(0), value))
+  expect_type(df$value, "double")
+})
+
+test_that("grouped mutate coerces factor + character -> character (WARN)", {
```
krlmlr: Please add the issue number to the test description, too.

tests/testthat/test-group-by.r
```diff
@@ -256,3 +256,33 @@ test_that("rowwise fails gracefully on raw columns (#1803)", {
...
+  df <- data_frame(
+    id = c(1, 4),
+    value = factor(c("blue", NA)),
+    group = c("A", "B")
+  )
+
+  expect_warning(
+    df <- df %>%
+      group_by(group) %>%
+      mutate(value = ifelse(id > 3, as.character("foo"), value))
```
krlmlr: I think the test should be precise about which verb throws the warning. I guess it's `mutate()`, so you might need to define `grouped_df` outside of `expect_warning()`.

tests/testthat/test-group-by.r
drknexus: That one doesn't have a issue number per se.  Do you want me to comment on in the context of #1892 or open a separate issue?

tests/testthat/test-group-by.r
drknexus: 'Guess' is key here.  I don't think it is knowable on either of our parts which of the two verbs will throw the warning until someone writes the fix.  

mutate() would have to know about the results of mutate run on other groups to throw the warning.  It seems more plausible for me that you'd put the conversion and checks in group_by itself.  

Regardless, it seems to me that the author of the fix can adjust this test to be more specific after the fix is written.

Respectfully I'm ask that you reconsider and accept this part as writ or escalate your 'guess' to an assertion.  If you or another dplyr contributor are willing to assert that mutate should be the warning throwing party, then the test can validly be likewise opinionated.

hadley: It is knowable - `group_by()` only sets up the indices. It doesn't do any computation by itself.

tests/testthat/test-group-by.r
krlmlr: Please refer to this PR.

tests/testthat/test-group-by.r
drknexus: Alright, in that case does this test belong under the mutate tests rather than the `group_by` tests?

krlmlr: Thanks. Yes, the mutate tests are a better home for these tests.

drknexus: @krlmlr Okay, moved in c3ee624.

tests/testthat/test-group-by.r
drknexus: Done in 47a7897.  Are there any other hanging threads that need to be addressed?

drknexus:
@krlmlr I believe all requested changes have been made.  Would you please re-review?

tests/testthat/test-mutate.r
```diff
@@ -555,6 +555,33 @@ test_that("grouped mutate errors on incompatible column type (#1641)", {
...
+test_that("grouped mutate coerces integer + double -> double (#1892)", {
+  skip("Currently failing")
+  df <- data_frame(
+    id = c(1, 4),
+    value = c(1L, NA),
+    group = c("A", "B")
+  ) %>%
+    group_by(group) %>%
+    mutate(value = ifelse(is.na(value), as.double(0), value))
+  expect_type(df$value, "double")
```
krlmlr: Thanks. Perhaps we should use expect_identical(..., c(1, 0)) ; similar below.

drknexus: Whereas the issue is really about variable types, shouldn't the syntax of the test should be explicit about the type expectation rather than have it simply be an implicit part of an expect_identical comparison?

drknexus: I'll go ahead and add the `expect_identical` comparison in addition to the `expect_type`.  The additional check doesn't seem harmful.

drknexus: @krlmlr Ready for another review pass I believe.

tests/testthat/test-mutate.r
```diff
@@ -555,6 +555,35 @@ test_that("grouped mutate errors on incompatible column type (#1641)", {
...
 
+test_that("grouped mutate coerces integer + double -> double (#1892)", {
+  skip("Currently failing")
+  df <- data_frame(
+    id = c(1, 4),
+    value = c(1L, NA),
+    group = c("A", "B")
+  ) %>%
+    group_by(group) %>%
+    mutate(value = ifelse(is.na(value), as.double(0), value))
```
hadley: The `as.double()` here and below is unnecessary. I'd expect readers of dplyr's test to know that numbers in R are doubles by default.

tests/testthat/test-mutate.r
hadley: I agree that an explicit test of type is useful here

tests/testthat/test-mutate.r
```diff
@@ -555,6 +555,35 @@ test_that("grouped mutate errors on incompatible column type (#1641)", {
...
+  skip("Currently failing")
+  df <- data_frame(
+    id = c(1, 4),
+    value = factor(c("blue", NA)),
+    group = c("A", "B")
+  ) %>%
+    group_by(group)
+  expect_warning(
+    df <- df %>%
+      mutate(value = ifelse(id > 3, as.character("foo"), value))
```
hadley: Similarly the `as.character()` here

hadley:
Couple of tiny comments, otherwise LGTM

drknexus:
@hadley I believe the above referenced comments have now been addressed in https://github.com/hadley/dplyr/pull/2249/commits/7751c0bf9e09ab754262517440057195d1ca5b0a.

hadley:
LGTM

krlmlr:
Thanks!

Issue #870, opened 2015-01-06
Rethink database testing


hadley:
Since after rewrite we should be able to load RMySQL, RPostgreSQL and RSQLite simultaneously

krlmlr:
Works best with rstats-db/RMySQL#90 and changing default values for `src_mysql()` to `NULL`.

hadley:
This was done a while back — see `test_load()` etc. It's not perfect, but it's a reasonable start.

PR #2263, opened 2016-11-23
fix and test for issue #2256

### LINKED ISSUE #2256: error when writing character fields containing NAs to mysql table

eduardgrebe:
1. encodeString before checking length of character vectors when writing to mysql source

1. rewrite "<NA>" to NA when reading from database tables with SQL

1. test that dataframe containing NAs in character columns writes correctly to database sources

Fixes #2256.

### DIFF
--- a/R/data-temp.r
+++ b/R/data-temp.r
@@ -20,13 +20,15 @@ NULL
 #' @export
 #' @rdname testing
 test_register_src <- function(name, src) {
-  message("Registering testing src: ", name, " ", appendLF = FALSE)
+  message("Registering testing src: ", name)
   tryCatch(
     {
       test_srcs$add(name, src)
-      message("OK")
+      message("OK: ", name, " registered")
     },
-    error = function(e) message(conditionMessage(e))
+    error = function(e) {
+      message("Failed: ", name, " not registered")
+    }
   )
 }
 
--- a/R/src-mysql.r
+++ b/R/src-mysql.r
@@ -145,7 +145,9 @@ db_has_table.MySQLConnection <- function(con, table, ...) {
 #' @export
 db_data_type.MySQLConnection <- function(con, fields, ...) {
   char_type <- function(x) {
-    n <- max(nchar(as.character(x), "bytes"))
+    x <- as.character(x)
+    x <- vapply(x, encodeString, character(1))
+    n <- max(nchar(x), "bytes")
     if (n <= 65535) {
       paste0("varchar(", n, ")")
     } else {
--- a/R/src-mysql.r
+++ b/R/src-mysql.r
@@ -145,9 +145,7 @@ db_has_table.MySQLConnection <- function(con, table, ...) {
 #' @export
 db_data_type.MySQLConnection <- function(con, fields, ...) {
   char_type <- function(x) {
-    x <- as.character(x)
-    x <- vapply(x, encodeString, character(1))
-    n <- max(nchar(x), "bytes")
+    n <- max(na.omit(nchar(x, "bytes")))
     if (n <= 65535) {
       paste0("varchar(", n, ")")
     } else {
@@ -183,23 +181,17 @@ db_commit.MySQLConnection <- function(con, ...) {
 #' @export
 db_insert_into.MySQLConnection <- function(con, table, values, ...) {
 
-  # Convert factors to strings
-  is_factor <- vapply(values, is.factor, logical(1))
-  values[is_factor] <- lapply(values[is_factor], as.character)
+  if (nrow(values) == 0)
+    return(NULL)
 
-  # Encode special characters in strings
-  is_char <- vapply(values, is.character, logical(1))
-  values[is_char] <- lapply(values[is_char], encodeString)
+  cols <- lapply(values, escape, collapse = NULL, parens = FALSE, con = con)
+  col_mat <- matrix(unlist(cols, use.names = FALSE), nrow = nrow(values))
 
-  tmp <- tempfile(fileext = ".csv")
-  utils::write.table(values, tmp, sep = "\t", quote = FALSE, qmethod = "escape",
-    row.names = FALSE, col.names = FALSE)
+  rows <- apply(col_mat, 1, paste0, collapse = ", ")
+  values <- paste0("(", rows, ")", collapse = "\n, ")
 
-  sql <- build_sql("LOAD DATA LOCAL INFILE ", encodeString(tmp), " INTO TABLE ",
-    ident(table), con = con)
+  sql <- build_sql("INSERT INTO ", sql(table), " VALUES ", sql(values))
   dbGetQuery(con, sql)
-
-  invisible()
 }
 
 #' @export
--- a/tests/testthat/helper-src.R
+++ b/tests/testthat/helper-src.R
@@ -15,10 +15,3 @@ skip_if_no_sqlite <- function() {
   if (!test_srcs$has("sqlite"))
     skip("No SQLite")
 }
-
-skip_if_no_mysql <- function() {
-  if (!test_srcs$has("mysql"))
-    skip("No MySQL")
-}
-
-print(test_srcs)
--- a/tests/testthat/helper-src.R
+++ b/tests/testthat/helper-src.R
@@ -5,11 +5,20 @@ if (identical(Sys.info()[["user"]], "hadley")) {
   test_register_src("postgres", src_postgres("test", host = "localhost"))
 } else if (identical(Sys.getenv("TRAVIS"), "true")) {
   test_register_src("postgres", src_postgres("test", user = "travis", password = ""))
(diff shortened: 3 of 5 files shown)


R/src-mysql.r
```diff
@@ -145,7 +145,9 @@ db_has_table.MySQLConnection <- function(con, table, ...) {
 #' @export
 db_data_type.MySQLConnection <- function(con, fields, ...) {
   char_type <- function(x) {
-    n <- max(nchar(as.character(x), "bytes"))
+    x <- as.character(x)
+    x <- vapply(x, encodeString, character(1))
+    n <- max(nchar(x), "bytes")
```
krlmlr: Wouldn't it be easier to check max(nchar(x[!is.na(x)], "bytes")) ? Please double-check the position of the "bytes" argument.

tests/testthat/helper-src.R
```diff
@@ -5,9 +5,11 @@ if (identical(Sys.info()[["user"]], "hadley")) {
   test_register_src("postgres", src_postgres("test", host = "localhost"))
 } else if (identical(Sys.getenv("TRAVIS"), "true")) {
   test_register_src("postgres", src_postgres("test", user = "travis", password = ""))
+} else if (identical(Sys.info()[["user"]], "eduardgrebe")) {
```
krlmlr: This doesn't look right to me.

tests/testthat/test-tbl-sql.r
```diff
@@ -9,3 +9,11 @@ test_that("can generate sql tbls with raw sql", {
     expect_equal(collect(tbl), collect(clone))
   }
 })
+
+test_that("NAs in character fields handled by db sources", {
+  df <- data.frame(x=c("a","aa",NA), stringsAsFactors=FALSE)
```
krlmlr: Please put a space after commas and around = . A test for NA first and NA in the middle (perhaps in other columns of this data frame) would be great, too.

tests/testthat/helper-src.R
```diff
@@ -5,9 +5,11 @@ if (identical(Sys.info()[["user"]], "hadley")) {
...
+  test_register_src("postgres", src_postgres("test", host = "localhost", user = "eduardgrebe", password = ""))
 }
 
-
+test_register_src("mysql", src_mysql("test", host = "localhost"))
```
krlmlr: What happens if the MySQL server is not available? It would be great to get a warning (not an error) in this case.

krlmlr:
Thanks. Just a few notes; tests seem to fail on Travis now, too -- we might need to set up MySQL testing on Travis along the lines of what [RMySQL does](https://github.com/rstats-db/RMySQL/blob/master/.travis.yml) (or simply deactivate MySQL on Travis for now).

R/src-mysql.r
eduardgrebe: That's what I did initially, but the encodeString applied to the column in the function that writes to the table with SQL results in the string "<NA>" being written - which may be the longest string in the vector - which causes a failure if there are no other strings longer than 4 characters.

tests/testthat/helper-src.R
eduardgrebe: Not sure how to address this, the ifs are false on my machine, meaning no postgres source is initialised.

tests/testthat/test-tbl-sql.r
eduardgrebe: Good point.

R/src-mysql.r
krlmlr: I don't think we should be writing `"<NA>"` for NA values, we should use a SQL `NULL`.

tests/testthat/helper-src.R
krlmlr: Maybe adapt `test_register_src()` to warn instead of throwing an error? Not sure what the implications are.

eduardgrebe:
The main worry I have is that as currently implemented with encodeString in  (I left this unchanged), the string `"<NA>"` is what is stored in the db table, rather than true NULLs. This feels odd to me. Should we be checking for NA values and ensure that NULLs are stored instead?

R/src-mysql.r
eduardgrebe: snap. agreed.

eduardgrebe:
I've changed db_insert_into to work like on postgres - NAs now become NULL. Errors are already caught in test_register_src - made the messages clearer, but it shouldn't produce an error when a db engine is not available.

krlmlr:
Thanks. The testthat errors are unrelated (RcppCore/Rcpp#596), but could you please take a look at the other R CMD check messages that Travis shows: https://travis-ci.org/hadley/dplyr/jobs/178453180#L1153 .

R/src-mysql.r
```diff
@@ -145,7 +145,7 @@ db_has_table.MySQLConnection <- function(con, table, ...) {
 #' @export
 db_data_type.MySQLConnection <- function(con, fields, ...) {
   char_type <- function(x) {
-    n <- max(nchar(as.character(x), "bytes"))
+    n <- max(na.omit(nchar(x, "bytes")))
```
krlmlr: max(..., na.rm = TRUE) is slightly better.

R/src-mysql.r
```diff
@@ -181,23 +181,17 @@ db_commit.MySQLConnection <- function(con, ...) {
...
 
-  tmp <- tempfile(fileext = ".csv")
-  utils::write.table(values, tmp, sep = "\t", quote = FALSE, qmethod = "escape",
-    row.names = FALSE, col.names = FALSE)
+  rows <- apply(col_mat, 1, paste0, collapse = ", ")
+  values <- paste0("(", rows, ")", collapse = "\n, ")
 
-  sql <- build_sql("LOAD DATA LOCAL INFILE ", encodeString(tmp), " INTO TABLE ",
-    ident(table), con = con)
+  sql <- build_sql("INSERT INTO ", sql(table), " VALUES ", sql(values))
```
krlmlr: Why are you switching to `INSERT INTO`? Wouldn't that be a lot slower, especially without prepared queries?

I think you could use `\N` as a placeholder for NULL values: http://dev.mysql.com/doc/refman/5.7/en/load-data.html.

R/data-temp.r
```diff
@@ -20,13 +20,15 @@ NULL
 #' @export
 #' @rdname testing
 test_register_src <- function(name, src) {
-  message("Registering testing src: ", name, " ", appendLF = FALSE)
+  message("Registering testing src: ", name)
```
krlmlr: Sorry, I forgot how test_register_src() works. The original implementation looks okay to me, I think.

- Doesn't fail
- Prints success and failure messages
- Output: one line only

tests/testthat/helper-src.R
```diff
@@ -1,13 +1,13 @@
 test_register_src("df", src_df(env = new.env(parent = emptyenv())))
 test_register_src("sqlite", src_sqlite(":memory:", create = TRUE))
 
-if (identical(Sys.info()[["user"]], "hadley")) {
-  test_register_src("postgres", src_postgres("test", host = "localhost"))
-} else if (identical(Sys.getenv("TRAVIS"), "true")) {
+if (identical(Sys.getenv("TRAVIS"), "true")) {
```
krlmlr: Could you please leave this unchanged?

krlmlr:
I've updated master to work around the Rcpp problem, could you please merge/rebase so that testthat tests pass?

krlmlr:
Sorry: I forgot #1975 already resolves this problem. The test, and the enabling of MySQL tests (#870), are still very useful.

krlmlr:
Thanks. I've added your test to #2381.

I see that the current implementation of `db_insert_into()` doesn't work with the development version of RMySQL, due to (at least) rstats-db/RMySQL#117. But this will be fixed eventually, and it currently works with the stable version of RMySQL. I'd rather not switch to using `INSERT INTO`.

PR #2282, opened 2016-12-01
setOldClass to allow S4 to work with grouped_df better

joranE:
I just whipped this up very fast and modeled it after what it appears was done in tibble, I hope I got it right.


R/dataframe.R
```diff
@@ -1,3 +1,5 @@
+methods::setOldClass(c("grouped_df", "tbl_df", "data.frame"))
```
krlmlr: Technically, we should declare the "methods" import in `DESCRIPTION`. I can only assume that the check didn't fail because we have it in "Suggests".

joranE: As usual, I think I totally borked the git commands for trying to add the methods import in `DESCRIPTION`. If you're willing to be patient and bear with my stumbling around in git I will try to fix it.

joranE: Ok, I think I managed to patch up that hot mess I created. Phew.

DESCRIPTION
```diff
@@ -14,6 +14,7 @@ BugReports: https://github.com/hadley/dplyr/issues
 Depends: 
     R (>= 3.1.2)
 Imports: 
+    methods,
```
krlmlr: Is methods still in "Suggests"?

joranE: Sigh.

krlmlr:
Thanks, LGTM.

hadley:
LGTM

krlmlr:
Thanks!

PR #2285, opened 2016-12-02
Stop converting colnames to UTF-8 in select()

### LINKED ISSUE #2284: select() mistakenly converts colnames to UTF-8
I've faced this behavior with R 3.3.2 and dplyr 0.5.0 on my Windows machine.

```r
d <- data.frame("φ" = 1:10)

Encoding(colnames(d))
#> [1] "unknown"

d2 <- d %>% select(φ)
Encoding(colnames(d2))
#> [1] "UTF-8"
```

This is due to this tricky behavior of `c()` function in the base R; The names of named vectors are converted to UTF-8 accidentally.

```r
x <- "φ"
names(x) <- "φ"

Encoding(names(x))
#> [1] "unknown"

Encoding(names(c(x)))
#> [1] "UTF-8"
```

This problem itself is not so problematic and not dplyr's fault, but until https://github.com/hadley/dplyr/issues/1950 is solved we will continue to see the error with functions like `group_by()` and `distict()`. (e.g. #2277, #2005)

```r
d %>%
  select(φ) %>%
  group_by(φ)
#> Error in grouped_df_impl(data, unname(vars), drop) : unknown column 'φ' 
```

So, could you consider to stop using `c()` here?: https://github.com/hadley/dplyr/blob/1ba25c03826372f41e7d7b850a174d881f0ac15b/R/select-vars.R#L86

yutannihilation:
This PR fixes #2284 by using the temporal workaround for `c()`'s tricky behavior. I am sure that [vctrs](https://github.com/hadley/vctrs) will provide some nice function for this, right? 😄 But, for now, we need some tweak like this.

### DIFF
--- a/R/select-vars.R
+++ b/R/select-vars.R
@@ -83,7 +83,7 @@ select_vars_ <- function(vars, args, include = character(), exclude = character(
 
   # Include/exclude specified variables
   sel <- setNames(vars[incl], names(incl))
-  sel <- c(setdiff2(include, sel), sel)
+  sel <- combine_named_vector(setdiff2(include, sel), sel)
   sel <- setdiff2(sel, exclude)
 
   # Ensure all output vars named
@@ -100,6 +100,11 @@ select_vars_ <- function(vars, args, include = character(), exclude = character(
 
 setdiff2 <- function(x, y) {
   x[match(x, y, 0L) == 0L]
+}
+
+# workaround for the problem that `c` converts names to UTF-8
+combine_named_vector <- function(x, y) {
+  setNames(c(x, y), c(names(x), names(y)))
 }
 
 #' @export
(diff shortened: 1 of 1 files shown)


krlmlr:
This should be unnecessary now after #2382.

yutannihilation:
@krlmlr Thanks!!! Now I close this. But, I feel I'm still seeing the problem with `group_by()`... I will confirm and report later today or tomorrow.

yutannihilation:
Ah, I got it. #2382 seems perfectly fine. Still, the problem occurs since now dplyr somehow skips setting Encoding of colnames as UTF-8.

krlmlr:
@yutannihilation: This is worrisome. Do you have a reprex? The locale setting ("LC_CTYPE") and OS (Windows?) is important.

Issue #1176, opened 2015-05-28
select(): negation of non-match gives no vars, instead of all vars


gunapemmaraju:
In dplyr, I want to exclude columns which contain the word "junk" but, there may not be any column that contain the word "junk". In that case, dplyr should return all columns. But it returns none. See unit test case below.

```R
df<-data.frame(name=paste("name",1:5), age=1:5)
str(df)
 'data.frame': 5 obs. of  2 variables:
 $ name: Factor w/ 5 levels "name 1","name 2",..: 1 2 3 4 5
 $ age : int  1 2 3 4 5
df1<-df%>%select(-contains("junk"))
str(df1)
 'data.frame': 5 obs. of  0 variables
```

More at

http://stackoverflow.com/questions/30496930/in-dplyr-select-with-a-drop-does-not-work

blasern:
This is because `contains` returns `integer(0)` if it does not find any match. `-integer(0)` is still `integer(0)`. An easy fix would be to return `-seq_along(vars)` instead.  

    contains <- function (vars, match, ignore.case = TRUE) 
      {
        stopifnot(is.string(match), nchar(match) > 0)
        if (ignore.case) {
          vars <- tolower(vars)
          match <- tolower(match)
        }
        res <- grep(match, vars, fixed = TRUE)
        if (length(res) == 0) res <- -seq_along(vars)
        res
      }

The same problem occurs also in `starts_with`, `ends_with`, `contains` and `matches`.

drewgendreau:
Just wanted to add that I tried doing this earlier today and was surprised to find that 

df %>% select(-contains("this_string_is_definitely_not_in_any_of_the_column_names") )

returns an empty data.frame.  The use case is that I'm running a program where I don't know all the column names ahead of time, but if there is one with a particular string I'd like to drop it.

PR #2289, opened 2016-12-04
change initial columns selected based on first selector (fix #2275)

### LINKED ISSUE #2264: No tolerance in dplyr's `all.equal.tbl_df()`
I've run into an issue where the calling `all.equal()` (which then get dispatched to dplyr or tibble's `all_equal()`) on a tibble with dplyr attached does not have any tolerance for minorly different values like `all.equal()` on a dataframe has. However, with just tibble attached, minor differences are ignored. In both cases, converting to a dataframe results in the expected match with tolerance. This was discussed before [#332](https://github.com/hadley/dplyr/issues/332), and I can understand if the dplyr version does an exact comparison for speed's sake, but this behavior took me some time to diagnose exactly what's going on.

```R
library(tibble)
tbl1 <- tibble(foo=1, bar=2)
tbl2 <- tibble(foo=(1)-1e-16, bar=2)

all.equal(data.frame(tbl1), data.frame(tbl2)) # TRUE
all.equal(tbl1, tbl2) # TRUE

library(dplyr)
all.equal(data.frame(tbl1), data.frame(tbl2)) # TRUE
all.equal(tbl1, tbl2) # does not work
```

I'm not sure if the best course of action would be adding the tolerance into the dplyr version of `all_equal()` or just clarifying in the documentation that the behavior of tibble's `all_equal()` and dplyr's `all_equal()` are different. (This is with R 3.3.2, Tibble 1.2, and dplyr 0.5.0)

### LINKED ISSUE #2275: select helpers dropping all before FALSE entry
The "select helper" functions appear to be a bit too aggressive about *excluding* everything before it. This may be a documentation thing, but I expect positive helper functions to be *additive* (which appears to be the case), and "none-found" helper functions to *not be subtractive*.

Note: in all examples, I'm using `summarize_at` (and `mean)` primarily for terse output. The behavior is the same with other helpers (e.g., `matches`, `starts_with`) and within other verbs (e.g., `select`, `mutate_at`).

## Currently-Correct Behavior

For the record, "all matches found" works fine, as does a not-found match when it is the first one:

```r
### expecting am, gear, vs
summarize_at(mtcars, vars(contains("am"), contains("gear"), contains("vs")), funs(mean))
#        am   gear     vs
# 1 0.40625 3.6875 0.4375

### expecting gear, vs
summarize_at(mtcars, vars(contains("FOO"), contains("gear"), contains("vs")), funs(mean))
#     gear     vs
# 1 3.6875 0.4375
```

## Expected Behavior

I have been interpreting `select(mtcars, contains("am"), contains("foo"))` to return all columns that contain `"am"` or `"foo"`, regardless of the order I provide the calls to `contains` (and other helpers).

## Errant Behavior

Everything *before* a failing conditional is dropped, regardless of match:

```r
### expecting am, gear
summarize_at(mtcars, vars(contains("am"), contains("gear"), contains("FOO")), funs(mean))
# data frame with 0 columns and 0 rows

### expecting am, vs
summarize_at(mtcars, vars(contains("am"), contains("FOO"), contains("vs")), funs(mean))
#       vs
# 1 0.4375
```

This appears to be because the helper functions, when nothing is found, return the negative vector of column indices:

```r
contains("a", vars = colnames(mtcars))
# [1]  5  9 10 11
contains("FOO", vars = colnames(mtcars))
#  [1]  -1  -2  -3  -4  -5  -6  -7  -8  -9 -10 -11
```

r2evans:
Fixes #2275.


tests/testthat/test-select-helpers.R
```diff
@@ -59,3 +59,38 @@ test_that("one_of tolerates but warns for unknown variables", {
 test_that("one_of converts names to positions", {
   expect_equal(one_of("a", "z", vars = letters), c(1L, 26L))
 })
+
+# first-selector ----------------------------------------------------------
+
+test_that("mid-selector fails don't clear previous selectors (issue #2264)", {
```
krlmlr: Please double-check issue number.

tests/testthat/test-select-helpers.R
```diff
@@ -59,3 +59,38 @@ test_that("one_of tolerates but warns for unknown variables", {
 test_that("one_of converts names to positions", {
   expect_equal(one_of("a", "z", vars = letters), c(1L, 26L))
 })
+
+# first-selector ----------------------------------------------------------
```
krlmlr: Do we still have explicit tests for #1176? We might also need to test combination with inclusion and exclusion of single columns ( `data %>% select(starts_with(...), column, -column2)` ).

R/select-vars.R
```diff
@@ -66,8 +66,11 @@ select_vars_ <- function(vars, args, include = character(), exclude = character(
   args <- lazyeval::as.lazy_dots(args)
   names_list <- setNames(as.list(seq_along(vars)), vars)
 
-  ind_list <- lazyeval::lazy_eval(args, names_list)
-  names(ind_list) <- names2(args)
+  # if the first selector is exclusive (negative), start with all columns
+  initial_case <- if (as.character(args[[1]]$expr)[1] == "-") list(seq_along(vars)) else integer(0)
```
krlmlr: Could you test `` identical(args[[1]]$expr, quote(`-`)) `` ?

tests/testthat/test-select-helpers.R
r2evans: Oops. Duh.

tests/testthat/test-select-helpers.R
r2evans: Related to my comment https://github.com/hadley/dplyr/issues/2275#issuecomment-264922468, I thought I'd go with exhaustive testing of all combinations of number and type of selectors. Currently it looks like this (rather verbose):

```r
  cn <- setNames(colnames(mtcars), nm = colnames(mtcars))

  ### Single Column Selected
  # single columns (present), explicit
  expect_equal(select_vars(cn, mpg), cn["mpg"])
  expect_equal(select_vars(cn, -mpg), cn[ cn != "mpg" ])
  # single columns (present), matched
  expect_equal(select_vars(cn, contains("mpg")), cn["mpg"])
  expect_equal(select_vars(cn, -contains("mpg")), cn[ cn != "mpg" ])
  # single columns (not present), explicit
  expect_error(select_vars(cn, foo), "object 'foo' not found")
  expect_error(select_vars(cn, -foo), "object 'foo' not found")
  # single columns (not present), matched
  expect_named(res <- select_vars(cn, contains("foo")))
  expect_length(res, 0)
  expect_equal(select_vars(cn, -contains("foo")), cn)

  ### Multiple Columns Selected
  # explicit(present) + matched(present)
  expect_equal(select_vars(cn, mpg, contains("vs")), cn[c("mpg", "vs")])
  expect_equal(select_vars(cn, mpg, -contains("vs")), cn["mpg"])
  expect_equal(select_vars(cn, -mpg, contains("vs")), cn[ cn != "mpg" ])
  expect_equal(select_vars(cn, -mpg, -contains("vs")), cn[ ! cn %in% c("mpg", "vs") ])
  # explicit(present) + matched(not present)
  expect_equal(select_vars(cn, mpg, contains("foo")), cn["mpg"])
  expect_equal(select_vars(cn, mpg, -contains("foo")), cn["mpg"])
  expect_equal(select_vars(cn, -mpg, contains("foo")), cn[ cn != "mpg" ])
  expect_equal(select_vars(cn, -mpg, -contains("foo")), cn[ cn != "mpg" ])
  # matched(present) + explicit(present)
  expect_equal(select_vars(cn, contains("vs"), mpg), cn[c("vs", "mpg")])
  expect_equal(select_vars(cn, contains("vs"), -mpg), cn["vs"])
  expect_equal(select_vars(cn, -contains("vs"), mpg), cn[cn != "vs"])
  expect_equal(select_vars(cn, -contains("vs"), -mpg), cn[ ! cn %in% c("mpg", "vs") ])
  # matched(not present) + explicit(not present)
  expect_error(select_vars(cn, contains("foo"), bar), "object 'bar' not found")
  expect_error(select_vars(cn, contains("foo"), -bar), "object 'bar' not found")
  expect_error(select_vars(cn, -contains("foo"), bar), "object 'bar' not found")
  expect_error(select_vars(cn, -contains("foo"), -bar), "object 'bar' not found")
  # matched(present) + matched(present)
  expect_equal(select_vars(cn, contains("vs"), contains("mpg")), cn[c("vs", "mpg")])
  expect_equal(select_vars(cn, contains("vs"), -contains("mpg")), cn["vs"])
  expect_equal(select_vars(cn, -contains("vs"), contains("mpg")), cn[cn != "vs"])
  expect_equal(select_vars(cn, -contains("vs"), -contains("mpg")), cn[! cn %in% c("vs", "mpg")])
  # matched(present) + matched(not present)
  expect_equal(select_vars(cn, contains("vs"), contains("foo")), cn["vs"])
  expect_equal(select_vars(cn, contains("vs"), -contains("foo")), cn["vs"])
  expect_equal(select_vars(cn, -contains("vs"), contains("foo")), cn[cn != "vs"])
  expect_equal(select_vars(cn, -contains("vs"), -contains("foo")), cn[cn != "vs"])
  # matched(not present) + matched(present)
  expect_equal(select_vars(cn, contains("foo"), contains("mpg")), cn["mpg"])
  expect_named(res <- select_vars(cn, contains("foo"), -contains("mpg")))
  expect_length(res, 0)
  expect_equal(select_vars(cn, -contains("foo"), contains("mpg")), cn)
  expect_equal(select_vars(cn, -contains("foo"), -contains("mpg")), cn[cn != "mpg"])
  # matched(not present) + matched(not present)
  expect_named(res <- select_vars(cn, contains("foo"), contains("bar")))
  expect_length(res, 0)
  expect_named(res <- select_vars(cn, contains("foo"), -contains("bar")))
  expect_length(res, 0)
  expect_equal(select_vars(cn, -contains("foo"), contains("bar")), cn)
  expect_equal(select_vars(cn, -contains("foo"), -contains("bar")), cn)
```

This is rather busy/loud, but it definitely encapsulates all of the problems in both #1176 and #2275. Though some of them are relatively identical to some in `test-select.r`, I suggest keeping these here and together helps contrast the different expectations.

Is this too much? I can remove the "obvious" ones if need be, though there are varying levels of "obvious".

I also have a three-selector to ensure the first (match) isn't removed by the second (no match).

```r
  expect_equal(
    select_vars(colnames(mtcars), contains("am"), contains("FOO"), contains("vs")),
    c(am = "am", vs = "vs")
  )
```

R/select-vars.R
r2evans:
```r
debug(select_vars)
select_vars(cn, -mpg)
# ... step into select_vars_, past line 66 where args is assigned
# Browse[3]> 
args[[1]]
# <lazy>
#   expr: -mpg
#   env:  <environment: R_GlobalEnv>
as.character(args[[1]]$expr)[1]
# [1] "-"
as.character(args[[1]]$expr)[1] == "-"
# [1] TRUE
identical(args[[1]]$expr, quote(`-`))
# [1] FALSE
identical(as.character(args[[1]]$expr)[1], quote(`-`))
# [1] FALSE
identical(as.character(args[[1]]$expr)[1], "-")
# [1] TRUE
microbenchmark(
a = as.character(args[[1]]$expr)[1] == "-",
b = identical(as.character(args[[1]]$expr)[1], "-")
)
# Unit: microseconds
#  expr   min    lq   mean median    uq    max neval
#     a 2.553 2.918 3.6000  3.282 3.283 22.246   100
#     b 3.282 3.648 4.4606  4.012 4.376 45.585   100
```

tests/testthat/test-select-helpers.R
krlmlr: I think this is a good start, but should be split in at least two test_that() calls.

@hadley: Should `select_vars(cn, contains("foo"))` throw an error if `foo` doesn't exist?

R/select-vars.R
r2evans: Not sure what you mean. I explicitly tested your expression and it returned false when it should have been true. (I then compared a different command for timing, but that's extraneous and probably should not have been there. I was intending to compare yours with the others, but when I saw it was false I removed it.)

By *result set*, do you mean the actual return value from `contains()` (et al)? I don't think checking that for positive/negative will work (`-integer(0)` ?).

(BTW: how did you get a single backtick within an inline code block? I can't get it to render.

tests/testthat/test-select-helpers.R
hadley: I can't remember. I think we decided it shouldn't.

R/select-vars.R
krlmlr: Sorry, missed that. Hadley's test should work better; `-integer(0)` is a problem indeed.

hadley: Why is that a problem? 

```R
quote(-integer(0))[[1]]
`-`
```

krlmlr: @hadley: The literal expression is not a problem. If we wanted to check if the first selector is inclusive or exclusive by looking at its return value, then the `integer(0) == -integer(0)` case is a problem.

r2evans: Copy all, thanks.

hadley: Yeah, that's why we can't just look at the return value.

r2evans: @hadley, I wasn't suggesting checking against `-integer(0)` within `args`. I was suggesting that checking the return value from a helper function would not be fruitful. I believe by the time `select_vars_` can look at the return values from each of the `args`'s calls, they have already lost the ability to check against ``quote(`-`)``. (The values are already realized by that time, right? Is `ind_list` initially full of promises not-yet-realized?)

I guess `select(mtcars, -integer(0))` could work (and does with this code). Though it seems a bit obscure (when `select(mtcars, everything())` is much clearer).

r2evans: Unfortunately @hadley, I can't get your call working:

```r
select_vars(cn, -mpg, mpg)
### stepping into select_vars_ and once args is assigned (line 66)
# Browse[3]> 
args
# [[1]]
# <lazy>
#   expr: -mpg
#   env:  <environment: R_GlobalEnv>
# [[2]]
# <lazy>
#   expr: mpg
#   env:  <environment: R_GlobalEnv>
# attr(,"class")
# [1] "lazy_dots"
# Browse[3]> 
identical(args[[1]]$expr[[1]], quote(`-`))
# [1] TRUE
# Browse[3]> 
identical(args[[2]]$expr[[1]], quote(`-`))
# Error in args[[2]]$expr[[1]] (from select-vars.R#70) : object of type 'symbol' is not subsettable
# Browse[3]> 
args[[2]]$expr
# mpg
```

Is there a an alternate way to check negativity?

hadley: I'd recommend something like:

```R
is_negated <- function(x) {
  if (!is.call(x)) return(FALSE)
  if (length(x) != 2) return(FALSE)

  identical(x[[1]], quote(`-`))
)

is_negated(args[[1]]$expr)
```

If you want to test interactively, use quote to generate the language objects:

```R
is_negated(quote(-1))
is_negated(quote(-a))
is_negated(quote(a))
is_negated(quote(a - b))
```

r2evans:
@krlmlr, do you see any further changes that need to be implemented? (I still see the "changes requested" tag, implying I did not resolve your suggestions. I *think* the testing of #1176 is covered by the updates.)

tests/testthat/test-select-helpers.R
```diff
@@ -59,3 +59,84 @@ test_that("one_of tolerates but warns for unknown variables", {
...
+  expect_equal(select_vars(cn, mpg), cn["mpg"])
+  expect_equal(select_vars(cn, -mpg), cn[ cn != "mpg" ])
+  # single columns (present), matched
+  expect_equal(select_vars(cn, contains("mpg")), cn["mpg"])
+  expect_equal(select_vars(cn, -contains("mpg")), cn[ cn != "mpg" ])
+  # single columns (not present), explicit
+  expect_error(select_vars(cn, foo), "object 'foo' not found")
+  expect_error(select_vars(cn, -foo), "object 'foo' not found")
+  # single columns (not present), matched
+  expect_named(res <- select_vars(cn, contains("foo")))
```
krlmlr: Could you please extract the assignment out of the function call? I saw other instances below.

tests/testthat/test-select-helpers.R
```diff
@@ -59,3 +59,84 @@ test_that("one_of tolerates but warns for unknown variables", {
 test_that("one_of converts names to positions", {
   expect_equal(one_of("a", "z", vars = letters), c(1L, 26L))
 })
+
+# first-selector ----------------------------------------------------------
+
+test_that("initial (single) selector defaults correctly (issue #2275)", {
+  cn <- setNames(colnames(mtcars), nm = colnames(mtcars))
```
krlmlr: `setNames(nm = ...)` would be shorter.

tests/testthat/test-select-helpers.R
r2evans: I can do both of these (`setNames` and `res`), but neither is consistent with the current coding style. All references (in `dplyr`) to `setNames` explicitly sets `order`. I can find no use of `expect_*` here where the assignment is outside of the code.

@hadley @krlmlr These changes are certainly easy to do, but they are inconsistent with the rest of the package. Are you changing styles?

krlmlr: I'm not sure about @hadley's preference for setNames(), but the only instances where assignment `<-` is used inside an `expect_()` function is for `expect_warning()`, `expect_message()` and such (where there seems to be no other way to retain the value *and* check for warning/message).

```bash
grep -r "expect_.*[<]-" tests
```

r2evans: I see your point. My gut feeling has been that `res <- expect_*` *implies* a return value from `expect` which is not the case, ergo using `res <-` *inside* was more explicit. Patch coming.

I don't mean to be argumentative, honestly. I know that code style needs to be consistent with the target repo and authors, so I am trying to avoid nitnoid changes that diverge from this effort.

tests/testthat/test-select-helpers.R
```diff
@@ -59,3 +59,84 @@ test_that("one_of tolerates but warns for unknown variables", {
...
+  expect_equal(select_vars(cn, mpg), cn["mpg"])
+  expect_equal(select_vars(cn, -mpg), cn[ cn != "mpg" ])
+  # single columns (present), matched
+  expect_equal(select_vars(cn, contains("mpg")), cn["mpg"])
+  expect_equal(select_vars(cn, -contains("mpg")), cn[ cn != "mpg" ])
+  # single columns (not present), explicit
+  expect_error(select_vars(cn, foo), "object 'foo' not found")
+  expect_error(select_vars(cn, -foo), "object 'foo' not found")
+  # single columns (not present), matched
+  res <- expect_named(select_vars(cn, contains("foo")))
```
krlmlr: Actually, I thought about:

```r
res <- select_vars(...)
expect_named(res)
expect_length(res, 0)
```

Sorry about the confusion.

krlmlr:
@hadley: PTAL.

krlmlr:
Thanks!

PR #2298, opened 2016-12-07
Constant folding and hybrid tests

### LINKED ISSUE #1691: with_hybrid helper

### LINKED ISSUE #2143: Constant folding or simplification in the hybrid evaluator

krlmlr:
- New `with_hybrid()`, `without_hybrid()` and `eval_dots()` allow evaluating an expression
    1. inside dplyr, asserting that hybrid evaluation is used
    1. inside dplyr, asserting that hybrid evaluation is not used
    1. outside dplyr

- Improved and simplified hybrid prototypes
    - using `match.call()` for proper argument matching
    - using constant folding where appropriate
    - better support for complex arguments

- Comprehensive (?) tests for functions that use hybrid evaluation

Fixes #1691.

Fixes #2143.

@hadley: PTAL.

### DIFF
--- a/R/nth-value.R
+++ b/R/nth-value.R
@@ -44,7 +44,7 @@ nth <- function(x, n, order_by = NULL, default = default_missing(x)) {
 
   # Negative values index from RHS
   if (n < 0) {
-    n <- length(x) + n + 1L
+    n <- length(x) + n + 1
   }
 
   if (is.null(order_by)) {
--- a/R/nth-value.R
+++ b/R/nth-value.R
@@ -35,18 +35,22 @@ nth <- function(x, n, order_by = NULL, default = default_missing(x)) {
   n <- trunc(n)
 
   if (n == 0 || n > length(x) || n < -length(x)) {
-    return(default)
+    if (is.list(x)) {
+      return(default)
+    } else {
+      return(as.vector(default, class(x)))
+    }
   }
 
   # Negative values index from RHS
   if (n < 0) {
-    n <- length(x) + n + 1
+    n <- length(x) + n + 1L
   }
 
   if (is.null(order_by)) {
     x[[n]]
   } else {
-    x[[order(order_by)[n]]]
+    x[[ order(order_by)[[n]] ]]
   }
 }
 
--- a/R/nth-value.R
+++ b/R/nth-value.R
@@ -38,7 +38,7 @@ nth <- function(x, n, order_by = NULL, default = default_missing(x)) {
     if (is.list(x)) {
       return(default)
     } else {
-      return(as.vector(default, class(x)))
+      return(as.vector(default, typeof(x)))
     }
   }
 
--- a/R/utils.r
+++ b/R/utils.r
@@ -172,19 +172,3 @@ is_1d <- function(x) {
 }
 
 is_bare_list <- function(x) is.list(x) && !is.object(x)
-
-# Returns the value of x if it is a constant, otherwise returns NULL
-# The compile() call with the optimize = 3 option performs constant folding
-# (among other things), it also processes calls like c(), list(), :, and [[.
-# It returns byte code, which then needs to be evaluated.  If evaluation in
-# a new empty environment succeeds, the result is a constant and can be returned.
-# Simply calling eval(x, emptyenv()) won't find c() et al.
-constfold <- function(x) {
-  tryCatch(
-    {
-      compiled <- compiler::compile(x, options = list(optimize = 3, suppressAll = TRUE))
-      eval(compiled, new.env(parent = emptyenv()))
-    },
-    error = function(e) NULL
-  )
-}
--- a/R/utils.r
+++ b/R/utils.r
@@ -179,6 +179,6 @@ constfold <- function(x) {
       compiled <- compiler::compile(x, options = list(optimize = 3, suppressAll = TRUE))
       eval(compiled, new.env(parent = emptyenv()))
     },
-    error = function(e) x
+    error = function(e) NULL
   )
 }
--- a/R/utils.r
+++ b/R/utils.r
@@ -173,6 +173,12 @@ is_1d <- function(x) {
 
 is_bare_list <- function(x) is.list(x) && !is.object(x)
 
+# Returns the value of x if it is a constant, otherwise returns NULL
+# The compile() call with the optimize = 3 option performs constant folding
+# (among other things), it also processes calls like c(), list(), :, and [[.
+# It returns byte code, which then needs to be evaluated.  If evaluation in
+# a new empty environment succeeds, the result is a constant and can be returned.
+# Simply calling eval(x, emptyenv()) won't find c() et al.
 constfold <- function(x) {
   tryCatch(
     {
--- a/R/utils.r
+++ b/R/utils.r
@@ -172,3 +172,13 @@ is_1d <- function(x) {
 }
 
 is_bare_list <- function(x) is.list(x) && !is.object(x)
+
+constfold <- function(x) {
+  tryCatch(
+    {
+      compiled <- compiler::compile(x, options = list(optimize = 3, suppressAll = TRUE))
+      eval(compiled, new.env(parent = emptyenv()))
+    },
+    error = function(e) x
+  )
+}
--- a/inst/include/tools/constfold.h
+++ b/inst/include/tools/constfold.h
@@ -17,7 +17,11 @@ namespace dplyr {
 
   inline SEXP r_constfold(SEXP x) {
     static RConstFold constfold;
-    return constfold(x);
+    SEXP ret = constfold(x);
+    if (Rf_isNull(ret))
+      return x;
+
+    return ret;
   }
 
 }
--- a/inst/include/tools/constfold.h
+++ b/inst/include/tools/constfold.h
@@ -5,12 +5,17 @@
 namespace dplyr {
 
   inline SEXP r_constfold(SEXP x) {
-    static Function constfold("constfold", "dplyr");
-    SEXP ret = constfold(x);
-    if (Rf_isNull(ret))
+    if (!Rf_isLanguage(x)) {
       return x;
+    }
 
-    return ret;
+    static Function force("force", R_BaseEnv);
+    try {
+      return force(x);
+    }
+    catch (...) {
+      return x;
+    }
   }
 
 }
--- a/inst/include/tools/constfold.h
+++ b/inst/include/tools/constfold.h
@@ -1,0 +1,25 @@
+#ifndef dplyr_tools_constfold_h
+#define dplyr_tools_constfold_h
+
+
+namespace dplyr {
+
+  class RConstFold {
+  public:
+    RConstFold() : constfold_fun("constfold", Environment::namespace_env("dplyr")) {}
+    SEXP operator()(SEXP x) {
+      return constfold_fun(x);
+    }
+
+  private:
+    Function constfold_fun;
+  };
+
+  inline SEXP r_constfold(SEXP x) {
+    static RConstFold constfold;
+    return constfold(x);
+  }
+
+}
+
+#endif
--- a/inst/include/tools/constfold.h
+++ b/inst/include/tools/constfold.h
@@ -4,19 +4,8 @@
 
 namespace dplyr {
 
-  class RConstFold {
-  public:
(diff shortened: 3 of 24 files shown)


krlmlr:
@hadley: This PR is largish, it would be great if you could look at the R code and the tests.

R/utils.r
```diff
@@ -172,3 +172,13 @@ is_1d <- function(x) {
 }
 
 is_bare_list <- function(x) is.list(x) && !is.object(x)
+
+constfold <- function(x) {
```
hadley: Could you add a brief comment describing why this does constant folding

R/nth-value.R
```diff
@@ -35,18 +35,22 @@ nth <- function(x, n, order_by = NULL, default = default_missing(x)) {
   n <- trunc(n)
 
   if (n == 0 || n > length(x) || n < -length(x)) {
-    return(default)
+    if (is.list(x)) {
+      return(default)
+    } else {
+      return(as.vector(default, class(x)))
```
hadley: Should that be `as.vector(default, typeof(x))`?

R/nth-value.R
```diff
@@ -35,18 +35,22 @@ nth <- function(x, n, order_by = NULL, default = default_missing(x)) {
...
+    }
   }
 
   # Negative values index from RHS
   if (n < 0) {
-    n <- length(x) + n + 1
+    n <- length(x) + n + 1L
```
hadley: If you want to be strict, I think you'll also need `as.integer(n)`. Maybe do that after the `if` block?

src/hybrid_count.cpp
```diff
@@ -23,15 +25,16 @@ Result* count_distinct_prototype(SEXP call, const ILazySubsets& subsets, int nar
   bool na_rm = false;
 
   for (SEXP p = CDR(call); !Rf_isNull(p); p = CDR(p)) {
-    SEXP x = CAR(p);
-    if (!Rf_isNull(TAG(p)) && TAG(p) == Rf_install("na.rm")) {
+    SEXP xe = CAR(p);
+    if (!Rf_isNull(TAG(p)) && TAG(p) == R_NaRmSymbol) {
```
hadley: Has `match.call()` already been called somewhere so that `na = TRUE` works as expected?

I suspect the whole `na.rm` check is sufficiently common that it should be pulled out into a function.

(This is really outside the scope of this review so I don't think you need to change here, just make a note if we need a fix elsewhere)

src/hybrid_nth.cpp
```diff
@@ -79,184 +90,158 @@ namespace dplyr {
...
-      // we only know how to handle the case where nidx is a length one
-      // integer or numeric. In any other case, e.g. an expression for R to evaluate
-      // we just fallback to R evaluation (#734)
+
+    SEXP nidxe = CAR(p);
+    SEXP nidx = r_constfold(nidxe);
+    // we only know how to handle the case where nidx is a length one
+    // integer or numeric. In any other case, e.g. an expression for R to evaluate
+    // we just fallback to R evaluation (#734)
+    int idx;
```
hadley: I'd prefer to do the usual TYPEOF + length check here

src/hybrid_nth.cpp
```diff
@@ -79,184 +90,158 @@ namespace dplyr {
...
-            return 0;
-          }
-
-        }
-      }
+    SEXP order_by = R_NilValue;
+    SEXP def    = R_NilValue;
 
+    // now get `order_by` and default
+    if (tag == Rf_install("order_by")) {
```
hadley: Doesn't this assume a specific argument order?

hadley: Again, this is probably be outside the scope of this PR, but it would be worth having a class that wraps around LANGSXP providing methods like `call.arg<bool>("na.rm")`

tests/testthat/test-hybrid.R
```diff
@@ -1,26 +1,229 @@
...
+                expected = 3L)
+  expect_hybrid(n_distinct(a, b), a = rep(1L, 3), b = c(1, 1, 2),
+                expected = 2L)
+  expect_hybrid(n_distinct(a, b), a = rep(1L, 3), b = c(1, 1, NA),
+                expected = 2L)
+  expect_hybrid(n_distinct(a, b, na.rm = TRUE), a = rep(1L, 3), b = c(1, 1, NA),
+                expected = 1L)
+
+  c <- 1:3
+  expect_not_hybrid(n_distinct(c), a = 1:5,
```
hadley: Do we really need an `expect_not_hybrid()` helper? Wouldn't `expect_equal()` be adequate?

For the hybrid tests, I think I'd prefer a syntax more like:

```R
expect_identical(hybrid(n_distinct(a), a = 1:5), 5L)
```

R/nth-value.R
hadley: I think using `class()` here is misleading because `as.vector()` works with types, not classes.

tests/testthat/test-hybrid.R
krlmlr: The `expect_hybrid()` and `expect_not_hybrid()` helpers do the following:

- Make sure that hybrid evaluation is used (or not)
- Also test standard evaluation
- Compare both hybrid and standard results to a known result

This is how I found the inconsistencies in `nth()`.

Strictly speaking we don't need `expect_not_hybrid()`, but this is a searchable reminder of what the hybrid evaluator currently can't do.

src/hybrid_nth.cpp
krlmlr: `match.call()` makes sure the order of the arguments matches that of the declaration.

```r
match.call(function(a = 1, b = 2){}, quote(fun(b = 3, 4)))
## fun(a = 4, b = 3)
```

`match.arg()` does very much of the heavy lifting, but walking through the argument list could certainly be better encapsulated.

src/hybrid_nth.cpp
krlmlr: Are you fine with `is<int>()` and `is<double>()`? Those don't throw.

krlmlr:
@hadley: Ready for a second round if we can keep the `expect_hybrid()` and related expectations.

R/utils.r
```diff
@@ -172,3 +172,19 @@ is_1d <- function(x) {
...
+# The compile() call with the optimize = 3 option performs constant folding
+# (among other things), it also processes calls like c(), list(), :, and [[.
+# It returns byte code, which then needs to be evaluated.  If evaluation in
+# a new empty environment succeeds, the result is a constant and can be returned.
+# Simply calling eval(x, emptyenv()) won't find c() et al.
+constfold <- function(x) {
+  tryCatch(
+    {
+      compiled <- compiler::compile(x, options = list(optimize = 3, suppressAll = TRUE))
+      eval(compiled, new.env(parent = emptyenv()))
```
hadley: I wonder if this should be `baseenv()`? That way you'd also evaluate more complicated constants like `sin(2 * pi)`

In that case, would you still even need to use the byte code compiler? The most important thing is evaluating in a scope where the variables are not bound.  I guess you'd have a mild problem if you'd defined `pi` inside your data frame as it would be evaluated from the wrong context.

src/hybrid_count.cpp
```diff
@@ -18,27 +21,41 @@ Result* count_prototype(SEXP args, const ILazySubsets&, int) {
...
+  call = r_match_call(get_n_distinct(), call);
+
   for (SEXP p = CDR(call); !Rf_isNull(p); p = CDR(p)) {
-    SEXP x = CAR(p);
-    if (!Rf_isNull(TAG(p)) && TAG(p) == Rf_install("na.rm")) {
-      if (TYPEOF(x) == LGLSXP && Rf_length(x) == 1) {
-        na_rm = LOGICAL(x)[0];
+    SEXP xe = CAR(p);
+    if (!Rf_isNull(TAG(p)) && TAG(p) == R_NaRmSymbol) {
+      SEXP x = r_constfold(xe);
```
hadley: Actually I think it might be worth while to pull this out an `na_rm` function because it used twice in this PR and the behaviour is inconsistent.

src/hybrid_nth.cpp
hadley: Yeah, assuming you're confident they do what you expect.

hadley: Also that comment needs updating in light of the changes to the code - now we can handle most constant expressions.

src/hybrid_nth.cpp
hadley: Got it. I just missed the call to `match_call`.

tests/testthat/test-hybrid.R
```diff
@@ -1,26 +1,308 @@
...
-  id = c(1L, 2L, 2L),
-  a = 1:3,
-  b = as.numeric(1:3),
-  c = letters[1:3],
-  d = c(TRUE, FALSE, NA)
-)
+test_that("n() and n_distinct() work", {
+  expect_hybrid(n(), a = 1:5, expected = 5L, test_eval = FALSE)
+
+  expect_hybrid(n_distinct(a), a = 1:5,
```
hadley: Ok, thanks for the explanation, I think that makes sense. But can you please match my usual indentation style? i.e.

```R
expect_hybrid(
  n_distinct(a), a = 1:5,
  expected = 5L
)
```

And maybe `expect_hybrid` needs a better name? Here you are simultaneously testing both hybrid and non-hybrid.

R/utils.r
krlmlr: Using the compiler is a tad safer:

```r
constfold(quote(sin(2 * pi)))
## [1] -2.449294e-16
constfold(quote(any))
## NULL
eval(quote(any), env = baseenv())
## function (..., na.rm = FALSE)  .Primitive("any")
```

hadley: Although we could check to see if the result was an atomic...

But it seems like this is a big step forward for now, and we can reconsider alternative approaches in the future.

hadley: I do wonder if it's worth making clear the implicit dependence of `compile()` on the `globalenv()`.

And maybe had a handful of tests just so we can catch if the compiler behaviour changes in the future?

hadley: Maybe global is ok though? That would allow us to catch constants defined in vars at a higher level.

krlmlr: Not sure how:

```r
a <- 5
constfold(quote(a))
## NULL
```

And even then -- shouldn't we be using "our" env and not global?

hadley: Ok, I'd assumed the compiler would figure it out. Base env is fine.

krlmlr:
Currently, match.call() is called for each group in groupwise operations. Need to wait until #2190 is merged (will add necessary changes there).

inst/include/tools/constfold.h
```diff
@@ -0,0 +1,23 @@
+#ifndef dplyr_tools_constfold_h
+#define dplyr_tools_constfold_h
+
+
+namespace dplyr {
+
+  inline SEXP r_constfold(SEXP x) {
```
hadley: So we're actually evaluating the argument - are you sure we even need to wrap in `force()`? Could we just `Rf_eval()` in `baseenv()`?

krlmlr: Maybe. But I've found that it doesn't work for all cases we need -- in particular `T`and `F` cannot be processed this way for some obscure reason. This needs some further investigation.

I've also thought about simply evaluating it in an environment where column access triggers an error -- this might be the most comprehensive constant folding we can get.

hadley: Yeah, that might be the best thing, and consistent with the new hybrid approach.

krlmlr:
Closing in favor of #2306.

krlmlr:
@lionel-: This is an attempt to add constant folding and proper call matching to the hybrid evaluator; currently too slow because c.f. + c.m. would be run for each group/row. https://github.com/hadley/dplyr/issues/2143#issuecomment-266312572 shows the current road map for refactoring the code so that c.f. + c.m. are run only once.

PR #2306, opened 2016-12-10
Hybrid tests

### LINKED ISSUE #1691: with_hybrid helper

krlmlr:
Extracted from #2298, will implement argument matching and constant folding after this is merged. @hadley: PTAL.

Fixes #1691.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -74,6 +74,7 @@ Collate:
     'group-indices.R'
     'group-size.r'
     'grouped-df.r'
+    'hybrid.R'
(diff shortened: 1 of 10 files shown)


hadley:
LGTM

krlmlr:
@lionel-: This tests what the hybrid evaluator currently can and cannot do.

PR #2307, opened 2016-12-10
Address issue https://github.com/hadley/dplyr/issues/2302

JohnMount:
This pull request should fix issue https://github.com/hadley/dplyr/issues/2302 .

The root cause (as diagnosed by Kirill Müller (krlmlr)) is dplyr SQL was of the form "`a + ('x') IS NULL`" which is potentially ambiguous as to are we checking if `('x')` is null, or checking if `a + ('x')` is null.  Probably this is specified in the SQL grammar- but with at least SQLite it was checking if the overall statement `a + ('x')` is null, which was not the desired behavior in the issue example calculation (and not what the dplyr SQL translator thinks it is emitting).

I have added a test that illustrates the problem, modified another test to accept the new translation, and patched translate-sql-base.r.  I ran check before and after the changes and see no other issues.

I assume the dplyr acceptance suite is much larger than the testthat coverage, so expect more tests will be necessary to accept this pull request.


tests/testthat/test-is-na.R
```diff
@@ -0,0 +1,24 @@
+context("is_na")
+
+
+test_that("is_na on databses as non-confusing IS NULL clause", {
```
krlmlr: Please check spelling (is_na, databses).

tests/testthat/test-is-na.R
```diff
@@ -0,0 +1,24 @@
...
+  skip_if_no_sqlite()
+  db <- src_sqlite(":memory:", TRUE)
+
+  d <- data.frame(x=c(1,2,2),
```
krlmlr: Could you please reformat the code? Start a new line after opening paren, spaces around operators and after comma, ... Try Ctrl + A, then Ctrl + Shift + A in RStudio.

JohnMount:
Yes, sorry about that- that was sloppy in my part.  Just tested and committed fixed test.  It looks like to me (and I am new at Github pull requests) my pushing to my fork put the changes into the pull request (neat!).

tests/testthat/test-is-na.R
```diff
@@ -0,0 +1,26 @@
...
+    x = c(1, 2, 2),
+    y = c(3, 5, NA),
+    z = c(NA, 'a', 'b'),
+    rowNum = 1:3,
+    stringsAsFactors = FALSE
+  )
+
+  dfn1 <- copy_to(db, d , "dfn1")
+
+  dfn1 %>% mutate(nna = 0) %>%
```
krlmlr: Thanks. Forgot to ask about using `<-` instead of `->` here.

krlmlr:
Thanks. @hadley: PTAL.

NEWS entry:

* Add missing parentheses to SQL translation of `is.na()` and `is.null()` (#2307, @JohnMount).

tests/testthat/test-is-na.R
```diff
@@ -0,0 +1,26 @@
+context("is_na")
+
+
+test_that("is_na on databases as non-confusing IS NULL clause", {
```
hadley: The test title is a bit confusing

hadley: But what does this entire test add that testing the generated sql directly does not? I think you'd be better off deleting this file altogether and adding another simple test to `test-sql-translation.r`

krlmlr: I'd argue that the acceptance test here is useful: it tests a different aspect than the SQL rendering. In the future, someone might revisit the code and alter the expected SQL; it's more difficult to justify adaptation of a working acceptance test. We might still test that the result is the same with a data frame.

hadley: In that case I'd rather make sure that we test next to each other so it's obvious that one parallels the other, and both tests should be as simple possible.

JohnMount: I'll simplify and fix the new test and place it near the old one.  

I originally no idea what the semantic intent of the old test was, so I had no idea if I could group my test in with that, or if that would cause pull rejection.  I was trying to maintain the purity of the old test by not changing it any more than necessary and not introducing the SQLite db of the acceptance test near it.  I feel without the acceptance test somebody will decide we have too many parenthesis for no reason and revert out the change.

tests/testthat/test-sql-translation.r
```diff
@@ -41,10 +41,33 @@ test_that("between translated to special form (#503)", {
...
 test_that("is.na and is.null are equivalent",{
-  expect_equal(translate_sql(!is.na(x)), sql('NOT(("x") IS NULL)'))
-  expect_equal(translate_sql(!is.null(x)), sql('NOT(("x") IS NULL)'))
+  expect_equal(translate_sql(!is.na(x)), sql('NOT((("x") IS NULL))'))
+  expect_equal(translate_sql(!is.null(x)), sql('NOT((("x") IS NULL))'))
 })
 
+test_that("is.na binds to term and not across expression (#2302)", {
+  skip_if_no_sqlite()
+  db <- src_sqlite(":memory:", TRUE)
```
hadley: Still need to use `memdb_frame` here

tests/testthat/test-sql-translation.r
```diff
@@ -41,10 +41,33 @@ test_that("between translated to special form (#503)", {
...
+  d <- data.frame(
+    y = c(3, 5, NA),
+    z = c(NA, 'a', 'b'),
+    rowNum = 1:3,
+    stringsAsFactors = FALSE
+  )
+
+  dfn1 <- copy_to(db, d , "dfn1")
+
+  dres <- dfn1 %>% mutate(nna = 0) %>%
```
hadley: I'd prefer something that more directly matched the generated sql above - i.e. something that would more clearly fail with a missing parens.

JohnMount:
This is my last commit on this, sorry it has run long.

Obviously we are not in agreement on the structure of the acceptance test.  I am not saying I disagree or have standing; I am saying I can't efficiently anticipate enough of the project preferences and mores to write the test that project members feel they would have written.

I copied my original test form from another test, so I had hoped it was a compliant use of SQLite.  A (presumably unshared) concern I have is if I simplify the test too much then it no longer resembles any task anyone would actually want to perform, and thus looks silly on the face of it (which would likely trigger its own objections).

Now we have the fix and I have restricted myself to merely updating the "exact translation" test that the fix breaks by changing the SQL.  There is now no test accepting the change (i.e. a test referring to the issue that is fails before the change and after).  So the original title of the test I altered stands- as it is just testing what it always did.

hadley:
I don't understand why you can't make a simple test that uses `memdb_frame()`? You just needed to remove some of the complication from your existing test.

hadley:
@JohnMount are you sure you don't want to finish this off? It's so close to being done!

JohnMount:
As I said, I am done submitting code updates to the pull.

hadley:
Ok, your choice obviously.

PR #2309, opened 2016-12-11
Template class for LazyRowwiseSubset and LazyGroupedSubset

krlmlr:
Main reason: Need to add `get()` to `ILazySubset` as a prerequisite for simplifications in the hybrid evaluator.


hadley:
LGMT

krlmlr:
@lionel-: This is an attempt to simplify the code base, similar in spirit to #2157: Here a template replaces two classes that were very similar before.

PR #2315, opened 2016-12-12
Refactor GroupedHybridCall

krlmlr:
- Split in three classes
- Introduce VariableResult, a hybrid handler that returns a data frame variable
- Simplify mutate_grouped()

Performance doesn't seem to be affected. @hadley: PTAL. CC @lionel-.

### DIFF
--- a/tests/testthat/test-mutate.r
+++ b/tests/testthat/test-mutate.r
@@ -517,7 +517,7 @@ test_that("grouped mutate does not drop grouping attributes (#1020)", {
 
 test_that("grouped mutate errors on incompatible column type (#1641)", {
   df <- data.frame(ID = rep(1:5, each = 3), x = 1:15) %>% group_by(ID)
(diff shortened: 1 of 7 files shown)


tests/testthat/test-mutate.r
```diff
@@ -517,7 +517,7 @@ test_that("grouped mutate does not drop grouping attributes (#1020)", {
 
 test_that("grouped mutate errors on incompatible column type (#1641)", {
   df <- data.frame(ID = rep(1:5, each = 3), x = 1:15) %>% group_by(ID)
-  expect_error( mutate(df, foo = mean), 'Unsupported type CLOSXP for column "foo"')
+  expect_error( mutate(df, foo = mean), 'Not a vector')
```
hadley: Do we lose the column name?

hadley:
LGTM

PR #2334, opened 2016-12-18
Fix groups when joining grouped data frames with duplicates (#2330)

### LINKED ISSUE #2330: Incorrect group-by columns when joining grouped data frames with overlapping columns
`join` updates overlapping column names, but does not update corresponding group column names in attribute `vars`. This becomes an issue when a group column is not used in the join. The resulting data frame causes errors in `mutate`.

```
library(dplyr)


df1 <- data.frame(x = 1:10, y = 1:10)
df2 <- df1

df1g <- df1 %>% group_by(x, y)

df3 <- inner_join(df1g, df2, by = "x")

df3 %>%
  mutate(a = 1)
```

Output:

```
Error in mutate_impl(.data, dots) : unknown column 'y.x' 
```

I have a fix and will submit a pull request.

davidkretch:
* Fix subset_join to update group column names in attribute
vars when they are duplicate column names and get renamed.

* Add a test to verify that the group column names after
join are all columns in the data frame.

* Fix build_index_cpp to report correct missing group
column name. Currently when a group column name does not
exist in the data frame, it reports a name from the names 
vector (all columns) instead of the vars vector (group columns).

Fixes #2330.

### DIFF
--- a/tests/testthat/test-group-indices.R
+++ b/tests/testthat/test-group-indices.R
@@ -22,3 +22,12 @@ test_that("group_indices can be used in mutate (#2160)", {
   res2 <- mtcars %>% mutate(group_idx = as.integer(factor(cyl)))
   expect_equal(res1, res2)
 })
+
+test_that("group indices are updated correctly for joined grouped data frames (#2330)", {
+  d1 <- data.frame(x = 1:2, y = 1:2) %>% group_by(x, y)
+  expect_equal(group_indices(d1), d1$x)
+
+  d2 <- expand.grid(x = 1:2, y = 1:2)
+  res <- inner_join(d1, d2, by = "x")
+  expect_equal(group_indices(res), res$x)
+})
--- a/tests/testthat/test-joins.r
+++ b/tests/testthat/test-joins.r
@@ -370,6 +370,21 @@ test_that("joins between factor and character coerces to character with a warnin
 
 })
 
+test_that("group column names reflect renamed duplicate columns (#2330)", {
+  d1 <- data_frame(x = 1:5, y = 1:5) %>% group_by(x, y)
+  d2 <- data_frame(x = 1:5, y = 1:5)
+  res <- inner_join(d1, d2, by = "x")
+  expect_equal(groups(d1), list(quote(x), quote(y)))
+  expect_equal(groups(res), list(quote(x), quote(y.x)))
+})
+
+test_that("group column names are null when joined data frames are not grouped (#2330)", {
+  d1 <- data_frame(x = 1:5, y = 1:5)
+  d2 <- data_frame(x = 1:5, y = 1:5)
+  res <- inner_join(d1, d2, by = "x")
+  expect_null(groups(res))
+})
+
 # Guessing variables in x and y ------------------------------------------------
 
 test_that("unnamed vars are the same in both tables", {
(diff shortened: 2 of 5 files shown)


krlmlr:
Thanks. Are the group indices updated correctly? We'll need a test for that (unless it exists, I haven't found any), and also a test for the error message buglet you discovered.

@hadley: Is joining a grouped data frame an operation we support?

hadley:
I don't know what joining a grouped data frame should do. Probably just preserve the grouping of `x`?

krlmlr:
This pull request addresses the case when columns are renamed. I just checked, group indices look ok after an expanding join (replace `df2 <- data.frame(...)` by `df2 <- expand.grid(...)`).

davidkretch:
Thanks. I'll add those tests asap.

It looks to me that the existing behavior when joining grouped data frames is to preserve the grouping of `x` -- at the bottom of `subset_join`, it copies `vars` from `x`: 

```
SEXP vars = x.attr("vars");
if (!Rf_isNull(vars))
  out.attr("vars") = vars;
```

davidkretch:
I've added tests for:
* expanding join group indices
* missing group column error message
* attribute `vars` in the joined data frame is null when the original df was not grouped, i.e. the new code doesn't introduce any

tests/testthat/test-group-indices.R
```diff
@@ -22,3 +22,10 @@ test_that("group_indices can be used in mutate (#2160)", {
   res2 <- mtcars %>% mutate(group_idx = as.integer(factor(cyl)))
   expect_equal(res1, res2)
 })
+
+test_that("group indices are updated correctly for joined grouped data frames (#2330)", {
+  d1 <- data.frame(x = 1:2, y = 1:2) %>% group_by(x, y)
+  d2 <- expand.grid(x = 1:2, y = 1:2)
+  res <- inner_join(d1, d2, by = "x") %>% group_indices()
```
krlmlr: I think testing the group indices as an invariant will make the test clearer (and easily extensible/adaptable).

```r
expect_equal(group_indices(d1), d1$x)

...

res <- inner_join(d1, d2, by = "x")
expect_equal(group_indices(res), res$x)
```

tests/testthat/test-joins.r
```diff
@@ -370,6 +370,20 @@ test_that("joins between factor and character coerces to character with a warnin
 
 })
 
+test_that("group column names get updated when they are duplicates (#2330)", {
+  d1 <- data_frame(x = 1:5, y = 1:5) %>% group_by(x, y)
+  d2 <- data_frame(x = 1:5, y = 1:5)
+  res <- inner_join(d1, d2, by = "x")
```
krlmlr: Here, an explicit check for an expected result will give slightly nicer output in case of failure. We can be explicit  about the suffixes the join uses, too.

davidkretch:
Hi, thank you for your review. I've updated the pull request. Hopefully any future PRs have fewer issues.

I added logic to give an error if `subset_join` somehow gets an `x` grouped data frame with non-existent group columns.  I'm not sure how to test the error without directly modifying the `vars` attribute though.  Without this check, it would try to access `names[NA_INTEGER]` and hang. I think the alternatives are to give an error, output with no groups, or not check assuming they're always valid.

I also fixed a serious bug in my previous commit, which accidentally modified `x.attr("vars")`  by reference, which is now `const`. Sorry about that.

krlmlr:
Thanks for the edits. You can still rip out the column in a grouped data frame:

```r
iris %>% group_by(Species) %>% { .[["Species"]] <- NULL; . }
```

We could forbid this, but creative users will find a way to break it anyway. I think anything that takes (grouped) data frames should check validity of the input, but this is too much for this PR. I'll merge this as is, the safety net you added will protect us from crashes.

Issue #2166, opened 2016-10-10
Summarise nth & order_by crashes R (on Windows)


avsmith:
A combination of 'summarise', 'nth', and 'order_by' crashes R on Windows (but not Mac).

Make sample data

``` r
library(dplyr)

set.seed(42)

# Simulate data
# obs is the total number of observations
# ninds is the number of unique individuals
obs <- 25000
ninds <- 5000

# Simulate BP data
id <- sample(1:ninds,obs,replace=TRUE)
SBP <- as.integer(rnorm(obs, 120, 15))
DBP <- as.integer(rnorm(obs,90,10))

# Put all into a data frame, adding a value for the visit number
bppheno <- data_frame(id,SBP,DBP) %>% 
  group_by(id) %>% 
  mutate(visit=row_number())
```

The following statement crashes R 3.3.1 on Windows (but not Mac). [dplyr 0.5.0]

``` r
bppheno_characteristics <- bppheno %>%
  dplyr::group_by(id) %>%
  dplyr::summarise(N_BP=n(), FIRST_SBP_DPLYR=first(SBP, order_by=visit))
```

The following variant without the wrapper does not crash R (on Windows)

``` r
bppheno_characteristics <- bppheno %>%
  dplyr::group_by(id) %>%
  dplyr::summarise(N_BP=n(),FIRST_SBP_RAW=SBP[[order(visit)[1]]])
```

Also, the following variant (explicitly calling dplyr::first) does not crash R (on Windows)

``` r
bppheno_characteristics <- bppheno %>%
  dplyr::group_by(id) %>%
  dplyr::summarise(N_BP=n(), FIRST_SBP_DPLYR=dplyr::first(SBP, order_by=visit))
```

kevinushey:
Are you running this code within RStudio? There's a known bad interaction between R 3.3.x and RStudio that can cause crashes like this. You might want to try updating to the [preview release](https://www.rstudio.com/products/rstudio/download/preview/), which contains a fix for this.

avsmith:
No, I can get the behavior independently of RStudio.

krlmlr:
Thanks, confirmed, I'm getting a SIGSEGV on Ubuntu 16.04. 

Would you like to contribute a testthat test?

krlmlr:
Other hybrid functions are affected by similar problems, #2299 is the canonical issue.

stephematician:
I assume I am having the same problem. I find that the following works fine in Windows:

[...]

PR #2342, opened 2016-12-25
Add test for crash in issue #2166.

karldw:
Summarise nth and order_by crashes R in both Windows and Ubuntu.
This commit triggers that crash, with code pulled from @avsmith's issue.

Apparently this issue does not affect Macs.

### DIFF
--- a/tests/testthat/test-nth-value.R
+++ b/tests/testthat/test-nth-value.R
@@ -40,3 +40,10 @@ test_that("default value errors for complicated structures", {
   expect_error(default_missing(factor("a")), "generate default for object")
   expect_error(default_missing(mtcars), "generate default for object")
 })
+
+test_that("nth and order_by doesn't crash", {
+    # Bug from issue #2166
+    data_frame(SBP = rnorm(2)) %>%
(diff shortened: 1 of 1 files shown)


krlmlr:
Thanks. Could you please unindent the first two lines of the test body by two spaces, and add spaces around `=`? Do we really need to use random numbers here?

Please also add `skip("Currently failing")` at the top of the test so that we can merge right away without affecting tests.

tests/testthat/test-nth-value.R
```diff
@@ -40,3 +40,10 @@ test_that("default value errors for complicated structures", {
   expect_error(default_missing(factor("a")), "generate default for object")
   expect_error(default_missing(mtcars), "generate default for object")
 })
+
+test_that("nth and order_by doesn't crash (#2166)", {
+  skip("Currently failing")
+  data_frame(SBP = c(1, 1)) %>%
```
hadley: I think you could use variable names `x` and `y`

krlmlr:
Thanks!

PR #2344, opened 2016-12-29
move copy_most_attributes to header

jayhesselberth:
- fixes linking error in external package using DataFrameSubsetVisitors (https://github.com/rnabioco/valr/issues/163#issuecomment-269687419)


krlmlr:
@hadley: PTAL. This prevents other packages from using the `DataFrameSubsetVisitor` which we are publishing in `inst/include` .

hadley:
LGTM, although I admittedly don't know exactly why the previous definition was a problem (and I'm happy to remain ignorant)

krlmlr:
Thanks!

Issue #2218, opened 2016-10-29
Document row_number()


krlmlr:
Currently, the documentation of row_number() only gives the version with one argument; but inside mutate() et al., row_number() without arguments works fine (and may be faster).

CarrKnight:
row_number() without arguments stopped working within mutate. Was it replaced by anything else?

krlmlr:
@CarrKnight: Could you please provide a reproducible example? `row_number()` definitely should work.

CarrKnight:
@krlmlr 
My bad, sorry, i must have screwed up the library loading order before calling it since I can't reproduce it again. Sorry for the false alarm.

holgerbrandl:
Even with the updated documentation I still  don't see where `row_number` gets its `x` argument from. Could you point me to the corresponding code bit?  

Is there a listing of dplyr methods that are formally defined of requiring an argument but work without?

krlmlr:
There are two versions of `row_number()`: one in R code with a mandatory argument, and one in the hybrid evaluator that doesn't require arguments. Look for `row_number_prototype` in the C++ code.

holgerbrandl:
Thanks. Are the handlers listed under https://github.com/tidyverse/dplyr/blob/14b5d128c303fd9af0fca4ab6f1fa17ede5c9bf4/src/hybrid_window.cpp#L158-L163
the only methods that have this duality in dplyr?
Is this a common R approach or something very specific for/in dplyr?

Issue #2205, opened 2016-10-27
dplyr::count returns count column named as 'nn' instead of 'n'


kdaily:
I'm having trouble replicating, except with one specific dataset, but I get something like:

```
> queryData %>% count(teamName, dateGrouping)
Source: local data frame [8 x 3]
Groups: teamName [?]

    teamName dateGrouping    nn
      <fctr>       <date> <int>
1       None   2016-08-01   633
2       None   2016-09-01     9
3  Anonymous   2016-08-01  2049
....
```

If I manually make a `data.frame` and run the same kind of function, my count column comes back as `n`:

```
> data.frame(teamName=sample(c('1','2'), 1000, replace=T), 
                  date=sample(c(lubridate::today(), 
                                lubridate::today() - months(1)), 1000, 
                              replace=T), 
                  foo='a') %>% 
  dplyr::count(teamName, date)
Source: local data frame [4 x 3]
Groups: teamName [?]

  teamName       date     n
    <fctr>     <date> <int>
1        1 2016-09-26   238
2        1 2016-10-26   267
3        2 2016-09-26   250
4        2 2016-10-26   245
```

sessionInfo:

```
R version 3.3.1 (2016-06-21)
Platform: x86_64-redhat-linux-gnu (64-bit)
Running under: Fedora 23 (Workstation Edition)

locale:
 [1] LC_CTYPE=en_US.UTF-8       LC_NUMERIC=C              
 [3] LC_TIME=en_US.UTF-8        LC_COLLATE=en_US.UTF-8    
 [5] LC_MONETARY=en_US.UTF-8    LC_MESSAGES=en_US.UTF-8   
 [7] LC_PAPER=en_US.UTF-8       LC_NAME=C                 
 [9] LC_ADDRESS=C               LC_TELEPHONE=C            
[11] LC_MEASUREMENT=en_US.UTF-8 LC_IDENTIFICATION=C       

attached base packages:
[1] stats     graphics  grDevices utils     datasets  methods   base     

other attached packages:
[1] dplyr_0.5.0   RMySQL_0.10.9 DBI_0.4-1

loaded via a namespace (and not attached):
 [1] lazyeval_0.2.0  magrittr_1.5    R6_2.1.3        assertthat_0.1 
 [5] tools_3.3.1     tibble_1.1      Rcpp_0.12.6     lubridate_1.5.6
 [9] stringi_1.1.1   stringr_1.1.0  
```

krlmlr:
[...]

Issue #2152, opened 2016-09-27
help file for arrange needs updating


Koantig:
The current version states that string ordering doesn't know about the locale but this closed issue: https://github.com/hadley/dplyr/issues/1175 and a simple test show that it's not the case anymore.

This paragraph: https://github.com/hadley/dplyr/blob/master/R/manip.r#L180 can be removed or updated.

Thank you

Issue #1175, opened 2015-05-27
Feature or bug? arrange sorts strings differently than sort (and characters other than factors)


skranz:
Consider this example:
```r
library(dplyr)
df = data.frame(x=c("B","a","A","b"))
tbl = data_frame(x=c("B","a","A","b"))

# all 3 commands sort x alphabetically
sort(df$x)
dplyr::arrange(df,x)
sort(tbl$x)

# sorts x differently: puts uppercase letters before lowercase letters
dplyr::arrange(tbl,x)

class(df$x)
class(tbl$x)
```
I just wondered whether it is on purpose that the last arrange sorts the characters in a different fashion or a bug?

Here is the output:
```r
> library(dplyr)
> df = data.frame(x=c("B","a","A","b"))
> tbl = data_frame(x=c("B","a","A","b"))
> 
> # all sort alphabetically
> sort(df$x)
[1] a A b B
Levels: a A b B
> dplyr::arrange(df,x)
  x
1 a
2 A
3 b
4 B
> sort(tbl$x)
[1] "a" "A" "b" "B"
> 
> # sorts differently: puts uppercase letters before lowercase letters
> dplyr::arrange(tbl,x)
Source: local data frame [4 x 1]

  x
1 A
2 B
3 a
4 b
> 
> class(df$x)
[1] "factor"
> class(tbl$x)
[1] "character"
```

hadley:
`arrange()` now respects your locale, so this is now fixed.

PR #2369, opened 2017-01-20
documentation fixes - #2152, #2205, #2218

cb4ds:
Fixed 3 documentation issues (#2152, #2205, #2218), and created one test in response to #2152 (based on #1175 fix)


R/rank.R
```diff
@@ -39,6 +39,12 @@
 #'
 #' ntile(x, 2)
 #' ntile(runif(100), 10)
+#' 
+#' # row_number can be used with single table verbs without
+#' # an explicit x (as long as the db supports windowing)
```
krlmlr: This confuses me because your examples operate on data frames, not databases. How about `(for data frames, and databases that support windowing)` ?

R/tally.R
```diff
@@ -31,6 +31,12 @@
 #' batting_tbl %>% count(playerID, wt = G)
 #' batting_tbl %>% count(playerID, wt = G, sort = TRUE)
 #' }
+#'
+#' @note
+#' The column name in the returned data is 'n', however if the data already
```
krlmlr: Can you please use `\code{n}`?

tests/testthat/test-arrange.r
```diff
@@ -133,6 +133,21 @@ test_that("arrange respects locale (#1280)", {
...
+test_that("arrange respects locale (#1175)", {
+  vector <- c("B","a","A","b")
+  df  <- data.frame(x = vector, stringsAsFactors = F)
+  tbl <- data_frame(x = vector)
+
+  master.asc <- sort(vector)
+  master.dsc <- sort(vector, decreasing = T)
+
+  expect_equivalent(master.asc, arrange(df, x)$x)
+  expect_equivalent(master.asc, arrange(tbl, x)$x)
```
krlmlr: - I think the test will be easier to understand later on if you structure it like this:
    1. `vector <- ...` and creation of `master.asc`
    2. `df <- ...` and tests
    3. `tbl <- ...` and tests

- The order of arguments in the expectations should be reversed, `master.*` are the expected values and not the values under test

R/rank.R
cb4ds: will update to the suggested wording - makes a lot of sense.

R/tally.R
cb4ds: yes, will update

tests/testthat/test-arrange.r
cb4ds: thanks for the catch on the ordering, I will change that and rearrange the function as well to make it clearer what is going on.

krlmlr:
@hadley: Okay to merge? I'd like to work through all PRs that affect roxygen2 documentation so that we can convert to Markdown.

tests/testthat/test-arrange.r
```diff
@@ -133,6 +133,20 @@ test_that("arrange respects locale (#1280)", {
 
 })
 
+test_that("arrange respects locale (#1175)", {
```
hadley: This is not a good test because you rely on the system locale.   But making it better is going to be hard because different systems have different names for the locales. I would rather simply drop this test.

krlmlr: @hadley: The test makes sure that arrange() is consistent with sort() in the current locale. It would detect if the implementation of arrange() suddenly changes to use the "C" locale. The test is not perfect, but I think it's better than no test at all.

hadley: It assumes that the current locale is not already the C locale, which seems risky to me.

krlmlr: This may be one of the occasions where it's easier to let the user test. We could explicitly state that arrange() respects the current locale, just like sort(); users will alert us if this isn't true anymore. Would that work better?

hadley: Yeah, I think that's reasonable.

cb4ds: so, if I'm following this correctly, I'll pull the test out and put back in the section on locales (taken out for #2152) with wording something like:

```
#' @section Locales
#' The sort order for character vectors will depend on the collating 
#' sequence of the locale in use: see \link{locales}.
```

krlmlr: Yes, I like the wording.

krlmlr:
Thanks!

PR #2372, opened 2017-01-21
Early exit in case_when

jrnold:
If all observations have been replaced, then stop the for loop.


krlmlr:
This might take longer in some cases. @hadley: Are we rewriting `case_when()` in C++ eventually?

jrnold:
Yeah, I was thinking about that, since it has to check each time. And presumably in most cases someone is using a `case_when` function they have an expectation that all cases are at least likely to occur. I should have checked it across a variety of sizes and configurations, but I wanted to see what your thoughts were.

hadley:
@krlmlr Yeah, that's the plan, so I'd prefer to leave optimisations until we do that.

@jrnold Sorry I couldn't merge your PR. I really appreciate the effort 😁

jrnold:
It was a "generous" three lines, so I wouldn't say it's effort :-)

PR #2383, opened 2017-01-27
Get rid of grouped_indices_impl()

krlmlr:
The implementation looks pretty similar to build_index_cpp(), I guess there's not much difference if we compute the index with a regular grouping operation.

The new implementation is a tad slower for `nycflights13::flights %>% group_indices(year, month, day)` (~95 ms vs. ~91 ms), I suspect this is because `group_by()` always computes labels. Do we need to care about that?

@hadley: PTAL.


hadley:
LGTM — we can always take another pass at optimisation in the future (which will be much easier once we have good R/C++ profiler integration 😉)

PR #2382, opened 2017-01-26
Fix group_by() for column names in UTF-8 on Windows

krlmlr:
Fixes #2284. Similar to #2058 but at a lower level.

Fixes #2005 and #2277.

PR #2384, opened 2017-01-27
New group_names() to return group names as character vector

krlmlr:
in `groups()`. This is a breaking change, in preparation for #1950.

@hadley: PTAL.


hadley:
Hmmmm, this is going to break every backend, and coordinating the changes is going to be hard. What if we made a new generic `group_vars()` (?) instead? It could have a default method that did `lapply(groups(x), as.character)` with a message/warning.

hadley:
@krlmlr do you have thoughts on how your API diffing package might handle this? It seems like you'd need some manual tags to flag changes in data types.

krlmlr:
@hadley: Added group_vars() generic instead, also using it internally. Could you please also look at #2382?

krlmlr:
Added tests. @hadley: PTAL.

Issue #2243, opened 2016-11-08
Column names with spaces or other special characters


krlmlr:
Various verbs have issues if column names contain spaces or other non-alphanumeric characters. 
They work only if all column names are valid R identifiers.

- [x] mutate_if
- [x] mutate_at
- [x] summarise_if
- [x] summarise_at
- [x] select_if
- [x] rename
- [x] summarize_all
- [ ] slice

markriseley:
I may have found a fix for some of this. It involves quoting character vector vars in probe_colwise_names() and select_colwise_names(), which should resolve the _if and _at functions at least.

See this commit in my fork of dplyr:
https://github.com/markriseley/dplyr/commit/6a4d495078c10116968aeba9035b109871ef6d4c

I added a couple of basic tests and ran R CMD check, and checked all the help page examples for summarise_all {dplyr} worked if you changed the column "Petal.Width" to "Petal Width".

I hope this helps, please do more thorough checking, I don't know whether this would cause any issues with databases etc.

mexindian:
Just a bit of experimenting leads to even some verbs showing the bug, others not:

```
iris$`1_brandNewColumn` <- runif(n = nrow(iris))
iris %>% group_by(Species) %>% summarise_all(mean, trim = 1) ## No bug
# A tibble: 3 × 6
     Species Sepal.Length Sepal.Width Petal.Length Petal.Width `1_brandNewColumn`
      <fctr>        <dbl>       <dbl>        <dbl>       <dbl>              <dbl>
1     setosa          5.0         3.4         1.50         0.2          0.5055038
2 versicolor          5.9         2.8         4.35         1.3          0.5388621
3  virginica          6.5         3.0         5.55         2.0          0.5186122
> 
iris %>% group_by(Species) %>% summarize_if(is.numeric,mean) ## Bug
Error in parse(text = x) : <text>:1:2: unexpected input
1: 1_
     ^
```

tchakravarty:
Not sure if this is related to spaces in the names of the columns variants that are collected in this issue, but I ran into this error when trying to answer [this](https://stackoverflow.com/questions/42595566/conditionally-fill-missing-values-according-to-subset-of-character-string-in-r): 

```{r}
r1 = c('', 'abc def', '')
r2 = c('1', 'ghi jkl', '2')
r3 = c('', 'mno pqr', '')
df = as.data.frame(rbind(r1, r2, r3), stringsAsFactors = FALSE)

df %>% 
  mutate_each(
[...]

PR #2386, opened 2017-01-27
WIP: A more consistent way to specify query arguments

krlmlr:
This defines a class `qspec` with subclasses that provide more control for programmatic use of dplyr verbs. Currently, an example implementation for `select()` is provided. Each verb (or class of verbs) would get its own `qspec` subclass. This solves two problems:

1. Allows passing raw column names or column indexes to `select_()`
2. Simplifies the definition of a consistent API across backends

## Raw column names

Currently, the user can

- use NSE and pass unquoted column names, column indexes, column ranges, selection predicates, etc. in `...`
- use SE and pass quoted column names, quoted column indexes etc. in `...`
- use SE, construct a `lazy_dots` object and pass it to `.dots`

With `qspec` and its `as.qspec_select.character()` and `as.qspec_select.integer()` methods, the user now can also pass named character or integer vectors. (Before, passing character vectors resulted in a call to `parse()`, so that non-syntactic names could not be used directly.)

Passing character vectors also circumvents encoding problems.

The implementation currently addresses #2243 for the `select_if()` case:

```r
> data_frame("a b c" = 1) %>% select_if(TRUE)
  `a b c`
    <dbl>
1       1
```

## Consistent API

For each backend, the implementation of `select_()` basically has to do the same steps:

- call `as.lazy_dots()`
- call `select_vars_()`
- in the grouped case, call `ensure_grouped_vars()`

With `qspec` it is sufficient to call `as.qspec_select()`, and the implementation will figure out what to do, independent of the backend and of the current grouping. This also simplifies future changes/enhancements to the argument processing.

## TODO

- Implement `qspec` for all other verbs
- Use a better name than `qspec`.

@hadley: PTAL.


krlmlr:
@lionel-: Hadley suggests that most of this will be covered by rlang/tidyeval. Would you like to comment?

lionel-:
yes, rlang will feature tidy evaluation which makes it possible and easy to program with any NSE functions. We are even considering dropping the idiom of pairing NSE functions with a SE version.

You'll be able to do:

```r
column <- ~cyl
select(mtcars, !! column)
```

or:

```r
var <- 10
dots <- list(
  cyl = ~ 2 * cyl,
  new = ~ disp + var
)

mutate(mtcars, !!! dots)
```

Wrapping around a NSE function will be handled like this:

```r
wrapper <- function(df, ...) {
  mutate(df, !!! tidy_dots(...))
}
```

`!!` is a shortcut for unquoting and `!!!` for unquote-splicing. The whole thing works because the evaluation environment treats formulas as self-evaluating promises. See https://github.com/hadley/rlang/pull/19

krlmlr:
Thanks. I still feel that we need something like this PR, but probably after converting dplyr to use tidyeval. Are you planning to implement this any time soon?

lionel-:
I'm not sure it would still be needed. E.g. you can have all of:

```r
var1 <- ~3

foo <- "d"
var2 <- tidy_quote(starts_with(!! foo))

var3 <- ~drat

select(mtcars, !!var1, !!var2, !!var3)
```

Also the `:=` operator will make it possible to have programmable dots names. E.g.:

```r
new_col <- "foo"
select(mtcars, new_col := cyl, toupper(new_col) := drat)
```

We should be able to start porting the tidyverse to tidyeval sometime this month, hopefully next week.

krlmlr:
Can you give an example of how the new implementation of `mutate_if()` would look like?

lionel-:
probably something like:

```r
select_if <- function(.data, .predicate, ...) {
  if (inherits(.data, "tbl_lazy")) {
    stop("Selection with predicate currently require local sources",
      call. = FALSE)
  }
  vars <- probe_colwise_names(.data, .predicate, ...)
  vars <- ensure_grouped_vars(vars, .data, notify = FALSE)

  vars <- lapply(vars, as.name)
  select(.data, !!! vars)
}
```

krlmlr:
Using `as.name()` will give us encoding headaches on Windows (#1950). Is there a way without converting to language symbols?

lionel-:
We could convert the names to column positions.

krlmlr:
I'm not convinced yet. To me, it looks like the new tools are great for interactive use, but I feel that for programming an extra layer like I'm suggesting is helpful. Let's discuss this again after converting dplyr to tidyeval.

PR #2388, opened 2017-01-28
Use String instead of Symbol

### LINKED ISSUE #1950: Encoding problems on Windows caused by character -> symbol -> character roundtrip

### LINKED ISSUE #2387: mutate()/transmute() fails to handle column names in UTF-8 on Windows
As suggested in https://github.com/hadley/dplyr/pull/2285#issuecomment-275795425, the current GitHub version of `mutate()` and `transmute()` behaves strangely for column names in UTF-8 on Windows. Though I couldn't find the cause yet, I'm afraid this is a regression bug related to #1950.

Here are the error details with reprexes:

## Case 1) Error with a data.frame that contains non-ASCII colnames

I found that `mutate()` adds a strange column for non-ASCII columns.

```r
df1 <- data_frame(Φ = 1)
df1
#> # A tibble: 1 × 1
#>      Φ
#>   <dbl>
#> 1     1

df1_mutated <- df1 %>% mutate(Φ = Φ * 2)
df1_mutated
#> # A tibble: 1 × 2
#>      Φ    ﾎｦ
#>   <dbl> <dbl>
#> 1     1     2
```

## Case 2) Error with a data.frame that contains non-ASCII colnames in UTF-8

If the non-ASCII colname is UTF-8-encoded, `mutate()` does not add but replaces the existing columns with the strange column.

```r
df2 <- df1
colnames(df2) <- enc2utf8(colnames(df2))

df2_mutated <- df2 %>% mutate(Φ = Φ * 2)
df2_mutated
#> # A tibble: 1 × 1
#>      ﾎｦ
#>   <dbl>
#> 1     2
```

## Details

Then, what is this mysterious character `ﾎｦ`?

This is actually a UTF-8-converted `Φ`, but unfortunately it lost Encoding attribute. This is why the non-ASCII columns are mistakenly handled in non-UTF-8 environments.

```r
charToRaw(colnames(df2_mutated))
#> [1] ce a6
charToRaw(enc2utf8("Φ"))
#> [1] ce a6

Encoding(colnames(df2_mutated))
#> [1] "unknown"
```

So if I set `Encoding()` as `"UTF-8"`, it starts to work fine again.

```r
colnames(df2_mutated) <- `Encoding<-`(colnames(df2_mutated), "UTF-8")
df2_mutated
#> # A tibble: 1 × 1
#>      Φ
#>   <dbl>
#> 1     2
```

## My environment

```r
library(dplyr)

packageVersion("dplyr")
#> [1] '0.5.0.9000'

# the installed version is at the point of this commit:
#    https://github.com/hadley/dplyr/commit/8aa1bdb8fe95b741fb9411dbccd1b3af2f631dfc
packageDescription("dplyr")$GithubSHA1
#> [1] "8aa1bdb8fe95b741fb9411dbccd1b3af2f631dfc"

Sys.getlocale()
#> [1] "LC_COLLATE=Japanese_Japan.932;LC_CTYPE=Japanese_Japan.932;LC_MONETARY=Japanese_Japan.932;LC_NUMERIC=C;LC_TIME=Japanese_Japan.932"
```

krlmlr:
~~Currently breaks tests.~~

Fixes #1950.

Fixes #2387.


krlmlr:
@lionel-: Would you mind reviewing this, especially the definition of the new `SymbolString` and `SymbolVector` classes? The purpose of these classes is to clearly indicate the strings that are supposed to occur in the context of a column name, or a list of column names. This should get rid of encoding problems with column names once and forever.

lionel-:
I'm not sure this helps solving the problem. What we need is:

- to convert symbols to the native encoding before transforming them to strings.
- to tag strings as UTF-8 before returning them to R in case we reencode them.

This PR does not seem to address the second issue (but I don't know if that's currently a problem in dplyr). As for the first issue, it should be mostly resolved on the R side by using rlang's symbol constructors, and could be handled on the C++ side with a simple `Symbol make_symbol(SEXP x)` function.

krlmlr:
Thanks. To me, it looks like `make_symbol()` is not sufficient, because it doesn't allow storing column names that cannot be represented in the native encoding (#1950). My approach allows this by avoiding conversion to `Symbol` altogether to the greatest extent possible. The R side should be doing the same, i.e., convert to `character` as soon as possible and avoid converting to `name` if possible. I'm happy to look at your WIP concerning the conversion to rlang if you want to share it.

I'm wrapping `String` and `CharacterVector`, because

1. Rcpp's implementation currently doesn't handle encoding very well
2. this clearly shows the intent of a variable that holds column name(s)

Tagging strings on input (during construction of `SymbolVector` and `SymbolString`) sounds like a good idea. This makes sure they are always tagged on output. I'd like to approach this after merging this PR.

krlmlr:
We should also supply `get_vars()` and `set_vars()` helpers instead of `.attr("vars")`.

```sh
sed -r -i 's/ ([^. ]+)[.]attr[(]"vars"[)] *= *(.*);/ set_vars(\1, \2);/' src/*.cpp inst/include/dplyr/*.h
sed -r -i 's/([^. (]+)[.]attr[(]"vars"[)]/get_vars(\1)/' src/*.cpp inst/include/dplyr/*.h inst/include/dplyr/Result/*.h
```

lionel-:
>  To me, it looks like make_symbol() is not sufficient, because it doesn't allow storing column names that cannot be represented in the native encoding (#1950).

ohh, I was assuming we could convert to native and back to UTF-8 but you're right that this is not the case. UTF-8 to native is potentially a lossy conversion. So the problem is harder: we should never convert UTF-8 strings to native, even when we convert them to symbols. But we do need a way to get back a string correctly tagged as UTF-8 when we convert the symbol back to a string...

Would it be reasonable to add an encoding attribute when we construct symbols? And `as_character()` would know how to deal with them, and your wrapper classes would use this info as well.

lionel-:
> Would it be reasonable to add an encoding attribute when we construct symbols?

This would change the symbol by reference, so not an option. Wrapping the symbol in a list does not feel right either, but we may not have any other option. Maybe we should try providing an upstream patch.

lionel-:
> This would change the symbol by reference, so not an option. 

hmm, maybe not a bad thing, actually?

krlmlr:
To be honest, a list of symbols where each element has an attribute doesn't look very appealing to me. With a character vector, I can simply call `match()` and other functions, and R knows how to deal with it. 

Why is it crucial to store column names as symbols?

I'm also not sure how to patch upstream so that it remains backward compatible with the current behavior. Maybe with a new type `SYMSXP2`?

lionel-:
> To be honest, a list of symbols where each element has an attribute doesn't look very appealing to me. 

That's not conceptually different from the regular encoding tag. But it's not clear whether that approach would be robust.

> Why is it crucial to store column names as symbols?

Because tidy evaluation is based on the manipulation of expressions, and within the dplyr scope, columns are referred to by their symbol. For instance `select_vars()` returns a character vector, and if we are to simplify NSE by deprecating SE versions, we need to be able to convert back to symbols. The transition `select_vars()` → `select()` occurs in the implementation of `tidyr::nest()` for instance.

There's nothing decided yet on the deprecation of SE versions but it would be desirable to remove this layer of complexity in the tidyverse DSLs. The approach is sound, but this bug in R poses problem.

krlmlr:
> within the dplyr scope, columns are referred to by their symbol

You could also use `get("symbol")` or `.data[["symbol"]]`, and there any UTF-8 string.

I'd be curious to see how deprecation of the SE version looks like. For instance, how would you get rid of `select_()`?

lionel-:
you can program any expression supplied to an NSE function by unquoting right into it. If you unquote a formula, it is inlined as is in the captured expression. Then the evaluation environment gives a special definition to `~` to make the formula self-evaluate.

In this setting formulas really act like reified promises. They obey regular R evaluation semantics, e.g. they evaluate lazily (won't evaluate unless they are actually called). They should enable arbitrary programming with NSE functions.

lionel-:
At least that's the theory, but the port of the tidyverse to tidyeval is going well so I'm hopeful that the approach is sound :)

lionel-:
the more I think about it, the more it feels right to add an attribute to symbols. If the symbol is overwritten it will lose the attribute, unless the new symbol is also generated by rlang. But that seems ok. Your class could read this attribute to know what kind of string to output. Am I missing something?

krlmlr:
I still don't see a compelling argument for using symbols in the first place. But we can easily change the internal representation later, on once this PR is merged.

krlmlr:
@hadley: PTAL.

lionel-:
> I still don't see a compelling argument for using symbols

We are deprecating all unparsed inputs to NSE functions.

krlmlr:
@lionel-: Checked your tibble changes, I think I understand the problem now. Perhaps `symbol()` could add the attribute only if the string is not representable in the native encoding?

lionel-:
The problem is that as far as I know we have no legal way of getting the values of `latin1locale` and `utf8locale` :/ So we'd have to parse the locales ourselves. I think always adding the encoding of tagged strings is an expedient and reasonable solution. Let's see what Hadley thinks.

krlmlr:
We need to convert to native anyway. For strings that aren't native yet, we just check if the result of the conversion is different. For strings that are already native we know that they are representable.

krlmlr:
On the other hand, always having the UTF-8 version in the attribute isn't bad either.

lionel-:
But conversion to native is lossy. I think we should leave the string as it is, like R does. We just need to tag the string back to UTF-8 once we're done.

krlmlr:
I don't mind the lossy conversion if we can retrieve the original information later on. I'd prefer to have symbols in the native encoding, because this is consistent with what R expects.

lionel-:
but we can't retrieve the original information, since the conversion is lossy.

krlmlr:
Didn't you suggest to store the original information as an UTF-8 encoded string in an attribute?

lionel-:
yes, and we shouldn't convert to native. R does not convert to native it just forgets the UTF-8 tag.

krlmlr:
I'm reasonably certain we should convert to native, because R or the user might rely on it at some point. Not converting is calling for trouble -- imagine a user will assume that he can call `eval()` on a symbol we create.

lionel-:
>  imagine a user will assume that he can call eval() on a symbol we create.

Maybe I've been too much focused on the `latin1` case where most unicode characters will be incorrectly translated, whereas the iconv translation would be fine for most CKJ users? Presumably they got UTF-8 strings that are representable in their locale encoding. This problem is hairy :s

lionel-:
I wonder if we should convert to native, but save the original utf-8 string in the attribute?

krlmlr:
Yes, I think this is the safest option. (Even if the original is not UTF-8, we should convert it to UTF-8.)

lionel-:
on hindsight I think we should leave the symbol alone. All reencoding options have deep issues:

- As you mentioned, converting to UTF-8 is not ideal because a user might already have created symbols with the locale encoding in an environment (e.g. with a simple assignment). If we convert to UTF-8, the symbols won't correspond.

- Converting to the native encoding will cause havoc for western latin1 users working with UTF-8 encoded data.

But I have found some workarounds to make R remember the tag. Will post a PR soon on rlang

krlmlr:
We must convert to the native encoding (your option 2), otherwise R won't find the variable that corresponds to the symbol.

lionel-:
I would think this is the user responsibility to convert to native if she had imported UTF-8 data. We cannot make dplyr incompatible with UTF-8 datasets for the whole latin1 Windows population...

Anyway, I found workarounds. A sane one that works 80% of the time, and an insane that works 100% of the time.

I wonder if R will take encodings into account when it performs symbol lookup? I would assume so. If that is the case, our problem is solved.

lionel-:
@krlmlr 

To be more precise, we cannot in general coerce to native because this will create corrupted strings for latin1 users. On the other hand, when it's time to evaluate, it might be reasonable to fixup the expression before evaluation and convert all UTF-8 or latin1 encoded symbols to the native encoding. I could add a C function for doing this in rlang. But this also means that data added to the evaluation environments should also be converted to native symbols before evaluation.

On the whole, this seems like a lot of trouble for a weak payoff. Most of the time the UTF-8 strings in a non-UTF-8 environment are coming from a dataset. Most likely, symbol lookups will refer to variables in that dataset and not the environment. Does this problem really need fixing?

It might still make sense to fix this issue in base R. Tomas is going to fix the encoding tag issue in R soon. While he is working on that, would it make sense to suggest to him that symbol lookup should first convert any UTF-8 encoded symbols to native encoding? This should not be very costly in ordinary situations, and this might handle most problems.

krlmlr:
In R, character encoding is transparent for most uses, except system API interactions (console, file system, ...) and language symbols. Once the data is in R, the user shouldn't be caring about encoding issues at all.

R converts *all* strings to native before coercing to symbol, please look up `installTrChar()` in the R sources. This is not exported, but `Rf_install(Rf_translateChar(.))` is a failsafe alternative. Why should rlang behave any different?

CC @kalibera: Ideally, base R would use UTF-8 consistently to encode all symbols. But looking at the R sources and the comments there leaves little hope that we can change it without breaking existing code. (I suspect there might be code that relies on a particular encoding for installed symbol strings.) I'd be happy to discuss this with you at your convenience.

lionel-:
> R converts all strings to native before coercing to symbol

I don't think that's true. `as.name()` ultimately calls `coerceToSymbol()`, which calls `installChar()`. The latter just forgets about the tag and does not perform any conversion. See https://github.com/wch/r-source/blob/14ee5c17dfab4bed5bc138cc4ef70a4c1e9fb7b9/src/main/coerce.c#L394


> Why should rlang behave any different?

In any case rlang and the tidyverse cannot work like this because they perform this roundtrip all the time: string → symbol → string. Converting to native is too risky because a lossy operation.

krlmlr:
Moving discussion to https://github.com/hadley/rlang/pull/34, because it doesn't seem to affect this PR anymore.

PR #2389, opened 2017-01-28
Change references to functionmap2 to pkgapi

jrnold:
Since `functionmap2` was renamed to `pkgapi` https://github.com/r-pkgs/pkgapi/commit/b8342cc58a241dbb5ed6f2ccc779a7cb9dd97c18

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -33,6 +33,7 @@ Suggests:
     microbenchmark,
     ggplot2,
     mgcv,
+    pkgapi,
     Lahman (>= 3.0-1),
     nycflights13,
     rmarkdown,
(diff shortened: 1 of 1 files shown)


DESCRIPTION
```diff
@@ -33,6 +33,7 @@ Suggests:
     microbenchmark,
     ggplot2,
     mgcv,
+    pkgapi,
```
krlmlr: Why do you want to add it to Suggests: ?

jrnold: I thought it was needed. Is it not?

krlmlr: roxygen2 isn't there either, pkgapi is required to create the API file only as part of creating the documentation with `devtools::document()`.

krlmlr:
Thanks!

Issue #1367, opened 2015-08-31
feature request:  A way to count missing/non-missing rows with n() or similar


rpruim:
There is the tricky bit of deciding what missing means, but perhaps columns could be supplied and only rows which are complete in all of those columns would count as non-missing.  Possible syntax:

``` r
data %>% 
  summarise(
    n = n(),
    incomplete = n - n(col1 : col3, my_other_column)
  )
```

Since `n()` currently has not arguments, `...` could be used to specify the columns in the same way that we currently specify columns for `select()`.

Alternatively, this could be put into additional helper functions `complete()` and/or `incomplete()`.  

``` r
data %>% 
  summarise(
    n = n(),
    incomplete = incomplete(col1 : col3, my_other_column),
    complete = complete(col1 : col3, my_other_column)
  )
```

With no additional arguments, these would act much like 

``` r
nrow(data %>% filter(complete.cases(.)))
```

matthieugomez:
+1

hadley:
I'd prefer not to introduce "variable" ranges inside verbs - `n()` is already a little icky.

matthieugomez:
I agree for column range btw. We juste need a faster
```R
count <- function(...){
    sum(complete.cases(...))
  }
```

hadley:
I think there are two factors at play here:

1.  `n()` needs a convenient way to count non-missing values.

2.  All aggregation functions need a way to work across multiple columns.

I think there are two possible approaches:

1.  Define a series of aggregation functions that take `...` and work like select: `row_n(a:f)`, `row_mean(a:f)`, `row_sd(a:f)`. This would require coming up with a naming convention that doesn't clash with existing functions, and implementing row-wise aggregates ourselves.

1. Define a single function that gives you select style column selection and returns a data frame (or matrix?) that can be fed to other functions: `n(cols(a:f))`, `rowMeans(cols(a:f))`, ... 

Related to #968

matthieugomez:
I like approach 2 because it keeps a clear separation between the function definition and its use within a `dplyr` chain. base R functions like `complete.cases` would work out of the box. Any user-defined function would work as long as it has a syntax similar to complete.cases :  ... a sequence of vectors, matrices and data frames.

Now, I'm not sure why we need to have a new verb `cols`. Why does not `select` work in this example?
``` 
df <- data_frame(
n1 = 1:5,
n2 = c(NA, 1:4),
n3 = c(1:4, NA),
n4 = 1:5
)
[...]

PR #2390, opened 2017-01-28
Add filter_na function

jrnold:
This function filters rows that have any `NA`, or all `NA` values. It does so by generating the expressions to pass to `filter_()` which means that it will work for any backend supported by dplyr.

This is somewhat related to #1367.
There are also a few stack overlow questions asking how to do this: [here](http://stackoverflow.com/questions/26665319/removing-na-in-dplyr-pipe), [here](http://stackoverflow.com/questions/28857653/removing-na-observations-with-dplyrfilter), and [here](http://stackoverflow.com/questions/22353633/filter-for-complete-cases-in-data-frame-using-dplyr-case-wise-deletion). The answers given are to use `na.omit` or `complete.cases` which don't extend to other backends supported by dplyr.


tests/testthat/test-filter-na.R
```diff
@@ -0,0 +1,24 @@
+context("filter_na")
+
+x <- tibble(a = c(1, 2, NA, NA),
```
krlmlr: Using tribble() for construction might make the tests easier to understand.

The first test would be stronger if c[4] was non-NA.

R/filter-na.r
```diff
@@ -0,0 +1,40 @@
...
+  }
+  vars <- select_vars_(names(.data), dots)
+  # construct expression for filter
+  # !is.na(var1), !is.na(var2), ...
+  args <- lapply(vars, function(i) call("!", call("is.na", as.name(i))))
+  if (.all) {
+    # !is.na(var1) | !is.na(var2) | ...
+    if (length(args) > 1) {
+      args <- Reduce(function(x, y) call("|", x, y),
+                     args[2:length(args)], args[[1]])
```
krlmlr: Does `args[-1]` work here?

DESCRIPTION
```diff
@@ -11,9 +11,9 @@ Authors@R: c(
     )
 URL: https://github.com/hadley/dplyr
 BugReports: https://github.com/hadley/dplyr/issues
-Depends: 
+Depends:
```
krlmlr: Could you please undo the unnecessary changes in this file?

R/filter-na.r
```diff
@@ -0,0 +1,40 @@
...
+  if (length(dots) == 0) {
+    dots <- list(~ everything())
+  }
+  vars <- select_vars_(names(.data), dots)
+  # construct expression for filter
+  # !is.na(var1), !is.na(var2), ...
+  args <- lapply(vars, function(i) call("!", call("is.na", as.name(i))))
+  if (.all) {
+    # !is.na(var1) | !is.na(var2) | ...
+    if (length(args) > 1) {
```
krlmlr: Do we need the special case here?

R/filter-na.r
```diff
@@ -0,0 +1,40 @@
...
+#' @param ... Comma separated list of unquoted expressions. Select variables
+#'   which to test for missing values using the same semantics as
+#'   \code{\link[dplyr]{select}}.
+#' @param .all If \code{TRUE}, then remove rows in which all columns are
+#'   missing, else select rows in which any column has a missing value.
+#' @param .dots use \code{filter_na_()} to do standard evaluation. See
+#'   \code{vignette("nse")} for details.
+#' @return An object of the same class as \code{.data}.
+#' @export
+filter_na <- function(.data, ..., .all = FALSE) {
```
krlmlr: Can you give a practical use case for `.all = FALSE`? How do I show only the rows that have NA in at least one column?

krlmlr:
Thinking about it -- maybe it's a better fit for tidyr?

hadley:
`filter_na()` feels to specific to be dplyr to me. But maybe dplyr could have `filter_if()` + `filter_all()` that worked similarly to the summarise and mutate equivalents?

krlmlr:
So the syntax would be `data %>% filter_all(funs(!is.na(.)))` ?

hadley:
Yeah or `data %>% filter_all(not_na)` if we also provided a helper.

R/filter-na.r
jrnold: It is removing all the rows with an `NA` in at least one column, as in `na.omit` or `complete.cases`

R/filter-na.r
jrnold: Yes, or I need to use `init = TRUE`, otherwise `Reduce` generates `NULL | ...`. `purrrr::reduce` is much nicer :-)

R/filter-na.r
jrnold: yeah, my bad.

jrnold:
Are a `filter_if` and `filter_any`  in the works?

jrnold:
The other feature that I had added in [my original version ](https://github.com/jrnold/rubbish/blob/master/R/filter_na.R) that I didn't include here was the option of using `is.finite` instead of `!is.na` for local tables and numeric variables.

R/filter-na.r
jrnold: And for the case of `NA` in at least one column, I hadn't thought about that.

krlmlr:
I'm not aware of any implementation attempts. Would it be difficult to change this PR to implement `filter_if()` and `filter_any()`?

jrnold:
I think that the pull request would likely have to be separate.  I didn't write this function with too much of a view for generality, focusing more on having a usable function for a common case. So I'd more about what the general set of functions would look like and rewrite completely to account for that.   

I'm not sure whether it makes sense to go with generating a `lazy_list` that is passed to `filter` or in this more general case to have to customize the backend.

So would it be a set of functions directly analogous to those for `summarise` and `mutate`? 
What set of functions were you thinking? 
```
filter_if(.tbl, .predicate, .funs,  .or = FALSE, ...)
filter_all(.tbl, .funs, .or = FALSE, ...)
filter_at(.tbl, .cols, .funs, .or = FALSE, ... )
```
where if `.or = TRUE` then the calls are combined by `|`.

hadley:
@lionel- might have API suggestions too. We should make sure we have a solid API before doing any more coding (and we're happy to program this if you don't have the time/motivation)

I think the general functions belong in dplyr, and the specific filter_na probably belongs in tidyr.

lionel-:
If I understand correctly, the suffix in `filter_all()`, `filter_at()` and `filter_if()` imply a column specification? If that's the case I think that's a very good idea. For modelling I often ignore NAs in a specific set of columns corresponding to model predictors, so I can explore the data based on the rows that are used in a model. The `filter_na()` variant could have select semantics for dots.

Maybe the `.or` thing should be a parameter in `funs()`? Or a variant of the `funs()` function might be even better, and could also be used for the predicate in other `_if()` verbs. Maybe an `ifs()` function as suggested here: https://github.com/hadley/dplyr/issues/1977, paired with an `ifs_any()`.

hadley:
I think `.or` is out of scope here - it's not supported by `filter()` so it shouldn't be available in a filter helper. 

I'm going to close this PR in favour a more general issue on `filter_if()` etc. To achieve what you want, you'll eventually be able to just do `filter_all(data, ~ !is.na(.))`, and then that could potentially become a helper in tidyr.

jrnold:
The idea behind `.or` was that in a manual `filter` someone is already writing it, so it's not much extra work to combine the expressions with a binary operator,
```r
filter(data, is.na(foo) | bar > 0 | baz == 1)
```
But since `filter_if` and the other filter helpers will be generating multiple expressions actually having two optional functions would be useful: a binary operator to combine expressions, and a function that takes all the other expressions as arguments. 
Getting a row with at least one missing value either requires an or operator
```
filter(data, !is.na(foo) | !is.na(bar))
```
or a function applied to the results of all the expressions
```
filter(!(is.na(foo) & is.na(bar))
```
So it may make sense to actually have to additional arguments for the filter helpers: `.c` for a binary operator used to combine the expressions and `.f` for a function applied to all of them: 
```
filter_all(data, ~ is.na(.), .c = `|`)
```
or 
```
filter_all(data, ~ is.na(.), .c = `&`, .f = `!`)
```

Maybe that isn't fully thought out, but that was my thought process when suggesting an `.or` argument for the `filter` helpers, even though it isn't necessary for `filter()`.

PR #2412, opened 2017-02-08
Fix style for source files

krlmlr:
1. With an automated code.
2. Added line breaks for clarity.


krlmlr:
I have double-checked that the parsed code is identical before and after the transformation.

PR #2266, opened 2016-11-24
Update test-select-helpers.R (added test for issue #2184)

ngr-t:
added test for issue #2184 
The problem occurs when `one_of()` is called within `select()` or `select_vars()`.
Given below, `one_of()` alone seems to work as expected.
```{R}
> library(dplyr)
> Sepal.Width <- "Species"
> dplyr:::one_of(Sepal.Width, vars = names(iris))
[1] 5
```
Then I write test for the situation `one_of()` is called within `select_vars()` and passed variable name matches an element of the `vars` parameter.

Issue #2184, opened 2016-10-19
one_of() with arguments whose names are used in the columns of results an error


ngr-t:
When I use a variable name used as a column name of .data, select() and one_of() raises an error.
Below is an example. 

``` R
> library(dplyr)
> col <- Species <- "Sepal.Width"
> # variable name "Species" (of course it's one of a column name of iris data) is invalid
> iris %>% select(one_of(Species))
Error: `c(...)` must be a character vector
> iris %>% select(one_of(col)) %>% head
  Sepal.Width
1         3.5
2         3.0
3         3.2
4         3.1
5         3.6
6         3.9
```

krlmlr:
Thanks, confirmed. Would you like to contribute a testthat test?

ngr-t:
I'd gladly to do so.
I feel that the test should be included in [tests/testthat/test-select-helpers.R](https://github.com/hadley/dplyr/blob/master/tests/testthat/test-select-helpers.R), is it OK?

krlmlr:
I was confused. I think you need to do `one_of("Species")`.

ngr-t:
Not. Sorry for poor explanation.
In the example above, the expected return value is a data.frame which has one column of "Sepal.Width" because the content of `Species` is "Sepal.Width".
My point is that accessing with a variable whose name is included by the column names causes an error.
It might be problematic when we write some function which takes data.frames as its parameters and don't know the column names of them.

krlmlr:
I see:

```r
iris %>% select(one_of(Species))
```

doesn't work (and isn't supposed to) if there is no `Species` variable. But as soon as you assign a value to `Species`, it's supposed to work, but doesn't:

```r
Species <- "Petal.Width"
iris %>% select(one_of(Species))
```

It works if you use a variable that doesn't match the column name:

```r
col <- "Petal.Width"
iris %>% select(one_of(col))
```

I think test-select-helpers.R is a good place.

krlmlr:
Are other select helpers affected?

hadley:
It's not actually clear that this is a bug. Generally, you should avoid naming the variables you use inside `one_of()` the same as variables in your data frame. But @lionel- will take a deeper look.

lionel-:
I fixed it by evaluating all calls supplied to `select_vars()` in the calling environment. Except for calls known to work with columns like `:`.

lionel-:
Now fixed on master

PR #2415, opened 2017-02-09
Testing select helpers

krlmlr:
Tests for #2184.

Closes #2266 (=includes it), CC @ngr-t.


ngr-t:
@krlmlr Thanks for fixing my mistake you mentioned in #2266.  Your fix is in the line as what I meant and more polished.

PR #2425, opened 2017-02-14
Combine strips unknown classes. Support integer64.

### LINKED ISSUE #2406: bind_rows() with columns of class "table"
now works, with a warning. Is this intended? CC @zeehio.

Stable version, and 8aa1bdb8fe (the one before #2209):

```
> bind_rows(data_frame(a = table(1:3)), data_frame(a = table(4:6)))
# A tibble: 6 × 1
      a
  <int>
1     1
2     1
3     1
4     1
5     1
6     1
```

Current `master`, and 0833d5b8 (which contains #2209):

```
> bind_rows(data_frame(a = table(1:3)), data_frame(a = table(4:6)))
# A tibble: 6 × 1
            a
  <S3: table>
1           1
2           1
3           1
4           1
5           1
6           1
Warning message:
In bind_rows_(x, .id) :
  Vectorizing 'table' elements may not preserve their attributes
```

zeehio:
This commit closes #2406.

- If we `collect` objects of unknown classes there is risk of losing attributes, so we raise a warning unless we know it is safe.
- `bit64::integer64` is now supported by `dplyr::combine` and `dplyr::bind_rows` (partially deals with https://github.com/hadley/dplyr/issues/1092)
- Tests updated to reflect current status

### DIFF
--- a/inst/include/dplyr/Collecter.h
+++ b/inst/include/dplyr/Collecter.h
@@ -10,23 +10,33 @@
 namespace dplyr {
 
   static inline bool is_class_known(SEXP x) {
-    CharacterVector known = CharacterVector::create(
-      "POSIXct", "factor", "Date", "AsIs", "integer64", "table");
-    int num_known = known.length();
-    if (OBJECT(x)) {
-      CharacterVector classes(Rf_getAttrib(x, R_ClassSymbol));
-      int num_classes = classes.length();
-      for (int i=0;i<num_classes;i++) {
-        for (int j=0;j<num_known;j++) {
-          if (classes[i] == known[j]) {
-            return true;
-          }
-        }
-      }
-    } else {
+    /* C++11 (need initializer lists)
+    static std::set<std::string> known {
+      "POSIXct", "factor", "Date", "AsIs", "integer64", "table"
+    };
+    */
+    /* Begin C++98 workaround */
+    static std::vector<std::string> known;
+    if (known.empty()) {
+      known.push_back("POSIXct");
+      known.push_back("factor");
+      known.push_back("Date");
+      known.push_back("AsIs");
+      known.push_back("integer64");
+      known.push_back("table");
+      std::sort(known.begin(), known.end());
+    }
+    /* End of C++98 workaround */
+    std::vector<std::string> classes_in_x, known_in_x;
+    if (!OBJECT(x)) {
       return true;
     }
-    return false;
+    classes_in_x = Rcpp::as< std::vector<std::string> >(Rf_getAttrib(x, R_ClassSymbol));
+    std::sort(classes_in_x.begin(), classes_in_x.end());
+    std::set_intersection(classes_in_x.begin(), classes_in_x.end(),
+                          known.begin(), known.end(),
+                          std::back_inserter(known_in_x));
+    return not known_in_x.empty();
   }
 
   static inline void warn_loss_attr(SEXP x) {
--- a/inst/include/dplyr/Collecter.h
+++ b/inst/include/dplyr/Collecter.h
@@ -9,19 +9,33 @@
 
 namespace dplyr {
 
-  static inline bool has_classes(SEXP x) {
-    SEXP classes;
-    int num_classes;
+  static inline bool is_class_known(SEXP x) {
+    Rcpp::CharacterVector known = Rcpp::CharacterVector::create(
+      "POSIXct", "factor", "Date", "AsIs", "integer64", "table");
+    int num_known = known.length();
     if (OBJECT(x)) {
-      classes = Rf_getAttrib(x, R_ClassSymbol);
-      num_classes = Rf_length(classes);
-      if (num_classes > 0) {
-        return true;
-      } else {
-        return false;
+      Rcpp::CharacterVector classes(Rf_getAttrib(x, R_ClassSymbol));
+      int num_classes = classes.length();
+      for (int i=0;i<num_classes;i++) {
+        for (int j=0;j<num_known;j++) {
+          if (classes[i] == known[j]) {
+            return true;
+          }
+        }
       }
+    } else {
+      return true;
     }
     return false;
+  }
+
+  static inline void warn_loss_attr(SEXP x) {
+    /* Attributes are lost with unknown classes */
+    if (!is_class_known(x)) {
+      SEXP classes = Rf_getAttrib(x, R_ClassSymbol);
+      Rf_warning("Vectorizing '%s' elements may not preserve their attributes",
+                 CHAR(STRING_ELT(classes, 0)));
+    }
   }
 
   static inline bool all_logical_na(SEXP x, SEXPTYPE xtype) {
@@ -90,6 +104,7 @@ namespace dplyr {
     }
 
     void collect_sexp(const SlicingIndex& index, SEXP v) {
+      warn_loss_attr(v);
       Vector<RTYPE> source(v);
       STORAGE* source_ptr = Rcpp::internal::r_vector_start<RTYPE>(source);
       for (int i=0; i<index.size(); i++) {
@@ -105,6 +120,7 @@ namespace dplyr {
     Collecter_Impl(int n_): data(n_, NA_REAL) {}
 
     void collect(const SlicingIndex& index, SEXP v) {
+      warn_loss_attr(v);
       NumericVector source(v);
       double* source_ptr = source.begin();
       for (int i=0; i<index.size(); i++) {
@@ -118,8 +134,8 @@ namespace dplyr {
 
     inline bool compatible(SEXP x) {
       int RTYPE = TYPEOF(x);
-      return (RTYPE == REALSXP && !has_classes(x)) ||
-             (RTYPE == INTSXP && !has_classes(x)) ||
+      return (RTYPE == REALSXP && !Rf_inherits(x, "POSIXct") && !Rf_inherits(x, "Date")) ||
+             (RTYPE == INTSXP && !Rf_inherits(x, "factor")) ||
              all_logical_na(x, RTYPE);
     }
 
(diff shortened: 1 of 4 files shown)


tests/testthat/test-combine.R
```diff
@@ -160,5 +160,14 @@ test_that("combine works with NA and complex (#2203)", {
   expect_equal(works3, expected_result)
 })
 
+test_that("combine works with integer64", {
+  if (!test_srcs$has("bit64"))
```
krlmlr: What triggers this test?

tests/testthat/helper-combine.R
```diff
@@ -183,14 +199,17 @@ prepare_table_with_coercion_rules <- function() {
     )
 
     pairs$warning[i] <- give_a_warning(
-      item1, item2, class1, class2,
-      all_na1, all_na2, known_to_dplyr1, known_to_dplyr2
-    )
+      item1, item2,
+      class1, class2,
+      known_to_dplyr1, known_to_dplyr2,
+      can_be_combined = pairs$can_combine[i])
```
krlmlr: Could you please keep the closing paren on a separate line?

tests/testthat/test-combine.R
```diff
@@ -160,5 +160,14 @@ test_that("combine works with NA and complex (#2203)", {
...
+    skip("No bit64")
+  expect_equal(combine(bit64::as.integer64(2^34),
```
krlmlr: Could you please format function calls as described in https://github.com/krlmlr/r-pkgs/blob/39555708cb0a4aee19711c85963ad9a33d15defd/style.rmd#indentation?

tests/testthat/test-combine.R
zeehio: Given that this PR adds support for `integer64` objects I thought it made sense to check that numbers were combined as the `bit64` package expects.

I can replace the `bit64::as.integer64` calls with `structure(some number, class="integer64")` and drop the dependency, if you prefer.

zeehio: It triggers the line: https://github.com/hadley/dplyr/pull/2425/files#diff-3c72da1285a2222bfe76e4ea4d89a48dR456

krlmlr: It's just that I can't see how `test_srcs$has("bit64")` can be true. If the package is suggested, you shouldn't need to test for its presence anyway.

zeehio: I misunderstood how `test_srcs` and `Suggests` need to work. 
As I understand it now there are 3 options to implement this test:

1. Suggest: bit64. With that `if (!test_srcs$has("bit64"))` is not required *(current solution)*
2. Do not suggest `bit64`. Then I need to use `if (!test_srcs$has("bit64"))` and the test is skipped if `bit64` is not available.
3. Avoid `bit64` by using `structure(a number, class="integer64")`.

- Option 1) is bad because it adds another dependency (although it is just a suggestion). 
- Option 2) is bad because the test is not always run, so it may be skipped in some builds inadvertently.
- Option 3) is bad because if `bit64` changes and we do not follow then we may not be checking the right thing.

I chose `option 1` because it is the safest one, but feel free to tell me whatever you prefer.

inst/include/dplyr/Collecter.h
```diff
@@ -429,24 +447,14 @@ namespace dplyr {
...
       if (Rf_inherits(model, "Date"))
         return new TypedCollecter<REALSXP>(n, get_date_classes());
-      if (has_classes(model)) {
-        SEXP classes = Rf_getAttrib(model, R_ClassSymbol);
-        Rf_warning("Vectorizing '%s' elements may not preserve their attributes",
-                   CHAR(STRING_ELT(classes, 0)));
-        return new TypedCollecter<REALSXP>(n, classes);
-      }
+      if (Rf_inherits(model, "integer64"))
+        return new TypedCollecter<REALSXP>(n, Rcpp::CharacterVector::create("integer64"));
```
krlmlr: Please remove `Rcpp::` here and everywhere else. We have `using namespace Rcpp` practically everywhere.

tests/testthat/test-combine.R
```diff
@@ -160,5 +160,11 @@ test_that("combine works with NA and complex (#2203)", {
   expect_equal(works3, expected_result)
 })
 
+test_that("combine works with integer64", {
+  expect_equal(
+    combine(bit64::as.integer64(2^34), bit64::as.integer64(2^35)),
+            c(bit64::as.integer64(2^34), bit64::as.integer64(2^35)))
```
krlmlr: Indent looks off.  Also, I think `bit64::as.integer64(c(2^34, 2^35))` is a bit clearer.

tests/testthat/test-combine.R
```diff
@@ -160,5 +160,11 @@ test_that("combine works with NA and complex (#2203)", {
   expect_equal(works3, expected_result)
 })
 
+test_that("combine works with integer64", {
```
krlmlr: Can you please refer to the issue in the test title?

tests/testthat/helper-combine.R
```diff
@@ -156,7 +172,8 @@ prepare_table_with_coercion_rules <- function() {
     num_with_class = structure(4.5, class = "num_with_class")
   )
 
-  special_non_vector_classes <- c("factor", "POSIXct", "Date")
+  special_non_vector_classes <- c("factor", "POSIXct", "Date",
```
krlmlr: Please put a newline after the opening and before the closing parens.

krlmlr:
@hadley: PTAL.

inst/include/dplyr/Collecter.h
```diff
@@ -9,21 +9,35 @@
 
 namespace dplyr {
 
-  static inline bool has_classes(SEXP x) {
-    SEXP classes;
-    int num_classes;
+  static inline bool is_class_known(SEXP x) {
+    CharacterVector known = CharacterVector::create(
```
hadley: This really should be a list of character vector (or `std::vector< std::vector<std:string> >`). That would simplify the comparison below

inst/include/dplyr/Collecter.h
```diff
@@ -9,21 +9,35 @@
...
+      return true;
     }
     return false;
   }
 
+  static inline void warn_loss_attr(SEXP x) {
+    /* Attributes are lost with unknown classes */
+    if (!is_class_known(x)) {
+      SEXP classes = Rf_getAttrib(x, R_ClassSymbol);
+      Rf_warning("Vectorizing '%s' elements may not preserve their attributes",
```
hadley: Needs to mention column name?

inst/include/dplyr/Collecter.h
krlmlr: Or perhaps a static `std::map<std::string>`?

inst/include/dplyr/Collecter.h
krlmlr: This looks like it's difficult to achieve, the `Collecter` currently has no notion of the column name.?

inst/include/dplyr/Collecter.h
zeehio: Your C++ skills and your Rcpp skills are higher than mine.

The equivalent `R` code for the function would be:
```r
is_class_known <- function(x) {
  known <- c("POSIXct", "factor", "Date", "AsIs", "integer64", "table")
  length(intersect(class(x), known)) > 0
}
```

Using a list of character vectors would be useful if we wanted to match all the `x` classes. Here, in order to allow users to specify additional custom classes, I considered `x` of a "known class" if *any* of its classes is "known".

For this reason, I believe using `std::vector <std::string>` is good enough (with respect to `std::vector< std::vector<std::string> >`. 

I have replaced the double `for` loop with `std::set_intersection`, relying more on the STL. Is that better?

inst/include/dplyr/Collecter.h
zeehio: Yes, for instance `dplyr::combine(structure(4L, class="dummy_class"), 4.3)` has no column name to report.

```r
dplyr::combine(structure(4L, class="dummy_class"), 4.3)
[1] 4.0 4.3
Warning message:
In combine_all(args) :
  Vectorizing 'dummy_class' elements may not preserve their attributes
```

inst/include/dplyr/Collecter.h
```diff
@@ -9,19 +9,43 @@
...
+      known.push_back("factor");
+      known.push_back("Date");
+      known.push_back("AsIs");
+      known.push_back("integer64");
+      known.push_back("table");
+      std::sort(known.begin(), known.end());
+    }
+    /* End of C++98 workaround */
+    std::vector<std::string> classes_in_x, known_in_x;
+    if (!OBJECT(x)) {
```
hadley: Maybe extract this out into separate `inherits_from(SEXP x, const std::vector<std::string>& classes)` ?

inst/include/dplyr/Collecter.h
```diff
@@ -9,19 +9,43 @@
...
+    /* Begin C++98 workaround */
+    static std::vector<std::string> known;
+    if (known.empty()) {
+      known.push_back("POSIXct");
+      known.push_back("factor");
+      known.push_back("Date");
+      known.push_back("AsIs");
+      known.push_back("integer64");
+      known.push_back("table");
+      std::sort(known.begin(), known.end());
```
hadley: Maybe use `std::set<std::string>` instead?

inst/include/dplyr/Collecter.h
hadley: I guarantee that people will want this. 

I'm ok with not tackling it here, but we should create a meta-issue for ensuring that all error messages (where possible) mention the column name.

zeehio:
All comments covered (except for the column name message, will be dealt on #2448)

hadley:
Thanks! This should be the last change related to column combining incorporated into dplyr. All the other problems will have to wait for vctrs.

PR #2430, opened 2017-02-15
fix translate sql

mtreadwell:
In commit 421b569 the function `win_over()` was added. Today, I used `lag()` with `src_postgres()` on the master and I got the following error:

    Error in lag(value, order = "date") : could not find function "over"

When I looked in translate-base-sql.R, it looks like you changed all of the other base_win functions to use `win_over()` but you might have just missed `lag()`.

### DIFF
--- a/R/translate-sql-base.r
+++ b/R/translate-sql-base.r
@@ -211,7 +211,7 @@ base_win <- sql_translator(
     )
   },
   lag = function(x, n = 1L, default = NA, order = NULL) {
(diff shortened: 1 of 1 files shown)


hadley:
Oops, thanks

krlmlr:
I'm about to update NEWS, will include this.

hadley:
Fixes for bugs introduced only in dev need news entries

PR #2434, opened 2017-02-19
minor documentation edit

jrnold:
Make the description of the `...` argument in `recode` more clear.
Currently the documentation does not state whether the names or values of the arguments are the replacement, leaving the reader to infer it from the examples.
This is particularly confusing since `forcats::fct_recode` uses the opposite convention (names are replacements, values are the values to be replaced).


hadley:
Thanks!

PR #2445, opened 2017-02-19
Allow one join suffix to be the empty string

### LINKED ISSUE #2228: 'join' causes R to crash on empty-string suffix
Two tibbles with identical column names are joined. If one uses an _empty string_ as one of the two suffixes for the column names, `left_join` runs into an end-less loop, causing R to crash:

```
library(  dplyr )

tbl1 <- tibble( key = LETTERS[1:10], val = 1:10 )
tbl2 <- tibble( key = LETTERS[1:10], val = 11:20 )

# This here works as expected
left_join( tbl1, tbl2, by="key", suffix = c( ".A", ".B" ) )

# This here crashes R
left_join( tbl1, tbl2, by="key", suffix = c( "", ".B" ) )
```

I guess the implementation somewhere assumes that a column without suffix is the original one, but not the joined one, and so misses some stop condition.

---

Session info:
```
R version 3.3.2 (2016-10-31)
Platform: x86_64-pc-linux-gnu (64-bit)
Running under: Ubuntu 14.04.5 LTS

locale:
 [1] LC_CTYPE=fi_FI.UTF-8       LC_NUMERIC=C              
 [3] LC_TIME=en_GB.UTF-8        LC_COLLATE=en_GB.UTF-8    
 [5] LC_MONETARY=en_GB.UTF-8    LC_MESSAGES=en_GB.UTF-8   
 [7] LC_PAPER=en_GB.UTF-8       LC_NAME=C                 
 [9] LC_ADDRESS=C               LC_TELEPHONE=C            
[11] LC_MEASUREMENT=en_GB.UTF-8 LC_IDENTIFICATION=C       

attached base packages:

other attached packages:
[1] dplyr_0.5.0

[1] magrittr_1.5   R6_2.1.2       assertthat_0.1 tools_3.3.2    DBI_0.4-1     
[6] tibble_1.1     Rcpp_0.12.6   
```

krlmlr:
Avoids endless loop.

Fixes #2228.

@hadley: PTAL.

### DIFF
--- a/src/join_exports.cpp
+++ b/src/join_exports.cpp
@@ -26,6 +26,10 @@ DataFrame subset_join(DataFrame x, DataFrame y,
                       CharacterVector by_x, CharacterVector by_y ,
                       const std::string& suffix_x, const std::string& suffix_y,
                       CharacterVector classes) {
+  if (suffix_x.length() == 0 && suffix_y.length() == 0) {
+    stop("Cannot use empty string for both x and y suffixes");
+  }
+
   // first the joined columns
   DataFrameJoinVisitors join_visitors(x, y, by_x, by_y, false);
   int n_join_visitors = join_visitors.size();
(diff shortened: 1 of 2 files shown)


src/join_exports.cpp
```diff
@@ -75,12 +79,16 @@ DataFrame subset_join(DataFrame x, DataFrame y,
...
 
-      while (
-        (std::find(y_columns.begin(), y_columns.end(), col_name.get_sexp()) != y_columns.end()) ||
-        (std::find(names.begin(), names.begin() + i, col_name.get_sexp()) != names.begin() + i)
-      ) {
-        col_name += suffix_x;
+      // we suffix by .x if this column is in y_columns (and if the suffix is not empty)
+      if (suffix_x.length() > 0) {
+        while (
+          (std::find(y_columns.begin(), y_columns.end(), col_name.get_sexp()) != y_columns.end()) ||
```
hadley: This seems a bit complicated as a while condition

krlmlr: Yes. There is also duplication with the R function `unique_names()`, ideally we'd have one C++ routine for creating column names. Can we merge as-is and postpone cleanup?

hadley: Sounds good

Issue #2462, opened 2017-02-22
Travis & AppVeyor Reproduced: `mutate_if()` w/ single column


wibeasley:
This is the #2114 problem, but reproduced with Travis and AppVeyor.  I hope this isn't seen as unnecessary pestering --I'm concerned that this corner case isn't being caught in the existing dplyr test suite.  When this issue was created in September, it was failing only on my machines with r-devel.  Now it's on r-release (on both AppVeyor and Travis).

Here's a branch off a [simple package](https://github.com/OuhscBbmc/referral-exposure/blob/3e4de6ad09641b954640756591bd7600408b0320/R/function-1.R), and its test results in  [Travis](https://travis-ci.org/OuhscBbmc/referral-exposure/builds/204262554) and [AppVeyor](https://ci.appveyor.com/project/wibeasley/referral-exposure/build/1.0.42).

The core problematic function is:
```r
  ds_1a <- data.frame(
    id               = 1L:5L,
    stringsAsFactors = FALSE
  )

  ds_1b <- dplyr::mutate_if(
    ds_1a,
    is.character,
    function(x) dplyr::coalesce(x, "-") #Replace NAs with blanks
  )
#> Error in evalq(sys.calls(), <environment>): Vector 1 has type 'character' not 'integer'
```

The error `Vector 1 has type 'character' not 'integer'` is thrown by all the cells in the test matrices, including:
* https://travis-ci.org/OuhscBbmc/referral-exposure/jobs/204262555#L750
* https://travis-ci.org/OuhscBbmc/referral-exposure/jobs/204262556#L762
* https://ci.appveyor.com/project/wibeasley/referral-exposure/build/1.0.42/job/es41v7ci7dbwhkh1#L253
* https://ci.appveyor.com/project/wibeasley/referral-exposure/build/1.0.42/job/idhbr46sco43arky#L257

---
@hadley, you recommended a [slightly different reprex](https://github.com/hadley/dplyr/issues/2114#issuecomment-281652615).  I [tried that](https://github.com/OuhscBbmc/referral-exposure/blob/3e4de6ad09641b954640756591bd7600408b0320/R/function-3.R) (after qualifying the package of `coalesce()`).  That produced the message [`Error: Additional arguments should be named`](
[...]

Issue #2114, opened 2016-09-08
`mutate_if()` w/ single column breaks in r-devel


wibeasley:
In the recent r-devel, `mutate_if()` isn't working under some conditions when the dataset has only one column.

``` r
ds_1a <- data.frame(
  id               = 1L:5L,
  stringsAsFactors = FALSE
)

ds_1b <- dplyr::mutate_if(
  ds_1a,
  is.character,
  function(x) dplyr::coalesce(x, "-") #Replace NAs with blanks
)
```

The error is

```
Error in evalq(sys.calls(), <environment>) : 
  Vector 1 has type 'character' not 'integer'
```

But everything is fine with a second column:

``` r
ds_2a <- data.frame(
  id               = 1L:5L,
  name             = c("a", "b", NA_character_, "c", "d"),
  stringsAsFactors = FALSE
)
ds_2b <- dplyr::mutate_if(
  ds_2a,
  is.character,
  function(x) dplyr::coalesce(x, "-") #Replace NAs with blanks
)
```

output:

```
> ds_2b
  id name
1  1    a
2  2    b
3  3    -
4  4    c
5  5    d
```

I think the problem is with the predicate.  That error is identical to if coalesce is called directly (when the two arguments have different data types):

``` r
> dplyr::coalesce(1L:5L, "")
Error: Vector 1 has type 'character' not 'integer'
```

The problem is avoided on Linux & Windows with 3.3.1.  But it's occurring [Travis](https://travis-ci.org/OuhscBbmc/REDCapR/jobs/158287495) and a local Windows machine:

```
> sessionInfo()
R Under development (unstable) (2016-09-06 r71220)
Platform: x86_64-w64-mingw32/x64 (64-bit)
Running under: Windows >= 8 x64 (build 9200)

locale:
[1] LC_COLLATE=English_United States.1252 
[2] LC_CTYPE=English_United States.1252   
[3] LC_MONETARY=English_United States.1252
[4] LC_NUMERIC=C                          
[5] LC_TIME=English_United States.1252    

attached base packages:
[1] stats     graphics  grDevices utils     datasets  methods  
[7] base     

other attached packages:
[1] REDCapR_0.9.5.9004

loaded via a namespace (and not attached):
 [1] httr_1.2.1     lazyeval_0.2.0 magrittr_1.5   assertthat_0.1
 [5] R6_2.1.3       DBI_0.5        tools_3.4.0    dplyr_0.5.0   
[...]

PR #2463, opened 2017-02-22
test for mutate-if w/ single column

wibeasley:
#2114, #2462

### DIFF
--- a/tests/testthat/test-colwise.R
+++ b/tests/testthat/test-colwise.R
@@ -69,6 +69,23 @@ test_that("empty selection does not select everything (#2009, #1989)", {
   expect_equal(mtcars, mutate_if(mtcars, is.factor, as.character))
 })
 
+test_that("mutate_if w/ single column (#2114, #2462)", {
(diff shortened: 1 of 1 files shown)


hadley:
This would be more compelling if it actually failed.

wibeasley:
> This would be more compelling if it actually failed.

Sorry.  I referenced the reason in https://github.com/hadley/dplyr/issues/2462#issuecomment-281835027, and it would probably make sense to give context here.

I added it to the suite for two reasons:
1. It verifies that this basic dplyr scenario currently passes, and therefore adds value for the sake of regression testing.
1. Its passing state here contrasts with the environment in which #2462 fails.  I am guessing this highlights the unexpectedly different behavior with *vs* without the dplyr package being attached.

hadley:
Unfortunately this doesn't really seem useful to me. I'm in a big push for dplyr right now, so I recommend reopening in a couple of months when I don't have so much on my plate.

wibeasley:
Ok.  Thanks for your time and all the wonderful packages.

PR #2468, opened 2017-02-23
Minor fixes

krlmlr:
- tbl_sql() works with string argument
- memdb_frame() returns visible output

### DIFF
--- a/R/db-sqlite.r
+++ b/R/db-sqlite.r
@@ -16,7 +16,8 @@
 #'   unlikely to conflict with an existing table.
 #' @export
 memdb_frame <- function(..., .name = random_table_name()) {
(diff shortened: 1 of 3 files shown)


R/db-sqlite.r
```diff
@@ -16,7 +16,8 @@
 #'   unlikely to conflict with an existing table.
 #' @export
 memdb_frame <- function(..., .name = random_table_name()) {
-  copy_to(src_memdb(), data_frame(...), name = .name)
+  x <- copy_to(src_memdb(), data_frame(...), name = .name)
```
hadley: `(copy_to(src_memdb(), data_frame(...), name = .name))` would be slightly simpler

PR #2478, opened 2017-02-27
Combine character and factor show warning. Closes #2317

### LINKED ISSUE #2317: no warning message about factor to character coercion shown when binding dataframes
With [dplyr v0.4.0](https://github.com/hadley/dplyr/blob/master/NEWS.md#minor-improvements) warning messages were introduced when a coercion from type factor to character is needed to produce a valid result using `bind_rows()` and other operations.

If I run the following, the warning message is shown as expected:
```R
> bind_rows(data_frame(a = factor("a")), data_frame(a = "a"))
# A tibble: 2 × 1
      a
  <chr>
1     a
2     a
Warnmeldung:
In bind_rows_(x, .id) :
  binding factor and character vector, coercing into character vector
```
Now if I switch the order of the dataframes, no warning message appears although the same coercion from type `factor` to `character` is applied to the second dataframe:
```R
> bind_rows(data_frame(a = "a"), data_frame(a = factor("a")))
# A tibble: 2 × 1
      a
  <chr>
1     a
2     a
```

This behaviour seems confusing and in my eyes the warning message should appear in both cases. 

I'm using `dplyr 0.5.0` with `R 3.3.2` on `Ubuntu Linux 16.04 LTS`.

zeehio:
When we collect a factor into a character vector we should warn as well of the coercion of the factor to a character for consistency. This PR adds the warning and fixes the test case.

Closes #2317

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -347,6 +347,9 @@
 
 * `combine()` accepts `NA` values (#2203, @zeehio)
 
+* `combine()` and `bind_rows()` with character and factor types now always warn
+  about the coercion to character (#2317, @zeehio)
+
 # dplyr 0.5.0
 
 ## Breaking changes
--- a/inst/include/dplyr/Collecter.h
+++ b/inst/include/dplyr/Collecter.h
@@ -228,6 +228,7 @@ namespace dplyr {
 
     void collect_factor(const SlicingIndex& index, IntegerVector source) {
       CharacterVector levels = get_levels(source);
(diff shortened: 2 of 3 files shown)


hadley:
Thanks! Would you mind adding a bullet to `NEWS.md` too?

zeehio:
Done! (I amended the commit) :smiley:

hadley:
Thanks!

PR #2483, opened 2017-02-28
Adds mutate support to reference new left-side expressions in lazy-tbl

javierluraschi:
See https://github.com/hadley/dplyr/issues/2481. This adds support for:

```
df <- memdb_frame(x = runif(100), y = runif(100))
df %>% mutate(x1 = x, x2 = x1)
```

Notice that this won't work in `data.frame` nor `lazy-tbl`:

```
df <- memdb_frame(x = runif(100), y = runif(100))
df %>% mutate(x2 = x1, x1 = x)
```

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,8 @@
 # dplyr 0.5.0.9000
 
+* The SQL translation of `mutate()` now accepts references to new
+  expressions (#2481).
+
 * `bind_cols()` now calls `tibble::repair_names()` to ensure that all
   names are unique (#2248).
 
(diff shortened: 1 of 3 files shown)


hadley:
This will need to wait until @lionel-'s tidyeval work is done. 

It also doesn't feel like quite the right approach to me, but I can take what you have an finish it off.

javierluraschi:
The change only wraps around in `SELECT` statements each variable that has internal dependencies, so for instsance:

```
sql_render(df %>% mutate(cost2 = x, cost3 = x, cost = x, monthly = cost / 12, weekly = monthly / 4, cost4 = x, cost5 = x))
```

Renders as,

```
<SQL> SELECT `x`, `cost`, `monthly`, `x` AS `cost2`, `x` AS `cost3`, `monthly` / 4.0 AS `weekly`, `x` AS `cost4`, `x` AS `cost5`
FROM (SELECT `x`, `cost`, `cost` / 12.0 AS `monthly`
FROM (SELECT `x`, `x` AS `cost`
FROM `yeffewzjdi`))
```

which is much better than the `sparklyr` implementation which was wrapping each column.

javierluraschi:
@hadley sounds good, I'll leave it as it is then. Thanks!

PR #2486, opened 2017-03-02
Add difftime support to combine and bind_rows

zeehio:
- We are already supporting mutate with difftime variables.
- I'm preparing a PR to make mutate use `Collecter.h` (sharing the coercion rules and algorithm used by `combine` and `bind_rows`).
- To preserve backwards compatibility in mutate with difftime objects I need `Collecter.h` to support `difftime` objects.

This is the first out of two PRs for #1892. The next PR will depend on this one but this can be reviewed (and merged) already (improves dplyr compatibility).

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -350,6 +350,8 @@
 * `combine()` and `bind_rows()` with character and factor types now always warn
   about the coercion to character (#2317, @zeehio)
 
+* `combine()` and `bind_rows()` accept `difftime` objects.
+
 # dplyr 0.5.0
 
 ## Breaking changes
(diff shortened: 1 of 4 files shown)


hadley:
Can you take a look at the failure on oldrel please? You might need to generate the expected output directly, rather than relying on c

zeehio:
Done

krlmlr:
I reviewed the other PR before this. Can you rebase so that this PR uses the same commits as the other, so that the comments are carried over here?

krlmlr:
Oh, it does already :-(

zeehio:
Let's work on the other PR directly.

PR #2487, opened 2017-03-02
Make mutate use collecter h

### LINKED ISSUE #1892: Integer not automatically coerced to double

zeehio:
This PR is a continuation to #2486 and closes #1892.

`mutate(col2 = fun(col1))` on a grouped data frame calls `fun` once per group.

It used to require that `fun` returns the exact same type on each call. That is not desirable in functions that may return different (but compatible) types, such as integer and numeric.

This PR changes that behavior, so the returned vectors from each of the `fun` calls are combined using the same coercion rules than `combine` and `bind_rows`, defined in `Collecter.h`.

Comments are welcome. Feel free to be picky, so I can improve a bit my C++ and Rcpp skills. :smiley:

### DIFF
--- a/inst/include/dplyr/Collecter.h
+++ b/inst/include/dplyr/Collecter.h
@@ -337,8 +337,6 @@ namespace dplyr {
     }
 
   private:
-    RObject tz;
-
     void update_tz(SEXP v) {
       RObject v_tz(Rf_getAttrib(v, Rf_install("tzone")));
       // if the new tz is NULL, keep previous value
@@ -357,6 +355,7 @@ namespace dplyr {
       }
     }
 
+    RObject tz;
   };
 
   class DifftimeCollecter : public Collecter_Impl<REALSXP> {
@@ -396,19 +395,23 @@ namespace dplyr {
     }
 
   private:
-    RObject units;
-    SEXP types;
-
     bool has_valid_time_unit(SEXP x) {
-      SEXP x_units(Rf_getAttrib(x, Rf_install("units")));
-
-      if (Rf_isNull(x_units)) {
-        return false;
+      static std::set<std::string> valid_units;
+      if (valid_units.empty()) {
+        valid_units.insert("secs");
+        valid_units.insert("mins");
+        valid_units.insert("hours");
+        valid_units.insert("days");
+        valid_units.insert("weeks");
       }
 
+      SEXP x_units(Rf_getAttrib(x, Rf_install("units")));
+      if (TYPEOF(x_units) != STRSXP) {
+        return false;
+      }
       std::string x_units_c = CHAR(STRING_ELT(x_units, 0));
-      if (x_units_c == "secs" || x_units_c == "mins" || x_units_c == "hours" ||
-          x_units_c == "days" || x_units_c == "weeks") {
+
+      if (valid_units.find(x_units_c) != valid_units.end()) {
         return true;
       } else {
         return false;
@@ -460,25 +463,27 @@ namespace dplyr {
       }
     }
 
-    double time_conversion_factor(RObject v_units) {
+    double time_conversion_factor(SEXP v_units) {
       // Acceptable units based on r-source/src/library/base/R/datetime.R
-      double factor = 1;
-      std::string v_units_c = Rcpp::as<std::string>(v_units);
-      if (v_units_c == "secs") {
-        factor = 1;
-      } else if (v_units_c == "mins") {
-        factor = 60;
-      } else if (v_units_c == "hours") {
-        factor = 60*60;
-      } else if (v_units_c == "days") {
-        factor = 60*60*24;
-      } else if (v_units_c == "weeks") {
-        factor = 60*60*24*7;
-      } else {
-        stop("Cannot convert %s to seconds", v_units_c.c_str());
+      static std::map<std::string, double> unit_factor_map;
+      if (unit_factor_map.empty()) {
+        unit_factor_map["secs"] = 1;
+        unit_factor_map["mins"] = 60;
+        unit_factor_map["hours"] = 60*60;
+        unit_factor_map["days"] = 60*60*24;
+        unit_factor_map["weeks"] = 60*60*24*7;
       }
-      return factor;
+      std::map<std::string, double>::iterator unit_factor =
+        unit_factor_map.find(CHAR(STRING_ELT(v_units,0)));
+      if (unit_factor != unit_factor_map.end()) {
+        return unit_factor->second;
+      } else {
+        stop("Cannot convert %s to seconds", CHAR(STRING_ELT(v_units, 0)));
+      }
     }
+
+    RObject units;
+    SEXP types;
 
   };
 
--- a/inst/include/dplyr/Collecter.h
+++ b/inst/include/dplyr/Collecter.h
@@ -23,14 +23,9 @@ namespace dplyr {
   }
 
   static bool is_class_known(SEXP x) {
-    /* C++11 (need initializer lists)
-    static std::set<std::string> known_classes {
-      "difftime", "POSIXct", "factor", "Date", "AsIs", "integer64", "table"
-    };
-    */
-    /* Begin C++98 workaround */
     static std::set<std::string> known_classes;
     if (known_classes.empty()) {
+      known_classes.insert("hms");
       known_classes.insert("difftime");
       known_classes.insert("POSIXct");
       known_classes.insert("factor");
@@ -39,7 +34,6 @@ namespace dplyr {
       known_classes.insert("integer64");
       known_classes.insert("table");
     }
-    /* End of C++98 workaround */
     if (OBJECT(x)) {
       return inherits_from(x, known_classes);
     } else {
@@ -369,8 +363,8 @@ namespace dplyr {
   public:
     typedef Collecter_Impl<REALSXP> Parent;
 
-    DifftimeCollecter(int n, SEXP units_) :
-      Parent(n), units(units_) {}
+    DifftimeCollecter(int n, SEXP units_, SEXP types_) :
+      Parent(n), units(units_), types(types_) {}
 
     void collect(const SlicingIndex& index, SEXP v) {
       if (Rf_inherits(v, "difftime")) {
@@ -381,7 +375,7 @@ namespace dplyr {
     }
 
     inline SEXP get() {
-      set_class(Parent::data, "difftime");
+      set_class(Parent::data, types);
       if (!units.isNULL()) {
         Parent::data.attr("units") = units;
       }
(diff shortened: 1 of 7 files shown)


tests/testthat/test-binds.R
```diff
@@ -457,3 +457,10 @@ test_that("bind_rows rejects data frame columns (#2015)", {
     fixed = TRUE
   )
 })
+
+test_that("bind_rows accepts difftime objects", {
```
krlmlr: Test case for `"hms"`:

```r
  df1 <- data.frame(x = hms::hms(hours = 1))
  df2 <- data.frame(x = as.difftime(1, units = "mins"))
  res <- bind_rows(df1, df2)
  expect_equal(res$x, hms::hms(hours = 1, minutes = 1))
```

inst/include/dplyr/Collecter.h
```diff
@@ -364,6 +365,103 @@ namespace dplyr {
...
+    void collect(const SlicingIndex& index, SEXP v) {
+      if (Rf_inherits(v, "difftime")) {
+        collect_difftime(index, v);
+      } else if (all_logical_na(v, TYPEOF(v))) {
+        Parent::collect(index, v);
+      }
+    }
+
+    inline SEXP get() {
+      set_class(Parent::data, "difftime");
```
krlmlr: There is also the `"hms"` class that inherits from `"difftime"`, I guess we'd like to return `"hms"` objects if the first object is of that class.

inst/include/dplyr/Collecter.h
```diff
@@ -364,6 +365,103 @@ namespace dplyr {
...
+        if (STRING_ELT(units, 0) == STRING_ELT(v_units,0)) {
+          Parent::collect(index, v);
+        } else {
+          // If units are different convert the existing data and the new vector
+          // to seconds (following the convention on
+          // r-source/src/library/base/R/datetime.R)
+          double factor_data = time_conversion_factor(units);
+          if (factor_data != 1.0) {
+            for (int i=0; i<Parent::data.size(); i++) {
+              Parent::data[i] = factor_data*Parent::data[i];
```
krlmlr: Do we need to copy the data before in-place modification?

inst/include/dplyr/Collecter.h
```diff
@@ -364,6 +365,103 @@ namespace dplyr {
...
+            }
+          }
+          units = wrap("secs");
+          double factor_v = time_conversion_factor(v_units);
+          NumericVector v_sec(v);
+          double* v_sec_ptr = v_sec.begin();
```
krlmlr: This looks unsafe, better use an iterator and advance that (in addition to `i`).

inst/include/dplyr/Collecter.h
```diff
@@ -364,6 +365,103 @@ namespace dplyr {
...
+        // assume seconds
+        v_units = wrap("secs");
+      }
+      if (units.isNULL()) {
+        // if current unit is NULL, grab the new one
+        units = v_units;
+        // then collect the data:
+        Parent::collect(index, v);
+      } else {
+        // We already units, is the new vector with the same units?
```
krlmlr: Please check grammar.

inst/include/dplyr/Collecter.h
```diff
@@ -364,6 +365,103 @@ namespace dplyr {
...
+
+    void collect_difftime(const SlicingIndex& index, SEXP v) {
+      RObject v_units(Rf_getAttrib(v, Rf_install("units")));
+      if (v_units.isNULL()) {
+        // assume seconds
+        v_units = wrap("secs");
+      }
+      if (units.isNULL()) {
+        // if current unit is NULL, grab the new one
+        units = v_units;
```
krlmlr: Can we simply reject `"difftime"` objects that don't have a `"units"` attribute?

inst/include/dplyr/Collecter.h
```diff
@@ -25,12 +25,13 @@ namespace dplyr {
   static bool is_class_known(SEXP x) {
     /* C++11 (need initializer lists)
     static std::set<std::string> known_classes {
-      "POSIXct", "factor", "Date", "AsIs", "integer64", "table"
+      "difftime", "POSIXct", "factor", "Date", "AsIs", "integer64", "table"
```
krlmlr: We shouldn't need to carry around code in comments. The old-style code also works in C++11, we'll rewrite that when we are on C++11 *and* when we touch the code again.

inst/include/dplyr/Gatherer.h
```diff
@@ -11,341 +11,273 @@
...
-    typedef typename Rcpp::traits::storage_type<RTYPE>::type STORAGE;
-    typedef GroupedCallProxy<Data,Subsets> Proxy;
-
-    GathererImpl(RObject& first, SlicingIndex& indices, Proxy& proxy_, const Data& gdf_, int first_non_na_) :
-      gdf(gdf_), proxy(proxy_), data(gdf.nrows(), Vector<RTYPE>::get_na()), first_non_na(first_non_na_)
-    {
-      if (first_non_na < gdf.ngroups())
-        grab(first, indices);
-      copy_most_attributes(data, first);
+class Gatherer {
```
krlmlr: Could you please indent like the original, to make review simpler?

krlmlr:
@zeehio: If you address the feedback here and backport to #2486, we can continue the discussion there.

zeehio:
@krlmlr I have taken care of all the issues you mentioned. Apologies for the grammar error and for all the indentation issues, thanks for all the comments and suggestions :+1:

inst/include/dplyr/Collecter.h
krlmlr: So, if a copy has been made, I'm missing it ;-)

inst/include/dplyr/Collecter.h
```diff
@@ -364,6 +359,130 @@ namespace dplyr {
...
+          units = wrap("secs");
+          double factor_v = time_conversion_factor(v_units);
+          if (Rf_length(v) < index.size()) {
+            stop("Wrong size of vector to collect");
+          }
+          if (TYPEOF(v) == REALSXP) {
+            for (int i=0; i<index.size(); i++) {
+              Parent::data[index[i]] = factor_v * REAL(v)[i];
+            }
+          } else if (TYPEOF(v) == INTSXP) {
```
krlmlr: How do you create a difftime of mode integer?

inst/include/dplyr/Collecter.h
```diff
@@ -364,6 +359,130 @@ namespace dplyr {
...
+            stop("difftime must be integer or numeric");
+          }
+        }
+      }
+    }
+
+    double time_conversion_factor(RObject v_units) {
+      // Acceptable units based on r-source/src/library/base/R/datetime.R
+      double factor = 1;
+      std::string v_units_c = Rcpp::as<std::string>(v_units);
```
krlmlr: I wounder if we could use a map (as a static variable in a function) that allows lookup by SEXP, both here and in has_valid_time_unit().

inst/include/dplyr/Collecter.h
```diff
@@ -364,6 +359,130 @@ namespace dplyr {
...
+    inline bool can_promote(SEXP x) const {
+      return false;
+    }
+
+    std::string describe() const {
+      return collapse<STRSXP>(types);
+    }
+
+  private:
+    RObject units;
```
krlmlr: Can you please move data members to the bottom?

inst/include/dplyr/Collecter.h
zeehio: I can only think of `structure(4L, units="secs", class = "difftime")` and I agree it is forcing it. I can drop the `else if` if you prefer.

inst/include/dplyr/Collecter.h
zeehio: In the last commit I used an `std::string` as key. If that is not good enough I will try to fix it tomorrow.

inst/include/dplyr/Collecter.h
```diff
@@ -362,8 +355,139 @@ namespace dplyr {
...
+              Parent::data[index[i]] = factor_v * INTEGER(v)[i];
+            }
+          } else {
+            stop("difftime must be integer or numeric");
+          }
+        }
+      }
+    }
+
+    double time_conversion_factor(SEXP v_units) {
```
hadley: This feels overworked to me now - I preferred the previous version with explicit if statements. Might be better to make the argument `std::string()`

inst/include/dplyr/Collecter.h
```diff
@@ -362,8 +355,139 @@ namespace dplyr {
...
+
+      if (valid_units.find(x_units_c) != valid_units.end()) {
+        return true;
+      } else {
+        return false;
+      }
+    }
+
+    void collect_difftime(const SlicingIndex& index, SEXP v) {
+      RObject v_units(Rf_getAttrib(v, Rf_install("units")));
```
hadley: Can't you do `v.attr("units")`? I'm pretty sure there's a C++ api here.

inst/include/dplyr/Collecter.h
```diff
@@ -362,8 +355,139 @@ namespace dplyr {
...
+      if (v_units.isNULL()) {
+        stop("Can't collect difftime without units");
```
hadley: I think here, and for the non REALSXP case below, you can simply do `stop("Invalid difftime object")`.

inst/include/dplyr/Collecter.h
```diff
@@ -362,8 +355,139 @@ namespace dplyr {
...
+      std::map<std::string, double>::iterator unit_factor =
+        unit_factor_map.find(CHAR(STRING_ELT(v_units,0)));
+      if (unit_factor != unit_factor_map.end()) {
+        return unit_factor->second;
+      } else {
+        stop("Cannot convert %s to seconds", CHAR(STRING_ELT(v_units, 0)));
+      }
+    }
+
+    RObject units;
```
hadley: Would all this code be simpler if `units` was a `std::string`?

zeehio:
I took care of the comments, feel free to do another review if you want

hadley:
Looks good, thanks!

PR #2488, opened 2017-03-02
Port dplyr to rlang and tidy evaluation

lionel-:
TODO:

- A vignette on tidy evaluation (probably for another PR)

- Solve `.env` within magrittr pipeline.

- Should we have `rlang::quo_text()` in addition to `rlang::f_text()`? The latter serialises the RHS of formulas and is used for default names of expressions passed to dots. However, with quasiquotation dplyr will get expressions containing inner formulas. `quo_text()` would splice the formulas before serialising the expression. It would also understand formula guards created by `UQF()` (they show up as `_F()` in quosures).

- ~~Maybe rlang should export C level functions for installing and cleaning the dynamic scope, so there's no callback from C++ to R in the grouped hybrid class. That probably wouldn't gain us much in terms of performance though.~~

- [X] Clean dynamic scope more thoroughly in rlang, then we can remove cleaning code from dplyr


krlmlr:
I have no idea why the AppVeyor build is failing...

lionel-:
yeah, I was wondering.

NEWS.md
```diff
@@ -37,6 +37,8 @@
 
 * `tbl_df` gains `rbind()` and `cbind()` methods that call `bind_rows()` and
   `bind_cols()` respectively (#2138)
+* `column()` has been removed without deprecation (it was never
```
krlmlr: Please double-check NEWS differences (incorrect conflict resolution?)

inst/include/dplyr/Result/GroupedHybridCall.h
```diff
@@ -95,6 +107,8 @@ namespace dplyr {
 
     mutable Environment eval_env;
     mutable bool has_eval_env;
+
+    const Environment rlang;
```
krlmlr: We could use

```
static Function get_rlang_fun(const char* name) {
  static Environment rlang = ...;
  return rlang[name];
}
```

so that we do only one lookup. Or perhaps a global variable?

inst/include/dplyr/Result/GroupedHybridCall.h
```diff
@@ -69,7 +73,15 @@ namespace dplyr {
         return;
 
       Environment active_env = eval_env.parent();
+
+      // That's unreliable, but cleanup should move to rlang anyway
+      active_env = active_env.parent();
```
krlmlr: Do we still need the initialization at declaration?

inst/include/dplyr/Result/GroupedHybridCall.h
```diff
@@ -53,7 +54,10 @@ namespace dplyr {
       // call to Rcpp_eval() triggered by active_env.new_child()
       eval_env = active_env.new_child(true);
       eval_env[".data"] = active_env;
-      eval_env[".env"] = env;
+
+      // Install definitions for formula self-evaluation and unguarding
+      Function dyn_scope_install = rlang["dyn_scope_install"];
```
krlmlr: `static Function` is better (same below).

inst/include/dplyr/Result/GroupedHybridCall.h
```diff
@@ -53,7 +54,10 @@ namespace dplyr {
...
+      eval_env = dyn_scope_install(eval_env, active_env, env);
```
krlmlr: Do you have a writeup for dynamic scoping? What does it do?

lionel-:
> Do you have a writeup for dynamic scoping? What does it do?

Not yet, we'll need one vignette for tidyeval users and another for developers. However I'm not sure dynamic scoping is the right concept anymore.

The idea of the tidyeval framework is to perform NSE based on evaluation rules, as opposed to fiddling with expressions. The switch from the call-traversing code to the definition of active bindings for column slices fits that spirit. We give alternative definitions to symbols and we let R evaluate these definitions. In a dataframe DSL, this essentially means rescoping symbols corresponding to dataframe columns. In a modelling DSL we'd evaluate formulas with rescoping of operators like `+` and `:` (whereas the usual approach is to parse the formula manually).

At the same time, we want to ensure hygienic evaluation. That is, symbols should be scoped in the context where they appear in the code, even though they are evaluated later and deeper in the call stack. The tension between these two requirements is resolved by articultating two types of scoping in a principled way:

- lexical scoping with quosures (formula-quotes bundling a lexical enclosure). This captures the user context. These quosures self-evaluate in their own scope during tidy evaluation. They pretty much behave like promises.

- "dynamic scoping" for rescoped symbols. This can take the form of a single environment, or several of them. All we need is the top and bottom environments for this scope. Rlang will add one more layer at the bottom to install all the necessary definitions for tidy evaluation (definition for `~` so that quosures self-evaluate, etc). The trick is that when quosures self-evaluate, we rechain their enclosure on top of the dynamic scope.

Side note: the final piece of this framework is the quasiquotation mechanism. All capturing functions support quasiquotation, allowing the user to inline whatever she wants in the captured argument: values or expressions (quotes and quosures). This gives full programmability of NSE functions. Quosures are a good choice here because they ensure hygienic resolution of symbols at evaluation time, thanks to the mechanism described above.

However I think dynamic scope is not the right term, because only the captured expressions are rescoped. Dynamic scoping would imply that all evaluation frames from that point onward are rescoped, which is not the case at all.

But then I don't know what prefix to use for all these `rlang::dyn_scope_` functions :'(

lionel-:
copycat/sneaky scoping?

lionel-:
intervened / mediated scoping?

inst/include/dplyr/Result/GroupedHybridCall.h
lionel-: We need the grandparent here because rlang creates a child of the active env. But really this should all be moved in rlang. I think just wiping out all bindings in all environments of the mediator scope should do it.

inst/include/dplyr/Result/GroupedHybridCall.h
lionel-: Good idea. The advantage of a function over a global variable is that the former can live in that file thanks to the `inline` keyword.

lionel-:
hooked scoping? `hook_scope_install()`, `hook_scope_env()`

R/count-tally.R
```diff
@@ -107,75 +107,70 @@ n_name <- function(x) {
...
-    n <- lazyeval::interp(quote(sum(wt, na.rm = TRUE)), wt = wt)
-  }
   n_name <- n_name(tbl_vars(x))
-  out <- mutate_(x, .dots = setNames(list(n), n_name))
+  out <- mutate(x, !! n_name := !! n)
 
   if (sort) {
-    desc_n <- lazyeval::interp(quote(desc(n)), n = as.name(n_name))
-    out <- arrange_(out, desc_n)
+    desc_n <- tidy_quote(desc(!! symbol(n_name)))
```
hadley: Feels like an intermediate variable is no longer needed here.

R/distinct.R
```diff
@@ -31,9 +31,8 @@
 #' # You can also use distinct on computed variables
 #' distinct(df, diff = abs(x - y))
 distinct <- function(.data, ..., .keep_all = FALSE) {
-  distinct_(.data, .dots = lazyeval::lazy_dots(...), .keep_all = .keep_all)
+  UseMethod("distinct")
 }
-
 #' @export
 #' @rdname distinct
```
hadley: I think all these should be `@rdname nse-deprecated`

R/dbi-s3.r
```diff
@@ -142,7 +142,7 @@ db_create_table <- function(con, table, types, temporary = FALSE, ...) {
 #' @export
 db_create_table.DBIConnection <- function(con, table, types,
                                           temporary = FALSE, ...) {
-  assert_that(is.string(table), is.character(types))
+  assert_that(is_scalar_character(table), is.character(types))
```
hadley: `is_string()`

R/manip.r
```diff
@@ -346,25 +344,21 @@ select_ <- function(.data, ..., .dots) {
 #' iris %>% select_if(function(col) is.numeric(col) && mean(col) > 3.5)
 select_if <- function(.data, .predicate, ...) {
   if (inherits(.data, "tbl_lazy")) {
-    stop(
-      "Selection with predicate currently require local sources",
-      call. = FALSE
-    )
+    abort("Selection with predicate currently require local sources")
```
hadley: It's clear now that this warning should really go inside `probe_colwise_names()` but we can deal with that later

R/pull.R
```diff
@@ -39,23 +39,23 @@ pull.data.frame <- function(.data, var = -1) {
...
   .data <- collect(.data)
   .data[[1]]
 }
 
 find_var <- function(var, vars) {
-  if (is.character(var) && length(var) == 1) {
+  if (is_string(var)) {
     if (!var %in% vars) {
-      stop("Unknown variable '", var, "'", call. = FALSE)
+      abort(glue("Unknown variable '{var}'", var = var))
```
hadley: Do you need `var = var` here?

R/translate-sql-helpers.r
```diff
@@ -110,7 +110,7 @@ copy_env <- function(from, to = NULL, parent = parent.env(from)) {
 #' @rdname sql_variant
 #' @export
 sql_infix <- function(f) {
-  assert_that(is.string(f))
+  assert_that(is_scalar_character(f))
```
hadley: `is_string`

R/translate-sql-window.r
```diff
@@ -216,24 +216,33 @@ common_window_funs <- function() {
...
-
-      window_where(
-        expr = expr,
-        comp = unlist(lapply(args, "[[", "comp"), recursive = FALSE)
-      )
-    }
-  } else {
-    stop("Unknown type: ", typeof(expr))
-  }
+  # FIXME: Should this return a quosure?
```
hadley: I don't think so - this is always called after partial evaluation, so there should be no environment to preserve.

R/ts.R
```diff
@@ -1,7 +1,8 @@
 #' @export
-filter_.ts <- function(.data, ..., .dots) {
-  stop(
-    "dplyr::filter() called with ts object. Do you want stats::filter()?",
-    call. = FALSE
-  )
+filter.ts <- function(.data, ...) {
+  abort("dplyr::filter() called with ts object. Do you want stats::filter()?")
+}
+#' @export
```
hadley: You can delete the method for `filter_`

inst/include/tools/TidyQuote.h
```diff
@@ -0,0 +1,102 @@
...
+#define dplyr__TidyQuote_h
+
+#include <tools/SymbolString.h>
+
+
+namespace dplyr {
+
+class TidyQuote {
+ public:
+  TidyQuote(Formula data_, SymbolString name__) :
```
krlmlr: Should we use `const&` here?

inst/include/tools/TidyQuote.h
```diff
@@ -0,0 +1,102 @@
...
+  TidyQuote(const TidyQuote& other) :
+      data(other.data),
+      name_(other.name_)
+  {}
+
+  inline SEXP expr() const {
+    return Rf_duplicate(CADR(data));
+  }
+  inline SEXP env() const {
+    return Rf_getAttrib(data, Rf_install(".Environment"));
```
krlmlr: I wonder if we could use `static Symbol Environment = ".Environment"`.

inst/include/tools/TidyQuote.h
```diff
@@ -0,0 +1,102 @@
...
+
+  inline const TidyQuote& operator[](int i) const {
+    return data[i];
+  }
+
+  inline int size() const {
+    return data.size();
+  }
+
+  inline bool single_env() const {
```
krlmlr: I wonder if we really need the special handling of "single environment" calls.

lionel-: yes it should go, but this needs a bit more refactoring of `filter()`. Also we need to refactor the grouped hybrid class so it takes a quosure rather than call + env. Let's merge this PR for the time being I'll send another one later.

Also the hybrid evaluator should use overscoping rather than search & replace.

R/utils.r
```diff
@@ -92,6 +63,8 @@ commas <- function(...) paste0(..., collapse = ", ")
 
 in_travis <- function() identical(Sys.getenv("TRAVIS"), "true")
 
+has_names <- rlang::have_names
```
krlmlr: @lionel-: Did you have a specific reason to keep the alias? I see only five hits for `has_names()`.

lionel-: hmm I was probably trying to avoid merge conflicts caused by search&replace renaming since we were three people working on the codebase.

Issue #2464, opened 2017-02-22
log() in postgres gives a type error for doubles


karldw:
Currently dplyr uses the two-argument `log(base, x)` form (see #2288), but this form raises a type error in postgres when `x` is a double.

```r
df <- data_frame(a = 1:5, b = a + 0.5)  # a is integer, b is double
df_postgres <- copy_to(src_postgres, df)
mutate(df_postgres, log_a = log(a))  # works because a is integer
mutate(df_postgres, log_b = ln(b))   # works because 'ln' gets passed through
mutate(df_postgres, log_b = log(b))  # error
```

Both [mysql](https://dev.mysql.com/doc/refman/5.7/en/mathematical-functions.html#function_log) and [postgres](https://www.postgresql.org/docs/current/static/functions-math.html#FUNCTIONS-MATH-FUNC-TABLE) have a `ln()` function, which would work for the default case of natural log.
In postgres, the two-argument form requires numeric or integer  `x` ('numeric' is postgres' arbitrary-precision type).  Additionally `log()` in postgres is base 10, but I don't think that matters here.

Proposal:
- if `base` is *e*, use `ln` in translate-sql-base.r
- Otherwise, calculate `log(x) / log(base)` for postgres, [like src-sqlite.r does](https://github.com/hadley/dplyr/commit/a6ea410ecced3f209a57a761077bcfdcb96364b3#diff-fcfefa82c4b028eeed00b6f149b3a6f1R163).

I can submit a pull request for this.

hadley:
A PR sounds good

PR #2491, opened 2017-03-03
Add translation for log in postgres (#2464)

karldw:
This gets around the issue that postgres doesn't support two-argument
`log(base, x)` for floating point `x`.

- For default SQL: if log base is *e*, use `ln()`.
- For postgres: if log base is *e*, use `ln()`.
- For postgres: if log base isn't *e*, use `log(x) / log(base)`.
- Add tests.

### DIFF
--- a/tests/testthat/test-sql-translation.r
+++ b/tests/testthat/test-sql-translation.r
@@ -151,6 +151,11 @@ test_that("log base comes first", {
   expect_equal(translate_sql(log(x, 10)), sql('log(10.0, "x")'))
 })
 
+test_that("log becomes ln", {
+  expect_equal(translate_sql(log(x)), sql('ln("x")'))
+  expect_equal(translate_sql(log(x, exp(1))), sql('ln("x")'))
+})
+
 test_that("sqlite mimics two argument log", {
   translate_sqlite <- function(...) {
     translate_sql(..., con = src_memdb()$con)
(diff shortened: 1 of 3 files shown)


karldw:
The checks fail on appveyor because I've used a plain `src_postgres()`, with no arguments.  Should I use [CI-specific `src_postgres` calls ](https://github.com/hadley/dplyr/blob/master/tests/testthat/helper-src.R#L6) to connect, or is that too much for this test?

tests/testthat/test-sql-translation.r
```diff
@@ -160,6 +164,16 @@ test_that("sqlite mimics two argument log", {
   expect_equal(translate_sqlite(log(x, 10)), sql('log(`x`) / log(10.0)'))
 })
 
+test_that("postgres mimics two argument log", {
+  translate_postgres <- function(...) {
+    translate_sql(..., con = src_postgres()$con)
```
hadley: I think you might be able to do something hacky like `structure(list(), class = "PostgreSqlConnection")` because I don't think any of the methods use any fields in the object

hadley:
Thanks!

In the future, can you please include `Fixes  #2464` in the body of a commit message, so that the related issue gets closed automatically?

karldw:
Definitely, and thanks!

PR #2501, opened 2017-03-07
Minor cleanup after rlang changes

krlmlr:
Follow-up to #2488.

### DIFF
--- a/R/bind.r
+++ b/R/bind.r
@@ -141,7 +141,7 @@ is_data_list <- function(x) {
     return(TRUE)
 
   # With names
-  if (any(!has_names(x)))
+  if (any(!have_names(x)))
     return(FALSE)
 
   # Where each element is an 1d vector or list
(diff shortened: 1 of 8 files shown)


lionel-:
Of course it's better if the whole codebase uses the same style, and the ship has sailed, but I'd recommend not indenting over namespaces. Doing so makes indentation less meaningful and the code harder to parse visually. This should be paired with closing namespaces like this: `} // namespace dplyr`.

krlmlr:
We still can change easily, but at this point `astyle` thinks that namespaces should be indented. See helper-astyle.R.

I'm happy to revise this once the C++ code reaches a stable state; also, I'd like to pad operators with spaces with the `--pad-oper` switch.

jimhester:
FWIW If we are going to standardise on using `astyle`, we should have a `astylerc` file we can share between projects rather than passing command line arguments.

eddelbuettel:
I have no horse in this race ...  but don't all the cool kids use [clang-format](https://clang.llvm.org/docs/ClangFormat.html) ?

krlmlr:
Thanks for the pointer. I don't remember why I preferred astyle at that time, maybe it was just good (and noninvasive) enough and I was familiar with it. We should consider clang-format if astyle becomes insufficient.

PR #2509, opened 2017-03-08
Change 'sql_translate' to 'translate_sql' in doc

shabbybanks:
I believe these were in error or from older versions, and the documentation should read `translate_sql()` not `sql_translate()`.


hadley:
Thanks!

PR #2511, opened 2017-03-08
Add missing con = win_current_con()

ianmcook:
Queries with ORDER BY in OVER() failed with a SQL backend that uses non-standard identifier quoting because double quotes were inserted around identifiers.


ianmcook:
The failure on AppVeyor appears to be caused by an unrelated issue.

hadley:
Thanks!

PR #2515, opened 2017-03-09
Fix very rare case of false match during join

krlmlr:
Under the following conditions, a false match during a join could have occurred theoretically:

- Joining an integer and a numeric column
- Two numeric values have the same integer component but different fractional components, **and** hash to the same value

It is unlikely that this has been ever observed, but it might be possible to construct an artificial example.


krlmlr:
Merging, because further changes depend on this.

PR #2517, opened 2017-03-09
Don't match NA when joining

### LINKED ISSUE #2033: joins match rows on NA across datasets
Below, `inner_join()` matches two rows that each have an `NA` for the key. This seems to violate the intuition about NA's (and leads to an incorrect result). 

This seems to happen for all of the joins.

```
library(dplyr)
​
songs <- data_frame(song = c("Do-Re-Mi", "A Spoonful of Sugar"), movie = c("The Sound of Music", NA))
songs
##                  song              movie
##                 <chr>              <chr>
## 1            Do-Re-Mi The Sound of Music
## 2 A Spoonful of Sugar               <NA>
​
singers <- data_frame(movie = c(NA, "The Sound of Music"), singer = c("Arnold Schwarzenegger", "Julie Andrews"))
singers
##                movie                singer
##                <chr>                 <chr>
## 1               <NA> Arnold Schwarzenegger
## 2 The Sound of Music         Julie Andrews
​
songs %>% inner_join(singers, by = "movie")
##                  song              movie                singer
##                 <chr>              <chr>                 <chr>
## 1            Do-Re-Mi The Sound of Music         Julie Andrews
## 2 A Spoonful of Sugar               <NA> Arnold Schwarzenegger
```

krlmlr:
All individual commits except the last are refactorings that pass all tests.

The new `DualVector` class encapsulates the logic of a vector with two underlying types, and both positive and negative indexes.

If NAs are treated as non-matching, the hash function is evaluated as the index in the "dual vector" to avoid hash collisions and potentially O(n²) behavior.

We should think about unit tests for dplyr, perhaps using Catch.

Fixes #2033.


NEWS.md
```diff
@@ -1,5 +1,11 @@
 # dplyr 0.5.0.9000
 
+* `NA` values never match when joining. This mimics the behavior of relational
+  databases, but differs from the default behavior of `merge()` --
+  the new behavior is closer to `merge(incomparables = NA)`. This is a breaking
```
hadley: I'm now wonder if we should have an option. It is convenient to match on NAs for certain types of task

src/test.cpp
```diff
@@ -18,3 +19,87 @@ LogicalVector test_comparisons() {
...
+  typedef dplyr::join_match<INTSXP, REALSXP, true> int_real_na;
+  typedef dplyr::join_match<REALSXP, INTSXP, true> real_int_na;
+  typedef dplyr::join_match<INTSXP, INTSXP, false> int_int;
+  typedef dplyr::join_match<REALSXP, REALSXP, false> real_real;
+  typedef dplyr::join_match<INTSXP, REALSXP, false> int_real;
+  typedef dplyr::join_match<REALSXP, INTSXP, false> real_int;
+  return
+    List::create(
+      LogicalVector::create(
+        int_int_na::is_match(1, 1),
```
hadley: I wonder if it would be useful to export `is_match()` to R? That would obviously make the tests easier, but it feels like it might be a useful primitive in some situations (I think it's basically a scalar `==`)

krlmlr: What should the R function be called? Happy to tackle this after merging this PR.

hadley: Maybe just `is_match()` ?

krlmlr:
@hadley: Ready, but I'm not sure the `na_match` and `NA_MATCH` argument names are a good choice. Perhaps `match_na`?

hadley:
Yeah, `match_na()` would be better, but still doesn't feel quite right. `na_matches_na` is a bit better, but rather long.

krlmlr:
`na_matching`?

hadley:
`na_can_match`?

krlmlr:
We could use different options for the public R interface and for the internal code. For the interface, we could also use `xxx_join(na = c("can_match", "incomparable"))`, with functions `na_can_match()` and `na_incomparable()`. Internally, `NA_MATCH` looks too similar to `NA_INTEGER` et al., can we choose a name that doesn't start with `NA`?

hadley:
How about `xxx_join(na_match = c("never", "na"))`?  I don't care about the names of the C++ components.

krlmlr:
Done, now `xxx_join(na_matches = c("na", "never"))` to preserve the current behavior as default.

hadley:
I think we should probably change the default behaviour even if it isn't backward compatible.

krlmlr:
@gaborcsardi: What's the best way to introduce a compatibility option via pkgconfig? Joins gain a `na_matches` argument with this PR, and the default should be different from the current behavior.

gaborcsardi:
You just do 
```r
pkgconfig::get_config("dplyr::na_matches", _the_default_comes_here_)
```
then other packages will be able to call `pkgconfig::set_config("dplyr::na_matches" = _whatever_)` so set their default behavior.

Or, if you want to support "regular" options as well, you can also do:
```r
pkgconfig::get_config("dplyr::na_matches", getOption("dplyr.na_matches", _the_default_comes_here_))
```
The benefit of this is that R script (i.e. not packages) can set the default, too.

krlmlr:
Thanks. What happens if users call `set_config()` from `.GlobalEnv`? I'd rather avoid offering two configuration paths.

gaborcsardi:
You are right, that works fine as well. E.g.:

krlmlr:
@hadley: Changed default, now configurable with pkgconfig.

PR #2523, opened 2017-03-13
Fix mutation of list-column names in mutate()

lionel-:
Fixes a bug introduced in #2512 

It would change the names of list-columns by reference:

```r
df <- data_frame(y = list(a = 1, b = 1:2))
names(df$y)
#> [1] "a" "b"

dplyr::mutate(df, y)
names(df$y)
#> NULL
```

### DIFF
--- a/inst/include/dplyr/NamedListAccumulator.h
+++ b/inst/include/dplyr/NamedListAccumulator.h
@@ -19,7 +19,8 @@ namespace dplyr {
       if (! Rcpp::traits::same_type<Data, RowwiseDataFrame>::value)
         check_supported_type(x, name);
 
-      Rf_setAttrib(x, R_NamesSymbol, R_NilValue);
+      if (TYPEOF(x) != VECSXP)
+        Rf_setAttrib(x, R_NamesSymbol, R_NilValue);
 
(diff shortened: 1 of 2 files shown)


hadley:
I think the names probably still need to be stripped, we're just missing an `Rf_shallow_duplicate()` somewhere

lionel-:
`unnest()` unit tests in tidyr check that the names are preserved. And `data_frame()` and `data.frame()` don't strip the names of list-columns either.

krlmlr:
Stripping names was the intent of #2512, also for list columns.

We could use a helper class that indicates caller-owned immutable data that needs to be duplicated before modification. I'm not aware of an existing class in Rcpp.

lionel-:
it seems just wrong to change list columns here. These two statements should produce identical results:

```r
mutate(df, x)
mutate(df)
```

Also in practice the column-list names carry useful information, and mapping over a list-column in a mutate should preserve it.

krlmlr:
dplyr currently just isn't prepared to handle names in filter() and summarise(). I'm not sure it's worth the effort to change it, you can always move the names to a second column.

PR #2535, opened 2017-03-15
Make codecov quiet

lionel-:
So a red mark means that the build is broken rather than the coverage has dropped.

### DIFF
--- a/codecov.yml
+++ b/codecov.yml
@@ -1,1 +1,12 @@
 comment: false
+
+coverage:
+  status:
+    patch:
+      default:
+        target: 0
+        threshold: 100
+    project:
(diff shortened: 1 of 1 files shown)


krlmlr:
We could also work on the coverage by dropping unused parts and adding tests, especially for the C++ code.

lionel-:
Making codecov quiet doesn't prevent us from working on the coverage, we still get the reports.

krlmlr:
If coverage is good, we should be skeptical of patches that reduce coverage.

hadley:
I'm with @lionel- on this one: codecov is currently adding too much noise

PR #2546, opened 2017-03-20
Update to latest rlang API

lionel-:
Accounts for the changes in https://github.com/hadley/rlang/pull/59 and https://github.com/hadley/rlang/pull/64

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -141,5 +141,5 @@ Collate:
 RoxygenNote: 6.0.1
 Roxygen: list(markdown = TRUE, roclets = c("rd", "namespace", "collate"))
 Remotes:
-    hadley/rlang#59,
+    hadley/rlang,
     tidyverse/glue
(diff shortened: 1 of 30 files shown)


krlmlr:
Can we please pin rlang to a revision or a tag?

lionel-:
> Can we please pin rlang to a revision or a tag?

Will this work with AppVeyor? I'm updating the rlang API in bulk to limit issues. There shouldn't be many changes from now on as the API is stabilising. Should we merge this now to fix the build?

hadley:
I think we should merge for now.

krlmlr:
We could ask `Remotes: hadley/rlang@abcdef01`, this should work via devtools. If rlang API updates are irregular, we could perhaps pin rlang in dplyr just before you start working on the API, and release the pin in the PR that makes dplyr compatible to the API again. Would that work?

PR #2551, opened 2017-03-21
Pin rlang to known working revision

lionel-:


lionel-:
Appveyor seems to work when pinned to a revision rather than a branch. Thanks for the tip @krlmlr. Rlang is now pinned.

PR #2554, opened 2017-03-21
Shallow copy if must strip names

krlmlr:
Closes #2523.


lionel-:
Named list-columns are a legal and useful feature of data frames, I don't understand why you're trying to get rid of them. If `as.matrix()` doesn't handle them, it's a bug in `as.matrix()`. Unnaming the list-column breaks a legitimate dplyr/tidyr/purrr workflow.

krlmlr:
dplyr currently doesn't handle named vectors very well. I see the benefit of lists that contain named lists (or arbitrary objects), but not of named lists themselves.

lionel-:
You're right that it's preferable to have the names stored in their own column. I guess that's also the point of the `.id` argument in `tidyr::unnest()`

PR #2558, opened 2017-03-22
Cleanup around subset()

krlmlr:
- remove unused LogicalVector methods
- use partial specialization instead of traits
- use preallocated integer vector instead of counting
    - this trades memory for speed; a more conservative solution could be used to estimate the final size of the vector (similarly to readr) and to reserve accordingly


inst/include/dplyr/DataFrameSubsetVisitors.h
```diff
@@ -117,6 +95,19 @@ namespace dplyr {
 
   };
 
+  template <>
+  inline DataFrame DataFrameSubsetVisitors::subset(const LogicalVector& index, const CharacterVector& classes) const {
+    int n = index.size();
+    std::vector<int> idx;
+    idx.reserve(index.length());
```
hadley: Aren't `index.size()` and `index.length()` the same value?

PR #2560, opened 2017-03-22
Fix performance gap with tidyeval

lionel-:
I go from 6s to 1.4s with the `mutate.dplyr_df` test case. This could still be divided by 2 by using `Rf_eval()` instead of `Rcpp_eval()` once we get proper stack unwinding.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -21,7 +21,7 @@ Imports:
     magrittr,
     methods,
     pkgconfig,
-    rlang (>= 0.0.0.9001),
+    rlang (>= 0.0.0.9002),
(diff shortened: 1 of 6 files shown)


krlmlr:
Will do. Faster eval is discussed in #2295.

krlmlr:
Run time much better now, but there still seems to be a minor overhead:

download 10
download 9
download 8
download 7
download 6

lionel-:
great! Some overhead seems inevitable because of interpolation, quosure self-evaluation (when an NSE verb calls another NSE verb with unquoted expressions), etc

Issue #2504, opened 2017-03-07
Revise formatting of C++ files


krlmlr:
- Don't indent namespaces (remove [`--indent-namespaces`](http://astyle.sourceforge.net/astyle.html#_indent-namespaces))
- Pad operators with spaces (add [`--pad-per`](http://astyle.sourceforge.net/astyle.html#_pad-oper)

https://github.com/hadley/dplyr/pull/2501#issuecomment-284779984

PR #2563, opened 2017-03-23
Port filter() to tidyeval

lionel-:
I'll probably export `quo_reduce()` in rlang for the scoped verbs PR.


R/funs-predicates.R
```diff
@@ -0,0 +1,30 @@
...
+##   quoted name, the default environment is the [base
+##   environment][rlang::base_env] unless you supply a
+##   [quosure][rlang::quosure].
+quo_reduce <- function(..., .op) {
+  stopifnot(is_symbol(.op) || is_function(.op))
+
+  dots <- dots_quosures(...)
+  if (length(dots) == 1) {
+    return(dots[[1]])
+  } else if (!length(dots)) {
```
hadley: I think this is easier to understand if it looks like:

```R
if (length(dots) == 0) {

} else if (length(dots) == 1) {

}
```

R/funs-predicates.R
```diff
@@ -0,0 +1,30 @@
+
+all_of <- function(..., .vectorised = FALSE) {
```
hadley: Could you please include a couple of simple tests for these functions too?

lionel-: oops yes, I was thinking of doing it in the scoped verbs PR but it makes more sense here

src/hybrid.cpp
```diff
@@ -168,7 +168,7 @@ namespace dplyr {
         try {
           data = env.find(sym.get_string());
         } catch (Rcpp::binding_not_found) {
-          stop("object '%s' not found\n", sym.get_utf8_cstring());
+          return constant_handler(R_NilValue);
```
lionel-: @krlmlr is this the right way to do it?

krlmlr: I think it's simply `return NULL`.

lionel-: does't work because of return type

lionel-: oh you mean `NULL` as in C `NULL`?

lionel-: done, thanks!

krlmlr:
@lionel-: I planned to reformat the codebase soon-ish (#2504), please let me know when it's a good time for this.

src/hybrid.cpp
krlmlr: I always wondered why it's `return 0` and not `return NULL`...

Issue #2273, opened 2016-11-29
`summarise_all` performance


ChadGoymer:
`summarise_all` & `summarise_if` are a great addition to `dplyr`, but there is a large performance difference between them and using an explicit `summarise`. The performance difference is illustrated on large and wide tibbles in the following examples:

```r
library(tibble)
library(dplyr)

iris_big <- iris %>% 
  as_tibble %>% 
  sample_n(1e8, replace = TRUE)

system.time(
  iris_mean <- iris_big %>% 
    group_by(Species) %>% 
    summarise_all(mean)
)

##    user  system elapsed 
##   13.06    3.28   16.97

system.time(
  iris_mean <- iris_big %>% 
    group_by(Species) %>% 
    summarise(
      Sepal.Length = mean(Sepal.Length),
      Sepal.Width = mean(Sepal.Width),
      Petal.Length = mean(Petal.Length),
      Petal.Width = mean(Petal.Width)
    )
)

##    user  system elapsed 
##    8.53    0.54    9.64

iris_wide <- iris_big %>% 
  mutate(
    Sepal.Length2 = Sepal.Length,
    Sepal.Width2 = Sepal.Width,
    Petal.Length2 = Petal.Length,
    Petal.Width2 = Petal.Width,
    Sepal.Length3 = Sepal.Length,
    Sepal.Width3 = Sepal.Width,
    Petal.Length3 = Petal.Length,
    Petal.Width3 = Petal.Width,
    Sepal.Length4 = Sepal.Length,
    Sepal.Width4 = Sepal.Width,
    Petal.Length4 = Petal.Length,
    Petal.Width4 = Petal.Width,
    Sepal.Length5 = Sepal.Length,
    Sepal.Width5 = Sepal.Width,
    Petal.Length5 = Petal.Length,
    Petal.Width5 = Petal.Width
  )

system.time(
  iris_mean <- iris_wide %>% 
    group_by(Species) %>% 
    summarise_all(mean)
)

##    user  system elapsed 
##   42.30   14.59   58.51

system.time(
  iris_mean <- iris_wide %>% 
    group_by(Species) %>% 
    summarise(
      Sepal.Length = mean(Sepal.Length),
      Sepal.Width = mean(Sepal.Width),
      Petal.Length = mean(Petal.Length),
      Petal.Width = mean(Petal.Width),
      Sepal.Length2 = mean(Sepal.Length2),
      Sepal.Width2 = mean(Sepal.Width2),
[...]

PR #2569, opened 2017-03-24
Fix NSE issues

### LINKED ISSUE #1039: Should dplyr commands gobble empty last argument?
It would be nice to allow a trailing comma in commands like the example below.  That would make it easier to rearrange the order or to comment out lines without needing to figure out whether the commas need adjusting.

```
diamonds %>%
  group_by(color, clarity) %>%
  summarise(
    depth = mean(depth),
    price = mean(price),
    carat = mean(carat),  #  <-- trailing comma here currently causes errors
  )
```

I'm imaging a function that could be called at the top of a function definition that would gobble this empty argument.  Or perhaps a wrapper function a la `Vectorize()` that could be used to convert any function into a function with this behavior.  Seems like this could be useful for many functions both within `dplyr` and elsewhere.

### LINKED ISSUE #2089: pkg::fun doesn't work in dplyr::funs
This works:

```
slci_msa_wide <- slci_msa %>%
  dplyr::select(principal_city_property, placement_care_type, value_msa) %>%
  tidyr::spread(placement_care_type, value_msa) %>%
  dplyr::left_join(
    dplyr::distinct(slci_msa, principal_city_property, pop_rank)
  ) %>%
  dplyr::arrange(-A) %>%
  dplyr::mutate(cost_rank = row_number()) %>%
  dplyr::arrange(A) %>%
  dplyr::mutate(affordability_rank = row_number()) %>%
  dplyr::select(principal_city_property, dplyr::contains("rank"), A, R, Z) %>%
  dplyr::mutate_at(dplyr::vars(A, R, Z), dplyr::funs(round)) %>%
  dplyr::mutate_at(dplyr::vars(A, R, Z), dplyr::funs(dollar))
```

But this doesn't, and the only think I changed is the last line, resulting in an error: "Error: cannot coerce type 'closure' to vector of type 'character'"

```
slci_msa_wide <- slci_msa %>%
  dplyr::select(principal_city_property, placement_care_type, value_msa) %>%
  tidyr::spread(placement_care_type, value_msa) %>%
  dplyr::left_join(
    dplyr::distinct(slci_msa, principal_city_property, pop_rank)
  ) %>%
  dplyr::arrange(-A) %>%
  dplyr::mutate(cost_rank = row_number()) %>%
  dplyr::arrange(A) %>%
  dplyr::mutate(affordability_rank = row_number()) %>%
  dplyr::select(principal_city_property, dplyr::contains("rank"), A, R, Z) %>%
  dplyr::mutate_at(dplyr::vars(A, R, Z), dplyr::funs(round)) %>%
  dplyr::mutate_at(dplyr::vars(A, R, Z), dplyr::funs(scales::dollar))
```

I think I understand why this is happening, but it causes me to violate best practice in calling functions, which is to call out the package, too.

### LINKED ISSUE #2297: Several functions (with prepending) won't work if you don't load dplyr
Because of namespace conflicts, I typically avoid `library` and instead prepend everything (e.g., either `dplyr::mutate` or `dp <- loadNamespace('dplyr')` and `dp$mutate`). If I don't want to prepend, I'll use `import` (e.g., ``import::from(magrittr, `%>%`)``). Sometimes this doesn't work because of unaccounted dependencies and such, but most of the tidyverse packages seem to work perfectly with these. The two issues I've run into so far are `dplyr::order_by` and `dplyr::data_frame`. I noticed that `dplyr::order_by` work as expected if I first run `import::from(dplyr, with_order)` and `dplyr::data_frame` will work as expected if I first run `import::from(tibble, as_tibble)`, but it would be nice if I could avoid that.

lionel-:
- `order_by()`, `top_n()`, `sample_n()` etc now use tidyeval. Fixes #2297.

- `funs()` now handles namespaced functions. Fixes #2089.

- depends on an rlang update where literals are not rewrapped on capture → simpler count/tally implementation. Also the last arguments of captured dots are now gobbled by default, which fixes #1039 

- bare functions are now transformed to a call in colwise functions. This allows the hybrid evaluator to kick in, and this slightly simplifies the lazy table backend. The performance issue of colwise functions is now back to where it was at the time of `mutate_each()` and `summarise_each()`: still slow with many columns, but much better than reported in #2273.

  [colwise-dplyr-purrr-nogroups.pdf](https://github.com/hadley/dplyr/files/867620/colwise-dplyr-purrr-nogroups.pdf)


R/count-tally.R
```diff
@@ -59,23 +61,19 @@
...
-  # Check for NULL lazily, because `wt` could be a tidy-quoted NULL if
-  # add_tally() is called from another function (e.g. add_count())
-  n <- quo(
-    if (is_null(!! wt)) {
-      n()
-    } else {
-      sum(!! wt, na.rm = TRUE)
-    }
-  )
+  if (is_empty_quosure(wt) || is_null(f_rhs(wt))) {
```
hadley: `quo_is_missing(wt) || quo_is_empty(wt)`?

R/count-tally.R
```diff
@@ -59,23 +61,19 @@
...
+    n <- ~n()
```
hadley: I think we should use `quo()` here, just to be consistent

R/funs.R
```diff
@@ -24,15 +24,20 @@
...
 
-  names(dots) <- names2(dots)
-  missing_names <- names(dots) == ""
-  default_names <- map_chr(dots[missing_names], as_name)
-  names(dots)[missing_names] <- default_names
+new_funs <- function(funs) {
+  names(funs) <- names2(funs)
+  missing_names <- names(funs) == ""
+  default_names <- map_chr(funs[missing_names], function(dot) {
+    quo_text(node_car(f_rhs(dot)))
```
hadley: Should we use `quo_name()` here?

R/count-tally.R
lionel-: I was thinking we should only use `quo()` when we need to unquote something, so this provides additional intent in the code. This rule would not apply in introductory texts.

R/funs.R
lionel-: I think we want to allow unquoting literal functions there.

R/count-tally.R
lionel-: that's not the naming scheme we've followed so far. It's not clear what is best to do in these cases. Same issue occurs with `set_` and `mut_` functions. Currently rlang does not type-prefix all these functions.

R/funs.R
hadley: Isn't this line just determining the default name?

R/count-tally.R
hadley: I thought we had been pretty consistent with a type prefix. I think it would help here.

R/count-tally.R
hadley: I think simplicity wins over the minor signal loss.

R/funs.R
lionel-: Yes but `quo_name()` will abort with literal inputs.

hadley: Only if they're length >1 ?

R/count-tally.R
lionel-: I think these would be the very first predicates prefixed with a type in rlang.

This is linked to the issue of automatically taking the RHS in all language predicates. E.g. `is_symbol()` is shortcut for `is_symbol(f_rhs(x))`. I think this works mostly fine but is confusing because not all rlang predicates behave this way. An alternative would be a set of quosure-specific predicates. Then it probably makes sense to prefix all of them with `quo`, e.g. `quo_is_null()`, `quo_is_symbol()`.

R/funs.R
lionel-: oh that makes sense, `expr_name()` needs to be changed then

R/count-tally.R
hadley: What if we combined them both into `quo_is_empty()`? Then it's a specific definition of emptiness that applies primarily to quosures.

lionel-: hmm I think we should be consistent with base R, otherwise this will be confusing.

lionel-: > hmm I think we should be consistent with base R, otherwise this will be confusing.

maybe `is_empty_quosure()` needs to be `quo_is_missing()` then?

hadley: Maybe it's just awkwardness in the design of `count()`: most functions would either check for missingness or `NULL`, not both. And missingness is probably what needs a special predicate.

lionel-: these functions are definitely awkardly designed, but at least they pointed our attention to missing quosures and literal rewrapping :)

PR #2574, opened 2017-03-24
Adding perl-compatible regex to matches.

hennerw:
Adding an option in the select utility function `matches()` to use perl-esqe regular expressions. This is already implemented in the base `grep()` function which is the work horse of `matches()`, but there is currently no way to pass the option through.

This change would add slight complexity, and another possibly confusing option.


hadley:
For now, I'm going to say this is out of scope. I might consider adding it in the future, but I'm probably going to pull the `select_vars()` code out of dplyr into another package, so I'd like to keep it as simple as possible for now.

PR #2577, opened 2017-03-25
tbl calls new tbl_method function

### LINKED ISSUE #2576: Add tbl method for raw connections
By automatically wrapping with `src_dbi()`.

edgararuiz:
Fixes #2576


edgararuiz:
Ok, couple of first tries didn't work. I need to find a solution outside of the PR.

hadley:
The correct solution is to use `tbl.DBIConnection`. But we need to figure out why this isn't working for odbc

edgararuiz:
Ok

PR #2578, opened 2017-03-26
Add coalesce option to right_join.tbl_lazy and full_join.tbl_lazy

ianmcook:
`right_join` and `full_join` currently return discrepant results for SQL sources compared to local data frames when there are rows in the right table with no match in the left table. For local data frames, `by` columns in the join result contain values from the right table for rows from the right table that have no match in the left table.  For SQL sources, `by` columns in the result contain `NA` values for rows from the right table that have no match in the left table.

This pull request provides a resolution to this discrepancy through the addition of a `coalesce` option to `right_join.tbl_lazy` and `full_join.tbl_lazy`. When `coalesce = TRUE`, each `by` column in the result is generated by coalescing the corresponding `by` columns in the left and right tables, using the SQL `COALESCE` function (which is supported by all notable SQL distributions).

Internally, `coalesce` is added to the `by` list in `add_op_join` and then handled in `sql_join.default`. The implementation allows `coalesce` to be specified either as a `logical(1)` or as a `logical` having the same length as `by$x` and `by$y`. The documentation describes only the prior case. In the latter case, each `by` column can be controlled separately.

I defaulted `coalesce = FALSE` for compatibility, but you might consider defaulting it to `TRUE`.


R/sql-generic.R
```diff
@@ -110,10 +110,22 @@ sql_join.default <- function(con, x, y, vars, type = "inner", by = NULL, ...) {
...
-  select <- sql_vector(c(
-    sql_as(con, names(vars$x), vars$x, table = "TBL_LEFT"),
-    sql_as(con, names(vars$y), vars$y, table = "TBL_RIGHT")
-  ), collapse = ", ", parens = FALSE)
+  select_x <- sql_as(con, names(vars$x), vars$x, table = "TBL_LEFT")
+  select_y <- sql_as(con, names(vars$y), vars$y, table = "TBL_RIGHT")
+
+  select_x[vars$x %in% by$x[by$coalesce]] <-
+    sql(
+      paste0(
```
hadley: I think you can use `build_sql()` here, to eliminate `sql()` + `paste()`

R/sql-generic.R
```diff
@@ -110,10 +110,22 @@ sql_join.default <- function(con, x, y, vars, type = "inner", by = NULL, ...) {
...
   )
 
-  select <- sql_vector(c(
-    sql_as(con, names(vars$x), vars$x, table = "TBL_LEFT"),
-    sql_as(con, names(vars$y), vars$y, table = "TBL_RIGHT")
-  ), collapse = ", ", parens = FALSE)
+  select_x <- sql_as(con, names(vars$x), vars$x, table = "TBL_LEFT")
+  select_y <- sql_as(con, names(vars$y), vars$y, table = "TBL_RIGHT")
+
+  select_x[vars$x %in% by$x[by$coalesce]] <-
```
hadley: The logic here seems a bit confusing. Isn't `by$coalesce` always going to be either a single `TRUE` or a single ` FALSE`?

R/sql-generic.R
```diff
@@ -110,10 +110,22 @@ sql_join.default <- function(con, x, y, vars, type = "inner", by = NULL, ...) {
...
+    sql(
+      paste0(
+        "COALESCE(",
+        sql_table_prefix(con, by$x[by$coalesce], table = "TBL_LEFT"),
+        ", ",
+        sql_table_prefix(con, by$y[by$coalesce], table = "TBL_RIGHT"),
+        ") AS ",
```
hadley: I think this would be clearer if you did the replacement on `select_as` so you could rely on `sql_as` to add the aliases.

hadley:
Actually, would it be better to implement this as an argument to `join_vars()`? i.e. should common variables be selected from the LHS or the RHS?

ianmcook:
My first thought was to implement this in `join_vars`, but `join_vars` returns a list of vectors of the `x` and `y`, column names, not SQL. I could not think of a clean way to implement it there without making more significant changes.

R/sql-generic.R
ianmcook: See the comment at the top about how `coalesce` can be either a `logical(1)` or a `logical` having the same length as `by$x` and `by$y`. Internally preserving the ability to do the latter seems worthwhile, but documenting it for users seemed too complicated. If you think it would be better to only allow `by$coalesce` to be a single `TRUE` or `FALSE`, that's an easy change.

ianmcook:
A challenge in implementing this elsewhere (instead of in `sql_join.default`) is that we want to preserve column order in the result. So for example if `vars` had a third list representing coalesced columns, then additional work would be required to preserve the order of those columns relative to the other columns in `var$x` and `var$y`.

R/sql-generic.R
hadley: I don't see any comment above this code?

hadley:
I'd rather improve the underlying data structure, rather than adding complicated code to work around limitations.  I think key to solving this problem is to think about what `join_vars()` should actually return.

R/sql-generic.R
ianmcook: I mean the comment at the top of the page for the pull request.

ianmcook:
Ok, agreed, I'll close this pull request, work on how to achieve this in `join_vars`, and then submit another.

I'll also remove the `coalesce` option from `right_join.tbl_lazy` and `full_join.tbl_lazy` and instead just make this the standard behavior. But I'm inclined to keep the `coalesce` option in `add_op_join` to allow for custom backends against obscure SQL sources that might lack the `COALESCE` function. Alternately we could add `coalesce` to `base_scalar` and route through `sql_translate_env`.

hadley:
I don't think we need to use `coalesce()` at all if we simply select the variables from the "primary" table in the join.

ianmcook:
That's true for right joins but not for full joins. Example:

```r
df_left <- data_frame(x = 1:6, y = 1:6)
df_right <- data_frame(x = c(1:5, 7), z = 1:6)
full_join(df_left, df_right, by = "x")
```
The `x` column in the result contains a `6` from the left table and a `7` from the right table. We need the SQL `COALESCE` function (or equivalent logic) to match that result for database sources.

hadley:
Ah true. It would be good to use `colesce` only in that scenario.

ianmcook:
Agreed. I'll add it to `base_scalar` if you don't object.

hadley:
Oops, yes, it should be there.

ianmcook:
This has been resolved in dbplyr. See https://github.com/hadley/dbplyr/pull/3.

Issue #2426, opened 2017-02-14
Move database related code to a separate package


hadley:
Will be called dbplyr, and imported by default. dplyr will re-export `src_dbi()` so others don't need to know about it. `src_mysql()` etc will remain in dplyr as shims over `src_dbi()`.

```R
library(dplyr)
con <- DBI::dbConnect(RMySQL::MySQL(), username = "...")
src <- src_dbi(con)
```

hadley:
This version of dplyr acknowledges the importance of DBI: all database connections in dplyr must go through the DBI interface. This has a few consequences:

* Can now create a tbl directly from a `DBIConnection`
* All sql translation methods dispatch on the DBI connection class, not a dplyr class

This means that all the database related code can move into a separate dbplyr package that is imported by dplyr.

Steps:

* [ ] clone dbplyr from dplyr, and delete non-database related code
* [ ] import dbplyr into dplyr, and remove all database related code (except `src_mysql()` etc)
* [ ] update database backend vignette
* [ ] tease apart news files
* [ ] write migration guide for existing backends
* [ ] write SQL translation vignette

PR #2581, opened 2017-03-28
Consistent join result when there are rows in right table with no match in left table

ianmcook:
This is #2578 take two.

The options are removed, and the important logic is now in `join_vars()`. The list returned by `join_vars()` has a different structure than before: the list (`vars`) contains three parallel character vectors named `alias`, `x`, and `y`. `vars$alias` contains the column names from both tables as they will appear in the join result. `vars$x` contains the names of the corresponding columns in the left table, and `vars$y` contains the names of the corresponding columns in the right table. 

The absence or presence of `NA` values in each position of `vars$x` and `vars$y` is used to indicate which table(s) each column in `vars$alias` should be returned from. If both values in a position are non-`NA` (which happens only in full joins), that triggers a `COALESCE()` in the SQL for that column.

`sql_join.default()` uses a vectorized `paste0()` to create the `select` SQL. It calls three vectorized helper functions which test for the `NA` conditions to determine whether or not to include the different parts of the SQL for each column.

### DIFF
--- a/R/sql-generic.R
+++ b/R/sql-generic.R
@@ -110,20 +110,7 @@ sql_join.default <- function(con, x, y, vars, type = "inner", by = NULL, ...) {
     stop("Unknown join type:", type, call. = FALSE)
   )
 
-  select <- sql_vector(
-    paste0(
-      sql_if_vars_x_and_y(vars, "COALESCE("),
-      sql_if_vars_x(vars, sql_table_prefix(con, vars$x, table = "TBL_LEFT")),
-      sql_if_vars_x_and_y(vars, ", "),
-      sql_if_vars_y(vars, sql_table_prefix(con, vars$y, table = "TBL_RIGHT")),
-      sql_if_vars_x_and_y(vars, ")"),
-      sql(" AS "),
-      sql_escape_ident(con, vars$alias)
-    ),
-    parens = FALSE,
-    collapse = ", ",
-    con = con
-  )
+  select <- sql_join_vars(con, vars)
 
   on <- sql_vector(
     paste0(
@@ -143,6 +130,43 @@ sql_join.default <- function(con, x, y, vars, type = "inner", by = NULL, ...) {
     "  ON ", on, "\n",
     con = con
   )
+}
+
+sql_join_vars <- function(con, vars) {
+  sql_vector(
+    mapply(
+      FUN = sql_join_var,
+      alias = vars$alias,
+      x = vars$x,
+      y = vars$y,
+      MoreArgs = list(con = con),
+      SIMPLIFY = FALSE,
+      USE.NAMES = TRUE
+    ),
+    parens = FALSE,
+    collapse = ", ",
+    con = con
+  )
+}
+
+sql_join_var <- function(con, alias, x, y) {
+  if (!is.na(x) & !is.na(y)) {
+    sql_coalesce(
+      sql_table_prefix(con, x, table = "TBL_LEFT"),
+      sql_table_prefix(con, y, table = "TBL_RIGHT")
+    )
+  } else if (!is.na(x)) {
+    sql_table_prefix(con, x, table = "TBL_LEFT")
+  } else if (!is.na(y)) {
+    sql_table_prefix(con, y, table = "TBL_RIGHT")
+  } else {
+    stop("No source for join column ", alias, call. = FALSE)
+  }
+}
+
+sql_coalesce <- function(...) {
+  vars <- sql_vector(list(...), parens = FALSE, collapse = ", ")
+  build_sql("coalesce(", vars, ")")
 }
 
 sql_as <- function(con, var, alias = names(var), table = NULL) {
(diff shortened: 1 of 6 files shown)


R/translate-sql-base.r
```diff
@@ -122,6 +122,10 @@ base_scalar <- sql_translator(
     build_sql("((", x, ") IS NULL)")
   },
   na_if = sql_prefix("NULL_IF", 2),
+  coalesce = function(...) {
+    vars <- sql_vector(list(...), parens = FALSE, collapse = ", ")
+    build_sql("coalesce(", vars, ")")
+  },
```
ianmcook: This is functionally independent of the other changes here, because I don't think there's any clean way to apply SQL translation unless you're working with expressions. Is that correct?

hadley: Correct.

tests/testthat/test-sql-join-right-full.R
```diff
@@ -0,0 +1,135 @@
+context("SQL: right and full joins")
+
+test_that("right join on key column with same name in both tables", {
+  test_f <- function(tbl_left, tbl_right) {
+    res <-
```
hadley: Can you please follow the tidyverse style for indenting? http://style.tidyverse.org/syntax.html#indenting

tests/testthat/test-sql-join-right-full.R
```diff
@@ -0,0 +1,135 @@
...
+      right_join(tbl_left, tbl_right, by = "x") %>%
+      arrange(x, y, z) %>%
+      collect()
+    expect_equivalent(as.data.frame(res),
```
hadley: Please use `tibble()` to create sample data

tests/testthat/test-sql-join-right-full.R
```diff
@@ -0,0 +1,135 @@
...
+                   z = c(1L, 2L, 3L, 4L)
+                 ))
+  }
+
+  df_left <- data_frame(x = 1L:4L, y = 1L:4L)
+  df_right <- data_frame(x = c(1L:3L, 5L), z = 1L:4L)
+  # SQLite doesn't support right joins
+  tbls_left <- test_load(df_left, ignore = c("sqlite"))
+  tbls_right <- test_load(df_right, ignore = c("sqlite"))
+  mapply(test_f, tbls_left, tbls_right)
```
hadley: Please use the existing `compare_tbls()` function

tests/testthat/test-sql-join-right-full.R
```diff
@@ -0,0 +1,135 @@
+context("SQL: right and full joins")
+
+test_that("right join on key column with same name in both tables", {
```
hadley: The test name should ideally read like a sentence

tests/testthat/test-sql-join-right-full.R
hadley: The current code makes it hard to see why the result is as expected. Can you please spend a little time thinking about how to make sure that it's obvious that the expected result is correct?

R/sql-query.R
```diff
@@ -103,7 +103,14 @@ join_vars <- function(x_names, y_names, by, suffix = c(".x", ".y")) {
   y_new <- y_names
   y_new[y_match] <- paste0(y_names[y_match], suffix$y)
 
-  list(x = setNames(x_new, x_names), y = setNames(y_new, y_names))
+  x_x <- x_names
```
hadley: This needs a brief comment explaining the logic

R/sql-generic.R
```diff
@@ -110,10 +110,20 @@ sql_join.default <- function(con, x, y, vars, type = "inner", by = NULL, ...) {
     stop("Unknown join type:", type, call. = FALSE)
   )
 
-  select <- sql_vector(c(
-    sql_as(con, names(vars$x), vars$x, table = "TBL_LEFT"),
-    sql_as(con, names(vars$y), vars$y, table = "TBL_RIGHT")
-  ), collapse = ", ", parens = FALSE)
+  select <- sql_vector(
+    paste0(
+      sql_if_vars_x_and_y(vars, "COALESCE("),
```
hadley: It's currently a bit hard to see the logic here. It needs to be more clear that for left joins we prefer shared variables from `x`, for right joins we prefer shared variables from `y`, and for full joins we have to coalesce. I think this amount of complexity here suggests that the join vars data structure needs to be different.

R/sql-generic.R
```diff
@@ -145,6 +132,43 @@ sql_join.default <- function(con, x, y, vars, type = "inner", by = NULL, ...) {
...
+  } else {
+    stop("No source for join column ", alias, call. = FALSE)
+  }
+}
+
+sql_coalesce <- function(...) {
+  vars <- sql_vector(list(...), parens = FALSE, collapse = ", ")
+  build_sql("coalesce(", vars, ")")
+}
+
```
ianmcook: I think these changes improve the semantics. Instead of using vectorized a `paste0()` to build the SQL, this applies a function over the parallel vectors in `vars$alias`, `vars$x`, and `vars$y`.

ianmcook: Is the logic here clear enough? If so I'll make the other changes you requested. If not, then some more thinking is needed about what the `vars` data structure should be.

ianmcook:
Looks like most of this code is in dbplyr now (#2426).

hadley:
Would you mind moving this pull request to dbplyr please?

ianmcook:
Yes, will do, thanks.

ianmcook:
See https://github.com/hadley/dbplyr/pull/3.

PR #2597, opened 2017-03-30
Take dots with splicing

### LINKED ISSUE #2477: Dots splicing semantics for SE functions
Using `rlang::splice()`.

One issue is whether we allow bare list splicing or if we require the use of a modifier/adverb for signalling splicing. Bare-list splicing would apply most of the time, since most functions expect either atomic vectors or S3 objects in `...`. It is less verbose:

```r
bind_cols(mtcars, list(mtcars, mtcars), mtcars)
```

On the other hand, some functions also expect lists in `...`. A replacement for `structure()` would be an example. In this case an adverb like `spliced()` is necessary:

```r
struct(obj,
  attr1 = "attr",
  attr2 = list("attr"),
  spliced(list(attr3 = "attr"))
)
```

We'd need a standardised way of documenting those semantics in roxygen. But even then, would it be confusing to have these two kinds of splicing in the tidyverse?

lionel-:
Closes #2477 

Build fails because of dbplyr's rlang revision but passes locally.


hadley:
I've just removed the pinning in dbplyr, so it'll be implicitly pinned to the same version as dplyr.

NEWS.md
```diff
@@ -86,6 +86,7 @@ If you've implemented a database backend for dplyr, you'll need to re-read the v
   by `group_by_prepare()` now has a new `group_names` component (#1950, #2384).
 
 ## Colwise functions
+* `recode()` now supports argument splicing with `splice()`.
```
hadley: Needs a newline

NEWS.md
```diff
@@ -226,6 +227,7 @@ If you've implemented a database backend for dplyr, you'll need to re-read the v
 * `bind_rows()` explicitly rejects data frame columns (#2015, #2446).
 
 * `combine()` accepts `NA` values (#2203, @zeehio)
+* Fix `group_by()` for data frames that have UTF-8 encoded names (#2284, #2382).
```
hadley: Mistake?

PR #2598, opened 2017-03-30
Get rid of skip() calls in tests

krlmlr:
Now only two remaining. We could also move them to the corresponding GitHub issues.


hadley:
Thanks!

PR #2606, opened 2017-03-31
Error messages now mostly include column name or position

### LINKED ISSUE #2448: All error messages should mention the affected column(s)
https://github.com/hadley/dplyr/pull/2425#discussion_r102037470

krlmlr:
filter() is one of the few exceptions (#2605).

For review, the "test tweaks" commit contains most wording chainges.

Fixes #2448.


krlmlr:
@hadley: Could you please review the wording before I sweep over the code base to change the quoting?

hadley:
Here's a reasonably complete set of all the error messages:

```cpp
"%s at position %d"
"Can not automatically convert from %s to %s at position %d."
"can't arrange by a matrix at position %d"
"cannot arrange column of class '%s' at position %d"
"cannot group column '%s', of class %s"
"cannot modify grouping variable '%s'"
"column '%s' has unsupported class %s
"Column '%s' must be a vector, not a %s"
"incorrect size (%d) at position %d, expecting : %d"
"mutate does not support POSIXlt results for column '%s'"
"Unsupported vector type %s for column '%s'"
```

There are some basics. I think we should:

* should standardise on "can't" not "cannot" or "cannot" (similarly for "doesn't)

* Always start with "Col %s:", or "Row %d" (I think row is better than position). Don't end in with a full stop (i.e. an error message is a bullet point)

* either always include the verb or never include. I'd lean towards not including: if it's not obvious from the context (i.e. it's in the middle of the pipeline) you should be able to figure it out from the stack trace.

This leads to:

```
"Row %d: can't order by a %s"
"Row %d: can't automatically convert from %s to %s"
"Row %d: can't arrange by a matrix"
"Row %d: can't arrange by a %s"
"Col `%s`: can't group by a %s"
"Col `%s`: can't modify a grouping variable"
"Col `%s`: must be a vector, not a %s"
"Col `%s`: must be a vector, not a %s"
"Row %d: must be size %d, not %d"
"Col `%s`: unsupported vector type (POSIXct)"
"Col `%s`: unsupported vector type (%s)"
```

I think that illustrates a few other principles: 

* if we can say what the expected thing is, we should say it up front.
* types/classes should get an "a" in front

I think that should be enough so that you can fix the existing error messages. It would be useful to include a helper function where these principles can live as documentation.

krlmlr:
Thanks. Just one thing: "position" refers to the position in the argument list in verbs such as arrange() and filter() that take unnamed arguments. Would `"Pos %d"` work?

hadley:
If it's position in arguments, I think `Argument %d` would be better (and then use `Column` for consistency)

krlmlr:
So, in `arrange(mtcars, cyl, mpg)` the `cyl` column is still argument 1? This makes more sense for `mtcars %>% arrange(cyl, mpg)`.

Is there a special name for the arguments passed to the ellipsis?

hadley:
Yeah, I'd still call it argument 1. I can't think of a better name.

krlmlr:
This is waiting for #2608.

hadley:
@krlmlr do you have time to finish this off next week? No problem if not, @lionel- can take over

krlmlr:
Yes, I'll take care of this and the remaining issues in the project.

krlmlr:
~90 untested instances of stop() remain in the C++ code (not covered by tests), many can be formatted consistently using the `bad_xxx()` functions.

PR #2608, opened 2017-03-31
Revise all error messages for consistency

krlmlr:
in the R code. A prerequisite to #2606, the C++ code should be integrated with the formatting decisions made here.

I still need to look at the codecov report, which part of my patch hasn't been covered by the test.

### DIFF
--- a/R/bind.r
+++ b/R/bind.r
@@ -78,7 +78,8 @@ bind_rows <- function(..., .id = NULL) {
 
   if (!is.null(.id)) {
     if (!(is.character(.id) && length(.id) == 1)) {
-      stop(".id is not a string", call. = FALSE)
+      gabort("{hdr_args(~.id)} expected scalar string, ",
+        "got {typeof(.id)} of length {length(.id)}")
     }
     names(x) <- names(x) %||% seq_along(x)
   }
--- a/R/bind.r
+++ b/R/bind.r
@@ -78,7 +78,7 @@ bind_rows <- function(..., .id = NULL) {
 
   if (!is.null(.id)) {
     if (!(is.character(.id) && length(.id) == 1)) {
-      glubort("{hdr_args(~.id)} must be a scalar string, ",
+      glubort(args = ~.id, "must be a scalar string, ",
         "got {typeof(.id)} of length {length(.id)}")
     }
     names(x) <- names(x) %||% seq_along(x)
--- a/R/bind.r
+++ b/R/bind.r
@@ -78,7 +78,7 @@ bind_rows <- function(..., .id = NULL) {
 
   if (!is.null(.id)) {
     if (!(is.character(.id) && length(.id) == 1)) {
-      glubort("{hdr_args(~.id)} expected scalar string, ",
+      glubort("{hdr_args(~.id)} must be a scalar string, ",
         "got {typeof(.id)} of length {length(.id)}")
     }
     names(x) <- names(x) %||% seq_along(x)
--- a/R/bind.r
+++ b/R/bind.r
@@ -79,7 +79,7 @@ bind_rows <- function(..., .id = NULL) {
   if (!is.null(.id)) {
     if (!(is.character(.id) && length(.id) == 1)) {
       glubort(args = ~.id, "must be a scalar string, ",
-        "got {type_of(.id)} of length {length(.id)}")
+        "not {type_of(.id)} of length {length(.id)}")
     }
     names(x) <- names(x) %||% seq_along(x)
   }
--- a/R/bind.r
+++ b/R/bind.r
@@ -78,7 +78,7 @@ bind_rows <- function(..., .id = NULL) {
 
   if (!is.null(.id)) {
     if (!(is.character(.id) && length(.id) == 1)) {
-      gabort("{hdr_args(~.id)} expected scalar string, ",
+      glubort("{hdr_args(~.id)} expected scalar string, ",
         "got {typeof(.id)} of length {length(.id)}")
     }
     names(x) <- names(x) %||% seq_along(x)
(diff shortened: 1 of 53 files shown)


R/error.R
```diff
@@ -0,0 +1,83 @@
+check_pkg <- function(name, reason) {
+  if (requireNamespace(name, quietly = TRUE))
+    return(invisible(TRUE))
+
+  gabort('The {name} package is required to {reason}.
+    Please install it with `install.packages("{name}")`'
+  )
+}
+
+gabort <- function(..., .envir = parent.frame()) {
```
hadley: I think `glubort()` would pronounce better and is only a couple of characters longer

R/error.R
```diff
@@ -0,0 +1,83 @@
...
+  abort(glue(..., .envir = .envir))
+}
+
+hdr_args <- function(...) {
+  x <- parse_args(...)
+  if (length(x) == 1) args <- "Argument"
```
hadley: `ngettext(1, "Argument", "Arguments")` is a cleaner approach

R/error.R
```diff
@@ -0,0 +1,83 @@
...
+  gabort('The {name} package is required to {reason}.
+    Please install it with `install.packages("{name}")`'
+  )
+}
+
+gabort <- function(..., .envir = parent.frame()) {
+  abort(glue(..., .envir = .envir))
+}
+
+hdr_args <- function(...) {
```
hadley: What does `hdr` stand for?

R/colwise-filter.R
```diff
@@ -61,7 +61,8 @@ apply_filter_syms <- function(pred, syms, tbl) {
   } else if (inherits(pred, "any_vars")) {
     joiner <- any_exprs
   } else {
-    abort("`.vars_predicate` should be called with `all_vars()` or `any_vars()`")
+    gabort("{hdr_args(~.vars_predicate)} expected call to `all_vars()` or `any_vars()`, ",
```
hadley: I think I prefer "must" rather than "expected"

R/join.r
```diff
@@ -145,20 +145,12 @@ common_by_from_vector <- function(by) {
 common_by.list <- function(by, x, y) {
   x_vars <- tbl_vars(x)
   if (!all(by$x %in% x_vars)) {
-    stop(
-      "Join column not found in lhs: ",
-      paste(setdiff(by$x, x_vars), collapse = ", "),
-      call. = FALSE
-    )
+    gabort("{hdr_args(~by)} join column {fmt_obj(setdiff(by$x, x_vars))} not found in lhs")
```
hadley: I don't love such big chunks of code embedded in text. Maybe pull out into local variable or named arg to `glubort()`?

R/bind.r
```diff
@@ -78,7 +78,8 @@ bind_rows <- function(..., .id = NULL) {
 
   if (!is.null(.id)) {
     if (!(is.character(.id) && length(.id) == 1)) {
-      stop(".id is not a string", call. = FALSE)
+      gabort("{hdr_args(~.id)} expected scalar string, ",
```
hadley: `{hdr_args(~.id)}` seems considerably worse than `` `.id` `` to me

R/error.R
krlmlr: Header.

R/bind.r
krlmlr: This expands to ```Argument `.id` ```. I'll add a second argument to `glubort()` that will accept an optional header expression, this will get a nicer syntax.

hadley: In general, error message shouldn't get an "Argument" in front, because it's obvious from the context. You only need argument when you're referring to an index

krlmlr:
glubort() gained a bunch of arguments that specify the header of the error message.

R/grouped-df.r
```diff
@@ -282,7 +282,8 @@ sample_frac.grouped_df <- function(tbl, size = 1, replace = FALSE,
     warn("`.env` is deprecated and no longer has any effect")
   }
   if (size > 1 && !replace) {
-    abort("Sampled fraction can't be greater than one unless replace = TRUE")
+    glubort(args = c(~size, ~replace), "sampled fraction must be less or equal to one, ",
+      "pass TRUE for sampling with replacement")
```
hadley: This still needs `` `replace = TRUE` `` I think

R/bind.r
```diff
@@ -78,7 +78,8 @@ bind_rows <- function(..., .id = NULL) {
 
   if (!is.null(.id)) {
     if (!(is.character(.id) && length(.id) == 1)) {
-      stop(".id is not a string", call. = FALSE)
+      glubort(args = ~.id, "must be a scalar string, ",
```
hadley: I think I'd prefer 

```R
something_arg(
  .id, "must be a scalar string, not {type_of(.id)} of length {length(.id)}"
)
```

* I think `something` no longer needs to reference glue or abort since this is clearly customised for dplyr
* "got" -> "not"
* closing parens on own line

krlmlr:
Done, fd1e7f2 is best for review.

R/error.R
lionel-: Should `rlang::abort()` etc take dots, use `paste0()` by default, and check if `glue()` is imported in the search path and use it if it's there?

You'd get different behaviour depending on whether you `@importFrom glue glue`, but that seems reasonable since unit tests should catch fairly quickly that you've forgotten to include `glue()`. Might be a bit more problematic when using rlang in scripts, though functions involving error messages should typically live in packages.

lionel-:
why do you quote with a formula rather than in a string? I may have missed it but it doesn't look like you need to evaluate these arguments.

lionel-:
@jimhester should `glue()` capture dots as quosures? That would make it easier to construct wrappers around it because there would be no need to forward the environment. Also this would make it possible to have several layers of wrappers with arguments evaluated in different frames, which is currently not supported (though admittedly will not be a common use case).

hadley:
Yeah, strings would be nicer I think.

hadley:
Otherwise API and error messages look good.

lionel-:
@jimhester nevermind my former suggestion, it won't work because passing strings is now optimised by the compiler and don't carry environment info.

krlmlr:
re strings: this is an artefact of a former implementation, will revise and merge.

PR #2621, opened 2017-04-06
Implement vector bind

### LINKED ISSUE #1676: Better semantics for dataframeable objects
After thinking about https://github.com/hadley/purrr/issues/179, I wonder if we could improve the semantics of dataframeable objects in `bind_rows()` by taking a vector viewpoint rather than a list viewpoint.

If we think about vectors, the current behaviour is not consistent because `bind_rows()` effectively binds columns together:

``` r
col_vectors <- list(
  a = c(1, 2),
  b = c(3, 4)
)

row_vectors <- list(
  c(a = 1, b = 3),
  c(a = 2, b = 4)
)
```

``` r
bind_rows(row_vectors)
#> Error: cannot convert object to a data frame

bind_rows(col_vectors)
#> Source: local data frame [2 x 2]
#>
#>       a     b
#>   (dbl) (dbl)
#> 1     1     3
#> 2     2     4
```

I think it'd be more intuitive to go the `rbind()` and `cbind()` way, and treat vectors differently depending on the direction of the binding. `bind_rows()` would require vectors with inner names and `bind_cols()` would require vectors with outer names.

Advantages:
- Arguably more intuitive if we think about vectors rather than lists.
- Better behaviour down the line, e.g. `map_df()`
- This would be compatible with splicing the input as suggested in PR #992.

The latter would add a lot of flexibility. `bind_cols()` and `bind_rows()` would accept indistinctly data frames, named vectors, and lists of data frames and named vectors.

cc @jennybc

lionel-:
Closes #1676.

`bind_rows()` and `bind_cols()` now handle vectors. Vectors are considered to be rows by `bind_rows()` and columns by `bind_cols()`.

Lists are still treated as dataframes, so they require explicit splicing: `bind_rows(!!! mylist)`. Should they now yield a warning?

The input is preprocessed with `rlang::flatten_if()`, called with a native symbol object containing a pointer to a C-level predicate. This seems to work well and after reading `.Call()`'s implementation, it seems that casting external pointers to function pointers is portable. If you prefer I'll copy `flatten_if()` implementation over to dplyr until the next release.


src/bind.cpp
```diff
@@ -8,64 +8,175 @@
...
 
-template <typename Dots>
-List rbind__impl(Dots dots, SEXP id = R_NilValue) {
+extern "C"
+bool bind_spliceable(SEXP x) {
+  if (TYPEOF(x) != VECSXP)
+    return false;
+
+  if (OBJECT(x)) {
+    if (Rf_inherits(x, "spliced"))
```
hadley: `return Rf_inherits(x, "spliced")` ?

lionel-: must have been tired ;)

PR #2626, opened 2017-04-08
New interface registration, and Rcpp::interfaces

### LINKED ISSUE #2146: Use Rcpp::interfaces instead of GRAB_CALLABLE
https://www.rdocumentation.org/packages/Rcpp/versions/0.12.6/topics/interfacesAttribute

krlmlr:
I've been using the excellent [instructions](https://ironholds.org/registering-routines/) by @ironholds (thanks!) to add registration. It's a pity Rcpp doesn't do it automatically (yet).

Fixes #2146.


R/bind.r
```diff
@@ -106,7 +106,7 @@ NULL
 #' @export
 #' @rdname bind
 bind_rows <- function(..., .id = NULL) {
-  x <- flatten_if(dots_values(...), get_is_bind_spliceable())
+  x <- flatten_if(dots_values(...), get_is_bind_spliceable()[[1L]])
```
krlmlr: @hadley: I'm having second thoughts, is this actually safe?

`get_is_bind_spliceable()` returns an xptr wrapped in a list. With `[[1L]]`, the naked xptr is wrapped to `flatten_if()`.

CC @lionel-.

lionel-: This should be completely safe. 

But I'm going to look into this issue today because I'm not sure I like how the registration of flatten_if plugins looks like with the new system. The purpose of using native symbols was to make it easy and safe to pass pointers around, but we may have to find another way if that doesn't play nice with the new registration system.

lionel-: > This should be completely safe.

hmm I think I see what you mean. The list is not bound in `flatten_if()`, only the pointer is, so the list might get GC'd and the pointer with it. In that case it might be safer to return the pointer directly: it would be bound in `flatten_if()` but nowhere else, so it won't get GC'd in another place. These pointers copy semantics are crazy, I'm going to look into these as well.

krlmlr: @lionel-: I have removed hadley/rlang@76a6621b from the rlang PR, and will change this PR to pass around the xptr wrapped in a list. This may be not the nicest way, but it's safer than working with naked xptrs. Would you like to take over both PRs?

lionel-: yes if I find a better way I'll send changes to your branches, otherwise we can use your modifications for the time being, the PRs look good.

src/init_register.c
```diff
@@ -0,0 +1,119 @@
...
+};
+
+/* Raw function pointers */
+extern bool dplyr_is_bind_spliceable(SEXP x);
+
+void R_init_dplyr(DllInfo *dll)
+{
+  R_registerRoutines(dll, NULL, CallEntries, NULL, NULL);
+  R_useDynamicSymbols(dll, FALSE);
+  rlang_register_pointer("dplyr", "is_bind_spliceable", (DL_FUNC) &dplyr_is_bind_spliceable);
```
krlmlr: @lionel-: Do we still need this line?

lionel-: We're going to switch back to `flatten_if()` after the release, I guess it doesn't hurt to leave it there.

PR #2627, opened 2017-04-08
Strict matching for .data

### LINKED ISSUE #2591: .data and .env need strict matching semantics
Using `$.data_source` from rlang

krlmlr:
I think strict matching for .env needs to be fixed in rlang, CC @lionel-.

Fixes #2591.


krlmlr:
@lionel-: I'm also wondering why both grouped and ungrouped tests appear to hit the same code path. Is this new?

lionel-:
@krlmlr could you use the rlang ctor I provided in https://github.com/hadley/rlang/pull/92 please?

The class is now `dictionary` and has a `read_only` field.

PR #2628, opened 2017-04-10
dbplyr compatibility

hadley:
These are some helper to make it easier to write backends that work with both dplyr 0.5 and 0.6.


R/compat-dbplyr.R
```diff
@@ -0,0 +1,69 @@
...
+
+#' @export
+#' @rdname check_dbplyr
+dbplyr_fun <- function(fun_name) {
+  args <- formals(getExportedValue("dbplyr", fun_name))
+  pass_on <- map(set_names(names(args)), as_symbol)
+  fun <- as_symbol(fun_name)
+
+  body <- expr({
+    if (utils::packageVersion("dplyr") > "0.5.0") {
```
krlmlr: I don't understand the purpose of the check here. It looks like it will hit the `else` branch only if `dplyr` was > 0.5.0 during *installation* of the third-party backend package, and then switched back to `dplyr` <= 0.5.0. What am I missing?

Instead of exporting the three functions here, we should perhaps provide means to install the compat functions in the backend's source tree (like `use_dbplyr_compat()`).

hadley: The expected usage is like this:

```R
src_sql <- dbplyr_fun("src_sql")
src_sql
#> function (subclass, con, ...) 
#> {
#>     if (utils::packageVersion("dplyr") > "0.5.0") {
#>         dplyr::check_dbplyr()
#>         dbplyr::src_sql(subclass = subclass, con = con, ... = ...)
#>     }
#>     else {
#>         dplyr::src_sql(subclass = subclass, con = con, ... = ...)
#>     }
#> }
```

i.e. the check is definitely done at run-time.

hadley: `expr()` is the equivalent of `quote()` with support for quasiquotation

R/compat-dbplyr.R
```diff
@@ -0,0 +1,73 @@
...
+#' # cache the value at build time, not run time, causing bugs that
+#' # extremely hard to track down.
+#' }
+check_dbplyr <- function() {
+  check_pkg("dbplyr", "communicate with database backends")
+}
+
+#' @export
+#' @rdname check_dbplyr
+dbplyr_fun <- function(fun_name) {
```
krlmlr: Let's assume a dplyr backend that wants to support both dplyr 0.5.0 and 0.6.0. What happens if it uses `dbplyr_fun()` and is installed when currently dplyr 0.5.0 is installed?

IIUC, the backend would have a call like

```r
xxx <- dbplyr_fun("xxx")
```

in one of its source files, but how does it access `dbplyr_fun()` if dplyr 0.5.0 is installed?

hadley: Doh. I didn't think that through. I'll change the docs to suggest that they use these as code generation functions, then copy the results in their own package.

PR #2629, opened 2017-04-11
Programming vignette typo

ijlyttle:
Really enjoying this vignette!

It will likely take me a few evenings to get through the entire thing, but I thought I would at least start a PR. 

I can send a signal when this is "done", but happy to take feedback in the meantime.


yutannihilation:
AFAIK, tidyeval, or tidy evaluation, is a concept and rlang is a package, which is not just about tidyeval :)

lionel-:
yeah, I don't think this is a good idea to change tidyeval to rlang ;)

It shouldn't be mentioned at all. We may extract a tidyeval package from rlang in the future.

yutannihilation:
@lionel- By the way, it would be better if the vignette has some clear sentences about "what tidyeval is", IMHO. Can I expect this for somewhere else? (e.g. vignette of rlang?)

ijlyttle:
I'll go through and make another pass at tidyeval vs. rlang - I get that there may be other things going on with the immediate future of these packages. 

Overall, my goal is simply to clean things up a bit so it might be easier for others to make the next big changes. :)

vignettes/programming.Rmd
```diff
@@ -78,7 +78,7 @@ This vignette has two goals:
     
 *   To teach you the underlying theory including __quosures__, the data 
     structure that stores both an expression and an environment, and 
-    and __tidyeval__, the underlying toolkit.
+    and __rlang__, the underlying toolkit.
```
hadley: Can you please undo the tidyeval -> rlang translation

ijlyttle: will do

ijlyttle:
I think this is ready-to-go other than the question on 4a70d9a, above.

hadley:
Yeah, they should definitely use `quo()`.

ijlyttle:
Really ready-to-go now

hadley:
Thanks!

PR #2636, opened 2017-04-12
Retire deprecated rbind_list() and rbind_all()

lionel-:


hadley:
When were these deprecated?

lionel-:
0.5.0. Too soon?

hadley:
That was almost a year ago, so I think should be ok.

PR #2637, opened 2017-04-12
Fix hybrid handling of summarised symbols

### LINKED ISSUE #2632: summarize-defined columns in lists
Problem in dplyr 0.5.0 when using summarize-defined column in list.
Sorry, I couldn't get the github::install_github to build the latest dplyr and check if this has already been fixed.

```r
library(dplyr)
library(testthat)
df = mtcars %>%
  group_by(cyl) %>%
  summarize(x = min(cyl), y = list(x), z = x)

expect_equal(df[["y"]], list(2, 4, 8))
expect_false(identical(df[["y"]], list(8, 8, 8)))
```

lionel-:
So this works: `summarise(group_by(mtcars, cyl), x = 1, z = x)`

Closes #2632

### DIFF
--- a/src/hybrid.cpp
+++ b/src/hybrid.cpp
@@ -105,7 +105,7 @@ public:
   VariableResult(const ILazySubsets& subsets_, const SymbolString& name_) : subsets(subsets_), name(name_)  {}
 
   SEXP process(const GroupedDataFrame& gdf) {
-    if (subsets.count(name) && subsets.is_summary(name)) {
+    if (subsets.is_summary(name)) {
       // No need to check length since the summary has already been checked
       return subsets.get_variable(name);
     } else {
--- a/src/hybrid.cpp
+++ b/src/hybrid.cpp
@@ -105,8 +105,13 @@ public:
   VariableResult(const ILazySubsets& subsets_, const SymbolString& name_) : subsets(subsets_), name(name_)  {}
 
   SEXP process(const GroupedDataFrame& gdf) {
-    check_length(gdf.max_group_size(), 1, "a summary value", name);
-    return process(*gdf.group_begin());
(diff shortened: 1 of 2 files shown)


src/hybrid.cpp
```diff
@@ -105,8 +105,13 @@ class VariableResult : public Result {
   VariableResult(const ILazySubsets& subsets_, const SymbolString& name_) : subsets(subsets_), name(name_)  {}
 
   SEXP process(const GroupedDataFrame& gdf) {
-    check_length(gdf.max_group_size(), 1, "a summary value", name);
-    return process(*gdf.group_begin());
+    if (subsets.count(name) && subsets.is_summary(name)) {
```
krlmlr: Why is it relevant to check `subsets.count(name)`?

lionel-: I guess it's superfluous.

lionel-:
> does an existing test fail if you omit that? 

yes, we don't get proper error message when the result length isn't correct.

tests/testthat/test-summarise.r
```diff
@@ -947,3 +947,7 @@ test_that("proper handling of NA factors (#2588)", {
   ret <- df %>% group_by(x) %>% summarise(y = y[1])
   expect_identical(as.character(ret$y), c(NA, NA, "3"))
 })
+
+test_that("can refer to previously summarised symbols", {
+  expect_identical(summarise(group_by(mtcars, cyl), x = 1, z = x)[2:3], tibble(x = c(1, 1, 1), z = x))
```
krlmlr: Actually, could you double-check with a more complicated expression for `x`, say, `x = n()`?

lionel-: yup it works, just added a test for it.

PR #2639, opened 2017-04-12
Fixed typo in programming vignette

jlegewie:


lionel-:
Thanks, I think this has been fixed already.

PR #2641, opened 2017-04-13
Remove -Wall -Wextra warnings

krlmlr:
- Unused arguments
- One instance of potentialy wrong behavior detected

### DIFF
--- a/inst/include/dplyr/Collecter.h
+++ b/inst/include/dplyr/Collecter.h
@@ -93,7 +93,7 @@ public:
     return RTYPE == TYPEOF(x) || all_logical_na(x, TYPEOF(x));
   }
 
-  bool can_promote(SEXP x) const {
+  bool can_promote(BOOST_ATTRIBUTE_UNUSED SEXP x) const {
     return false;
   }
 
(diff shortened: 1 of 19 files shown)


lionel-:
do we really need to remove all warnings?

krlmlr:
Yes.

lionel-:
> Yes.

if you tell me why is that the case that we need to address esoteric warnings, I could add the relevant flags to my local builds and you wouldn't need to track warnings after I push code. These boost unused attributes etc seem to obscure the code without real benefit.

krlmlr:
I agree with your last point, but that esoteric warning helped pinpoint a potential bug and unnecessarily complicated code.

There are other options:

- Define a shorter macro (e.g., `UNUSED`)
- Use the `(void)arg;` idiom
- Don't name arguments we don't use (e.g., `int func(SEXP)`)

In my g++ builds, I'm now using `-Wall -Wextra -pedantic`, and I hope to add `-Wconversion` soon; in addition, `-Werror` treats warnings as errors (they are easy to overlook otherwise) and `-Wfirst-error` aborts on first error (which is the only one relevant in many cases).

lionel-:
I think I prefer unnaming the unused arguments.

hadley:
I'd prefer unnaming the unused arguments too.

PR #2642, opened 2017-04-13
Create Makevars on Travis

krlmlr:
that set additional warning flags. The build is aborted in case of warnings.

This requires RcppCore/Rcpp#671 or a workaround.


krlmlr:
Rcpp PR now merged, but let's wait until the next Rcpp release mid-May before merging this PR. I have this setup locally, so no warnings will get unnoticed when I compile.

krlmlr:
Confirmed that this actually detects compilation warnings and treats them as errors: https://travis-ci.org/krlmlr/dplyr/builds/253138104#L930. Upcoming Rcpp will even support `-Wconversion`.

PR #2672, opened 2017-04-16
Join accepts again POSIXct with different timezones

### LINKED ISSUE #2643: Joins are too strict about tz attributes
I think this is going to cause major pain for people

``` r
library(nycflights13)

hourly_delay <- flights %>% 
  group_by(time_hour, origin) %>% 
  summarise(delay = mean(dep_delay, na.rm = TRUE))
    
hourly_delay %>% 
  inner_join(weather, by = c("time_hour", "origin"))
#> Error in inner_join_impl(x, y, by$x, by$y, suffix$x, suffix$y, accept_na_match): attributes are different


attributes(flights$time_hour)
#> $class
#> [1] "POSIXct" "POSIXt" 
#> 
#> $tzone
#> [1] "UTC"
attributes(weather$time_hour)
#> $class
#> [1] "POSIXct" "POSIXt" 
#> 
#> $tzone
#> [1] ""
```

@zeehio would this be easy to fix?

@krlmlr could we get a better error message here? Ideally needs to give variable names on both sides of the join

zeehio:
Closes #2643

This was working on dplyr 0.5.0 so I did not add a news entry for this.

### DIFF
--- a/src/join.cpp
+++ b/src/join.cpp
@@ -46,6 +46,10 @@ void check_attribute_compatibility(SEXP left, SEXP right) {
   SEXP att_right = ATTRIB(right);
   int n_left = count_attributes(att_left);
   int n_right = count_attributes(att_right);
+  
+  if (Rf_inherits(left, "POSIXct") &&  Rf_inherits(right, "POSIXct")) {
+    return;
+  }
 
   if (n_left != n_right)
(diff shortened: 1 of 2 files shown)


hadley:
Thanks!

PR #2674, opened 2017-04-17
Better doc for deprecated _each() functions

lionel-:
Also removes the `_each_q()` functions from the package (they were deprecated in 0.3).


lionel-:
@hadley done

PR #2679, opened 2017-04-17
Only strip names of atomic vectors

### LINKED ISSUE #2575: Mutate return length where statistics are inside ifelse conditionals
---

In order to set the values of a particular group to NA based on a conditional, one must pre-compute and assign the summation to a new field, like so:

```r
group_by(clinic, measure, month) %>%
mutate(pre_value_sum = sum(value),
           value = ifelse(pre_value_sum < min_score_denominator, NA, value)) %>%
select(-pre_value_sum)
```

I would expect the following code to work, however value[1] is returned for *all* rows instead of the *original* vector of values.

```r
group_by(clinic, measure, month) %>%
mutate(value = ifelse(sum(value) < min_score_denominator, NA, value))
```

### LINKED ISSUE #2675: mutate() should preserve names in list-columns
To fix https://github.com/tidyverse/dplyr/pull/1689, we now strip names from all vectors in summarise and mutate (https://github.com/tidyverse/dplyr/pull/2512).

I think this is slightly too strong, and we should only strip names from atomic vectors. The current behaviour breaks `tidyr::unnest()` which assumes `transmute()` preserves the names (and there is a unit test that checks it). I'd prefer not to change the API for `unnest()` because it seems likely that someone will have relied on this behaviour.

@krlmlr do you think it's ok to make the stripping conditional (i.e. only for atomic vectors)? I've tried it out locally: the tests now pass in tidy, and only one test fails in dplyr.

```R
test_that("mutate strips names of list-columns, but keeps names in original data (#2523)", {
  vec <- list(a = 1, b = 2)
  data <- data_frame(x = vec)
  data <- mutate(data, x)
  expect_identical(names(vec), c("a", "b"))
  expect_null(names(data$x))
})
```

hadley:
Fixes #2675

### DIFF
--- a/inst/include/dplyr/NamedListAccumulator.h
+++ b/inst/include/dplyr/NamedListAccumulator.h
@@ -2,6 +2,7 @@
 #define dplyr_NamedListAccumulator_H
 
 #include <tools/SymbolMap.h>
+#include <tools/utils.h>
 
 #include <dplyr/checks.h>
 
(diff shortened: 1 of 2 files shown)


krlmlr:
I'd like to fix handling of names instead. Surprisingly, this is also related to #2680.

PR #2689, opened 2017-04-19
Implement tbl_sum.grouped_df() instead of print.grouped_df()

krlmlr:
tibble already defined the `tbl_sum()` generic, but it hasn't been used anywhere until now. I wonder if we should keep the naming (because it's similar to the existing `obj_sum()` and `type_sum()` generics) or if we should define and use new names. I also wonder if we really need to make printing of body and footer configurable in tibble.

PR to dbplyr follows. This is how the new printing looks like:

``` r
dplyr::group_by(iris, Species)
#> # A tibble: 150 × 5
#> # Groups: Species [3]
#>    Sepal.Length Sepal.Width Petal.Length Petal.Width Species
#>           <dbl>       <dbl>        <dbl>       <dbl>  <fctr>
#>  1          5.1         3.5          1.4         0.2  setosa
#>  2          4.9         3.0          1.4         0.2  setosa
#>  3          4.7         3.2          1.3         0.2  setosa
#>  4          4.6         3.1          1.5         0.2  setosa
#>  5          5.0         3.6          1.4         0.2  setosa
#>  6          5.4         3.9          1.7         0.4  setosa
#>  7          4.6         3.4          1.4         0.3  setosa
#>  8          5.0         3.4          1.5         0.2  setosa
#>  9          4.4         2.9          1.4         0.2  setosa
#> 10          4.9         3.1          1.5         0.1  setosa
#> # ... with 140 more rows
```


DESCRIPTION
```diff
@@ -23,7 +23,7 @@ Imports:
     rlang (>= 0.0.0.9017),
     R6,
     Rcpp (>= 0.12.6),
-    tibble (>= 1.2),
+    tibble (>= 1.3.0.9001),
```
hadley: Is this a hard dependency? I think it will still work with an older version?

krlmlr: Unfortunately, yes, because of the way the output of `tbl_sum()` is processed: tidyverse/tibble@e42da622e62ef30271f4f9f254c3d76fb333dc8d

krlmlr:
@hadley: Okay to merge?

hadley:
Yeah, lets go for it.

PR #2693, opened 2017-04-19
Remove names only for grouped summarise() and mutate()

### LINKED ISSUE #2675: mutate() should preserve names in list-columns

krlmlr:
Includes a refactoring, the second commit contains the essential change.

Fixes #2675.

Closes #2679.

Do we support unnesting grouped data frames in tidyr?

### DIFF
--- a/inst/include/dplyr/NamedListAccumulator.h
+++ b/inst/include/dplyr/NamedListAccumulator.h
@@ -9,20 +9,16 @@ namespace dplyr {
 
 template <typename Data>
 class NamedListAccumulator {
-public:
+private:
   SymbolMap symbol_map;
-  std::vector<RObject> data;
+  std::vector<RObject> data; // owns the results
 
+public:
   NamedListAccumulator() {}
 
   inline void set(const SymbolString& name, RObject x) {
     if (! Rcpp::traits::same_type<Data, RowwiseDataFrame>::value)
       check_supported_type(x, name);
-
-    if (!Rf_isNull(Rf_getAttrib(x, R_NamesSymbol))) {
-      x = Rf_shallow_duplicate(x);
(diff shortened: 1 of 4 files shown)


hadley:
We do, but the implementation is trivial:

```
> tidyr:::unnest_.grouped_df
function (data, unnest_cols, .drop = NA, .id = NULL, .sep = NULL) 
{
    out <- unnest_(dplyr::ungroup(data), unnest_cols, .drop = .drop, 
        .id = .id, .sep = .sep)
    regroup(out, data)
}
```

krlmlr:
I feel that data frames just shouldn't support names for any columns. For grouped operations this probably never worked correctly, but for now we're stuck with supporting them for ungrouped operations.

Perhaps we could warn, so that we can finally get rid of them in dplyr 0.7.0?

PR #2694, opened 2017-04-20
Avoid partial matching on select_vars() arguments

### LINKED ISSUE #2692: select() partial matches on select_vars() arguments
Something seems to have broken between 0.5.0.9000 and 0.5.0.9004 because this no longer works:

```r
data.frame(a = 1:4) %>% select(var = a)
Error in typeof(x) : object 'a' not found
```

I think it's because `var` is already assigned to an existing function. This works perfectly:

```r
data.frame(a = 1:4) %>% select(b = a)
```

lionel-:
Fixes #2692 

It is probably too late to rename `vars` to `.vars`, even if it is first argument.

@topepo did you copy `select_vars()` implementation for recipes? If so it's better to prefix arguments of functions taking data via dots with `.` to avoid this issue. Any variable named `var` passed to `select_vars()` will partial match on `vars` :/

### DIFF
--- a/R/dataframe.R
+++ b/R/dataframe.R
@@ -106,7 +106,8 @@ arrange_.data.frame <- function(.data, ..., .dots = list()) {
 
 #' @export
 select.data.frame <- function(.data, ...) {
-  vars <- select_vars(names(.data), ...)
+  # Pass via splicing to avoid matching select_vars() arguments
+  vars <- select_vars(names(.data), !!! quos(...))
   select_impl(.data, vars)
 }
 #' @export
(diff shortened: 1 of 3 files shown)


lionel-:
on second thought maybe the first commit is better as it allows selecting variables named `include` or `exclude`.

hadley:
Yeah, I think you're right.  Can you add a couple of tests for those vars too?

hadley:
And it probably needs a warning in the `select_vars()` docs

topepo:
Thanks for the heads up. I did make a copy but it doesn't use `vars` and I tend to be fairly explicit about using argument names in function calls.

PR #2707, opened 2017-04-26
Fix off-by-one error in bind_rows() error msg

jennybc:
~~The 2nd commit holds an unrelated fix I needed to pass tests. I can remove if that is somehow peculiar to me.~~ *removed* Also added `rlang::` while poking around this test. Some rlang calls are namespaced in the tests while others are not, so not sure how it should be.


lionel-:
Thanks! rlang function don't need to be namespaced.

lionel-:
just noticed the branch name and loved it ;)

Issue #2279, opened 2016-12-01
top_n crashes


burkovae:
Recently I was experimenting with top_n and this (wrong) application of `top_n` does fry all R sessions:

`data.frame(x = 1:10) %>% top_n(10, "x")`

```
 *** caught segfault ***
address 0x20000010, cause 'memory not mapped'

Traceback:
 1: .Call("dplyr_filter_impl", PACKAGE = "dplyr", df, dots)
 2: filter_impl(.data, dots)                                                                                                                                                                     
 3: filter_.tbl_df(tbl_df(.data), .dots = dots)                                                                                                                                                  
 4: filter_(tbl_df(.data), .dots = dots)                                                                                                                                                         
 5: as.data.frame(filter_(tbl_df(.data), .dots = dots))                                                                                                                                          
 6: filter_.data.frame(.data, .dots = lazyeval::lazy_dots(...))                                                                                                                                  
 7: filter_(.data, .dots = lazyeval::lazy_dots(...))                                                                                                                                             
 8: filter(x, min_rank(desc("x")) <= 10)                                                                                                                                                         
 9: eval(expr, envir, enclos)                                                                                                                                                                    
10: eval(call)                                                                                                                                                                                   
11: top_n(., 10, "x")                                                                                                                                                                            
12: function_list[[k]](value)                                                                                                                                                                    
13: withVisible(function_list[[k]](value))                                                                                                                                                       
14: freduce(value, `_function_list`)                                                                                                                                                             
15: `_fseq`(`_lhs`)                                                                                                                                                                              
16: eval(expr, envir, enclos)
17: eval(quote(`_fseq`(`_lhs`)), env, env)
18: withVisible(eval(quote(`_fseq`(`_lhs`)), env, env))
19: data.frame(x = 1:10) %>% top_n(10, "x")
```

krlmlr:
Thanks, confirmed.

PR #2710, opened 2017-04-26
Allow top_n() argument `wt` to be a call

### LINKED ISSUE #2705: Can no longer evaluate a function for weight in top_n()
The following works with the current version on CRAN but does not with the development version. This is pretty nice functionality for log ratios, absolute values, and such.

``` r
library(dplyr)

mtcars %>%
  top_n(5, abs(mpg))
#> Error: is_symbol(wt) is not TRUE
```

lionel-:
Closes #2705

The new tidyeval implementation of `filter()` has fixed https://github.com/tidyverse/dplyr/issues/2279. However it makes sense to check that `wt` evaluate to a vector as long as the number of rows.

@hadley I'm not sure what's the generic way of retrieving the number of rows. Should `tbl_rows_n()` (added in this PR) be exported at some point?

### DIFF
--- a/R/top-n.R
+++ b/R/top-n.R
@@ -68,5 +68,5 @@ top_n <- function(x, n, wt) {
 
 # Works with lazy tibbles
 tbl_rows_n <- function(tbl) {
-  pull(summarise(tbl, n = n()))
+  pull(summarise(ungroup(tbl), n = n()))
 }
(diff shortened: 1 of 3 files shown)


hadley:
Can we just omit the length check for now?  It's hard to compute nrows across all backends

PR #2711, opened 2017-04-26
Add scoped verbs for arranging

### LINKED ISSUE #2682: arrange_if, arrange_at, arrange_all?
(This issue is just out of curiosity. I don't find these useful yet)

Together with `filter()`, `mutate()`, `select()`, and `summarise()`, `arrange()` is referred to as

> provide the basis of a language of data manipulation.

in the vignette, "Introduction to dplyr"

While all of other basic verbs (plus `group_by()`) got their colwise forms, why are there no scoped `arrange()`? Are there any good reason to avoid implementing `arrange()` variants? Or, is it just because no one has requested the feature so far?

`arrange()` has a bit different semantics in that the order of arguments affects the result and it is difficult to deduce which columns are used in what order from the result. So, I feel the idea of colwise may not be suitable for this verb. But, I still wonder it should exist or not just for consistency.

lionel-:
Closes #2682

Arranging by descending order is trivial after all, since `arrange()` has mutate semantics.

```r
arrange_all(mtcars, `-`)
```


R/colwise-arrange.R
```diff
@@ -0,0 +1,43 @@
...
+#' @export
+#' @examples
+#' df <- as_tibble(mtcars)
+#' df
+#' arrange_all(df)
+#'
+#' # You can supply a function that will be applied before taking the
+#' # ordering of the variables. The variables of the sorted tibble
+#' # keep their original values.
+#' arrange_all(df, `-`)
```
hadley: I feel like just one example using `desc` is enough — I don't want to encourage people to use `-` because it only works with numerics

PR #2713, opened 2017-04-27
Tidy naming for pronouns

lionel-:
Requires https://github.com/hadley/rlang/pull/109 (Travis will fail due to rlang version mismatch).

That currently only deals with pronouns at the top of the call tree.

### DIFF
--- a/R/utils-expr.R
+++ b/R/utils-expr.R
@@ -29,3 +29,19 @@ expr_substitute <- function(expr, old, new) {
   )
   expr
 }
+
+is_pronoun <- function(expr) {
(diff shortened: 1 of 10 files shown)


R/utils-expr.R
```diff
@@ -29,3 +29,19 @@ expr_substitute <- function(expr, old, new) {
   )
   expr
 }
+
+is_pronoun <- function(expr) {
+  is_lang(expr, quote(`$`)) && identical(node_cadr(expr), quote(.data))
```
hadley: Probably should check for either `$` or `[[`

PR #2721, opened 2017-04-30
Fix indexing beyond vector length in nth() hybrid handler

### LINKED ISSUE #2673: Segmentation fault when using summarise()
The following code doesn't make much sense, but it creates a segmentation fault:

```r
df <- data.frame(id = c(1, 2, 3, 3, 4),
                 color = c(NA, "blue", "red", "blue", NA),
                 stringsAsFactors = FALSE)

df%>% 
  group_by(id) %>% 
  summarise(color = color[1],
            color2 = first(color))
```
I use R 3.3.3 and dplyr 0.5.0, but I'm told that it's the same with the dev version.

Original SO question: http://stackoverflow.com/q/43421152/6656269

lionel-:
Closes #2673

### DIFF
--- a/tests/testthat/test-hybrid.R
+++ b/tests/testthat/test-hybrid.R
@@ -275,6 +275,20 @@ test_that("first(), last(), and nth() work", {
   )
 })
 
+test_that("nth(), first(), etc can subset beyond the vector length (#2673)", {
+  df <- tibble(g = c(1, 1, 2), x = c("a", "b", "c"))
+  out <- group_by(df, g) %>%
+    summarise(
+      x = x[1],
+      x2 = first(x)
+    )
+  expect_identical(out, tibble(
+    g = c(1, 2),
+    x = c("a", "c"),
+    x2 = c("a", NA)
+  ))
+})
+
 test_that("lead() and lag() work", {
   check_hybrid_result(
     list(lead(a)), a = 1:5,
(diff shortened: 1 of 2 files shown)


tests/testthat/test-hybrid.R
```diff
@@ -275,6 +275,20 @@ test_that("first(), last(), and nth() work", {
...
+  out <- group_by(df, g) %>%
+    summarise(
+      x = x[1],
+      x2 = first(x)
+    )
+  expect_identical(out, tibble(
+    g = c(1, 2),
+    x = c("a", "c"),
+    x2 = c("a", NA)
+  ))
```
hadley: I don't get why the second value should be NA. What am I missing?

lionel-: oops I got confused at some point. We should update the group indices after each summarise operation. This will require a more complex fix.

lionel-: hrm, that's a tough design issue :/

The group indices are dataframe-wise, not column-wise. Maybe we should fix that as part of a rethinking of hybrid evaluation for dplyr 0.7? Or can you think of an easy fix @krlmlr?

krlmlr: The hybrid handlers already handle summary variables in many cases, but not in the case described here. Summary variables contain one item per group in the order of the groups, this is usually less than the original number of items. This triggers the segfault we were seeing here. A simple remedy is to fall back to regular evaluation, which now works thanks to the overscope. We could also fix the hybrid handlers to always support summary variables, but I think this unnecessarily complicates the proble (#2312); to the contrary, I think we should stop supporting them altogether.

lionel-: > to the contrary, I think we should stop supporting summary variables in the hybrid handlers, and always fall back to regular evaluation.

Thanks this seems like the right short term move.

PR #2728, opened 2017-05-02
Join warnings

hadley:
The main purpose of this PR is to switch from an error to a warning when the attributes don't match.  I also used this as an opportunity to pull down the join variable names so that we can start to make better warnings and errors.

### DIFF
--- a/src/join.cpp
+++ b/src/join.cpp
@@ -33,6 +33,31 @@ int count_attributes(SEXP x) {
   return n;
 }
 
+void warn_bad_var(const SymbolString& var_left, const SymbolString var_right,
+                  std::string message, bool warn = true) {
+  if (!warn)
+    return;
+
+  if (var_left == var_right) {
+    std::string var_utf8 = var_left.get_utf8_cstring();
+    Rf_warningcall(R_NilValue,
+      "Variable `%s` %s",
+      var_utf8.c_str(),
+      message.c_str()
+    );
+  } else {
+    std::string left_utf8 = var_left.get_utf8_cstring();
+    std::string right_utf8 = var_right.get_utf8_cstring();
+    Rf_warningcall(R_NilValue,
+      "Variable `%s`/`%s` %s",
+      left_utf8.c_str(),
+      right_utf8.c_str(),
+      message.c_str()
+    );
+  }
+
+}
+
 SEXP grab_attribute(SEXP name, SEXP x) {
   while (!Rf_isNull(x)) {
     if (TAG(x) == name) return CAR(x);
@@ -51,8 +76,10 @@ void check_attribute_compatibility(SEXP left, SEXP right, const SymbolString& na
     return;
   }
 
-  if (n_left != n_right)
-    stop("attributes of different sizes");
+  if (n_left != n_right) {
+    warn_bad_var(name_left, name_right, "has different attributes on RHS and LHS of join");
+    return;
+  }
 
   List list_left(n_left), list_right(n_left);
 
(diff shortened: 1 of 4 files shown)


src/join.cpp
```diff
@@ -51,8 +76,10 @@ void check_attribute_compatibility(SEXP left, SEXP right) {
     return;
   }
 
-  if (n_left != n_right)
-    stop("attributes of different sizes");
+  if (n_left != n_right) {
```
krlmlr: I think we should extract a function and just write `if (attributes_different(left, right)) warn_bad_var(...);` .

hadley: Yeah, and it's probably better to do that at the R level and use `all.equal()`.  Something like this:

```R
attributes_equal <- function(x, y) {
  attr_x <- attributes(x)
  attr_x <- attr_x[sort(names(attr_x))]
  
  attr_y <- attributes(y)
  attr_y <- attr_y[sort(names(attr_y))]
  
  isTRUE(all.equal(attr_x, attr_y)
}
```

src/join.cpp
```diff
@@ -33,6 +33,31 @@ int count_attributes(SEXP x) {
   return n;
 }
 
+void warn_bad_var(const SymbolString& var_left, const SymbolString var_right,
```
hadley: Ooops, thanks.

src/join.cpp
```diff
@@ -21,54 +21,42 @@ inline bool is_bare_vector(SEXP x) {
...
-      list_left[i]  = CAR(p_left);
-      list_right[i] = grab_attribute(name, att_right);
-    }
-    p_left = CDR(p_left);
-  }
-  RObject test = Language("all.equal", list_left, list_right).fast_eval();
-  if (!is<bool>(test) || !as<bool>(test)) {
-    stop("attributes are different");
+  // Otherwise rely on R function based on all.equal
+  static Function attr_equal = Function("attr_equal", Environment::namespace_env("dplyr"));
```
hadley: I think this is a more accurate approach than previously which would fail with an error if there was an attribute on the LHS but not the RHS. @krlmlr what do you think?

src/join.cpp
```diff
@@ -21,54 +21,42 @@ inline bool is_bare_vector(SEXP x) {
...
+    Rf_warningcall(R_NilValue,
+      "Variable `%s` %s",
+      var_utf8.c_str(),
+      message.c_str()
+    );
+  } else {
+    std::string left_utf8 = var_left.get_utf8_cstring();
+    std::string right_utf8 = var_right.get_utf8_cstring();
+    Rf_warningcall(R_NilValue,
+      "Variable `%s`/`%s` %s",
```
krlmlr: Didn't we use ``"Column `xxx`"`` in error messages?

src/join.cpp
krlmlr: Yes, the R code looks much cleaner, too. The one-time call into R should't be too much of an issue performance-wise.

tests/testthat/test-joins.r
```diff
@@ -830,3 +830,17 @@ test_that("join accepts tz attributes (#2643)", {
...
 })
+
+test_that("join takes LHS with warning if attributes inconsistent", {
+  df1 <- tibble(a = 1:2, b = 2:1)
+  df2 <- tibble(
+    a = structure(1:2, foo = "bar"),
+    c = 2:1
+  )
+
+  expect_warning(out1 <- left_join(df1, df2, by = "a"), "different attributes")
```
krlmlr: Do you want to check the precise wording here? Worked well for the error messages.

PR #2737, opened 2017-05-03
Fix typo in wrap_dbplyr_obj

ianmcook:
I believe the expression for `dplyr_call` should use `dplyr_sym` not `dbplyr_sym`.


hadley:
Oops, yes, thanks!

PR #2741, opened 2017-05-04
Fix join warnings

### LINKED ISSUE #2733: New Column class
From #2728 — should encapsulate a column + its name to reduce the size of argument lists.

### LINKED ISSUE #2739: Eliminate duplicate warning on joins
```R
df1 <- tibble(a = 1:2)
df2 <- tibble(
  b = structure(1:2, foo = "bar"),
)

out <- left_join(df1, df2, by = c("a" = "b"))
#> Warning messages:
#> 1: Column `b`/`a` has different attributes on LHS and RHS of join 
#> 2: Column `a`/`b` has different attributes on LHS and RHS of join 
```

@krlmlr could you take a look at this? I'm not familiar enough with the join code to figure out what's happening, but it looks like `DualVector` is constructed twice, once with left and right flipped.

krlmlr:
- New `Column` class
- Remove duplicate warnings
- Warn about inconsistencies in semi- and anti-joins
- Reverse order of column names in warnings for left join

Closes #2733. Fixes #2739.

### DIFF
--- a/src/join_exports.cpp
+++ b/src/join_exports.cpp
@@ -30,7 +30,7 @@ DataFrame subset_join(DataFrame x, DataFrame y,
   }
 
   // first the joined columns
-  DataFrameJoinVisitors join_visitors(x, y, SymbolVector(by_x), SymbolVector(by_y), false, false);
+  DataFrameJoinVisitors join_visitors(x, y, SymbolVector(by_x), SymbolVector(by_y), true, false);
   int n_join_visitors = join_visitors.size();
 
   // then columns from x but not y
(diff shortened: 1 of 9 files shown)


src/join_exports.cpp
```diff
@@ -30,7 +30,7 @@ DataFrame subset_join(DataFrame x, DataFrame y,
   }
 
   // first the joined columns
-  DataFrameJoinVisitors join_visitors(x, y, SymbolVector(by_x), SymbolVector(by_y), false, false);
+  DataFrameJoinVisitors join_visitors(x, y, SymbolVector(by_x), SymbolVector(by_y), true, false);
```
krlmlr: @hadley. We're now warning only from here, used to warn from the individual join functions below (with flipped col names for left join), and from both constructors for the new warnings you added.

PR #2746, opened 2017-05-06
Wrap ptrdiff_t instead of R_xlen_t

### LINKED ISSUE #2736: Fix wrapping of R_xlen_t
The current implementation causes problems on Win32 (https://github.com/krlmlr/dplyr/commit/3c5c203bd8b31c79261073cebecf5fdab0458dcb#commitcomment-21990414):

```r
Rcpp::sourceCpp(code = '
#include <Rcpp.h>

namespace Rcpp {

template <>
inline SEXP wrap(const R_xlen_t& x) {
  if (x < -2147483647 || x > 2147483647) {
    return Rf_ScalarReal(static_cast<double>(x));
  }
  else {
    return Rf_ScalarInteger(static_cast<int>(x));
  }
}

}

// [[Rcpp::export]]
SEXP test() {
  Rcpp::Function passthrough("passthrough");
  return passthrough(NA_INTEGER);
}

/*** R

passthrough <- function(...) list(...)
test()

*/

'
)
```

krlmlr:
because `R_xlen_t` is typedef'd to `int`, and our wrapper then takes precedence for *all* `int` values.

Fixes #2736.

### DIFF
--- a/inst/include/dplyr/workarounds/xlen.h
+++ b/inst/include/dplyr/workarounds/xlen.h
@@ -5,7 +5,7 @@ namespace Rcpp {
 
 template <>
 inline SEXP wrap(const ptrdiff_t& x) {
-  if (x < -R_SHORT_LEN_MAX || x > R_SHORT_LEN_MAX) {
+  if (x < -R_LEN_T_MAX || x > R_LEN_T_MAX) {
     return Rf_ScalarReal(static_cast<double>(x));
   }
   else {
--- a/inst/include/dplyr/workarounds/xlen.h
+++ b/inst/include/dplyr/workarounds/xlen.h
@@ -1,11 +1,13 @@
 #ifndef DPLYR_WORKAROUND_XLEN_H
 #define DPLYR_WORKAROUND_XLEN_H
 
+#ifdef LONG_VECTOR_SUPPORT
+
 namespace Rcpp {
 
 template <>
-inline SEXP wrap(const ptrdiff_t& x) {
-  if (x < -R_LEN_T_MAX || x > R_LEN_T_MAX) {
+inline SEXP wrap(const R_xlen_t& x) {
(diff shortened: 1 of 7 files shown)


inst/include/dplyr/workarounds/xlen.h
```diff
@@ -4,7 +4,7 @@
 namespace Rcpp {
 
 template <>
-inline SEXP wrap(const R_xlen_t& x) {
+inline SEXP wrap(const ptrdiff_t& x) {
   if (x < -2147483647 || x > 2147483647) {
```
lionel-: Should we use `R_SHORT_LEN_MAX` instead of hard-coded numbers?

Should the negative side be `-(R_SHORT_LEN_MAX + 1)` so that we wrap missing values as well on platforms where `INT_MIN` (the missing value) is exactly `2^31`?

lionel-: perhaps worth it to add a couple of unit tests to check the type of wrapped values and if NAs are still wrapped correctly.

krlmlr: NAs aren't supposed to be wrapped; this is for wrapping a length of a vector for creating error messages.

krlmlr:
Using named constant, added unit tests.

src/test.cpp
```diff
@@ -104,3 +104,15 @@ List test_matches() {
       )
     );
 }
+
+// [[Rcpp::export]]
+LogicalVector test_length_wrap() {
+  R_xlen_t small = R_SHORT_LEN_MAX / 2;
+  R_xlen_t large = (R_xlen_t)(R_SHORT_LEN_MAX * 2.0);
```
lionel-: I'd just add `R_xlen_t missing = INT_MIN`, otherwise looks good!

PR #2771, opened 2017-05-12
Support unquoting in mutate()

lionel-:


src/mutate.cpp
```diff
@@ -96,7 +96,8 @@ SEXP mutate_not_grouped(DataFrame df, const QuosureList& dots) {
       accumulator.rm(name);
       continue;
     } else {
-      stop("cannot handle");
+      // Unquoted values
+      variable = call;
```
krlmlr: Do we need to change anything for grouped `mutate()`?

krlmlr: Is there a way we can further discriminate between the intended case and possible misuse?

tests/testthat/test-mutate.r
```diff
@@ -706,6 +706,11 @@ test_that("mutate() names pronouns correctly (#2686)", {
   expect_named(mutate(tibble(x = 1), .data[["x"]]), "x")
 })
 
+test_that("mutate() supports unquoted values", {
+  expect_identical(mutate(tibble(x = 1), x = !! 2 * 3), tibble(x = 6))
```
krlmlr: Can you please test `y <- quote(x)`; mutate(..., x = y + 2)` or something related?

tests/testthat/test-summarise.r
```diff
@@ -952,3 +952,7 @@ test_that("can refer to previously summarised symbols", {
   expect_identical(summarise(group_by(mtcars, cyl), x = 1, z = x)[2:3], tibble(x = c(1, 1, 1), z = x))
   expect_identical(summarise(group_by(mtcars, cyl), x = n(), z = x)[2:3], tibble(x = c(11L, 7L, 14L), z = x))
 })
+
+test_that("summarise() supports unquoted values", {
+  expect_identical(summarise(mtcars, x = !! 1 + 2), data.frame(x = 3))
```
krlmlr: Same here.

tests/testthat/test-mutate.r
lionel-: hmm but that's not linked to unquoting. Do you mean `x = !! y + 2`? this would trigger a R error outside of mutate, when the quosures are captured.

src/mutate.cpp
lionel-: oh good point, I always forget to check for grouped cases.

About misuses, I added a test where we unquote an environment. Type checking seems to be performed at a later stage.

tests/testthat/test-mutate.r
krlmlr: What's the syntax of treating `y` as a variable name in the input again?

lionel-: that would be `!! quote(y + 2)` (or possibly with `quo()` if referring to contextual variables). But this creates quosures that are identical to the normal case. This is already plenty covered by the tests but I can add another just to be sure ;)

krlmlr: IIUC this is handled in another layer; in that case it's fine to test the simple cases here.

lionel-:
@krlmlr @hadley 

It's not clear what semantics we should have for unquoted values in grouped mutate and summarise.

I think it would probably make sense to require a vector as long as the final dataset, e.g. nrows for mutate and ngroups for summarise.

lionel-:
The rule is now that unquoted values are directly assigned to columns (whereas expressions are evaluated once per group). This makes sense because otherwise we could only unquote values when the groups have the same size.

@krlmlr I have improved type checking in `summarise()` in the last commits. Before this, the following would return an error about output size (as `Rf_length()` is defined over all containers, including envs):

```r
summarise(df, out = env(a = 1))
```

src/mutate.cpp
```diff
@@ -52,6 +52,14 @@ void check_not_groups(const QuosureList& quosures, const GroupedDataFrame& gdf)
   }
 }
 
+static
+SEXP validate_unquoted_value(SEXP value, int nrows, SymbolString& name) {
+  if (is_vector(value))
+    check_length(Rf_length(value), nrows, "the tibble size", name);
```
krlmlr: I remember we were using `"the number of rows"` elsewhere. Has this changed?

src/summarise.cpp
```diff
@@ -11,12 +11,27 @@
...
 #include <dplyr/NamedListAccumulator.h>
 #include <dplyr/Groups.h>
 
 using namespace Rcpp;
 using namespace dplyr;
 
+static
+SEXP validate_unquoted_value(SEXP value, int nrows, const SymbolString& name) {
+  int n = Rf_length(value);
+  check_length(n, nrows, "the summarised tibble size", name);
```
krlmlr: Same.

lionel-: The issue is that "the number of rows" seems too vague, and "the number of rows of the summarised tibble" seems too long.

is "the number of summary rows" accurate?

krlmlr: I remember using `"the number of groups"` in a similar context.

lionel-: oh of course

R/manip.r
```diff
@@ -162,6 +162,20 @@ slice_ <- function(.data, ..., .dots = list()) {
...
+#'
+#'
+#' # summarise() supports quasiquotation. You can unquote expressions:
+#' var <- quo(mean(cyl))
+#' summarise(mtcars, !! var)
+#'
+#' # Or values:
+#' summarise(mtcars, !! 10)
+#'
+#' # For grouped tibbles, unquoted values are handled a bit
```
hadley: Let's remove the value unquoting examples.

PR #2778, opened 2017-05-15
Remove `vars` attribute when empty

### LINKED ISSUE #2772: Dev Version of Filter adds a vars attribute
The new version of `filter` in dplyr (0.5.0.9004) adds a `vars` attribute that is new (does not show in dplyr 0.5.0).  This is causing some tests to fail when using `expect_identical()` because attributes have changed.

Expected behavior is to not have the `vars` attribute

In my tests, the attribute seems always to be `character(0)`, which suggests to me that it is unintentional.  I have struck out finding where it is coming from, though.

```r
## Using CRAN Version - expected and desired behavior
suppressMessages(library(dplyr))                          
packageVersion('dplyr')                                   
#> [1] '0.5.0'
data_frame(id=c(1,2,3)) %>% filter(id==1) %>% attributes()
#> $class
#> 
#> $row.names
#> [1] 1
#> 
#> $names
#> [1] "id"

## Using devtools::install_github('tidyverse/dplyr')
suppressMessages(library(dplyr))
packageVersion("dplyr")
#> [1] '0.5.0.9004'
data_frame(id = c(1, 2, 3)) %>% filter(id == 1) %>% attributes()
#> $class
#> 
#> $row.names
#> [1] 1
#> 
#> $names
#> [1] "id"
#> 
#> $vars
#> character(0)
```

lionel-:
Fixes #2772

### DIFF
--- a/src/utils.cpp
+++ b/src/utils.cpp
@@ -147,7 +147,12 @@ SymbolVector get_vars(SEXP x) {
 
 void set_vars(SEXP x, const SymbolVector& vars) {
   static SEXP vars_symbol = Rf_install("vars");
-  Rf_setAttrib(x, vars_symbol, vars.get_vector());
+
+  SEXP vars_sexp = vars.get_vector();
+  if (!Rf_length(vars_sexp))
+    vars_sexp = R_NilValue;
+
+  Rf_setAttrib(x, vars_symbol, vars_sexp);
 }
 
 void copy_vars(SEXP target, SEXP source) {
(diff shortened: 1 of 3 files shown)


src/utils.cpp
```diff
@@ -147,7 +147,12 @@ SymbolVector get_vars(SEXP x) {
 
 void set_vars(SEXP x, const SymbolVector& vars) {
   static SEXP vars_symbol = Rf_install("vars");
-  Rf_setAttrib(x, vars_symbol, vars.get_vector());
+
+  SEXP vars_sexp = vars.get_vector();
```
krlmlr: I'd prefer a new method `get_vector_or_nil()`, this pattern might occur for other verbs, too.

lionel-: A free function `nil_if_empty()` defined on SEXPs would probably be better then.

krlmlr: Yes, a free function could be used elsewhere, too.

PR #2786, opened 2017-05-16
Compatibility fix for deserialized grouped dfs

krlmlr:
for eechidna and perhaps others.

### DIFF
--- a/R/grouped-df.r
+++ b/R/grouped-df.r
@@ -55,12 +55,15 @@ n_groups.grouped_df <- function(x) {
 
 #' @export
 groups.grouped_df <- function(x) {
-  lapply(group_vars(x), as.name)
+  syms(group_vars(x))
 }
 
 #' @export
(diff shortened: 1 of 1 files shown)


R/grouped-df.r
```diff
@@ -55,12 +55,15 @@ n_groups.grouped_df <- function(x) {
...
 groups.grouped_df <- function(x) {
-  lapply(group_vars(x), as.name)
+  syms(group_vars(x))
 }
 
 #' @export
 group_vars.grouped_df <- function(x) {
-  attr(x, "vars")
+  vars <- attr(x, "vars")
+  # Need this for compatibility with existing packages that might
```
lionel-: that might?

Should this issue a warning so that packages converge to the right data structure?

krlmlr: Maybe later, we could add an issue to that effect.

PR #2787, opened 2017-05-16
Handle .data pronoun in rename()

lionel-:


tests/testthat/test-select.r
```diff
@@ -108,6 +108,10 @@ test_that("abort on unknown columns", {
   expect_error(select_vars(letters, c("a", "bar", "foo", "d")), "bar, foo")
 })
 
+test_that("rename() handles data pronoun", {
+  expect_identical(rename(tibble(x = 1), y = .data$x), tibble(y = 1))
```
krlmlr: Just curious: Why is this necessary?

lionel-: That's because it is our official way of avoiding a R CMD check note about unknown global variables.

krlmlr: I see. Does `y = sym("x")` also work?

lionel-: not currently, you'd have to unquote it. We'll have to wait until 0.7 for a tidyeval implementation of rename() :/

The difficulty is that we need to ignore evaluation errors about unbinded symbols when `strict = FALSE`.

hadley: Consistency with the other verbs + avoid R CMD check note

PR #2788, opened 2017-05-16
Handle vars list on the C++ side

lionel-:
Complements https://github.com/tidyverse/dplyr/pull/2786

This fixes the "Cannot handle STRSXP" revdep failure in highcharter.

### DIFF
--- a/src/utils.cpp
+++ b/src/utils.cpp
@@ -148,9 +148,47 @@ bool same_levels(SEXP left, SEXP right) {
   return character_vector_equal(get_levels(left), get_levels(right));
 }
 
+SEXP list_as_chr(SEXP x) {
+  int n = Rf_length(x);
+  CharacterVector chr(n);
+
+  for (int i = 0; i != n; ++i) {
+    SEXP elt = VECTOR_ELT(x, i);
+    switch (TYPEOF(elt)) {
+    case STRSXP:
+      if (Rf_length(chr) == 1) {
+        chr[i] = elt;
+        break;
+      } else {
+        stop("The tibble's `vars` attribute has unexpected contents");
+      }
+    case SYMSXP:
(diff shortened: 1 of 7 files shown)


src/utils.cpp
```diff
@@ -148,9 +148,47 @@ bool same_levels(SEXP left, SEXP right) {
...
+        chr[i] = elt;
+        break;
+      } else {
+        stop("The tibble's `vars` attribute has unexpected contents");
+      }
+    case SYMSXP:
+      chr[i] = PRINTNAME(elt);
+      break;
+    default:
+      stop("The tibble's `vars` attribute has unexpected contents");
```
krlmlr: We could use `continue` to get rid of the duplicated `stop()` call.

src/utils.cpp
```diff
@@ -148,9 +148,47 @@ bool same_levels(SEXP left, SEXP right) {
...
+    }
+  }
+
+  return chr;
+}
+
 SymbolVector get_vars(SEXP x) {
   static SEXP vars_symbol = Rf_install("vars");
-  return SymbolVector(Rf_getAttrib(x, vars_symbol));
+  SEXP vars = Rf_getAttrib(x, vars_symbol);
```
krlmlr: Do we need `RObject` here to protect against GC?

lionel-: not currently, but using `RObject` might make the code more robust to refactoring.

krlmlr: Oh, I see -- even if `list_as_chr()` allocates, we don't care, because we overwrite.

lionel-: I was thinking that we don't care because the next thing we do is to return, but we also pass the object to `SymbolVector()` so we are vulnerable to refactoring in both `get_vars()` and the `SymbolVector()` constructor.

PR #2789, opened 2017-05-16
tbl_df always inherits from tbl

krlmlr:
Fixes tidyverse/tibble#256.

Ideally, `classes_grouped()` and `classes_not_grouped()` should be exported functions and called throughout, maybe later.


krlmlr:
@hadley: We might want to add a safety net to tibble, but the CRAN submission is underway. If we don't, this affects users who create classes that inherit from `tbl_df` but not `tbl`. Please advise.

PR #2791, opened 2017-05-16
Deprecation messages

krlmlr:
instead of warnings, for cleaner revdep checks


lionel-:
Not very important but the rlang verb for messages is `inform()`.

krlmlr:
Thanks Lionel, I was looking for `msg()` but forgot about `inform()`.

PR #2794, opened 2017-05-17
Fix slice() error with gctorture()

krlmlr:
I suspect one of the destructors of the local objects we're creating performs an allocation. @lionel-: If you like, you can take over, I'll be away for the next two hours or so.


krlmlr:
Thanks for your input BTW, I was able to rule out the obvious right away ;-)

lionel-:
is there any countraindication to using gctorture within our tests by the way? We should probably not enable it for all unit tests but it'd be nice to have a few running with it.

hadley:
Just the extreme slowness

krlmlr:
We could set up a daily Travis run with a moderate `gctorture()` setting. If we run only `devtools::test()`, there should be frequent enough progress updates.

I'll finish this PR.

lionel-:
Is it really useful to use `RObject` as return type? Either the result is assigned to another RObject, or coerced to another preserving type, or coerced to a SEXP. In all cases the wrapping of the return value seems redundant.

lionel-:
> I'll finish this PR.

Thanks, I haven't had the time to look into it again today. Let me know if I should take over tomorrow.

krlmlr:
It prevents an unprotected return value, which can be overwritten by (I think) `~GroupedHybridEnv()` which calls `rlang::overscope_clean()`.

lionel-:
For the record, here is another perplexing bug that was linked to destruction semantics: https://github.com/tidyverse/dplyr/issues/1396

PR #2802, opened 2017-05-18
Revdep checks

krlmlr:
@hadley: For a few packages we could do better, and we might want to look at the segfaults again. I'l
l be away until much later tonight. ac7c18b7 shows the detailed comparison.

CC @lionel-.


cran-comments.md
```diff
@@ -14,5 +14,63 @@
 
 New problems that appeared after the emails sent on April 18:
 
-* Two packages (geoSpectral, valr) now fail with "Variable context not
+* Three packages (carpenter, geoSpectral, valr) now fail with "Variable context not
```
lionel-: carpenter was from before

hadley:
Thanks - submitting to CRAN now!

PR #2805, opened 2017-05-19
Update API description

krlmlr:
also create it automatically when documenting (but not with Cmd+Shift+D from RStudio).


hadley:
If you're going to add to the default document, I think you should also add `pkgapi` to suggests. But maybe it's not on CRAN yet?

PR #2816, opened 2017-05-23
Update Rcpp

krlmlr:
to avoid TLS failure for CRAN submission. Also update RcppExports as generated with the most recent version of Rcpp.

CC @hadley @lionel-.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,7 +15,7 @@ include:
 - r: devel
 
 before_script:
-- R -q -e 'devtools::install("hadley/dbplyr")'
+- R -q -e 'devtools::install("tidyverse/dbplyr")'
 
 after_success:
 - Rscript -e 'covr::codecov()'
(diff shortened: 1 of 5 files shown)


hadley:
CRAN is fixing the pandoc issue that lead to the badge issue, so I don't think there's a need to change.

krlmlr:
Wouldn't codecov's own badge be preferable to that of shields.io?

hadley:
When I tried it, I got the same TLS problem. We should fix upstream in usethis and then roll out more systematically. Here I think it just adds confusion to this commit (which is mostly about updating Rcpp)

krlmlr:
@hadley: Merged to fix Travis tests, happy to fix as necessary.

krlmlr:
Manual installation of dbplyr doesn't seem to be necessary anymore, reverted in 299aff051.

PR #2836, opened 2017-06-03
corrected very small spelling mistake

tonyfischetti:
Haha this is not to be pedantic, but because my test that tests have the error string in them and I don't want them to fail later when this gets corrected


hadley:
Thanks!

Issue #2811, opened 2017-05-22
Reproducible segfault


krlmlr:
The following code crashed twice in my test system on Ubuntu 16.04~~10~~:

```
R --debugger=valgrind -q -e 'library(vdmR); set.seed(123); gctorture2(1); example(vcmap)'
```

@lionel-: Can you replicate on OS X?

Output:

```
==24742==
==24742== Process terminating with default action of signal 11 (SIGSEGV)
==24742==    at 0x5472269: raise (pt-raise.c:35)
==24742==    by 0x547238F: ??? (in /lib/x86_64-linux-gnu/libpthread-2.23.so)
==24742==    by 0x1365CA9B: SymbolString (SymbolString.h:14)
==24742==    by 0x1365CA9B: dplyr::GroupedHybridEnv::hybrid_get_callback(Rcpp::String const&, bindrcpp::PAYLOAD) (GroupedHybridCall.h:88)
```

More details follow.

krlmlr:
Will work on speeding up the example.

The following tests did **not** exhibit faulty behavior, among others:

```sh
R --debugger=valgrind -q -e 'library(vdmR); library(dplyr); data(vsfuk2012); set.seed(123); gctorture2(1); res <- filter(vsfuk2012, CityCode<40110&CityCode>40100)'
R --debugger=valgrind -q -e 'library(vdmR); library(dplyr); set.seed(123); gctorture2(1); filter(tibble(a = 1:10), a > 3 & a < 7)'
```

The two failing runs of the example above had identical output (apart from process IDs):

<details>

```
$ R --debugger=valgrind -q -e library(vdmR); library(dplyr); data(vsfuk
2012); set.seed(123); gctorture2(1); res <- filter(vsfuk2012, CityCode<40110&CityCode>40100)
        Command being timed: "/bin/sh -c  'R' '--debugger=valgrind' '-q' '-e' 'library(vdmR); library
(dplyr); data(vsfuk2012); set.seed(123); gctorture2(1); res <- filter(vsfuk2012, CityCode<40110&CityC
ode>40100)' 2>&1"
        User time (seconds): 1808.47
muelleki@ubuntu:~$ mail-after R --debugger=valgrind -q -e 'library(vdmR); set.seed(123); gctorture2(1
); example(vcmap)'
==24742== Memcheck, a memory error detector
==24742== Copyright (C) 2002-2015, and GNU GPL'd, by Julian Seward et al.
[...]

PR #2844, opened 2017-06-08
Test: Reproducible segfaults

krlmlr:
To run: `R -q -e 'devtools::test(f = "hybrid-traverse")'` .

Related to #2811.

@lionel-: Would you mind reviewing bindr and bindrcpp, and the pull requests there? It's likely that these packages are responsible for the erratic failures we're seeing here.


krlmlr:
Need to try ASAN/UBSAN.

krlmlr:
FWIW, I can't seem to replicate this in R 3.3.3 built from source (but R 3.4.0 built from source crashes). Now trying the entire test suite with R 3.3.3 and R 3.4.0 with `R_ENABLE_JIT=0`.

krlmlr:
Update: Can reproduce with R 3.3.3. Will look into ASAN.

PR #2854, opened 2017-06-12
Fix a few out of date things in programming.rmd

gergness:
- Remove obsolete comment about error message

- `rlang::eval_tidy` doesn't work with base R's formulas (I think this is intentional). Before changing from `~` to `quo`, the output looked like: 
>When we evaluate those formulas using rlang::eval_tidy(), we see that they yield different values:
```
library(rlang)

eval_tidy(x1)
#> ~x
#> <environment: 0x0000000007a12958>
eval_tidy(x2)
#> ~x
#> <environment: 0x000000000c5e48e8>
```

### DIFF
--- a/vignettes/programming.Rmd
+++ b/vignettes/programming.Rmd
@@ -418,18 +418,24 @@ x1 <- f(10)
 x2 <- f(100)
 ```
 
-It might look like the expressions are the same if you print them out. But look carefully at the environments --- they're different.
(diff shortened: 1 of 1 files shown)


gergness:
Doh, `rlang:::print.quosure()` doesn't include environment, so I had to make a few more changes.

hadley:
Nice, thanks!

PR #2877, opened 2017-06-15
Fix protection error when constructing error messages

krlmlr:
I have no idea why I used `wrap()` originally, but this results in a naked `SEXP` which may be destroyed even before the function is called. Tests with gctorture are still running.


krlmlr:
I don't see any of the errors I saw before when running alll tests with gctorture.

PR #2886, opened 2017-06-19
Fix UTF-8 column name test for Windows

krlmlr:
and fix `rename()` with a hack.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -23,7 +23,7 @@ Imports:
     magrittr,
     methods,
     pkgconfig,
-    rlang (>= 0.1),
+    rlang (>= 0.1.1.9000),
     R6,
     Rcpp (>= 0.12.6),
     tibble (>= 1.3.1),
@@ -47,6 +47,8 @@ Suggests:
     RSQLite,
     testthat,
     withr
+Remotes:
+    tidyverse/rlang@ff87439
 VignetteBuilder: knitr
 LinkingTo: Rcpp (>= 0.12.0),
     BH (>= 1.58.0-1),
--- a/R/select-vars.R
+++ b/R/select-vars.R
@@ -190,6 +190,9 @@ rename_vars <- function(vars, ..., strict = TRUE) {
     abort("All arguments must be named")
   }
 
+  # Triggers unescaping of UTF-8 names:
+  exprs <- as_list(as_env(exprs))
+
   old_vars <- map2(exprs, names(exprs), switch_rename)
   new_vars <- names(exprs)
 
--- a/R/select-vars.R
+++ b/R/select-vars.R
@@ -190,9 +190,6 @@ rename_vars <- function(vars, ..., strict = TRUE) {
     abort("All arguments must be named")
   }
 
-  # Triggers unescaping of UTF-8 names:
-  exprs <- as_list(as_env(exprs))
-
   old_vars <- map2(exprs, names(exprs), switch_rename)
   new_vars <- names(exprs)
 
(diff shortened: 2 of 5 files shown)


lionel-:
> Perhaps exprs() and quo() should unescape names?

I don't think so. Why doesn't `rename_vars()` work with UTF-8 names? Is it an issue in `setdiff()`?

krlmlr:
No, foreign characters are mangled already by `exprs()`: https://github.com/tidyverse/dplyr/pull/2886/commits/1db519ce9bbcb6c1042bc1479c63ba07dc5e0d96#diff-736b31bba1292201e8910669af931883R188

lionel-:
I see this in a latin1 R session, which looks right:

```r
library("rlang")
x <- exprs("\u5e78" := 10)

names(x)
#> [1] "å¹¸"

Encoding(names(x))
#> [1] "UTF-8"
```

lionel-:
oops sorry, the shell was in latin1 but not the R session. This is actually:

```r
mut_latin1_locale()
x <- exprs("\u5e78" := 10)

Encoding(names(x))
#> [1] "UTF-8"

names(x)
#> "<U+5E78>"
```

krlmlr:
This is on Windows:

``` r
library(rlang)
x <- exprs("\u5e78" := 10)
charToRaw(names(x))
#> [1] e5 b9 b8
Encoding(names(x))
#> [1] "UTF-8"
```

This looks right, but the last commit seems to resolve the problems with `rename()`. Will investigate.

krlmlr:
This is after reverting 1db519c:

``` r
library(dplyr)
#> 
#> Attaching package: 'dplyr'
#> 
#>     filter, lag
#> 
data_frame(a = 1) %>% rename("\u5e78" := a) %>% names %>% charToRaw
#> [1] 3c 55 2b 35 45 37 38 3e
```

lionel-:
yes, just saw that the bytes are right and it's the print method that escapes unicode. I wish they gave more clues about what's going on in that method.

krlmlr:
Still, my last commit may be curing the wrong problem. I'll be looking into it.

krlmlr:
Could it be that uqs in dataframe.R:123 mangles foreign characters?

krlmlr:
It appears so:

``` r
library(rlang)
f <- function(...) {
  exprs <- exprs(...)
  charToRaw(names2(exprs))
}

g <- function(...) {
  f(!!! quos(...) )
}
g("\u5e78" := 10)
#> [1] 3c 55 2b 35 45 37 38 3e
```

lionel-:
good catch, looks like it:

```r
exprs <- exprs("\u5e78" := 10)
x <- quos(!!! exprs)

chr_encoding(names(x))
#> [1] "unknown"

as_bytes(names(x))
#> [1] 3c 55 2b 35 45 37 38 3e
```

krlmlr:
Back to my original question: Maybe demangle right in `exprs()` and `quos()`?

lionel-:
I think this needs to be done here: https://github.com/tidyverse/rlang/blob/d92dbde3886f65c7e197776174f095ba621739d8/R/dots.R#L191

krlmlr:
I don't know rlang well enough to comment on this, but we should be able to write tests that indicate if this works as expected. Perhaps we should also turn on testing on AppVeyor?

lionel-:
I think it's just a matter of running `mut_latin1_locale()` before running the test. It should work even if R is launched in a shell run with a UTF-8 locale. I'll look into it.

krlmlr:
I remember adding `with_` functions to dplyr test helpers for this purpose.

It's great to emulate an OS without UTF-8 support, but testing on that OS may reveal other problems.

Issue #2865, opened 2017-06-14
programming vignette: object named "group_by"


davharris:
In the dplyr programming vignette, the following code is suggested:

```
my_summarise <- function(df, group_by) {
  group_by <- enquo(group_by)
  print(group_by)

  df %>%
    group_by(!!group_by) %>%
    summarise(a = mean(a))
}
```

If I understand correctly, this code uses `group_by` to refer to a variable to use for grouping (the argument) and also to refer to the `group_by` verb. 

If so, wouldn't it be better to name the argument `group_var`, as in the preceding examples?  The current naming approach seems more confusing than it needs to be.

Sorry if I'm misunderstanding something.

hadley:
Yes, that would be great. Want to do a PR?

markdly:
I've added a pull request for this in #2901 and also updated the other examples where this occurred.  I'm rather new to this so apologies if this is poor etiquette / incorrect procedure...

PR #2901, opened 2017-06-23
programming vignette: rename object "group_by" to "group_var"

markdly:
This is an attempt to address issue #2865


hadley:
Thanks!  (In the future, can you please put `Fixes #2865` in a commit message so that the related issue is automatically closed?)

PR #2910, opened 2017-06-25
case_when fizzbuzz example: fix formula ordering

davechilders:
`TRUE ~ as.character(x)` should be the final formula to generate fizzbuzz


krlmlr:
Thanks, I've already merged #2917 before seeing this.

PR #2917, opened 2017-06-26
Show case_when()'s splicing support with arguments in order.

maurolepore:
Likely a typo resulted from copy-pasting code immediately above. The intent seems to be to show an example that works.


krlmlr:
Thanks!

PR #2956, opened 2017-07-10
Port to tidyselect

### LINKED ISSUE #2896: Use tidyselect

lionel-:
Closes #2896


NEWS.md
```diff
@@ -9,6 +9,12 @@
 
 * Move build-time vs. run-time checks out of `.onLoad()` and into `dr_dplyr()`.
 
+* dplyr now depends on the new tidyselect package to power `select()`,
```
hadley: This is in the wrong place now

lionel-:
@hadley Tests pass locally. Should we merge?

hadley:
I prefer to wait for a bigger release. We're just doing a patch right now.

lionel-:
I'll disable the dplyr check in tidyselect then. The sentinel value thing didn't work because the dplyr namespace is not populated when the tidyselect onLoad hook is run, so we'll have to check for a hardcoded version.

PR #2958, opened 2017-07-11
Fix symbol scoping in selecting verbs

### LINKED ISSUE #2904: `select(df, colname)` sometimes impersonates `select_(df, .dots=colname)`
`select(df, colname)` *should* issue an error message when `colname` is not a column in `df`. However, if there exists a (global) `colname` variable which is a character vector, the columns in `df` corresponding to the elements in `colname` are instead returned. Basically, `select(df, colname)` works like `select_(df, .dots=colname)` *iff* there is no column named `colname` in `df`.

``` r
# Load package and example data
library(dplyr)
df = as_data_frame(iris)

# Create two identical variables string variables
Species = c("Sepal.Width", "Petal.Length")
myvar = Species

df %>% select(Species) # OK, returns Species column
df %>% select(myvar)   # Not OK, returns Sepal.Width & Petal.Length columns
df %>% select(myvar2)  # OK, returns error message
```

The last command gives the error message `#> Error in overscope_eval_next(overscope, expr): object 'myvar2' not found`. The next to last line *should* have also given a similar error message, since there is no `myvar` colmn in `df`. But instead, it returns a tibble with the columns `Sepal.Width` and `Petal.Length`. This is unexpected and dangerous behaviour.

I observe this bug with dplyr 0.7.1 and the latest GitHub version (as of 2017-06-24).

lionel-:
Includes #2956 and depends on tidyverse/tidyselect#13

Closes #2904


hadley:
I'm going to close this - let's review one update that does everything.

Issue #2981, opened 2017-07-19
Feature request: strict parameter for select_vars


twolodzko:
`rename_vars` has `strict` parameter that enables user to turn off throwing errors when calling nonexistent variables. Similar parameter should be introduced for `select_vars` function. It is a reasonable to assume that if user asks for a column that does not exist in the dataset, he should expect the function to return nothing. Another reasonable behavior is to throw error, but introducing the `strict` parameter would make the function more flexible. 

Current behavior:

```
select_vars(colnames(mtcars), mpg, cyl, ff)
##  Error in overscope_eval_next(overscope, expr) : object 'ff' not found 
```

Expected behavior:

```
select_vars(colnames(mtcars), mpg, cyl, ff, strict = FALSE)
##   mpg   cyl 
## "mpg" "cyl" 
```

krlmlr:
This seems to belong in tidyselect.

PR #2982, opened 2017-07-19
strict parameter for select_vars (related to #2981)

twolodzko:
This commit introduces `strict` parameter in `select_vars` function (as discussed in #2981). Instead of using `eval_tidy`, `eval_safely` function is introduced that uses `safely(eval_tidy)` instead of it when `strict = FALSE`. For everything to work, next the `NULL` variables (i.e. nonexistent) are dropped from `ind_list` and `names2(quos)`.

### DIFF
--- a/R/select-vars.R
+++ b/R/select-vars.R
@@ -83,7 +83,7 @@
 #'
 #' # This won't match on `include`:
 #' wrapper(include = cyl)
-select_vars <- function(vars, ..., include = character(), exclude = character()) {
+select_vars <- function(vars, ..., include = character(), exclude = character(), strict = TRUE) {
   quos <- quos(...)
 
   if (is_empty(quos)) {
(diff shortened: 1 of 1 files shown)


lionel-:
These functions are now living in the tidyselect package.

I think we'll want to catch only the "symbol not found" errors and to do this cleanly will require some changes in R core. Catching and parsing the error message could be a stopgap for compatibility with old releases.

twolodzko:
@lionel- OK, but the `strict = FALSE` default would not produce any compatibility issues.

lionel-:
Thanks for the PR @twolodzko. We've reimplemented it in tidyselect with a different approach.

lionel-:
Here is the relevant commit if you want to check it out: https://github.com/tidyverse/tidyselect/commit/14356f23da0a079af38bcdff4c7824edfab6459b

PR #2999, opened 2017-07-27
Rcpp 0.12.12 compatibility

krlmlr:
Looks like we don't need to register `is_bind_spliceable()` ~~anymore~~for now.


lionel-:
I wouldn't say that we don't need it anymore, but we can remove it until we find the right solution.

PR #3000, opened 2017-07-28
Adapt to dev version of glue

krlmlr:
@jimhester: dplyr tests fail with tidyverse/glue@28aa7400, these are the necessary changes. I have seen changes related to NA, but is stripping of space really intended?

### DIFF
--- a/R/select-var.R
+++ b/R/select-var.R
@@ -45,6 +45,7 @@ select_var <- function(vars, var = -1) {
     pos <- match_var(var, vars)
   } else if (is_integerish(var, 1)) {
     if (is_na(var) || abs(var) > n || var == 0L) {
+      if (is_na(var)) var <- "NA"
       abort(glue(
         "`var` must be a value between {-n} and {n} (excluding zero), not {var}"
       ))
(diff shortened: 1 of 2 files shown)


jimhester:
Possibly the space stripping is incorrect, it is tricky to get the trimming to work in all expected cases.

krlmlr:
Thanks. I'll leave this open for now.

krlmlr:
@jimhester: Space are still trimmed in `glue()`, I guess this feature is to stay. Can you please confirm?

``` r
as.character(glue::glue(" "))
#> [1] ""
```

jimhester:
As of https://github.com/tidyverse/glue/commit/ed4b7c0fd142ab72a1afae95b240e7431e50006e single lines are no longer trimmed.

```r
as.character(glue::glue(" "))
#> [1] " "
```

krlmlr:
What about `NA`?

jimhester:
NA was working as intended, any string with an NA should be NA.

One way to fix the example rather than special casing NA is to use

```r
abort(glue(
  "`var` must be a value between {-n} and {n} (excluding zero), not {single_quote(var)}"
))
```

``` r
glue::single_quote(c("foo", NA, "bar"))
#> [1] "'foo'" "NA"    "'bar'"
```

krlmlr:
Don we still need to change the code in dplyr, either way?

hadley:
We're going to roll back this change in glue

jimhester:
It is rolled and on CRAN.

PR #3047, opened 2017-08-23
min() and max() always return integers in hybrid evaluation

### LINKED ISSUE #3036: summarise_all(max) does not preserve integer type
The results of `summarise_all(max)` and `summarise_all(min)` are converted to double, whereas `max()` and `summarise_all(sum)` preserve integer type as expected.

``` r
library(dplyr)
tbl = tibble(x=seq_len(3L))
tbl %>% summarise_all(max)
#> # A tibble: 1 x 1
#>       x
#>   <dbl>
#> 1     3
tbl$x %>% max() %>% class()
#> [1] "integer"
tbl %>% summarise_all(sum)
#> # A tibble: 1 x 1
#>       x
#>   <int>
#> 1     6
```

krlmlr:
`min(integer())` returns `NA_integer_`.

Closes #3036.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -17,6 +17,8 @@
 
 * `case_when()` accepts `NA` on the LHS (#2927).
 
+* In hybrid evaluation, `min()` and `max()` now always return integers for integer arguments. `min(integer())` is returned as `NA_integer_` (#3047).
+
 # dplyr 0.7.2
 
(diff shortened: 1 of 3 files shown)


lionel-:
Is it desirable to have different semantics than base R functions?

lionel-:
It seems consistency should be the primary goal for the same reason that a compiler should produce code that behaves just like the interpreter.

krlmlr:
Let's discuss this in the original issue.

PR #3048, opened 2017-08-23
Support zero-length vectors in case_when()

### LINKED ISSUE #3041: case_when fails in mutate with zero-row tbl_df
When using case_when in mutate, if the input tbl_df has zero rows, I receive the error: Error in Error in mutate_impl(.data, dots) : 
  Evaluation error: LHS of case 1 (`price > 900`) must be length 1 (the longest input), not 0.

```r
tibble::tibble(
	price = numeric(0)
) %>%
	dplyr::mutate(new_col = dplyr::case_when(price > 900 ~ "Super Expensive", 
									   price >= 500 ~ "Expensive", 
									   price >= 100 ~ "Mild",
									  TRUE ~ "Cheap"))
```

krlmlr:
Closes #3041.

The second commit is optional, it simplifies error handling but the message is less verbose.

### DIFF
--- a/tests/testthat/test-case-when.R
+++ b/tests/testthat/test-case-when.R
@@ -32,7 +32,7 @@ test_that("cases must yield compatible lengths", {
       c(TRUE, FALSE) ~ 1,
       c(FALSE, TRUE, FALSE) ~ 2
     ),
-    "LHS of case 1 (`c(TRUE, FALSE)`) must be length 3 (the longest input) or one, not 2",
+    "LHS of case 2 (`c(FALSE, TRUE, FALSE)`) must be length 2 (the first non-atomic input or value) or one, not 3",
     fixed = TRUE
   )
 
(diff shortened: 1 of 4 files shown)


tests/testthat/test-case-when.R
```diff
@@ -32,7 +32,7 @@ test_that("cases must yield compatible lengths", {
       c(TRUE, FALSE) ~ 1,
       c(FALSE, TRUE, FALSE) ~ 2
     ),
-    "LHS of case 1 (`c(TRUE, FALSE)`) must be length 3 (the longest input) or one, not 2",
+    "All conditions and values must be length 1 or 2 (the first non-atomic value), not 3",
```
hadley: I'd say remove the parenthetical ("(the first non-atomic value)") and be more precise about the failures:

All conditions and values must be length 1 or 2. 
Problems at the following locations: 2.

krlmlr:
@hadley: The error message now quotes the failing arguments verbatim, for consistency with other error messages.

PR #3049, opened 2017-08-23
Get rid of rlang remote

krlmlr:
Tests seem to pass with CRAN rlang.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -23,7 +23,7 @@ Imports:
     magrittr,
     methods,
     pkgconfig,
-    rlang (>= 0.1.1.9000),
+    rlang (>= 0.1.2),
     R6,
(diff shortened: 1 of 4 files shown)


lionel-:
Weird, 0.1.2 does not contain the fix implemented here https://github.com/tidyverse/rlang/commit/638db926bdee5b92e6330fc644a5957a59c46ca9

krlmlr:
Where is this fix tested? Do I need to run on Windows to see the problem? What do we need to do to be compatible with rlang 0.1.2?

lionel-:
This PR added the tests: https://github.com/tidyverse/dplyr/pull/2886

krlmlr:
Yes, I'll probably disable these tests for now so that we don't need to wait for an rlang release. Need to fix AppVeyor tests, too.

PR #3068, opened 2017-09-03
Better checking of list arguments to bind_rows()

### LINKED ISSUE #3024: Bad error message when binding an unnamed list
```r
dplyr::bind_rows(list(1))
#> Error in bind_rows_(x, .id) : 'getCharCE' must be called on a CHARSXP
```

krlmlr:
Fixes #3024.

Related: https://github.com/billdenney/pknca/issues/48#issuecomment-326768053

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -14,6 +14,8 @@
 
 * Semi- and anti-joins now preserve the order of left-hand-side data frame (#3089).
 
+* Improved error message for invalid list arguments to `bind_rows()` (#3068).
+
 * Grouping by character vectors is now faster (#2204).
 
 # dplyr 0.7.2
--- a/src/bind.cpp
+++ b/src/bind.cpp
@@ -207,7 +207,6 @@ List rbind__impl(List dots, const SymbolString& id) {
   SymbolVector names;
 
   k = 0;
-  Function enc2native("enc2native");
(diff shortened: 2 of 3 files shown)


lionel-:
> Argument 1 must have names

I think it should be: Argument 1 must be named

lionel-:
i.e. lists are treated as data frames so they should have outer names and the inner names of vectors are irrelevant.

krlmlr:
To me, in `dplyr::bind_rows(a = df)`, the `df` is a named argument (because of the name `a`) but has names itself (the names of the columns. Can you please clarify which call should give which error message?

lionel-:
If there are unused outer names it should probably be a warning.

lionel-:
oh I understand now: "Argument 1 must have names". Argument 1 refers to the list, got it.

lionel-:
> If there are unused outer names it should probably be a warning.

Though the outer names can be used as `.id` column right? So we should allow unused outer names.

krlmlr:
I'm not sure, this PR is just about fixing an annoyance just before the upcoming dplyr bugfix release, which may or may not be connected with a segfault only CRAN is seeing. Can we discuss wording of error messages after the release?

krlmlr:
Simplified the check.

Issue #2311, opened 2016-12-11
Ungrouped processing should always be a special case of grouped processing


krlmlr:
This would simplify the code base considerably.

Use the existing `FullDataFrame` class in place of `GroupedDataFrame` and `RowwiseDataFrame`, respectively. Probably need to add/extend/refactor.

- [ ] Add consistency checks for each data structure for each verb
- [ ] Specifically test for missing group variable (#2031)
- [ ] Convert `sample_frac()` and `sample_n()` to C++
- [ ] Use hybrid evaluation also in the ungrouped case for consistency (#3036)

krlmlr:
Mostly done now, the fourth bullet point is "hybrid all or nothing". Will open minor issues if necessary.

PR #3073, opened 2017-09-03
Fix slice() for negative indices

krlmlr:
if the range stretches over the number-of-rows boundary.

Fixes downstream problems with waccR.


src/slice.cpp
```diff
@@ -66,7 +66,7 @@ DataFrame slice_grouped(GroupedDataFrame gdf, const QuosureList& dots) {
   Call call(quosure.expr());
 
   std::vector<int> indx;
-  indx.reserve(1000);
```
hadley: Why does this need to change?

krlmlr: It doesn't, removed. We should take another look as part of #2311.

PR #3098, opened 2017-09-12
Fix for recent Fedora and ASAN errors

krlmlr:
because the original fix didn't help.

Fedora: https://www.r-project.org/nosvn/R.check/r-devel-linux-x86_64-fedora-clang/dplyr-00check.html

ASAN: https://www.stats.ox.ac.uk/pub/bdr/memtests/clang-ASAN/dplyr/00check.log

UBSAN: https://www.stats.ox.ac.uk/pub/bdr/memtests/clang-UBSAN/dplyr/00check.log

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,10 +10,10 @@ env:
   - _R_CHECK_FORCE_SUGGESTS_=false
   - MAKEFLAGS="-j 2"
 
-include:
-- r: release
-- r: oldrel
-- r: devel
+r:
+- release
+- oldrel
+- devel
 
 before_script:
 - mkdir -p ~/.R; echo 'PKG_CXXFLAGS := ${PKG_CXXFLAGS} -Wall -Wextra -pedantic -Werror' > ~/.R/Makevars
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,6 +1,6 @@
 Type: Package
 Package: dplyr
-Version: 0.7.3.9000
+Version: 0.7.4
 Title: A Grammar of Data Manipulation
 Description: A fast, consistent tool for working with data frame like objects,
     both in memory and out of memory.
--- a/NEWS.md
+++ b/NEWS.md
@@ -2,6 +2,8 @@
 
 * Fix recent Fedora and ASAN check errors (#3098).
 
+* Avoid dependency on Rcpp 0.12.10 (#3106).
+
 # dplyr 0.7.3
 
 * Fixed protection error that occurred when creating a character column using grouped `mutate()` (#2971).
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,3 +1,7 @@
+# dplyr 0.7.3.9000
+
+* Attempt to fix recent Fedora and ASAN check errors (#3098).
+
 # dplyr 0.7.3
 
 * Fixed protection error that occurred when creating a character column using grouped `mutate()` (#2971).
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,6 +1,6 @@
-# dplyr 0.7.3.9000
+# dplyr 0.7.4
 
-* Attempt to fix recent Fedora and ASAN check errors (#3098).
+* Fix recent Fedora and ASAN check errors (#3098).
 
 # dplyr 0.7.3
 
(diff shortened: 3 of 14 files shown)


lionel-:
Should `class OrderVectorVisitorImpl : public OrderVisitor` take`VECTOR` by const ref?

https://github.com/tidyverse/dplyr/blob/908e0a59e65fdf9237fc624affeb8d93e6fa9132/inst/include/dplyr/OrderVisitorImpl.h#L42

krlmlr:
`const&` members seem dangerous because the class that holds the reference may live longer than the referenced object: 0b40356.

lionel-:
Sorry I got confused while trying to decipher the stack trace. Most of the classes involved assign members by const ref, I wondered if having one class in the mix taking by value could create problems somehow but I guess that doesn't make sense.

The changes you are making in this patch seem to be in another method than the one reported in the SAN reports by the way. The report also indicates:

- The faulty memory access occurs while subsetting the Rcpp integer vector of the slicing index.

- The report says that "Memory access at offset 280 is inside this variable". This is a temporary object created while instantiating `slice`

- There's a warning about a pointer overflow

   ```
   pointer index expression with base 0x000000000000 overflowed to 0xfffffffffffffff8
   ```

I think the temporary object is created when constructing `VectorSliceVisitor<RTYPE>`. Its constructor takes the vector by const ref but it is passed a `SEXP`. The implicit coercion creates a temporary object.

lionel-:
So yes we should probably get rid of all const ref members because it seems all too easy to create hard-to-debug UB. The compiler will probably elide many copies anyway.

hadley:
Alternatively, you could make them pointers to make the ownership more explicit.

krlmlr:
I'd rather avoid naked pointers, I don't see how this makes ownership more explicit.

Thanks @lionel- for looking into this. I agree about removing `const&` members and replacing them by smart classes like `Vector` or perhaps `shared_ptr`, which are essentially pointers with reference count semantics. With C++11 the copy overhead should go away in many cases.

lionel-:
I think Hadley was thinking about something like "scope ownership", i.e. making sure that const refs refer to objects that belong to a scope, which is problematic with implicit coercions. Pointers could help here. I agree that overloading the concept of ownership for this case is a bit confusing especially since we are talking about pointers.

I wouldn't be surprised if shared pointers have more overhead than simple values (copy elisions vs counting). I think the best thing to do is to try both and measure performance.

lionel-:
An alternative or intermediate step would be to add an `explicit` keyword to constructors of all classes that wrap const refs.

lionel-:
I have reviewed the headers and added the `explicit` keyword to all classes that wrap refs: https://github.com/krlmlr/dplyr/compare/b-ub...lionel-:b-ub

In theory this should disallow temporary objects as well as copy initilisation (things like `Comparer x = Comparer(y)`). However in the second commit (https://github.com/krlmlr/dplyr/commit/1e0748b9d3e115e7fb03b95a8e6b6e1d9a6495b8) I reintroduce both of those to check that the build fails and it does not. Any idea why @krlmlr?

krlmlr:
I think you need, for the Compare_Single_OrderVisitor` template class, either a private copy constructor or (in >= C++11) a copy constructor marked with `= delete`.

Inheriting from [`boost::noncopyable`](http://www.boost.org/doc/libs/master/libs/core/doc/html/core/noncopyable.html) seems to be a safer way to forbid copying.

Also, there seems to be no need to mark constructors with != 1 arguments as `explicit`, because AFAIR this modifier only avoids automatic coercion.

krlmlr:
Anyway, I'd rather resolve these issues separately. At this point we have UB which needs to be handled with as few changes as possible, to speed up the release process. I couldn't yet replicate the UBSAN problems locally, so CRAN is currently the only way to check if our fixes work.

hadley:
CRAN accepted the last dplyr release so the fixed isn't needed urgently.

krlmlr:
That's true, but we've been fighting this particular problem long enough, and now is a window where we can simply push an update addressing this with very little effort.

krlmlr:
I could replicate the Fedora problems on Ubuntu + clang 5.0 (with `-O3` only).

krlmlr:
Okay to merge after 0.7.4 hits CRAN?

hadley:
Yup

PR #3102, opened 2017-09-14
added support for SQLServer

dantonnoriega:
Updated documentation by searching where I found instances of `src_mysql`. Tested new function `dplyr::src_sqlserver` to connect to an Azure server successfully.


hadley:
`src_` functions now only exist for backward compatibility - going forward you should just use the DBI connection.

PR #3123, opened 2017-09-24
Add warning if any of the selected columns for `distinct()` are of type `list`

### LINKED ISSUE #3088: distinct() does not work if a column of the tbl is a list column
If a data frame contains a list column then distinct is not working. It simply returns all rows.

```r
library(tibble)
library(dplyr)
testTbl <- tibble(a=c("1", "1", "2", "2", "3","3"), b=c(list("A"), list("A"), list("B"), list("B"), list("C"), list("C")))
testTblDistinct <- testTbl %>% distinct()
```

foo-bar-baz-qux:
Fixes #3088. Went with a warning so as not to introduce a breaking change, should probably be an error in next version that allows breaking changes (0.9 I'm assuming).

I added a new NEWS section as well as I assume 0.7.3 is already released.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -43,13 +43,13 @@
 
 * `select()` and `vars()` now treat `NULL` as empty inputs (#3023).
 
+* Add error for `distinct()` if any of the selected columns are of type `list` (#3088, @foo-bar-baz-qux).
 
 # dplyr 0.7.4
 
 * Fix recent Fedora and ASAN check errors (#3098).
 
 * Avoid dependency on Rcpp 0.12.10 (#3106).
-
 
 # dplyr 0.7.3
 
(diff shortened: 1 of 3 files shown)


hadley:
I think this can be an error because turning a silent failure into an error doesn't seem like an API breaking change to me

foo-bar-baz-qux:
Ok, I can change it to an error. Was just thinking about #3055 where throwing an error caused an issue on CRAN.

hadley:
Calling distinct with list columns seems less common. Plus this is going to a minor release, so we can have some minor API changes.

foo-bar-baz-qux:
Updated with error instead of warning, and rebased against master.

hadley:
Thanks!

jankislinger:
It throws an error even if the list-column is not selected when `.keep_all = TRUE`. I am not sure if this is desired behavior or not.
``` r
df <- data_frame(x = sample(1:2, 5, T), y = as.list(1:5))
distinct(df, x, .keep_all = T)
#> Error: distinct() does not support columns of type `list`
```

krlmlr:
@jankislinger: Thanks, this has been reported in #3226. Please note that comments on closed issues and pull requests are more difficult to follow.

PR #3127, opened 2017-09-27
use quiet = TRUE with src_dbi

mtmorgan:
- depends on pull request to tidyverse/dbplyr#39


hadley:
Fixed by @krlmlr's change to the `auto_disconnect` semantics.

PR #3142, opened 2017-10-17
Correct link title in README

takahisah:
In the README, it says the best place to start is the "data import chapter" when it is actually pointing to the "data transformation chapter" and I believe the transformation chapter is the correct one to reference for dplyr.


hadley:
Thanks!

PR #3177, opened 2017-10-29
add consistency to join documentation

### LINKED ISSUE #3097: na_matches argument not mentioned in documentation
The `na_matches` argument isn't mentioned in the `?dplyr::join` documentation.  Mentioned on this StackOverflow post: https://stackoverflow.com/a/46161655/4470365

sfirke:
To close #3097.  This tweaks a few parts of what users see in the `?inner_join` and related documentation:

1. Current documentation says "dplyr supports four join types" but then lists six joins.  This breaks them into two categories of 4 and 2, consistent with the examples where they are grouped into Mutating & Filtering.
2. Reorders them to match the examples
3. Current language says that mutating joins add variables to the LHS; rewrote this sentence to be true for `right_join` as well.
4. To close #3097, tells users to visit `join.tbl_df` to see additional arguments to the `_join` functions.

(I realize that to be a complete PR, this file needs to be turned into an .Rd file with `document()`, but I'm having problems doing so with this package & Windows so am hopeful someone else can run `document()` if this addition is desired)

### DIFF
--- a/R/join.r
+++ b/R/join.r
@@ -7,7 +7,9 @@
 #'
 #' @section Join types:
 #'
-#' Currently dplyr supports four join types:
+#' Currently dplyr supports four types of mutating joins and two types of filtering joins.
+#'
+#' \strong{Mutating joins} combine variables from the two data.frames:
 #'
 #' \describe{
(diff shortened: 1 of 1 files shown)


hadley:
Thanks!

PR #3182, opened 2017-11-01
Efficient and Adaptive Stratified Sampling

saurfang:
Improves `df %>% group_by(...) %>% sample_*(...)` performance by 10-100x for dataset with large number of groups.

Also allows sample size and replacement policy to be customizable for each group.

The motivation is that when performing stratified sampling using `group_by %>% sample_n` on 100k+ strata, it can take minutes or longer. A toy example shows every 1k groups increases runtime by ~2s. A quick profiling shows most of time is spent in `eval_tidy(weight, ...)` for each group. This PR performs the weight calculation using `mutate` instead to preserve the semantics but eliminates the repeated overscope lookup across groups.

In addition, `size, frac, replace` arguments now accepts a vector for all groups in addition to a constant value. Unit tests have been added for this new feature.

cc: @lemioneta

``` r
library(dplyr)
#>
#> Attaching package: 'dplyr'
#>
#>     filter, lag
#>

n_strata <- 1000
# number of original observations in each strata (follows normal distribution)
n_per_strata_mean <- 100
n_per_strata_sd <- 10

# how many to sample from each strata
n_sample <- 10

source_df <-
  data_frame(group = 1:n_strata) %>%
  group_by(group) %>%
  do(sample_n(iris, round(rnorm(1, n_per_strata_mean, n_per_strata_sd)), replace = TRUE)) %>%
  ungroup() %>%
  # shuffle
  sample_frac(1)

source_df
#> # A tibble: 100,151 x 6
#>    group Sepal.Length Sepal.Width Petal.Length Petal.Width    Species
#>    <int>        <dbl>       <dbl>        <dbl>       <dbl>     <fctr>
#>  1   406          5.0         3.6          1.4         0.2     setosa
#>  2   998          5.5         2.4          3.8         1.1 versicolor
#>  3   101          6.4         2.8          5.6         2.2  virginica
#>  4   803          5.6         2.7          4.2         1.3 versicolor
#>  5   897          6.7         3.3          5.7         2.5  virginica
#>  6   743          4.9         3.1          1.5         0.2     setosa
#>  7   885          6.2         2.2          4.5         1.5 versicolor
#>  8    37          5.0         3.6          1.4         0.2     setosa
#>  9   276          5.2         4.1          1.5         0.1     setosa
#> 10   225          6.9         3.2          5.7         2.3  virginica
#> # ... with 100,141 more rows

# dplyr master ###########################
## sample without replacement no weights
system.time({
  source_df %>%
    group_by(group) %>%
    sample_n(n_sample)
})
#>    user  system elapsed
#>   1.677   0.006   1.687
## sample without replacement with weights
system.time({
  source_df %>%
    group_by(group) %>%
    sample_n(n_sample, weight = Petal.Length)
})
#>    user  system elapsed
#>   1.827   0.017   1.857
# sample with replacement
system.time({
  source_df %>%
    group_by(group) %>%
    sample_n(n_sample, replace = TRUE)
})
#>    user  system elapsed
#>   1.860   0.015   1.895

# dplyr dev ##################################
devtools::load_all("~/workspace/dplyr")
#> Loading dplyr
## sample without replacement no weights
system.time({
  source_df %>%
    group_by(group) %>%
    sample_n(n_sample)
})
#>    user  system elapsed
#>   0.019   0.000   0.023
## sample without replacement with weights
system.time({
  source_df %>%
    group_by(group) %>%
    sample_n(n_sample, weight = Petal.Length)
})
#>    user  system elapsed
#>   0.065   0.002   0.070
# sample with replacement
system.time({
  source_df %>%
    group_by(group) %>%
    sample_n(n_sample, replace = TRUE)
})
#>    user  system elapsed
#>   0.017   0.000   0.018
```

Before (dplyr master)

image

After

image

### DIFF
--- a/R/grouped-df.r
+++ b/R/grouped-df.r
@@ -259,20 +259,29 @@ distinct_.grouped_df <- function(.data, ..., .dots = list(), .keep_all = FALSE) 
 #' @export
 sample_n.grouped_df <- function(tbl, size, replace = FALSE,
                                 weight = NULL, .env = NULL) {
+  index <- attr(tbl, "indices")
 
-  assert_that(is_scalar_integerish(size), size >= 0)
+  assert_that(
+    is_scalar_integerish(size) || is_integerish(size, length(index)),
+    all(size >= 0),
+    is_scalar_logical(replace) || is_logical(replace, length(index))
+  )
   if (!is_null(.env)) {
     inform("`.env` is deprecated and no longer has any effect")
   }
   weight <- enquo(weight)
+  weight <- mutate(tbl, w = !!weight)[["w"]]
 
-  index <- attr(tbl, "indices")
-  sampled <- lapply(index, sample_group,
-    frac = FALSE,
-    tbl = tbl,
+  sampled <- mapply(sample_group,
+    i = index,
     size = size,
     replace = replace,
-    weight = weight
+    MoreArgs = list(
+      frac = FALSE,
+      tbl = tbl,
+      weight = weight
+    ),
+    SIMPLIFY = FALSE
   )
   idx <- unlist(sampled) + 1
 
@@ -282,7 +291,14 @@ sample_n.grouped_df <- function(tbl, size, replace = FALSE,
 #' @export
 sample_frac.grouped_df <- function(tbl, size = 1, replace = FALSE,
                                    weight = NULL, .env = NULL) {
-  assert_that(is.numeric(size), length(size) == 1, size >= 0)
+  index <- attr(tbl, "indices")
+
+  assert_that(
+    is.numeric(size),
+    length(size) == 1 || length(size) == length(index),
+    all(size >= 0),
+    is_scalar_logical(replace) || is_logical(replace, length(index))
+  )
   if (!is_null(.env)) {
     inform("`.env` is deprecated and no longer has any effect")
   }
@@ -292,14 +308,18 @@ sample_frac.grouped_df <- function(tbl, size = 1, replace = FALSE,
     )
   }
   weight <- enquo(weight)
+  weight <- mutate(tbl, w = !!weight)[["w"]]
 
-  index <- attr(tbl, "indices")
-  sampled <- lapply(index, sample_group,
-    frac = TRUE,
-    tbl = tbl,
+  sampled <- mapply(sample_group,
+    i = index,
     size = size,
     replace = replace,
-    weight = weight
+    MoreArgs = list(
+      frac = TRUE,
+      tbl = tbl,
+      weight = weight
+    ),
+    SIMPLIFY = FALSE
   )
   idx <- unlist(sampled) + 1
 
@@ -315,9 +335,8 @@ sample_group <- function(tbl, i, frac, size, replace, weight) {
     check_size(size, n, replace)
   }
 
-  weight <- eval_tidy(weight, tbl[i + 1, , drop = FALSE])
   if (!is_null(weight)) {
-    weight <- check_weight(weight, n)
+    weight <- weight[i + 1]
   }
 
   i[sample.int(n, size, replace = replace, prob = weight)]
--- a/R/grouped-df.r
+++ b/R/grouped-df.r
@@ -36,7 +36,8 @@ is_grouped_df <- is.grouped_df
 
 #' @export
 tbl_sum.grouped_df <- function(x) {
-  grps <- if (is.null(attr(x, "indices"))) "?" else length(attr(x, "indices"))
+  index <- grouped_df_r_indices(x)
+  grps <- if (is_null(index)) "?" else length(index)
   c(
     NextMethod(),
     c("Groups" = paste0(commas(group_vars(x)), " [", big_mark(grps), "]"))
@@ -50,7 +51,7 @@ group_size.grouped_df <- function(x) {
 
 #' @export
 n_groups.grouped_df <- function(x) {
-  length(attr(x, "indices"))
+  length(grouped_df_r_indices(x))
 }
 
 #' @export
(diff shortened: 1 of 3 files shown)


R/grouped-df.r
```diff
@@ -259,20 +259,29 @@ distinct_.grouped_df <- function(.data, ..., .dots = list(), .keep_all = FALSE)
...
+  assert_that(
+    is_scalar_integerish(size) || is_integerish(size, length(index)),
+    all(size >= 0),
+    is_scalar_logical(replace) || is_logical(replace, length(index))
+  )
   if (!is_null(.env)) {
     inform("`.env` is deprecated and no longer has any effect")
   }
   weight <- enquo(weight)
+  weight <- mutate(tbl, w = !!weight)[["w"]]
```
saurfang: The `[["w"]]` would return `NULL` if `weight` itself is `NULL` because `mutate` would drop the column when the RHS evaluates into `NULL`. Let me know if you prefer spelling this special case out via 

```r
weight <- mutate(tbl, w = !!weight)
weight <- if ("w" %in% names(weight)) {
  weight$w
} else {
  NULL
}
```

I don't see any other convenient way to check if `weight` is NULL without evaluating first due to NSE of column names that would be passed otherwise if the argument is not NULL.

R/grouped-df.r
```diff
@@ -259,20 +259,29 @@ distinct_.grouped_df <- function(.data, ..., .dots = list(), .keep_all = FALSE)
...
 
-  index <- attr(tbl, "indices")
-  sampled <- lapply(index, sample_group,
-    frac = FALSE,
-    tbl = tbl,
+  sampled <- mapply(sample_group,
+    i = index,
     size = size,
     replace = replace,
```
saurfang: I haven't met any use cases where I would assign different replacement policies for different groups. However, given this is practically free, I am throwing it here. The associated test case is contrived and doesn't seem to be a good sampling practice. Let me know if you prefer forcing it to be constant.

hadley: I'd prefer it to be constant

hadley: And can you please use `pmap()` (there's a helper defined in `purrr-compat.R` that implements the purrr API using `mapply()`)

R/grouped-df.r
```diff
@@ -315,9 +335,8 @@ sample_group <- function(tbl, i, frac, size, replace, weight) {
     check_size(size, n, replace)
   }
 
-  weight <- eval_tidy(weight, tbl[i + 1, , drop = FALSE])
   if (!is_null(weight)) {
```
hadley: I don't think this function needs the `tbl` argument or `i` arguments anymore - instead you can do the indexing in the higher-level functions

R/grouped-df.r
```diff
@@ -315,10 +334,21 @@ sample_group <- function(tbl, i, frac, size, replace, weight) {
...
+# Utility ----------------------------------------------------------------------
+
+grouped_df_r_indices <- function(tbl) {
+  index <- attr(tbl, "indices")
+  if (is_null(index)) {
+    NULL
+  } else {
+    map(index, function(x) x + 1)
+  }
+}
```
saurfang: I added this helper function so we don't have the mental tax to remember converting it to R 1-based index here and there. Let me know if you think this is useful or whether you have a better name or location.

R/grouped-df.r
saurfang: Yes! `tbl` seems no longer needed but looks like we still need `i` for size check. Let me know if you prefer to do vectorized size check first and `pmap` over `sample.int` instead, which would break up the `sample_group` function.

Another thing I tried is currying on the `sample_group` function and usisng `map2` instead of `pmap`. It wasn't much cleaner. I originally hopped `map2_int` would eliminate the `unlist` but then realized it expects a single integer for each map operation.

R/grouped-df.r
```diff
@@ -259,66 +260,67 @@ distinct_.grouped_df <- function(.data, ..., .dots = list(), .keep_all = FALSE)
...
-    check_size(size, n, replace)
-  }
+# Utility ----------------------------------------------------------------------
 
-  weight <- eval_tidy(weight, tbl[i + 1, , drop = FALSE])
-  if (!is_null(weight)) {
-    weight <- check_weight(weight, n)
+grouped_df_r_indices <- function(tbl) {
+  index <- attr(tbl, "indices")
+  if (is_null(index)) {
```
saurfang: In the `do.group_df` function (L164) , you handled a case where this `indices` attribute could be missing. Can you clarify when that will happen and is this also a concern for `sample_*` function? If so, should we move the index calculation invoke to here instead?

R/grouped-df.r
```diff
@@ -259,66 +260,67 @@ distinct_.grouped_df <- function(.data, ..., .dots = list(), .keep_all = FALSE)
...
+  check_weight(weight, nrow(tbl))
 
-  index <- attr(tbl, "indices")
-  sampled <- lapply(index, sample_group,
-    frac = FALSE,
-    tbl = tbl,
-    size = size,
-    replace = replace,
-    weight = weight
+  sampled <- pmap(
```
hadley: I think both `sample_n` and `sample_frac` should use the same underlying helper function. I would prefer not to change the API of `sample_n()` to allow you to specify a number of samples per group.

saurfang:
Sure. I do think there is a legitimate use case for different sample sizes per group but I agree that the interface does not feel natural here given the vector relies on an internal knowledge of the group order. This PR is a little too messy and I started a fresh one at #3193.

I am happy to submit followup PRs if you think it is worth tidying up the R/C++ index difference and possible corner cases when `indices` are not available.

PR #3194, opened 2017-11-06
Update revdeps

lionel-:
Failures due to tidyselect:

- **assertr**: Unit test depends on tidyselect error message which has changed.

- **biobroom, switchde, TPP**: Warning about replacing previous import `dplyr::exprs` by `Biobase::exprs`. These packages should selectively import dplyr or refer to Biobase::exprs() with qualifier.

- **ddpcr**: Fixed in next tidyselect version by https://github.com/tidyverse/tidyselect/commit/1831a92bad3abdffca41370e4a7540822a77619e

- **desctable**: Uses datawise complex expression in `-`. This doesn't work anymore because expressions such as `- starts_with(foo)` are scoped in the context only. This seems like a minor breaking change.

- **modelr, sjlabelled, sjmisc, sjstats**: Warning about missing `select_helpers` link. I have added a topic alias to fix these.

- **neuropsychology**: There's a lonely `select_vars()` in the source. Selecting without arguments now fails.


revdep/email.md
```diff
@@ -0,0 +1,40 @@
+Hi,
+
+This is an automated email to let you know about the release of {{{ my_package }}}, which will be submitted to CRAN in the near future (on {{{ date }}}).
+
+The major change in this version is that dplyr now depends on the selecting
```
hadley: Can you please remove from here and instead putting in a "Breaking changes" section of the news?

lionel-: Here you go.

PR #3247, opened 2017-12-13
add example in case_when about consistent RHS

### LINKED ISSUE #3202: Assigning NA yields error in case_when
Not sure if this is a bug or feature-request... I've seen several `NA` errors related to `case_when`, like #2000, but I think they are all with the LHS, but what about assigning an NA? 

I know that technically NA is class `logical`, which explains the error I got... but shouldn't I be able to assign NAs to otherwise consistent vectors? Check out the reprex:

```r
library(dplyr)
iris %>% mutate(
  Sepal.Width=case_when(
    Sepal.Length == 5.1 ~ NA,
    TRUE ~ 2)) %>% head
#> Error in mutate_impl(.data, dots): Evaluation error: must be type logical, not double.

iris %>% mutate(
  Sepal.Width=case_when(
    Sepal.Length == 5.1 ~ 9999,
    TRUE ~ 2)) %>% head
#> 1          5.1        9999          1.4         0.2  setosa
#> 2          4.9           2          1.4         0.2  setosa
#> 3          4.7           2          1.3         0.2  setosa
#> 4          4.6           2          1.5         0.2  setosa
#> 5          5.0           2          1.4         0.2  setosa
#> 6          5.4           2          1.7         0.4  setosa
```

cderv:
closes #3202 

I added some examples in the documentation to show how consistent RHS works when using `NA` value. 

I think just adding example is enough but I can add precisions in documentation text also. 

Question : Can we add a sample of code that throws an error in a doc example ? 
something like
```r
x <- 1:50
case_when(
  x %% 35 == 0 ~ 35,
  x %% 5 == 0 ~ 5,
  x %% 7 == 0 ~ 7,
  TRUE ~ NA
)
```

I did not generate `Rd` file as you may prefer to do it on your side. Tell me if I had to.

### DIFF
--- a/R/case_when.R
+++ b/R/case_when.R
@@ -49,6 +49,15 @@
 #'   x %% 7 == 0 ~ 7,
 #'   TRUE ~ NA_real_
 #' )
+#' # This throws an error as NA is logical not numeric
+#' \dontrun{
+#' case_when(
+#'   x %% 35 == 0 ~ 35,
+#'   x %% 5 == 0 ~ 5,
+#'   x %% 7 == 0 ~ 7,
+#'   TRUE ~ NA
+#' )
+#' }
 #'
 #' # case_when is particularly useful inside mutate when you want to
 #' # create a new variable that relies on a complex combination of existing
--- a/R/case_when.R
+++ b/R/case_when.R
@@ -34,6 +34,22 @@
 #'   x %% 35 == 0 ~ "fizz buzz"
 #' )
 #'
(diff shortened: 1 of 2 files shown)


krlmlr:
Thanks. We can include examples that aren't supposed to be run with `\dontrun{}`. It's slightly better if `.Rd` updates are included in a PR, but it's not particularly important.

cderv:
I have just added a `\dontrun` example to show the difference between `NA` and `NA_real_` and included the Rd file in the PR

krlmlr:
Thanks!

PR #3249, opened 2017-12-13
Modify tally and friends to be consistent

lepennec:
Hi,

I've noticed that tally and their firends were not consistent. You could not apply consistently tally more than two times. The third time the count variable `nn` is not interpreted as a count anymore.

I suggest to use `n` only as a count for rows. Thus `tally` should (almost) always use this name `n` to store the counts. To be consistent, I suggest to use `nn` (and the `nnn` and so on) when using `add_tally` whether or not it exists a column `n`.

The only exception is when using `tally` when `n` is a grouping variable. In that case, I suggest using `nn` for the count with a warning message.

This modifies slightly the behavior of `add_tally` but I hope this may be acceptable.

Erwan

### DIFF
--- a/R/count-tally.R
+++ b/R/count-tally.R
@@ -6,19 +6,19 @@
 #' for the first time, or re-tallying. `count()` is similar but calls
 #' [group_by()] before and [ungroup()] after.
 #'
-#' `add_tally()` adds a column "n" to a table based on the number
+#' `add_tally()` adds a column "nn" to a table based on the number
 #' of items within each existing group, while `add_count()` is a shortcut that
 #' does the grouping as well. These functions are to [tally()]
 #' and [count()] as [mutate()] is to [summarise()]:
 #' they add an additional column rather than collapsing each group.
 #'
 #' @note
-#' The column name in the returned data is usually `n`, even if you
-#' have supplied a weight.
+#' The column name in the returned data is `n` for `tally` and `¢ount`,
+#' even if you have supplied a weight.
 #'
-#' If the data already already has a column named `n`, the output column
-#' will be called `nn`. If the table already has columns called `n` and `nn`
-#' then the column returned will be `nnn`, and so on.
+#' For `add_tally` (and `add_count`), the output column
+#' will be called `nn` even if no column `n` exists. If the table already
+#' has columns called `nn` then the column returned will be `nnn`, and so on.
 #'
 #' There is currently no way to control the output variable name - if you
 #' need to change the default, you'll have to write the [summarise()]
@@ -61,7 +61,7 @@
 tally <- function(x, wt, sort = FALSE) {
   wt <- enquo(wt)
 
-  if (quo_is_missing(wt) && "n" %in% names(x)) {
+  if (quo_is_missing(wt) && "n" %in% tbl_nongroup_vars(x)) {
     inform("Using `n` as weighting variable")
     wt <- quo(n)
   }
@@ -72,7 +72,14 @@ tally <- function(x, wt, sort = FALSE) {
     n <- quo(sum(!! wt, na.rm = TRUE))
   }
 
-  n_name <- n_name(tbl_vars(x))
+  if (("n" %in% tbl_vars(x))&(!("n" %in% tbl_nongroup_vars(x)))) {
+    n_name <- nn_name(tbl_vars(x))
+    warning(paste0("Using `",n_name,"` as count/tally variables"))
+  } else {
+    n_name <- "n"
+  }
+
+
   out <- summarise(x, !! n_name := !! n)
 
   if (sort) {
@@ -89,8 +96,8 @@ tally_ <- function(x, wt, sort = FALSE) {
   tally(x, wt = !! wt, sort = sort)
 }
 
-n_name <- function(x) {
-  name <- "n"
+nn_name <- function(x) {
+  name <- "nn"
   while (name %in% x) {
     name <- paste0(name, "n")
   }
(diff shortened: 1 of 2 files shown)


krlmlr:
Thanks for raising this. Could you please provide reproducible examples of the problems you are seeing? The implementation you are suggesting would be a breaking change, I feel we should discuss the problems and potential solutions before moving forward.

lepennec:
Here is the preview (incorrect behavior)
``` r
library(tidyverse)
#> ── Attaching packages ───────────────────────────────── tidyverse 1.2.1 ──
#> ✔ ggplot2 2.2.1     ✔ purrr   0.2.4
#> ✔ tibble  1.3.4     ✔ dplyr   0.7.4
#> ✔ tidyr   0.7.2     ✔ stringr 1.2.0
#> ✔ readr   1.1.1     ✔ forcats 0.2.0
#> ── Conflicts ──────────────────────────────────── tidyverse_conflicts() ──
#> ✖ dplyr::filter() masks stats::filter()
#> ✖ dplyr::lag()    masks stats::lag()
mtcars %>% group_by(gear, vs, am) %>% tally()
#> # A tibble: 7 x 4
#> # Groups:   gear, vs [?]
#>    gear    vs    am     n
#>   <dbl> <dbl> <dbl> <int>
#> 1     3     0     0    12
#> 2     3     1     0     3
#> 3     4     0     1     2
#> 4     4     1     0     4
#> 5     4     1     1     6
#> 6     5     0     1     4
#> 7     5     1     1     1
mtcars %>% group_by(gear, vs, am) %>% tally() %>% tally()
#> # A tibble: 6 x 3
#> # Groups:   gear [?]
#>    gear    vs    nn
#>   <dbl> <dbl> <int>
#> 1     3     0    12
#> 2     3     1     3
#> 3     4     0     2
#> 4     4     1    10
#> 5     5     0     4
#> 6     5     1     1
mtcars %>% group_by(gear, vs, am) %>% tally() %>% tally() %>% tally()
#> # A tibble: 3 x 2
#>    gear     n
#>   <dbl> <int>
#> 1     3     2
#> 2     4     2
#> 3     5     2
```

and now the one with the proposed correction.

``` r
library(tidyverse)
#> ── Attaching packages ───────────────────────────────── tidyverse 1.2.1 ──
#> ✔ ggplot2 2.2.1          ✔ purrr   0.2.4     
#> ✔ tibble  1.3.4          ✔ dplyr   0.7.4.9000
#> ✔ tidyr   0.7.2          ✔ stringr 1.2.0     
#> ✔ readr   1.1.1          ✔ forcats 0.2.0
#> ── Conflicts ──────────────────────────────────── tidyverse_conflicts() ──
#> ✖ dplyr::filter() masks stats::filter()
#> ✖ dplyr::lag()    masks stats::lag()
mtcars %>% group_by(gear, vs, am) %>% tally()
#> # A tibble: 7 x 4
#> # Groups:   gear, vs [?]
#>    gear    vs    am     n
#>   <dbl> <dbl> <dbl> <int>
#> 1     3     0     0    12
#> 2     3     1     0     3
#> 3     4     0     1     2
#> 4     4     1     0     4
#> 5     4     1     1     6
#> 6     5     0     1     4
#> 7     5     1     1     1
mtcars %>% group_by(gear, vs, am) %>% tally() %>% tally()
#> # A tibble: 6 x 3
#> # Groups:   gear [?]
#>    gear    vs     n
#>   <dbl> <dbl> <int>
#> 1     3     0    12
#> 2     3     1     3
#> 3     4     0     2
#> 4     4     1    10
#> 5     5     0     4
#> 6     5     1     1
mtcars %>% group_by(gear, vs, am) %>% tally() %>% tally() %>% tally()
#> # A tibble: 3 x 2
#>    gear     n
#>   <dbl> <int>
#> 1     3    15
#> 2     4    12
#> 3     5     5
```
I understand that this is breaking the previous behavior. A second possibility is to add one 'n' at each step (as this is supposed to be according to the doc) and using the variable with the least 'n' as a count. Note that the behavior of add_tally is less predictible in that case...

krlmlr:
Thanks, it's much clearer with the example.

There's also `add_tally()` that would keep both `n` and `nn` if `n` already exists. Perhaps `tally()` should be consistent with this? Code that depends on the broken behavior of a triple `tally()` is broken anyway, so it feels safe to look for a column that matches `"^n+$"`. However, I also feel that it may be safer to abort if multiple matches are found, the user still can indicates the weighting column manually. That would still be a breaking change, but maybe the least disruptive one?

#3075 is related, changes related to `tally()` should also fix that problem.

krlmlr:
On the other hand, your example work if you use `tally(nn)` instead of the third `tally()`, so I'm not sure how important this is. We may be better off just keeping the current implementation and documenting it, maybe using your example.

krlmlr:
@hadley: Should we use `nn` as weighting variable in `tibble(a = 1:3, nn = 1:3) %>% tally()`? If yes, how should `tibble(a = 1:3, n = 1:3, nn = 1:3) %>% tally()` then behave?

- throw an error?
- use `n` as weighting variable (current behavior)?
- use `nn` as weighting variable?

Currently, we can't apply `tally()` three times in a row, but we can do `. %>% tally() %>% tally() %>% tally(nn)`. If we change it, I suggest we throw an error and require manual specification -- otherwise we also need to ask what happens on `tibble(a = 1:3, n = 1:3, nnn = 1:3) %>% tally()` etc.. That would be a breaking change, though.

hadley:
I think the current behaviour is fine. It's not perfect, but I think at some point you have to manually specify the weighting variable and I'd rather not make `tally()` more complicate.

krlmlr:
Thanks. Then I think we should fix the documentation for the `wt` argument, explicitly mentioning that we're only looking for a column named `"n"` but not `"nn"`.

lepennec:
Sounds perfect to me. Do you want me to propose a documentation fix?

krlmlr:
This would be great.

PR #3261, opened 2017-12-23
conservative styling with styler

lorenzwalthert:
`style_pkg(scope = "spaces", strict = FALSE)`

### DIFF
--- a/tests/testthat/test-combine.R
+++ b/tests/testthat/test-combine.R
@@ -172,8 +172,8 @@ test_that("combine works with NA and complex (#2203)", {
 
 test_that("combine works with integer64 (#1092)", {
   expect_equal(
-    combine(bit64::as.integer64(2^34), bit64::as.integer64(2^35)),
-    bit64::as.integer64(c(2^34, 2^35))
+    combine(bit64::as.integer64(2 ^ 34), bit64::as.integer64(2 ^ 35)),
+    bit64::as.integer64(c(2 ^ 34, 2 ^ 35))
   )
 })
 
@@ -188,15 +188,15 @@ test_that("combine works with difftime", {
   )
   expect_equal(
     combine(as.difftime(1, units = "days"), as.difftime(1, units = "secs")),
-    as.difftime(c(24*60*60, 1), units = "secs")
+    as.difftime(c(24 * 60 * 60, 1), units = "secs")
   )
   expect_equal(
     combine(as.difftime(2, units = "weeks"), as.difftime(1, units = "secs")),
-    as.difftime(c(2*7*24*60*60, 1), units = "secs")
+    as.difftime(c(2 * 7 * 24 * 60 * 60, 1), units = "secs")
   )
   expect_equal(
     combine(as.difftime(2, units = "weeks"), as.difftime(3, units = "weeks")),
-    as.difftime(c(2,3), units = "weeks")
+    as.difftime(c(2, 3), units = "weeks")
   )
 
 })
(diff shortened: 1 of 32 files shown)


lionel-:
Looks great!

tests/testthat/test-combine.R
```diff
@@ -172,8 +172,8 @@ test_that("combine works with NA and complex (#2203)", {
 
 test_that("combine works with integer64 (#1092)", {
   expect_equal(
-    combine(bit64::as.integer64(2^34), bit64::as.integer64(2^35)),
-    bit64::as.integer64(c(2^34, 2^35))
+    combine(bit64::as.integer64(2 ^ 34), bit64::as.integer64(2 ^ 35)),
```
lionel-: I think there shouldn't be any spaces for exponentiation. The `^` operator has very high precedence, even higher than `:` so spacing should reflect that. E.g. `1:2 ^ 3` would be misleading about precedence.

lorenzwalthert: Thanks for reviewing. The tidyverse style guide requires a space between all infix operators according to [section 2.2](http://style.tidyverse.org/syntax.html#spacing) (although `^` is not mentioned explicitly), so we concluded to implement spacing around `^` too. Your example `1:2 ^ 3` makes sense to me though, but if the guiding principle for spacing around math tokens should be operator precedence, I am not sure whether quite a few rules had to be rethought (e.g. also `1 + 1 / 2`, `c(1:2) + 2 %>% sum()`). Do you think we should suggest a change to the style guide? 

There was a similar discussion about spacing around math tokens [here](https://github.com/tidyverse/lubridate/pull/577#issuecomment-331698854), which led me to implement the argument `math_token_spacing`. Hence, we could style with zero spaces around `^` with
```r
library("styler")
style_pkg(
  scope = "spaces",  
  strict = FALSE, 
  math_token_spacing = specify_math_token_spacing(zero = "'^'")
)
```
To achieve the styling you suggest. How should we proceed?

lionel-: > I am not sure whether quite a few rules had to be rethought

I don't think so. The way I see it there's several lines of reasoning and the whole thing does not necessarily need to be consistent with each of them. In this case reflecting the precedence helps readability, compare `1 ^ 2 * 3` to `1^2 * 3`. AFAICT no spaces around `^` is a strong convention across languages.

What do you think @hadley?

batpigandme: Not a data-driven answer, but, FTR, (though incorrect) I'd almost read `1 ^ 2 * 3` as 1 to the power of six, just because of the engrained conventions re spacing (or lack thereof) bw the base and its exponent.

lorenzwalthert: > The way I see it there's several lines of reasoning and the whole thing does not necessarily need to be consistent with each of them.

Ok, that's fine for me. I don't have strong feelings about how to do it, I just wanted to make sure that it is in line with the tidyverse style guide, and if the style guide is not precise enough at that point (because it does not mention `^` explicitly), I thought it might be worth considering adapting it and then (in turn) change the default option in styler.

tests/testthat/test-rbind.R
```diff
@@ -43,107 +43,107 @@ test_that("rbind_list reorders columns", {
...
+  ll <- list(c(1, 2, 3, 4, 5), c(6, 7, 8, 9, 10))
   expect_error(rbind_all_warn(ll))
 })
 
-test_that( "rbind propagates timezone for POSIXct #298", {
-  dates1 <- data.frame(ID=c("a", "b", "c"),
-                     dates=structure(c(-247320000, -246196800, -245073600),
+test_that("rbind propagates timezone for POSIXct #298", {
+  dates1 <- data.frame(ID = c("a", "b", "c"),
+                     dates = structure(c(-247320000, -246196800, -245073600),
```
krlmlr: @lorenzwalthert: This call looks odd, `dates2` looks right. Can you please try in a reprex?

tests/testthat/test-combine.R
krlmlr: Will tweak after merging.

tests/testthat/test-rbind.R
lorenzwalthert: What's odd about it? Indention is untouched with `scope = "spaces"`

lorenzwalthert:
Should we do a follow-up styling with scope = "indention" etc?

tests/testthat/test-rbind.R
krlmlr: This explains it. I can do the follow-up, need to run styler with the new gprofiler package which doesn't yet work on OS X unfortunately.

lorenzwalthert: Ok sure, I leave the follow-up to you then.

lorenzwalthert:
@krlmlr will take care of the follow-up. Reference:  https://github.com/tidyverse/dplyr/pull/3261#discussion_r159131497

PR #3271, opened 2017-12-30
Joins give better error message with duplicate column names

### LINKED ISSUE #3243: duplicate colnames gives uninformative error when joining with tibbles
Tibbles don't allow for duplicate column names at creation:

    > data_frame(x = 1:2, x = 3:4)
    Error: Column `x` must have a unique name

The `_join` functions throw informative errors when join key is duplicate and x is a regular `data.frame`

    > x <- data.frame(x = 1:2, x = 3:4)
    > colnames(x) <- c("x", "x")
    > inner_join(x, data.frame(x = 1:2))
    Error: Column `x` must have a unique name

However, via the same back door the column names of a `tibble` can get duplicated. When trying to join then the error is no longer informative.

    > x <- data_frame(x = 1:2, x2 = 3:4 )
    > colnames(x) <- c("x", "x")
    > inner_join(x, data_frame(x = 1:2))
    Joining, by = "x"
    Error in inner_join_impl(x, y, by$x, by$y, suffix$x, suffix$y, check_na_matches(na_matches)) : 
       Column `` is unknown

If possible, I think the best option here is close the back door route and throw an error at the second line of the last code block. Otherwise an informative error seems desired.

krlmlr:
- semi- and anti-joins give a warning if RHS has dupes

Fixes #3243.

I have no idea why the build fails when `~/.R/Makevars` is created on Travis.

### DIFF
--- a/src/utils.cpp
+++ b/src/utils.cpp
@@ -10,9 +10,6 @@ using namespace Rcpp;
 
 // [[Rcpp::export(name = "check_valid_colnames")]]
 void check_valid_colnames_export(const DataFrame& df, bool warn_only = false) {
-  if (df.size() == 0)
-    return;
-
   CharacterVector names(df.names());
   LogicalVector dup = duplicated(names);
   if (any(dup).is_true()) {
--- a/src/utils.cpp
+++ b/src/utils.cpp
@@ -8,6 +8,26 @@
 
 using namespace Rcpp;
 
+// [[Rcpp::export(name = "check_valid_colnames")]]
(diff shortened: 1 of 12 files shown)


src/utils.cpp
```diff
@@ -8,6 +8,26 @@
 
 using namespace Rcpp;
 
+// [[Rcpp::export(name = "check_valid_colnames")]]
+void check_valid_colnames_export(const DataFrame& df, bool warn_only = false) {
+  if (df.size()) {
```
hadley: I general think an early return is better style than a large block

PR #3273, opened 2017-12-31
fix cbind() error message for vectors

### LINKED ISSUE #3085: bind_cols Error message has odd output and maybe wrong values
I was trying to take all the columns of data in a tibble, split a text column on a certain string and then attach that column, now having 1 or more items, back onto the original data. To make it even simpler I did it with just one row of data. I can resolve the issue by using crossing from tidyr, or even cbind from base R. The issue here thought is related to the error message given. It seems to say it needs what it actually has, and even then when you meet that need it fails, the crazy case says this, Argument 2 must be length 1, not 1.

```r
library(tidyverse)
data <- tibble(a = 1, b = 2)
new_c <- 1:3


bind_cols(data, tibble(new_c = new_c))
# Error in cbind_all(x) : Argument 2 must be length 3, not 1
bind_cols(data, new_c = new_c)
# Error in cbind_all(x) : Argument 2 must be length 1, not 1

# This is probably the way to do this.
crossing(data, new_c)
```

krlmlr:
Fixes #3085.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -53,6 +53,8 @@
 
 * Better error message if dbplyr is not installed when accessing database backends (#3225).
 
+* Corrected error message when calling `cbind()` with an object of wrong length (#3085).
+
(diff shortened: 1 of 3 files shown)


krlmlr:
Thanks, I requested a review just to confirm the wording of the error message (again).

PR #3274, opened 2017-12-31
summarise() works for empty data frames

### LINKED ISSUE #3071: dplyr::tally does not work for data.frame with no columns
In attempting to find a good work-around for [`dplyr` issue 3069](https://github.com/tidyverse/dplyr/issues/3069) I ran into this.  If we form a `data.frame` that happens to have no columns (by dropping a column), then `dplyr::tally` does not return a count.

``` r
suppressPackageStartupMessages(library("dplyr"))
packageVersion("dplyr")
#> [1] '0.7.2.9000'

data.frame(n = as.raw(1:3)) %>% 
  select(-n) %>% tally
#> data frame with 0 columns and 3 rows
```

krlmlr:
Fixes #3071.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -51,6 +51,8 @@
 
 * `select()` and `vars()` now treat `NULL` as empty inputs (#3023).
 
+* Fix `summarise()` for empty data frames with zero columns (#3071).
+
 * Add error for `distinct()` if any of the selected columns are of type `list` (#3088, @foo-bar-baz-qux).
 
 * `sample_n()` and `sample_frac()` on grouped data frame are now faster especially for those with large number of groups (#3193, @saurfang).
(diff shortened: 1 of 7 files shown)


krlmlr:
Indeed, `mean(na.rm = TRUE)` doesn't handle empty vectors. Will review all handlers.

krlmlr:
I'm seeing an average 10% slowdown with this PR with `flights`, note the different scales for `has_na: TRUE` and `na_rm: FALSE` due to #3288.

I've started a writeup, http://rpubs.com/krlmlr/dplyr-spec-template, would that be suitable as an internal dplyr vignette?

rplot

krlmlr:
Update: A second measurement with `times = 100` gives a 1% speed-up.

download

PR #3283, opened 2018-01-04
Correct implicit units.

DataWookie:
The units in the `iris` data are cm. Dividing by 2.54 converts cm to inch, so it's more appropriate to label the new columns with a _in suffix.


krlmlr:
Thanks!

lionel-:
This causes a failure in the examples (which didn't show up in the checks because the doc was not updated with roxygen).

```r
quote(foo(in = bar))
#> Error: unexpected 'in' in "quote(foo(in"
```

krlmlr:
Thanks. Good argument for always updating `.Rd` files.

PR #3299, opened 2018-01-12
Update to rlang 0.2.0

lionel-:
On the R side:

- Fixes wrong assumptions about `!!` precedence
- Use proper quosure accessors

On the C++ side:

- Add an `.onLoad()` hook to initialise C callable pointers when dplyr is loaded.
- Import the dynamic rlang C API for quosures and data masks and use it in the relevant classes.


inst/include/tools/Quosure.h
```diff
@@ -31,18 +29,17 @@ class NamedQuosure {
...
-    return Rf_getAttrib(data, sym_dotenv);
+    return internal::rlang_quo_get_env(data);
   }
   SymbolString name() const {
     return name_;
   }
 
 private:
-  Formula data;
+  SEXP data;
```
krlmlr: Do we need protection here?

inst/include/tools/utils.h
```diff
@@ -45,6 +45,23 @@ inline void copy_most_attributes(SEXP out, SEXP data) {
   Rf_copyMostAttrib(data, out);
 }
 
-}
+
+namespace internal {
```
krlmlr: I wonder if we could use a class/struct here that initalizes its members on construction. That class could also live in rlang, and might help getting rid of `dplyr_onload()`.

tests/testthat/test-mutate.r
```diff
@@ -710,18 +710,18 @@ test_that("mutate() names pronouns correctly (#2686)", {
 
 test_that("mutate() supports unquoted values", {
   df <- tibble(g = c(1, 1, 2, 2, 2), x = 1:5)
-  expect_identical(mutate(df, out = !! 1), mutate(df, out = 1))
-  expect_identical(mutate(df, out = !! 1:5), mutate(df, out = 1:5))
-  expect_identical(mutate(df, out = !! quote(1:5)), mutate(df, out = 1:5))
-  expect_error(mutate(df, out = !! 1:2), "must be length 5 (the number of rows)", fixed = TRUE)
-  expect_error(mutate(df, out = !! get_env()), "unsupported type")
+  expect_identical(mutate(df, out = !!1), mutate(df, out = 1))
```
krlmlr: @lorenzwalthert: Did we agree on space after bang-bang or bang-bang-bang in styler?

tests/testthat/test-overscope.R
```diff
@@ -3,11 +3,11 @@ context("overscope")
 test_that(".data has strict matching semantics (#2591)", {
   expect_error(
     data_frame(a = 1) %>% mutate(c = .data$b),
-    "Column `b`: not found in data"
+    "Column `b` not found in data"
```
krlmlr: Does this error message come from rlang now? Then we shouldn't test the exact wording.

inst/include/tools/utils.h
lionel-: Hmm, I'm not sure we want any C++ in rlang. But that's something to think about.

I think we'd still need onload to initialise it though.

tests/testthat/test-overscope.R
lionel-: Good point.

tests/testthat/test-mutate.r
lorenzwalthert: I though the consensus was putting a space before bang-bang and bang-bang-bang. 
https://github.com/tidyverse/style/issues/38#issuecomment-324617558

lorenzwalthert: Ok, so do we need to change the style guide? And does it mean that bang-bang becomes space sensitive? I would like to understand why / what changed, can you point me to the resources where I can learn about that (if you have time)? 😊

krlmlr: I don't think bang-bang is case-sensitive, maybe the rlang NEWS contain pointers? (Haven't checked.)

inst/include/tools/utils.h
krlmlr: I'd prefer to have it encapsulated in a class implemented by rlang. If we have a C API, we can as well have an optional C++ class. We could use the following idiom to avoid the loader:

```cpp
rlang& get_rlang() {
  static rlang my_rlang;
  return rlang;
}

struct rlang {
  rlang() { /* initialize API pointers */ }
  ...
};
```

tests/testthat/test-mutate.r
lorenzwalthert: Ok thanks, what I understand from NEWS is it more natural now to write `!!a > b` than `!! a > b` because `!!` has now the same precedence as infix operators such as `>` so it only unquotes `a` in `rlang` 2.0 and not `a > b` in the above example.

lorenzwalthert: So @lionel- I suggest to then also update the style guide - and styler. I think it would make sense to commit to a rule for `!!!` also.

lionel-: You can update the guide for `!!` but we still haven't got a consensus for `!!!`. It makes sense for new syntax style to be in flux for a while.

inst/include/tools/utils.h
lionel-: so you'd do `get_rlang().mypointer()`?

krlmlr: Yes. Originally I thought about a global object with static members, but globals in static libraries don't seem to be specified very well.

tests/testthat/test-mutate.r
lorenzwalthert: Ok. So then I am also going to suggest to remove the rule for !!! in the style guide until it's decided on how handle it.

lorenzwalthert: Reference: https://github.com/r-lib/styler/issues/321

lionel-: Sounds good, thanks!

lorenzwalthert: https://github.com/tidyverse/style/issues/55

krlmlr:
Thanks for taking care of this. I assume the CRAN version of tidyverse packages will work with rlang 0.2?

FWIW, I haven't received notifications from your commits.

lionel-:
> I assume the CRAN version of tidyverse packages will work with rlang 0.2

There's some things to fix because of new operator precedence / cleaning of parentheses around !!. Overall it should mostly work without issue.

> FWIW, I haven't received notifications from your commits.

Which commits?

inst/include/tools/utils.h
lionel-: I'm on vacation this week but will implement the class the week after. I think it should live in dplyr for a while before being exported from rlang.

krlmlr:
The last two that addressed my remarks. Is this a repo or org setting?

PR #3320, opened 2018-01-24
Export more tidy eval functions

lionel-:


R/utils-tidy-eval.R
```diff
@@ -1,16 +1,42 @@
...
-#' These six functions provide tidy eval-compatible ways to capture
-#' symbols (`sym()`, `syms()`), expressions (`expr()`,
-#' `exprs()`, `enexpr()`), and quosures (`quo()`, `quos()`, `enquo()`).
-#' To learn more about tidy eval and how to use these tools, read
-#' <http://rlang.tidyverse.org/articles/tidy-evaluation.html>
+#' @description
+#'
+#' * \code{\link[rlang]{sym}()} creates a symbol from a string and
+#'   \code{\link[rlang]{syms}()} creates a list of symbols from a
+#'   character vectors.
```
krlmlr: vectors -> vector

lionel-:
@krlmlr I think the plan is still to reexport these functions just like we reexport `%<%`.

lionel-:
The tidyeval package will export non-user facing functions like `eval_tidy()` so shouldn't be loaded by the `tidyverse` package.

krlmlr:
I see we already reexport many other functions from rlang.

Do we have local tests for the reexported functions? We could just copy from rlang, I think. Strictly speaking, reexported functions are part of the interface, and any changes should be documented in dplyr's NEWS as well.

lionel-:
> Strictly speaking, reexported functions are part of the interface, and any changes should be documented in dplyr's NEWS as well.

That's something we've been discussing and we're not sure what's the right way yet. This would imply coordinated releases of all packages reexporting rlang functions, such as tidyr and soon ggplot2. That doesn't seem workable.

krlmlr:
To me, that "only" implies no breaking changes.

lionel-:
> To me, that "only" implies no breaking changes.

That's the goal. After rlang 0.2.0 tidy eval functions should be stable though, but probably still in maturing lifecycle for one more release.

Being an exported function does not imply stability, the lifecycle stage does. The important thing is not stability but clearer communication about what is stable. Hence the new `?rlang::lifecycle` topic we've been experimenting with and that we'd like to generalise in tidyverse packages.

R/utils-tidy-eval.R
```diff
@@ -1,16 +1,42 @@
...
+#'
+#' * \code{\link[rlang]{enexpr}()} and \code{\link[rlang]{enquo}()}
+#'   capture the expression supplied as argument by the user of the
+#'   current function (`enquo()` wraps this expression in a quosure).
+#'
+#'   \code{\link[rlang]{enexprs}()} and \code{\link[rlang]{enquos}()}
+#'   capture multiple expressions supplied as arguments, including
+#'   `...`.
+#'
+#' * \code{\link[rlang]{qq_show}()} shows the result of the
```
hadley: I think we can omit `qq_show()` and `quo_squash()` because they're still evolving a bit faster than the others (and are different in spirit).

Can you also do a PR to usethis?

Issue #3309, opened 2018-01-18
row_number() produces error when calling with namespace (dplyr::) in filter()


jimvine:
I would expect `dplyr::row_number()` to give the same output as I get with `row_number()`. But if I use it within `filter()` I get the expected results if I use the plain version but an error if I use the namespaced version.

```r
# Repex:
# This works fine:
filter(mtcars, row_number() == 6L)

# This gives an error:
filter(mtcars, dplyr::row_number() == 6L)
# Error in filter_impl(.data, quo) : 
#   Evaluation error: argument "x" is missing, with no default.
```

batpigandme:
If you're trying to select a row by position, the dplyr verb you want is [slice](http://dplyr.tidyverse.org/reference/slice.html):

``` r
suppressPackageStartupMessages(library(dplyr))

slice(mtcars, 6)
#> # A tibble: 1 x 11
#>     mpg   cyl  disp    hp  drat    wt  qsec    vs    am  gear  carb
#>   <dbl> <dbl> <dbl> <dbl> <dbl> <dbl> <dbl> <dbl> <dbl> <dbl> <dbl>
#> 1  18.1  6.00   225   105  2.76  3.46  20.2  1.00     0  3.00  1.00
```

Created on 2018-01-18 by the [reprex package](http://reprex.tidyverse.org) (v0.1.1.9000).

jimvine:
@batpigandme I simplified for the repex to generate the error, but I'm actually trying to selectively **drop** rows, so my actual use case is more akin to:

``` r
filter(mtcars, row_number() != 6L)
```

(As in the positive row selection case, this works with `row_names()` but fails with `dplyr::row_names()`.)

I'm not sure whether that's possible with `slice()`. This **doesn't** work:

``` r
slice(mtcars, !6)
```

hadley:
You want `slice(mtcars, -6)`

But I wanted to solicit this issue because it suggests there's a problem with the hybrid evaluation not recognising calls of the form `dplyr::`

jimvine:
@hadley That's very helpful - many thanks. Would that be worth mentioning the use of negative row values to drop in the `?slice` documentation? If so, would it be helpful for me to add that as an issue, or submitting a PR?

hadley:
@jimvine a PR would be awesome - the docs for `slice()` are pretty terrible right now

PR #3321, opened 2018-01-25
Documentation and errors for slice() with negative indices

jimvine:
Related to discussion on Issue 3309. I have added some documentation on using negative indices, and also updated the slice() error message that is triggered if a mix of positive and negative indices are used.

https://github.com/tidyverse/dplyr/issues/3309


krlmlr:
Thanks!

PR #3322, opened 2018-01-25
Added tpchr testing

hannesmuehleisen:
Check results for queries 1-10 as test case using a reasonably small scale factor


hannesmuehleisen:
Never mind, this would prevent a CRAN release  in the current form.

PR #3325, opened 2018-01-26
Clean data mask ourselves

lionel-:
Part of #382

### DIFF
--- a/tests/testthat/test-hybrid.R
+++ b/tests/testthat/test-hybrid.R
@@ -2,7 +2,7 @@ context("hybrid")
 
 test_that("hybrid evaluation environment is cleaned up (#2358)", {
   # Can't use pipe here, f and g should have top-level parent.env()
-  df <- data_frame(x = 1)
+  df <- data_frame(`_foo` = 1)
   df <- mutate(df, f = list(function() {}))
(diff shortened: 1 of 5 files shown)


tests/testthat/test-hybrid.R
```diff
@@ -7,9 +7,9 @@ test_that("hybrid evaluation environment is cleaned up (#2358)", {
   df <- mutate(df, g = list(quo(.)))
   df <- mutate(df, h = list(~.))
 
-  expect_environments_clean(df$f[[1]])
```
krlmlr: I don't understand the effect of this change. What do the environments of `f[[1]]` etc. contain?

lionel-: Objects installed by `rlang::new_data_mask()`: a tidyeval flag and the `~` binding.

krlmlr: Can we at least check that these environments don't contain any variables (`a:h` in our case)?

lionel-: I don't think we need to do that, the mask bottom is private to rlang.

lionel-: I'll add an `env_has(inherit = TRUE)` check just to be sure.

PR #3327, opened 2018-01-26
Gather rlang API ptrs in a struct initialised with static storage

lionel-:
Includes #3325


lionel-:
@krlmlr now rebased if you'd like to take another look.

src/rlang.cpp
```diff
@@ -0,0 +1,13 @@
+#include <dplyr.h>
+
+
+namespace dplyr {
+namespace internal {
+
+rlang_api_ptrs_t rlang_api() {
```
krlmlr: Will the compiler warn if we return a reference here?

lionel-: Seems it's permitted! I made it const.

PR #3342, opened 2018-02-05
Grammar tweak

Nick-Rivera:
“Syntactic uniformity” is a singular noun, so the verb hide should be changed to hides.


krlmlr:
Thanks!

PR #3363, opened 2018-02-15
Fix rename_all() on grouped data

### LINKED ISSUE #2947: Scoped rename functions don't work on grouped_df
In dplyr 0.7.1 these work:
```
mtcars %>% rename_all(toupper)
mtcars %>% group_by(vs) %>% rename(VS = vs)
```
But these don't (throw an error `All arguments must be named`):
```
mtcars %>% group_by(vs) %>% rename_all(toupper)
mtcars %>% group_by(vs) %>% rename_if(rlang::is_integerish, toupper)
mtcars %>% group_by(vs) %>% rename_at("mpg", toupper)
```

lionel-:
Closes #2947.

@hadley currently `rename_all()` does not rename the grouping variables. This is consistent with `mutate_all()` and `summarise_all()` but it seems that in the case of names it'd make sense to operate on the grouping vars as well. What do you think?

### DIFF
--- a/tests/testthat/test-colwise-select.R
+++ b/tests/testthat/test-colwise-select.R
@@ -71,3 +71,9 @@ test_that("select_if() handles quoted predicates", {
   expect_identical(select_if(mtcars, "is_integerish"), expected)
   expect_identical(select_if(mtcars, ~is_integerish(.x)), expected)
 })
+
+test_that("rename_all() works with grouped data", {
+  df <- data.frame(a = 1, b = 2)
+  out <- df %>% group_by(a) %>% rename_all(toupper)
+  expect_identical(out, group_by(data.frame(a = 1, B = 2), a))
+})
--- a/R/colwise-select.R
+++ b/R/colwise-select.R
@@ -95,6 +95,8 @@ vars_select_syms <- function(vars, funs, tbl, strict = FALSE) {
     bad_args(".funs", "must specify a renaming function")
   }
 
-  group_syms <- base::setdiff(syms(group_vars(tbl)), syms)
+  group_vars <- group_vars(tbl)
+  group_syms <- base::setdiff(syms(group_vars), syms)
+  group_syms <- set_names(group_syms, group_vars)
   c(group_syms, syms)
 }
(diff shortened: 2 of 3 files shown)


hadley:
Can you figure out what's failing?

lionel-:
The failure is unrelated to this PR as far as I can tell.

tests/testthat/test-colwise-select.R
```diff
@@ -71,3 +71,9 @@ test_that("select_if() handles quoted predicates", {
   expect_identical(select_if(mtcars, "is_integerish"), expected)
   expect_identical(select_if(mtcars, ~is_integerish(.x)), expected)
 })
+
+test_that("rename_all() works with grouped data", {
```
krlmlr: Adding a reference to the GH issue or PR in the test title helps understanding why and when this was added (faster than blame).

tests/testthat/test-colwise-select.R
```diff
@@ -71,3 +71,9 @@ test_that("select_if() handles quoted predicates", {
...
+  df <- data.frame(a = 1, b = 2)
+  out <- df %>% group_by(a) %>% rename_all(toupper)
+  expect_identical(out, group_by(data.frame(a = 1, B = 2), a))
```
krlmlr: Do we need to update documentation to describe the behavior?

lionel-: I don't know what should be the behaviour, cf my comment in the PR description. cc @hadley

krlmlr: In the first post in this PR, you're suggesting that the new behavior is consistent with `mutate_all()` and `summarize_all()`. You added a test, too, this qualifies as defined behavior. I think we may as well document it the way it is implemented and tested now.

krlmlr:
Thanks!

tests/testthat/test-colwise-select.R
lionel-: @krlmlr I'm not sure why but I'm not systematically getting email notifications, I missed that comment. I still think that it makes sense for names operations to apply on the whole tibble and not just the grouped columns.

PR #3387, opened 2018-03-01
combine() returns logical() (closes #3365)

### LINKED ISSUE #3365: combine(NULL) should return logical() instead of error
Given that `combine()` acts like `c()` I don't understand why `combine()` (without args) does not work like `c()` returning `NULL` and raises an error instead.

If I am combining a list that may contain all `NULL` elements, I'd rather get back a `NULL` than tryCatching or checking for that condition myself in advance. It's not a big deal, but it does not seem consistent with for instance `bind_rows()` that returns an empty tibble.

If you agreed I'd be happy to provide a pull request.

``` r
c(1, NULL, 2)
#> [1] 1 2
dplyr::combine(1, NULL, 2)
#> [1] 1 2
c()
#> NULL
dplyr::combine(NULL)
#> Error in combine_all(args): no data to combine, all elements are NULL
c(NULL, NULL)
#> NULL
dplyr::combine(NULL, NULL)
#> Error in combine_all(args): no data to combine, all elements are NULL
```

As always, thanks for your time

zeehio:
As discussed on #3365 :smile: 

As always, thanks for the code review.


src/bind.cpp
```diff
@@ -399,7 +399,7 @@ List cbind_all(List dots) {
 // [[Rcpp::export]]
 SEXP combine_all(List data) {
   int nv = data.size();
-  if (nv == 0) stop("combine_all needs at least one vector");
+  if (nv == 0) return LogicalVector();
```
krlmlr: This shortcut seems unnecessary now. Since we're touching the code now and we don't need to distinguish the two cases anymore, I prefer to get rid of it. `LogicalVector()` will be returned below in line 415 for `nv == 0`.

tests/testthat/test-combine.R
```diff
@@ -1,14 +1,10 @@
...
+test_that("combine handles NULL (#1596, #3365)", {
   expect_equal(combine(list(NULL, 1, 2)), c(1, 2))
   expect_equal(combine(list(1, NULL, 2)), c(1, 2))
   expect_equal(combine(list(1, 2, NULL)), c(1, 2))
-  expect_error(
-    combine(list(NULL, NULL)),
-    "no data to combine, all elements are NULL",
-    fixed = TRUE
-  )
+  expect_equal(combine(list(NULL)), logical())
```
krlmlr: Can you please add more test cases? Empty argument list, several lists of `NULL`, ...

zeehio:
I pushed a commit that includes your comments. Please note that:

- `combine(list(NULL))` now returns `logical()` (before, it returned an error)
- `combine(list(NULL), list(NULL))` still returns `list(NULL, NULL)` as before

(Combine does not flatten the lists)

krlmlr:
Thanks. This feels pretty inconsistent, I think `combine(list(NULL))` should return the same as `combine(list(NULL), NULL)`.

zeehio:
I agree that it seems inconsistent, but this is because there are two ways of using `combine`:

The "single argument" way accepts a single argument that must be a list whose elements will be combined. The implementation is roughly equivalent to: 

`combine <- function(arg) do.call(c, arg)`

The "multiple argument" way accepts more than one argument and combines its the arguments. The implementation is roughly equivalent to `combine <- c`

`dplyr::combine` chooses the first one if there is only one argument given and if that argument is a list. Otherwise it chooses the second one. https://github.com/tidyverse/dplyr/blob/9f707d2e6f64eb1eea58500b0bd478b721de79ed/R/bind.r#L145

`combine(list(NULL))` follows the single argument approach so it is equivalent to `do.call(c, list(NULL))` and therefore `c(NULL)`, that we agreed to map to `logical(0)` for consistency with `tibble::enframe` in #3365.

`combine(list(NULL), NULL)` follows the multiple argument approach so it is like `c(list(NULL), NULL)` and therefore should return `list(NULL)`.

Having two ways of using `combine` depending on the number of arguments is quite confusing and leads to apparent inconsistencies as you mention but I believe they actually make sense.

zeehio:
Sorry, I did not mean to close the pull request. Does my explanation make sense? I am willing to contribute any change (or no change at all). I think the best option is the one in this pull request. Alternatively I would suggest to split the two ways of using dplyr::combine in two separated functions, as you prefer... Other ideas are welcome

PR #3439, opened 2018-03-16
Improve back-compatibility for _each() functions

### LINKED ISSUE #3020: result of function mutate_each_ not consisit with old versions
```r
mutate_each_(data.frame(t=30:35,x=1:6),funs(CDD=ifelse(.<33,33-.,0)),vars="t")
```

with dplyr>0.5 it yield t, x, CDD
with dplyr>0.7.2 it yield t, x, t_CDD

```r
mutate_each_(data.frame(t=30:35,x=1:6),funs(CDD=ifelse(.<33,33-.,0)),vars=c("t","x"))
```

krlmlr:
Fix column name for special case of `mutate_each()`. Not sure if it matters anymore.

Closes #3020.


krlmlr:
I think the old naming behavior for `mutate_each()` is safer: We should always use a prefix when creating a new column, even if it's just one variable. I'm thinking about situations where `vars()` depends on the data (e.g. after `spread()`) and can be = 1 or > 1. But let's wait for user feedback until we start changing anything here.

PR #3445, opened 2018-03-19
Update `select_helpers()` link in docs.

batpigandme:
* Update roxygen comments for `select_at()` to use `tidyselect::select_helpers()`.
* As mentioned in #3444.

PR #3448, opened 2018-03-19
WIP: Render pkgdown site w/ select_helpers() workaround.

### LINKED ISSUE #3444: Bad link to `select_helpers()` in NEWS?
I'm trying to rebuild the pkgdown site, and am getting `Error: 'select_helpers' is not an exported object from 'namespace:dplyr'`.

I _think_ it's occurring in trying to build `select_at()` from `colwise-mutate.R`
https://github.com/tidyverse/dplyr/blob/a4cb8f2fea524852dd8171417a2dcc38c234bb72/R/colwise-mutate.R#L10-L12

batpigandme:
* Fixes #3444 by way of adding a fake `select_helpers` object (as described in https://github.com/tidyverse/dplyr/pull/3445). 
* Updates google analytics UA.
* Re-render pkgdown site with same.

TODO: Fix display of ORCID logo (currently gigantic).


krlmlr:
Thanks, looks good. It feels like a pkgdown problem if we can't refer to valid topics for which no R object exists.

Netlify would give us PR preview and HTTPS. Any plans to migrate?

batpigandme:
Oh, are we not building these on netlify?! Or do you mean the issue? Either way 👍 ! cc @hadley

krlmlr:
I just checked, https://dplyr.tidyverse.org/ now works and shows a "Let's encrypt" certificate. That didn't work for GitHub Pages before, seems we're using Netlify. I wonder why no preview option is offered from the pull request then.

batpigandme:
Hmm…not sure. I mean, it seems like it would be overkill to use a commit hook for _every_ commit to a repo, when the pkgdown site isn't built. But, were that not the case, I'm still not sure where the netlify site for this lives, so I don't know how to add it:
https://www.netlify.com/docs/webhooks/#outgoing-webhooks

krlmlr:
Are webhooks related to deploy previews? https://www.netlify.com/blog/2016/07/20/introducing-deploy-previews-in-netlify/

batpigandme:
Yeah, that's how I ended up on that link:
> You only need to add commit hooks to your Netlify sites, we take care of the rest. We’ve also integrated Deploy Previews with other notifications, like Slack’s incoming webhooks:

batpigandme:
OK, fixed in pkgdown, so no need for the workaround. I'm gonna wait to rebuild until end-of-day to ensure I don't end up with re-re-re-rebuilds as we tidy things up! 👍

PR #3457, opened 2018-03-24
Specify `[`-type base subsetting in filter docs.

### LINKED ISSUE #3379: Documentation: dplyr::filter note about base subsetting.
From `help(filter)` (for `dplyr`):

> [dplyr::filter] Unlike base subsetting, rows where the condition evaluates to NA are dropped.

I believe `base::subset()` does drop rows where the condition is `NA`.

```r
library("dplyr")
#> 
#> Attaching package: 'dplyr'
#> 
#>     filter, lag
#> 
subset(mtcars, NA)
#>  [1] mpg  cyl  disp hp   drat wt   qsec vs   am   gear carb
filter(mtcars, NA)
#>  [1] mpg  cyl  disp hp   drat wt   qsec vs   am   gear carb
```

The `[ ]` operator does not drop with `NA`, but I think many readers will take "base subsetting" to mean `base::subset`.

batpigandme:
* Fixes #3379.

### DIFF
--- a/R/manip.r
+++ b/R/manip.r
@@ -1,7 +1,8 @@
 #' Return rows with matching conditions
 #'
 #' Use `filter()` find rows/cases where conditions are true. Unlike
-#' base subsetting, rows where the condition evaluates to `NA` are dropped.
+#' base subsetting with `[`, rows where the condition evaluates to `NA` are
+#' dropped.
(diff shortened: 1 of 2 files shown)


krlmlr:
Thanks!

PR #3461, opened 2018-03-26
 Update `case_when()` param and return descriptions.

### LINKED ISSUE #3246: case_when: issue with zero length RHS values
dplyr::case_when (tested with dplyr 0.7.4), does not seem to behave correctly in situations where a part of the RHS returns a zero length character vector (see below).

```r
dplyr::case_when(
  0 == 1 ~ sprintf("%s", c()),
  TRUE   ~ "True"
)
#> character(0)
dplyr::case_when(
  0 == 1 ~  "False",
  TRUE   ~  "True"
)
#> [1] "True"
```

batpigandme:
* Fixes #3246.

### DIFF
--- a/R/case_when.R
+++ b/R/case_when.R
@@ -13,9 +13,9 @@
 #'
 #'   These dots support [tidy dots][rlang::tidy-dots] features.
 #' @export
-#' @return A vector as long as the longest LHS or RHS, with the type (and
-#'   attributes) of the first RHS.  Inconsistent lengths or types will
-#'   generate an error.
+#' @return A vector of length 1 or n, matching the length of the logical
+#'   input vectors, with the type (and attributes) of the first RHS.
+#'   Inconsistent lengths or types will generate an error.
 #' @examples
 #' x <- 1:50
 #' case_when(
--- a/R/case_when.R
+++ b/R/case_when.R
@@ -7,9 +7,9 @@
 #'   determines which values match this case. The right hand side (RHS)
 #'   provides the replacement value.
 #'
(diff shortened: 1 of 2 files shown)


R/case_when.R
```diff
@@ -7,15 +7,15 @@
...
+#'   have a the same length. All RHSs must evaluate to the same type of
+#'   vector.
 #'
 #'   These dots support [tidy dots][rlang::tidy-dots] features.
 #' @export
-#' @return A vector as long as the longest LHS or RHS, with the type (and
-#'   attributes) of the first RHS.  Inconsistent lengths or types will
-#'   generate an error.
+#' @return A vector of length 1 or n, matching the length of the logical
+#'   input vectors, with the type (and attributes) of the first RHS.
```
krlmlr: I think it's "input *or* output vectors", maybe double-check if there's an example for this?

R/case_when.R
```diff
@@ -7,15 +7,15 @@
 #'   determines which values match this case. The right hand side (RHS)
 #'   provides the replacement value.
 #'
-#'   The LHS must evaluate to a logical vector. Each logical vector can
-#'   either have length 1 or a common length. All RHSs must evaluate to
-#'   the same type of vector.
+#'   The LHS must evaluate to a logical vector. Each logical vector must
+#'   have a the same length. All RHSs must evaluate to the same type of
```
krlmlr: "a the"? I think length 1 is also supported (for the final `TRUE`).

"the LHS" and "all RHSs" feels inconsistent.

LHS and RHS must have the same length, too.

krlmlr:
Thanks!

PR #3474, opened 2018-03-30
improved documentation for set operations - fixes #3238

### LINKED ISSUE #3238: union the exactly same two copies gives different number of rows
tested with latest tidyverse 

setequal(union(iris,iris),iris)  
union(iris,iris)  (149) has different number of rows from iris (150)! How can this be?

edublancas:
Fixes #3238.

I added explicit comments on duplicates handling, also added one example.

I am new to R package documentation, do I also need to build the documentation?

Let me know what you think.

### DIFF
--- a/R/sets.r
+++ b/R/sets.r
@@ -2,7 +2,8 @@
 #'
 #' These functions override the set functions provided in base to make them
 #' generic so that efficient versions for data frames and other tables can be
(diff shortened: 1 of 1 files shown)


krlmlr:
Thanks! I'll document and tweak a bit.

PR #3506, opened 2018-04-12
Discussion: Duplicate call object before modifying it

krlmlr:
in `firstlast_prototype()`.

This fixes the underlying problem of #3498 in a different way. I don't understand why this works, though: we are setting the `CDR` of `CDR(new_call)`, has this been duplicated too?


krlmlr:
From the observation it looks like it's duplicating the entire call, but I haven't checked further.

I don't really like modifying calls in-place, but then I haven't been able to see any observable effect (other than the evaluation failure if `nth()` isn't exported).

PR #3517, opened 2018-04-17
fix typo: with->will

jbao:


krlmlr:
Thanks. This is a generated file, but the website should contain a link to the source. Would you like to resubmit a change to the source `.R` file that contains this documentation?

jbao:
Thanks for pointing out, I overlooked this one, didn't know that generated files are also version controlled. So I'll close this PR then.

krlmlr:
Closing in favor of #3518.

PR #3564, opened 2018-05-09
Changes from 0.7.5 release candidate

krlmlr:
Same as #3563, now against `master`.

We could leave this open until it's on CRAN, but it's better to merge this first so that these changes "win" in case of conflicts with other PRs.


hadley:
Yes, these should be merged ASAP.

Issue #52, opened 2013-10-12
Support Simple Data Format data packages


rufuspollock:
http://data.okfn.org/standards/simple-data-format

Already a basic R lib listed on http://data.okfn.org/tools

PR #3643, opened 2018-06-06
Add lock bot 🔒🤖

batpigandme:
* Add lock bot that goes into effect after 365 days.

### DIFF
--- a/.github/lock.yml
+++ b/.github/lock.yml
@@ -1,7 +1,7 @@
 # Configuration for lock-threads - https://github.com/dessant/lock-threads
 
 # Number of days of inactivity before a closed issue or pull request is locked
-daysUntilLock: 365
+daysUntilLock: 180
 
 # Issues and pull requests with these labels will not be locked. Set to `[]` to disable
 exemptLabels: []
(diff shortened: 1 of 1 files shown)


.github/lock.yml
```diff
@@ -0,0 +1,28 @@
+# Configuration for lock-threads - https://github.com/dessant/lock-threads
+
+# Number of days of inactivity before a closed issue or pull request is locked
+daysUntilLock: 365
```
hadley: This seems too long to me. Maybe 6 months?

.github/lock.yml
```diff
@@ -0,0 +1,28 @@
...
+
+# Issues and pull requests with these labels will not be locked. Set to `[]` to disable
+exemptLabels: []
+
+# Label to add before locking, such as `outdated`. Set to `false` to disable
+lockLabel: false
+
+# Comment to post before locking. Set to `false` to disable
+lockComment: >
```
hadley: This will cause an initial burst of notifications as we lock thousands of old issues. Maybe we should set it to false, wait until all old issues are locked, and then re-enable it?

batpigandme: Good call.

krlmlr:
Thanks!

krlmlr:
Example: #52 hasn't been locked yet, do we need to take any further action?

batpigandme:
According to the site (https://probot.github.io/apps/lock/):
> It will start scanning for closed issues and/or pull requests within an hour!
⏲

PR #3659, opened 2018-06-09
Link recoding functions together

### LINKED ISSUE #3656: Documentation Suggestion: Suggest link between coalesce, na_if and replace_na
It'd be helpful to have a link to show that `tidyr::replace_na`, `dplyr::coalesce`, and `dplyr::na_if` have complementary functions.  Can you please add a See Also section entry linking all of them?

And it would also be useful to see `dplyr::recode` in the See Also section for `dplyr::na_if` and `tidyr::replace_na`.

(The same was just suggested as an issue in tidyverse/tidyr#461.)

billdenney:
Fix #3656


hadley:
Hmmm, I forgot that these functions are in tidyr, and you get an R CMD check NOTE if you link to a function in a package that is not listed in the dependencies.

billdenney:
Maybe just some text then to "see tidyr::XXX for some more functions that may help" rather than the direct link?

krlmlr:
Would it be too bad to add _tidyr_ to `Suggests`?

R/coalesce.R
```diff
@@ -11,6 +11,7 @@
 #' @return A vector the same length as the first `...` argument with
 #'   missing values replaced by the first non-missing value.
 #' @seealso [na_if()] to replace specified values with a `NA`.
+#' \code{\link[tidyr]{replace_na}} to replace `NA` with a value
```
krlmlr: Can you please use Markdown: `[tidyr::replace_na()]` expands to `tidyr::replace_na()` with a link.

billdenney:
@krlmlr, when I made that change, I get an error trying to build the documentation:

```r
> roxygen2::roxygenize()
Loading required package: roxygen_devtest
Error in .requirePackage(package) : 
  unable to find required package ‘roxygen_devtest’
In addition: Warning messages:
1: Version of roxygen2 last used with this package is 6.0.1.9000.  You only have version 6.0.1 
2: In library(package, lib.loc = lib.loc, character.only = TRUE, logical.return = TRUE,  :
  there is no package called ‘roxygen_devtest’
> install.packages("roxygen_devtest")
Installing package into ‘C:/Users/Bill Denney/Documents/R/win-library/3.4’
(as ‘lib’ is unspecified)
Warning in install.packages :
  package ‘roxygen_devtest’ is not available (for R version 3.4.3)
> 
```

Would you prefer that I make the change (and not update the .Rd) or not make the change?

Also, adding a dependency seems like a large change for this request.  I'm going to leave that to you and @hadley to hash out.

krlmlr:
Can you please install the dev version of _roxygen2_ and use `devtools::document()` to update the documentation?

@hadley: Okay to suggest _tidyr_ so that documentation links work?

billdenney:
@krlmlr, Installing the development version of roxygen2 and then running devtools::document() crashed R on my system.  Can you please take the documentation update from here?

krlmlr:
Thanks!

PR #3666, opened 2018-06-12
Unused variable warnings cleanup

MichaelChirico:
My installation just now gave me the following warnings:

```
filter.cpp:386:20: warning: unused variable 'data' [-Wunused-variable]
  const DataFrame& data = gdf.data() ;
                   ^
filter.cpp:500:9: warning: unused variable 'nr' [-Wunused-variable]
    int nr = indices.size();
        ^
```

I took a glance and those variables are indeed not used in the scope of the functions in which they're declared. 

I might be missing something about how `<template>`s are used in C++, however.


MichaelChirico:
Travis keeps me in check 😅

krlmlr:
Thanks for the hint, removed manually.

PR #3689, opened 2018-06-20
Make a few doc sections more explicit

kenahoo:
In general, if a Description section jumped right into exceptions or special cases without first saying what the function does, I added something briefly describing what the function does.  

The rest of what's here is mostly small clarifications or corrections.

### DIFF
--- a/R/manip.r
+++ b/R/manip.r
@@ -31,7 +31,7 @@
 #'   Multiple conditions are combined with `&`. Only rows where the
 #'   condition evaluates to `TRUE` are kept.
 #'
-#'   These arguments are automatically [quoted][rlang::quo] and
+#'   The arguments in `...` are automatically [quoted][rlang::quo] and
 #'   [evaluated][rlang::eval_tidy] in the context of the data
 #'   frame. They support [unquoting][rlang::quasiquotation] and
 #'   splicing. See `vignette("programming")` for an introduction to
@@ -64,9 +64,9 @@ filter_ <- function(.data, ..., .dots = list(), .preserve = TRUE) {
   UseMethod("filter_")
 }
 
-#' Select rows by position
+#' Choose rows by position
 #'
-#' Select rows by their ordinal position in the tbl.  Grouped tbls use
+#' Choose rows by their ordinal position in the tbl.  Grouped tbls use
 #' the ordinal position within the group.
 #'
 #' Slice does not work with relational databases because they have no
@@ -77,9 +77,10 @@ filter_ <- function(.data, ..., .dots = list(), .preserve = TRUE) {
 #' @param .data A tbl.
 #' @param ... Integer row values.  Provide either positive values to keep,
 #'   or negative values to drop. The values provided must be either all
-#'   positive or all negative.
+#'   positive or all negative.  Indices beyond the number of rows in the
+#'   input are silently ignored.
 #'
-#'   These arguments are automatically [quoted][rlang::quo] and
+#'   The arguments in `...` are automatically [quoted][rlang::quo] and
 #'   [evaluated][rlang::eval_tidy] in the context of the data
 #'   frame. They support [unquoting][rlang::quasiquotation] and
 #'   splicing. See `vignette("programming")` for an introduction to
@@ -121,10 +122,9 @@ slice_ <- function(.data, ..., .dots = list()) {
 
 #' Reduce multiple values down to a single value
 #'
-#' Create one or more scalar variables summarizing the variables of an existing tbl.
-#'
-#' When `summarise()` is used on grouped data created by [group_by()],
-#' the output will have one row for each group.
+#' Create one or more scalar variables summarizing the variables of an
+#' existing tbl. Tbls with groups created by [group_by()] will result in one
+#' row in the output for each group.  Tbls with no groups will result in one row.
 #'
 #' `summarise()` and `summarize()` are synonyms.
 #'
@@ -149,7 +149,7 @@ slice_ <- function(.data, ..., .dots = list()) {
 #'   name of the variable in the result. The value should be an expression
 #'   that returns a single value like `min(x)`, `n()`, or `sum(is.na(y))`.
 #'
-#'   These arguments are automatically [quoted][rlang::quo] and
+#'   The arguments in `...` are automatically [quoted][rlang::quo] and
 #'   [evaluated][rlang::eval_tidy] in the context of the data
 #'   frame. They support [unquoting][rlang::quasiquotation] and
 #'   splicing. See `vignette("programming")` for an introduction to
@@ -209,11 +209,12 @@ summarize_ <- summarise_
 #' Create or transform variables
 #'
 #' `mutate()` adds new variables and preserves existing ones;
-#' `transmute()` adds new variables and drops existing ones.
+#' `transmute()` adds new variables and drops existing ones.  Both
+#' functions preserve the number of rows of the input.
 #'
 #' @section Useful functions available in calculations of variables:
 #'
-#' * [`+`], [`-`], [log()], etc., for their usual arithmetic meanings
+#' * [`+`], [`-`], [log()], etc., for their usual mathematical meanings
 #'
 #' * [lead()], [lag()]
 #'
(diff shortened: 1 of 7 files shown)


kenahoo:
Are the Travis & Appveyor failures expected?  The errors look unrelated to the changes in the PR.

R/manip.r
```diff
@@ -66,16 +66,18 @@ filter_ <- function(.data, ..., .dots = list(), .preserve = TRUE) {
...
 #'
-#' Positive values select rows to keep; negative values drop rows. The
-#' values provided must be either all positive or all negative.
-#'
 #' @family single table verbs
 #' @param .data A tbl.
-#' @param ... Integer row values.
+#' @param ... Integer row values.  Provide either positive values to keep,
+#'   or negative values to drop. The values provided must be either all
+#'   positive or all negative.
```
hadley: Mention what happens if the row doesn't exist?

R/manip.r
```diff
@@ -66,16 +66,18 @@ filter_ <- function(.data, ..., .dots = list(), .preserve = TRUE) {
 
 #' Select rows by position
```
hadley: Can we use something other than select here?

R/manip.r
```diff
@@ -116,10 +119,14 @@ slice_ <- function(.data, ..., .dots = list()) {
   UseMethod("slice_")
 }
 
-#' Reduces multiple values down to a single value
+#' Reduce multiple values down to a single value
+#'
+#' Create one or more scalar variables summarizing the variables of an existing tbl.
+#'
+#' When `summarise()` is used on grouped data created by [group_by()],
```
hadley: I think this needs to go in the description. Should also describe what happens when not grouped.

R/manip.r
```diff
@@ -199,16 +206,14 @@ summarize <- summarise
 summarize_ <- summarise_
 
 
-#' Add new variables
+#' Create or transform variables
 #'
-#' `mutate()` adds new variables and preserves existing;
-#' `transmute()` drops existing variables.
+#' `mutate()` adds new variables and preserves existing ones;
```
hadley: Mention that both preserve the number of rows?

R/manip.r
```diff
@@ -199,16 +206,14 @@ summarize <- summarise
...
+#' `transmute()` adds new variables and drops existing ones.
 #'
-#' @section Useful functions:
-#'
-#' * [`+`], [`-`] etc
+#' @section Useful functions available in calculations of variables:
 #'
-#' * [log()]
+#' * [`+`], [`-`], [log()], etc., for their usual arithmetic meanings
```
hadley: Arithmetic -> mathematical?

R/manip.r
```diff
@@ -233,7 +238,7 @@ summarize_ <- summarise_
 #' @inheritParams filter
 #' @inheritSection filter Tidy data
 #' @param ... Name-value pairs of expressions. Use `NULL` to drop
-#'   a variable.
+#'   a variable in `mutate`.
```
hadley: Mention (before `NULL`) that each expression must yield a result of length 1 or the same length as the group?

R/manip.r
```diff
@@ -344,13 +349,14 @@ transmute_.grouped_df <- function(.data, ..., .dots = list()) {
...
-#' @param ... Comma separated list of unquoted variable names. Use
-#'   [desc()] to sort a variable in descending order.
+#' @param ... Comma separated list of unquoted variable names, or expressions
+#'   involving variable names. Use [desc()] to sort a variable in descending order.
 #' @family single table verbs
 #' @return An object of the same class as `.data`.
 #' @examples
 #' arrange(mtcars, cyl, disp)
 #' arrange(mtcars, desc(disp))
+#' arrange(mtcars, (disp-150)^2)
```
hadley: I don't think arranging by computed variables is useful enough to include in an example.

R/manip.r
```diff
@@ -386,12 +392,13 @@ arrange.grouped_df <- function(.data, ..., .by_group = FALSE) {
 
 #' Select/rename variables by name
 #'
+#' Choose variables from a tbl, optionally renaming them.
```
hadley: That's only `select()`, not `rename()` (which you can't use to drop variables)

R/manip.r
```diff
@@ -386,12 +392,13 @@ arrange.grouped_df <- function(.data, ..., .by_group = FALSE) {
...
 #' `select()` keeps only the variables you mention; `rename()`
 #' keeps all variables.
 #'
 #' @section Useful functions:
 #' As well as using existing functions like `:` and `c()`, there are
-#' a number of special functions that only work inside `select`
+#' a number of special functions that only work inside `select`:
```
hadley: `select` -> `select()`

R/manip.r
```diff
@@ -416,17 +423,18 @@ arrange.grouped_df <- function(.data, ..., .by_group = FALSE) {
...
-#'   Positive values select variables; negative values to drop variables.
+#'   Positive values select variables; negative values drop variables.
 #'   If the first expression is negative, `select()` will automatically
 #'   start with all variables.
 #'
-#'   Use named arguments to rename selected variables.
+#'   Use named arguments, e.g. `new_name=old_name`, to rename selected variables.
 #'
-#'   These arguments are automatically [quoted][rlang::quo] and
+#'   The arguments in `...` are automatically [quoted][rlang::quo] and
```
hadley: Should be the same in all these functions

R/manip.r
```diff
@@ -416,17 +423,18 @@ arrange.grouped_df <- function(.data, ..., .by_group = FALSE) {
...
+#'   You can treat variable names like they are positions, so you can
+#'   use expressions like `x:y` to select ranges of variables.
 #'
-#'   Positive values select variables; negative values to drop variables.
+#'   Positive values select variables; negative values drop variables.
 #'   If the first expression is negative, `select()` will automatically
 #'   start with all variables.
 #'
-#'   Use named arguments to rename selected variables.
+#'   Use named arguments, e.g. `new_name=old_name`, to rename selected variables.
```
hadley: Needs space around `=`

R/manip.r
kenahoo: Indeed - I used "choose" elsewhere to avoid ambiguity, will do the same here.

R/manip.r
kenahoo: Will do.

R/manip.r
kenahoo: Will do - please check my wording to make sure it's correct.

R/manip.r
kenahoo: Sure.

R/manip.r
kenahoo: Wasn't sure whether that behavior was set yet - looks like it silently ignores indices outside the range of data, is that what we should document?

R/manip.r
```diff
@@ -66,16 +66,18 @@ filter_ <- function(.data, ..., .dots = list(), .preserve = TRUE) {
 
 #' Select rows by position
 #'
+#' Select rows by their ordinal position in the tbl.  Grouped tbls use
```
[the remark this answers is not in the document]
kenahoo: I'm not sure how consistent the existing docs are - there's precedent for "tbl" in the docs for `distinct`, `do`, `explain`, `group_by`, etc.  There are also other terms used, e.g. "tibble" in `group_data` (which I think is just trying to be nonspecific for the various table types it accepts, but might be specifically a `tbl_df`?).  `nest_by` uses both in the same sentence.

R/manip.r
kenahoo: I do it fairly often - adding an example was my sneaky way to make sure it didn't get accidentally removed in a future revision. =)

R/manip.r
kenahoo: Will do.

R/manip.r
kenahoo: Got it.

R/manip.r
kenahoo: Will do.

R/manip.r
kenahoo: I'll rework that.

I also noticed that the RHS of `select()` expressions can't usefully be computed expressions right now:

```
> data.frame(x=1:5, y=10) %>% select(z=x)
  z
1 1
2 2
3 3
4 4
5 5
> data.frame(x=1:5, y=10) %>% select(z=2)
   z
1 10
2 10
3 10
4 10
5 10
> data.frame(x=1:5, y=10) %>% select(z=x+1)
  z10 z9
1   1 10
2   2 10
3   3 10
4   4 10
5   5 10
```

I think I would have expected it to either add 1 to the values of `x` (but that's what `mutate`/`transmute` are for), or select the column one position to the right of `x`, but it does neither.

I could add the following to the `Details` section if desired, I'll include it in a patch and you can consider it:

```
#' These functions work by column index, not value; thus, an expression
#' like `select(data.frame(x = 1:5, y = 10), z = x+1)` does not create a variable
#' with values `2:6`. (In the current implementation, the expression `z = x+1`
#' wouldn't do anything useful.)  To calculate using column values, see
#' [mutate()]/[transmute()].
```

hadley: @lionel- do we have some text somewhere (maybe in a blog post) that describes this more precisely?

lionel-: There is some information at https://dplyr.tidyverse.org/articles/dplyr.html#patterns-of-operations in the "Selecting operations" subsection.

R/manip.r
krlmlr: @hadley: Do we keep this behavior that we allow arranging by arbitrary expressions, or do we want to remove it at some point? The documentation should reflect this; if we keep, we should have a test for it too -- so far I only see `-` and `desc()` in the tests.

hadley: We want to keep it, we just don't want to show it in the examples.

krlmlr:
Thanks! I'll remove the example that arranges by an expression for now.

PR #3693, opened 2018-06-22
Fix rchk errors

krlmlr:
https://raw.githubusercontent.com/kalibera/cran-checks/master/rchk/results/dplyr.out

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -9,6 +9,8 @@
 * Using namespaced calls to `base::sort()` and `base::unique()` from C++ code
   to avoid ambiguities when these functions are overridden (#3644).
 
+* Fix rchk errors (#3693).
+
 # dplyr 0.7.5
 
 ## Breaking changes for package developers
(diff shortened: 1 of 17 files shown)


krlmlr:
Unfortunately, these errors don't show up on rhub (but a new error does). Will merge, submit to CRAN and hope that at least some of the errors disappear.

https://builder.r-hub.io/status/dplyr_0.7.6.tar.gz-649269d951c84f8e98fffaebbcec3a71

lionel-:
With the bools created outside of `create()`?

krlmlr:
A completely different set of errors is reported on rhub, this particular error doesn't show at all on rhub.

@gaborcsardi: We're seeing different output from CRAN's rchk and rhub's rchk. What could be the reason? (Thanks for providing the service!)

gaborcsardi:
Rhub's rchk is older, and I could not update it so far.

PR #3738, opened 2018-08-03
Update window functions vignette

jasonmhoule:
I fix a few minor typos.

I also updated the lead()/lag() example. The text talks about using the order_by argument, while the example uses the order_by() helper function with cumsum() (which also appears later in the vignette).


hadley:
Thanks!

PR #3752, opened 2018-08-13
count-tally doc - adding rename option

suzanbaert:
Not a typo or grammatical errors, but a suggestion to add to the documentation for add_count/tally.

The current documentation suggests to summarise yourself if you want another variable name than n, but there is an even easier option which is to call dplyr::rename() afterwards.


krlmlr:
Thanks!

PR #3798, opened 2018-09-05
add missing "not"

sowla:


sowla:
no problem! I also saw that Jenny edited the .Rd file already, but I wasn't sure why the HTML file wasn't changed.. now I understand the HTML file wasn't generated after the edit - thanks for the explanation! :)

PR #3864, opened 2018-09-29
rlang compatibility

krlmlr:
for https://github.com/r-lib/rlang/issues/211.


lionel-:
Should we add backticks from the dplyr side around all calls? It seems like that makes the message more readable and I think that's part of our conventions.

PR #3898, opened 2018-10-11
Warn when mutate_all() and transmute_all() are used on grouped tibbles

### LINKED ISSUE #3473: mutate_all() and transmute_all() for grouped data frames
should give a message that grouping variables are not touched:

```r
"Not changing grouping variables `gr1`, `gr2`. To remove this message, use `mutate_at(vars(-group_vars()))`."
```

lionel-:
Closes #3473


lionel-:
* Warn when _all variants are used on grouped tibbles.

* Split documentation of colwise mutate and summarise to make it easier to document.

* Document group semantics of colwise mutate and summarise.

The alternative tidyselect syntax is not great:

```r
mutate_at(df, vars(-c(!!!groups(df))))
```

You have to splice inside `c()` and pass the data frame manually. The former issue should be avoidable once `-` works with arbitrary functions, then you can do `-group_vars(tbl)`. I've opened an issue: https://github.com/tidyverse/tidyselect/issues/78.

But to solve the second issue, I wonder if we need to provide the group variables dynamically with a special select helper? We could add the grouping variables as attribute to the tidyselect variables to make it work.

R/colwise-mutate.R
```diff
@@ -96,9 +93,90 @@ summarize_if <- summarise_if
...
+#'   names needed to uniquely identify the output. To force inclusion of a name,
+#'   even when not needed, name the input (see examples for details).
+#' @seealso [The other scoped verbs][scoped], [vars()], [funs()]
+#'
+#' @section Grouping variables:
+#'
+#' If applied on a grouped tibble, these operations normally apply to
+#' only to the non-grouping variables.
+#'
+#' * `mutate_if()` and `transmute_if()` silently ignore the grouping
```
hadley: Why don't we use the same principle as for summary?

lionel-: I don't know, I thought this was decided from a previous discussion: https://github.com/tidyverse/dplyr/issues/3473

It sort of makes sense to me. Here is how I'd frame the problem. There are implicit selections (with _if and _all variants) and explicit selections (with _at variants). When grouping variables are found in an explicit selection, we fail in all cases. When grouping variables are found in an implicit selection, we do what's intuitive and obvious.

It seems intuitive to me that grouping variables implicitly selected are ignored in a summarise operation. Somehow it's less obvious with mutate operations.

But this does suggest `mutate_if()` and `transmute_if()` should have the same behaviour than `mutate_all()` and `transmute_all()` (either a warning or silently ignore).

Do you prefer removing the warning? This would make the semantics more consistent (this consistency will be even more apparent in the docs of the next PR I'm working on).

hadley: I think it's clear that `summarise_if()` should only apply to non-grouping variables. It's obviously less clear for `mutate_if()` and friends because we've had a number of issues from confused users. I think that suggests we need at least a message. I think a warning might be too much, and I think suggesting `vars(-c(!!!groups(df)))` is going to confuse most people. That might be better left as advice for the docs.

R/colwise-mutate.R
```diff
@@ -39,35 +41,30 @@
...
-#' by_species %>% mutate_all(funs(. / 2.54))
+#' # When that's not good enough, you can also supply the names explicitly:
+#' by_species %>% summarise_all(list(min = min, max = max))
 #'
-#' # Function names will be included if .funs has names or multiple inputs
-#' by_species %>% mutate_all(funs(inches = . / 2.54))
-#' by_species %>% summarise_all(funs(med = median))
-#' by_species %>% summarise_all(funs(Q3 = quantile), probs = 0.75)
-#' by_species %>% summarise_all(c("min", "max"))
+#' # The function is always applied on new variables when the function
```
hadley: Do you mean "The function name is always added to" ?

lionel-: The function is applied to new variables, with the function name added. When there's only one unnamed function, it is applied on the existing variables.

lionel-:
Now includes #3901. The warning has been demoted to message and now mentions the grouping variables.

The `if_` variants now issue a message as well. No alternative is suggested as there isn't any obvious one.

```r
gdf <- group_by(mtcars, cyl, am)

mutate_all(gdf, `*`, 100)
#> `mutate_all()` ignored the following grouping variables:
#> Columns `cyl`, `am`
#> Use `mutate_at(df, vars(-group_cols()), myoperation)` to silence the message.
#> # A tibble: 32 x 11

transmute_all(gdf, `*`, 100)
#> `transmute_all()` ignored the following grouping variables:
#> Columns `cyl`, `am`
#> Use `transmute_at(df, vars(-group_cols()), myoperation)` to silence the message.
#> # A tibble: 32 x 11

mutate_if(gdf, is.numeric, `*`, 100)
#> `mutate_all()` ignored the following grouping variables:
#> Columns `cyl`, `am`
#> # A tibble: 32 x 11

transmute_if(gdf, is.numeric, `*`, 100)
#> `transmute_all()` ignored the following grouping variables:
#> Columns `cyl`, `am`
#> # A tibble: 32 x 11
```

hadley:
Can you please ping me for another review once the other prs are merged?

lionel-:
@hadley Rebased on master.

R/colwise-mutate.R
```diff
@@ -96,15 +109,102 @@ summarize_if <- summarise_if
...
+#' # The scoped variants of mutate() make it easy to apply the same
+#' # transformation to multiple variables. There are three variants:
+#' # * _all affects every variable
+#' # * _at affects variables selected with a character vector or vars()
+#' # * _if affects variables selected with a predicate function:
+#'
+#' # All variants can be passed functions and additional arguments,
+#' # purrr-style. The _at() variants directly support strings. Here
+#' # we'll divide two variables by 100:
+#' starwars %>% mutate_at(c("height", "mass"), `/`, 100)
```
hadley: Using an infix operator might be a bit too hard to understand for most people

R/colwise-mutate.R
```diff
@@ -96,15 +109,102 @@ summarize_if <- summarise_if
...
+#' # returns TRUE or FALSE) to determine the relevant subset of
+#' # columns. Here we divide all the numeric columns by 100:
+#' starwars %>% mutate_if(is.numeric, `/`, 100)
+#'
+#' # mutate_if() is particularly useful for transforming variables from
+#' # one type to another
+#' iris %>% mutate_if(is.factor, as.character)
+#' iris %>% mutate_if(is.double, as.integer)
+#'
+#'
```
hadley: Maybe add a heading like:

```R
#' # Multiple transformations ----------------------------
```

R/colwise-mutate.R
```diff
@@ -96,15 +109,102 @@ summarize_if <- summarise_if
...
+#' # If you want to apply multiple transformations, pass a list of
+#' # functions. When there are multiple functions, they create new
+#' # variables instead of modifying the variables in place:
+#' iris %>% mutate_if(is.numeric, list(`/`, `*`), 100)
+#'
+#' # Note how the new variables include the function name, in order to
+#' # keep things distinct. Passing purrr-style lambdas often creates
+#' # better default names:
+#' iris %>% mutate_if(is.numeric, list(~ . / 100, mul = ~ . * 100))
```
hadley: Isn't this supplying the names explicitly? I'm not sure we need a case showing unnamed `~` style

R/colwise-mutate.R
```diff
@@ -96,15 +109,102 @@ summarize_if <- summarise_if
...
+#'
+#' # When that's not good enough, you can also supply the names explicitly:
+#' iris %>% mutate_if(is.numeric, list(div = `/`, mul = `*`), 100)
+#'
+#' # When there's only one function in the list, it modifies existing
+#' # variables in place. Give it a name to create new variables instead:
```
hadley: "to instead create" sounds slightly better to my ear

PR #4003, opened 2018-12-08
Make documentation more precise for `count()` with grouped data.

jschelbert:
See issue #3910.


jschelbert:
Mmh, seems like the Travis CI build config for R:devel is broken. Seems to have nothing to do with this PR - all other builds run smooth through...

Issue #4072, opened 2019-01-02
Break up/enhance scoped verb documentation e.g. `summarise_all()`


batpigandme:
pkgdown page: https://dplyr.tidyverse.org/reference/summarise_all.html
Scoped variants of 

- `summarise()`
- `mutate()` 
- `transmute()`

thisisnic:
I'll take this!

PR #4114, opened 2019-01-19
Use mutate_all method documentation for scoped mutate and transmute verbs

sharlagelfand:
Re: #4072, change the `mutate` and `transmute` verbs to use the already written `mutate_all` method documentation.


hadley:
Thank you @sharlagelfand!

PR #4115, opened 2019-01-19
add cumall documentation (addressing issue 4071)

### LINKED ISSUE #4071: Add examples for `cumall()`
site page: https://dplyr.tidyverse.org/reference/cumall.html
file: https://github.com/tidyverse/dplyr/blob/master/R/RcppExports.R

tonyelhabr:
Fixes #4071


src/window.cpp
```diff
@@ -10,6 +10,42 @@
...
+//' # `cumall()` and `cumany()` return logicals, not numerics/integers.
+//' cumall(x)
+//' cumany(x)
+//'
+//'
+//' # An NA only affects all elements after it for `cumall()` and `cummean()`,
+//' # but NOT for `cumany()`
+//' x <- c(1, 3, 5, NA, 2, 2)
+//'
+//' # NAs in every position after the corresponding NA position in the vector.
```
hadley: This sections feels a bit duplicative because `cumall()`, `cummean()` etc have the same behaviour wrt `NA`. Could you please condense a bit?

src/window.cpp
```diff
@@ -10,6 +10,42 @@
...
 //'   `cummean()` an integer or numeric vector
 //' @export
+//' @examples
+//' x <- c(1, 3, 5, 2, 2)
+//'
+//' # `cummean()` returns a numeric/integer vector of the same length
+//' # as the input vector.
+//' cummean(x)
+//'
+//' # `cumall()` and `cumany()` return logicals, not numerics/integers.
```
hadley: It would be nice to include a realistic example of why you might using `cumany()`/`cumall()`.  I have some vague intuition that `cumany()` is useful for finding all records after some event has occurred. Maybe you could also poll twitter for ideas? (I can retweet for a signal boost)

src/window.cpp
```diff
@@ -10,6 +10,42 @@
...
+//' cumany(x)
+//'
+//' # An NA at the beginning of the vector causes all elements after to be NA.
+//' x <- c(NA, 1, 3, 5, 2, 2)
+//'
+//' cummean(x)
+//' cumall(x)
+//' cumany(x)
+//'
+//' # NULLs are thrown out.
```
hadley: They're thrown out by `c()`, not by the `cum*()` functions, so I think you can drop these examples.

src/window.cpp
tonyelhabr: I can give this a shot. I'm hesitating to poll twitter...

tonyelhabr:
Fixes #4071 (part 2)

src/window.cpp
```diff
@@ -7,9 +7,40 @@
...
+//'
+//' # `cummean()` returns a numeric/integer vector of the same length
+//' # as the input vector.
+//' # `cumall()` and `cumany()` return logicals, not numerics/integers.
+//' x <- c(1, 3, 5, 2, 2)
+//' cummean(x)
+//' cumall(x)
+//'
+//' # An NA only affects all elements after it for `cumall()` and `cummean()`,
+//' # but NOT for `cumany()`
```
hadley: It's probably worth a note to explain why — maybe just mention it's consistent with `NA || TRUE` vs `NA && TRUE`. 

Also I think you do get `NA` propagation: `cumany(c(NA, FALSE))`

src/window.cpp
```diff
@@ -7,9 +7,40 @@
...
+//' cummean(x)
+//' cumall(x)
+//'
+//' # An NA at the beginning of the vector causes all elements after to be NA.
+//' x <- c(NA, 1, 3, 5, 2, 2)
+//' cummean(x)
+//' cumall(x)
+//'
+//' # `cumall()` vs. `cumany()`
+//' # `cumall()` will return none of the rows..
```
hadley: I'd say: `cumany()` will return all rows after the first `TRUE`; `cumall()` will return all rows until the first `FALSE`.

src/window.cpp
hadley: Examples at the end are looking good. This conversation is super helpful for forcing me to thinking about why these functions are useful - thanks!

src/window.cpp
```diff
@@ -7,9 +7,44 @@
...
+//' x <- c(1, 3, 5, 2, 2)
+//' cummean(x)
+//' cumall(x)
+//'
+//' # An `NA only affects all elements after it for `cumall()` and `cummean()`,
+//' # but NOT for `cumany()`. This behavior is consistent with that of
+//' # `NA || TRUE` compared to `NA && TRUE`.
+//' x <- c(1, 3, 5, NA, 2, 2)
+//' cummean(x)
+//' cumall(x)
```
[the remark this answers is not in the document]
tonyelhabr: noted in most recent commit (#4)  #4071 .

src/window.cpp
```diff
@@ -7,9 +7,44 @@
...
+//' # An `NA` at the beginning of the vector causes all elements after to be `NA`.
+//' x <- c(NA, 1, 3, 5, 2, 2)
+//' cummean(x)
+//' cumall(x)
+//'
+//' # `cumall()` vs. `cumany()`
+//' # `cumall()` will return all rows until the first `FALSE`.
+//' # `cumany()` will return all rows after the first `TRUE`.
+//' dplyr::filter(storms, cumall(wind > 26))
+//' dplyr::filter(storms, cumany(wind > 26))
```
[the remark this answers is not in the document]
tonyelhabr: done in most recent commit (#4)  #4071 .

tonyelhabr:
Yeah, I caught that when playing around with examples a bit more. (Hence, the change to one of the comments, "... for `cumall()` **if all prior conditions evaluate to `TRUE`**".)

hadley:
We shouldn't document incorrect behavaiour, we should fix it. I'll make a few changes to the docs and finish this PR off.

hadley:
Thank you!

PR #4357, opened 2019-05-11
Use enquos() insted of quos()

### LINKED ISSUE #4277: Update functions using `quos()` to use `enquos()`
From the community thread here: https://community.rstudio.com/t/should-i-use-quos-or-enquos/25976

The following functions use `quos()`:
`dplyr:::filter.tbl_df`
`dplyr:::arrange.tbl_df`
`dplyr:::select.data.frame`

and should be updated to use `vars()`.

krlmlr:
Closes #4277.


krlmlr:
Thanks, I wasn't aware.

We're teaching that called functions should use `enquos()`, we should be consistent.

PR #4450, opened 2019-06-29
Move responsibility of an eval_tidy() call

krlmlr:
I was wondering about the `eval_tidy()` here: https://github.com/tidyverse/dplyr/pull/4382/files#diff-80918a3b3722d3c1bb88b2739d466cf0R208

Seems that inlined functions need to do both `eval_bare()` (to construct the quo) and `eval_tidy()` (to actually evaluate it).


hadley:
Closing since we're going to tweak the implementation of all these functions.

Issue #4455, opened 2019-07-01
null vehicles in `starwars` dataset construction


dhicks:
Most characters in the Starwars API don't have associated vehicles.  In `starwars.R`, this ultimately leads to passing `NULL` values to `flatten_chr()` on line 54, which raises an error.  

The `NULL`s appear because `map("vehicles")` replaces `list()` elements (i.e., empty lists) with `NULL` elements.  There's probably a more elegant way to handle the `NULL`s, but the simplest fix I've found is to substitute `map(~.[["vehicles"]])`.  

Current version (raises an error):  
```{r}
starwars <- tibble(
  name = people %>% map_chr("name"),
  height = people %>% map_chr("height") %>% parse_integer(na = "unknown"),
  mass = people %>% map_chr("mass") %>% parse_number(na = "unknown"),
  hair_color = people %>% map_chr("hair_color") %>% parse_character(na = "n/a"),
  skin_color = people %>% map_chr("skin_color"),
  eye_color = people %>% map_chr("eye_color"),
  birth_year = people %>% map_chr("birth_year") %>% parse_number(na = "unknown"),
  gender = people %>% map_chr("gender") %>% parse_character(na = "n/a"),
  homeworld = people %>% map_chr("homeworld") %>% planets[.] %>% unname(),
  species = people %>% map("species") %>% map_chr(1, .null = NA) %>% species[.] %>% unname(),
  films = people %>% map("films") %>% map(. %>% flatten_chr() %>% films[.] %>% unname()),
  vehicles = people %>% map("vehicles") %>% map(. %>% flatten_chr() %>% vehicles[.] %>% unname()),
  starships = people %>% map("starships") %>% map(. %>% flatten_chr() %>% starships[.] %>% unname())
)
```

Working version:  
```{r}
starwars <- tibble(
  name = people %>% map_chr("name"),
  height = people %>% map_chr("height") %>% parse_integer(na = "unknown"),
  mass = people %>% map_chr("mass") %>% parse_number(na = "unknown"),
  hair_color = people %>% map_chr("hair_color") %>% parse_character(na = "n/a"),
  skin_color = people %>% map_chr("skin_color"),
  eye_color = people %>% map_chr("eye_color"),
  birth_year = people %>% map_chr("birth_year") %>% parse_number(na = "unknown"),
  gender = people %>% map_chr("gender") %>% parse_character(na = "n/a"),
[...]

PR #4457, opened 2019-07-01
Gender sex

dhicks:
PR implementing my suggested fixes for issues #4455 and #4456


hadley:
Superseded by #4480 — I think the only key difference is that droids are given sex "none", rather than asexual. Let me know if you think "asexual" is more appropriate and I'll fix.

hadley:
(Droids are the only "species" that have sex = "none")

PR #4469, opened 2019-07-07
Preserve generic chains in add_count() & add_tally()

### LINKED ISSUE #3739: Make add_count() and add_tally() generic
Most **dplyr** verbs are generic, which enables efficient reuse by subclasses of `tbl_df`. However, `add_tally()` and `add_count()`, wrappers of a grouped mutate (#2078), are not generic, and both return the result of a `grouped_df()` call: a `grouped_df` or just a `tbl_df`.

https://github.com/tidyverse/dplyr/blob/95845968725831e6a4f42fc78fb018781be952d7/R/count-tally.R#L145

As a result, there isn't currently a way to pass a `tbl_df` subclass object to these functions and retain its class. 

``` r
library(dplyr)
library(tibble)

df <- new_tibble(head(mtcars), subclass = "foo")
class(df)
#> [1] "foo"        "tbl_df"     "tbl"        "data.frame"

out <- df %>% add_count(gear)
class(out)
```

I would propose to remedy the situation by either:
1) making `add_tally()`/`add_count()` generic, enabling custom methods for subclasses; or
2) replacing the `grouped_df()` return calls with `group_by()` calls instead: then both functions would pass the input data down a chain of generic verbs, enabling indirect control over the return type _via_ `mutate`/`arrange`/`summarize`/`group_by` methods (this is what already happens with `tally`/`count`).

I have a small branch ready for a PR implementing the former [here](https://github.com/mikmart/dplyr/tree/generic-add-tally). I wouldn't mind giving a shot at implementing a different method, either. Does this sound like a reasonable feature?

mikmart:
Fixes #3739.

Only touching the data with generic verbs allows preserving subclasses for the input.

### DIFF
--- a/R/count-tally.R
+++ b/R/count-tally.R
@@ -181,7 +181,7 @@ add_tally <- function(x, wt, sort = FALSE, name = "n") {
     out <- arrange(out, desc(!!sym(n_name)))
   }
 
-  grouped_df(out, group_vars(x), drop = group_by_drop_default(x))
+  out
 }
 #' @rdname se-deprecated
 #' @export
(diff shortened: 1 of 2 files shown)


hadley:
This looks great thanks! Can you please add a bullet to the top of `NEWS.md`? It should briefly [describe the change](http://style.tidyverse.org/news.html) and end with `(@yourname, #issuenumber)`.

hadley:
Accidentally did this in passing while generally cleaning up `count()` and `tally()`

PR #4472, opened 2019-07-08
Add example in recode about NA types

### LINKED ISSUE #3768: Documentation of NA types (NA_real_, etc.) in recode function
I was reading issue #3202, which relates to the need to use the correct type of NA on the RHS of the case_when function (i.e. NA_real_, NA_character_) etc. The author indicated that they have never heard of different types of NA and I hadn't either. They opened a PR to add an example to the documentation.

I think I'm observing the same type of issue with recode and I wondered if you would like me to open a PR for the documentation for that, too? NA_character_ appears once in the current documentation but it doesn't really say anything specific about matching up the types. Just thought it would be a bit clearer for people like me.

``` r
library(dplyr)
#> 
#> Attaching package: 'dplyr'
#> 
#>     filter, lag
#> 

starwars %>%
  select(name:mass, gender, species) %>%
  mutate(
    gender = recode(gender, "male" = 1, "female" = 2, .default = NA)
  )
#> Error in mutate_impl(.data, dots): Evaluation error: `.default` must be type double, not logical.
```
Created on 2018-08-23 by the [reprex package](http://reprex.tidyverse.org) (v0.2.0).

mpaulacaldas:
Closes #3768 

I added an example and modified slightly the wording of another to make it clearer that NAs need to be of the same type as replacement values, similarly to what was done for `case_when` via https://github.com/tidyverse/dplyr/pull/3247

### DIFF
--- a/R/recode.R
+++ b/R/recode.R
@@ -56,7 +56,8 @@
 #' recode(char_vec, a = "Apple", b = "Banana")
 #'
 #' # Use .default as replacement for unmatched values. Note that NA and
-#' replacement values need to be of the same type.
+#' # replacement values need to be of the same type. For more information, see
+#' # https://adv-r.hadley.nz/vectors-chap.html#missing-values
 #' recode(char_vec, a = "Apple", b = "Banana", .default = NA_character_)
 #'
 #' # This throws an error as the NA type is of type logical, not character.
--- a/R/recode.R
+++ b/R/recode.R
@@ -60,7 +60,7 @@
 #' # https://adv-r.hadley.nz/vectors-chap.html#missing-values
 #' recode(char_vec, a = "Apple", b = "Banana", .default = NA_character_)
 #'
-#' # This throws an error as the NA type is of type logical, not character.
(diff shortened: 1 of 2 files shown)


hadley:
Maybe it would be worthwhile to add a link to some other resource that describes the different types of `NA`? There might be something in R4DS, otherwise there's this section in [adv-r](https://adv-r.hadley.nz/vectors-chap.html#missing-values).

mpaulacaldas:
I did a [quick search of R4DS](https://github.com/hadley/r4ds/search?q=NA_&unscoped_q=NA_) and there doesn't seem to be any info on  NA types, so I will use the example you mention from adv-r.

Regarding placement, should I put the link in the "Details" section or directly above one of the examples?

mpaulacaldas:
Maybe it's easier with an example.

hadley:
Thanks @mpaulacaldas!

mpaulacaldas:
Thank you @hadley 😄

Issue #3575, opened 2018-05-14
Match error message formatting to tidyverse style guide.


batpigandme:
A rolling thought, obviously, it just came up in one of the style guide examples (described https://github.com/tidyverse/style/issues/80) — particularly the separation of the follow up question as a bullet on its own line. 

https://github.com/tidyverse/dplyr/blob/4c2487a9dae88f7968315b35f64de48ae957fd58/R/tbl-df.r#L55-L62

hadley:
Fixed by #4476

PR #4476, opened 2019-07-08
Fix error message formatting for bad_eq_ops

rundel:
Addresses #3575

### DIFF
--- a/tests/testthat/test-filter.r
+++ b/tests/testthat/test-filter.r
@@ -21,15 +21,27 @@ test_that("filter gives useful error message when given incorrect input", {
   )
 })
 
+
+
 test_that("filter complains in inputs are named", {
   expect_error(
     filter(mtcars, x = 1),
-    "`x` (`x = 1`) must not be named, do you need `==`?",
+    "Filter specifications must not be named\n* Did you mean `x == 1`?",
     fixed = TRUE
   )
   expect_error(
     filter(mtcars, x = 1 & y > 2),
-    "`x` (`x = 1 & y > 2`) must not be named, do you need `==`?",
+    "Filter specifications must not be named\n* Did you mean `x == 1 & y > 2`?",
+    fixed = TRUE
+  )
+  expect_error(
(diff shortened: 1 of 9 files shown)


hadley:
Could you please add a test that uses `expect_known_output()` (plus manually printing the condition) so we have a regression test for this? (And so I can see what the new output looks like?)

rundel:
I've updated the existing test_that checks for this type of error ("filter complains in inputs are named"), they used `expect_error()` so I've just adopted that approach but can switch over to `expect_known_output()` if that is preferred.

hadley:
Yeah, I'd prefer `expect_known_output()` — I think that's a better approach for "testing" that output is useful to a human.

rundel:
We've made an attempt at using `expect_known_output()` for the filter tests. Let us know if this looks good or you have a different structure in mind.

rundel:
@beatrizmilz

tests/testthat/test-filter.r
```diff
@@ -21,16 +21,31 @@ test_that("filter gives useful error message when given incorrect input", {
...
 })
 
+
+
 test_that("filter complains in inputs are named", {
-  expect_error(
-    filter(mtcars, x = 1),
-    "`x` (`x = 1`) must not be named, do you need `==`?",
-    fixed = TRUE
+  expect_known_output(
```
hadley: Basic structure looks good, but I think it would be simpler to output all of these to the same file (with a `\n` between them)

rundel: Makes sense, I hadn't quite grokked `expect_known_output()` - looking at vctrs more carefully helped me sort things out.

rundel:
Should be good now

tests/testthat/test-filter.r
hadley: Yeah, it's a bit hokey currently — see more discussion in https://github.com/r-lib/testthat/issues/782

hadley:
Thanks @rundel and @beatrizmilz!

PR #4485, opened 2019-07-12
Add dots to generic src_tbls() arguments

ianmcook:
I maintain a package ([implyr](http://github.com/ianmcook/implyr)) that has dplyr in its Depends. In that package, I'm hoping to implement a `src_tbls()` method that takes an additional argument. But the generic `src_tbls()` does not have `...` in its arguments, so doing this produces an S3 generic/method consistency warning. Might you consider adding `...` to the arguments? Thanks!


ianmcook:
P.S. The additional argument I hope to include would allow a user to specify a specific database (schema) so that `src_tbls()` would return only the names of the tables in that particular database.

P.S. I'm aware that putting dplyr in the Imports instead of the Depends would avoid this S3 generic/method inconsistency. I know there's a prevailing opinion that Depends are to be avoided in favor of Imports. In this particular case I made a conscious decision to put dplyr in the Depends after observing users forgetting to load both implyr _and_ dplyr then struggling with the resulting error messages.

hadley:
@ianmcook would you mind adding a NEWS bullet too?

hadley:
Just after merging this, I realised that this is technically a backward incompatible change since it will require any implementors of `src_tbls()` to change their package. That's unlikely to affect many people, but it's probably worth calling out a bit more clearly in the NEWS bullet.

ianmcook:
Thanks for merging! I'll open a separate PR with a sub-bullet under the NEWS bullet to call that out.

ianmcook:
P.S. `src_tbls.src_sql()` in dplyr already takes `...` so I think dbplyr backend packages should not be affected by this change, but I will add the note to NEWS nonetheless.

https://github.com/tidyverse/dbplyr/blob/master/R/src-sql.R#L24-L26

ianmcook:
Added in #4672

PR #4500, opened 2019-07-18
Allow mutate keep original row.names

iago-pssjd:


hadley:
Given the changes in dplyr, this is now the responsibility of vctrs, but it is relatively low priority because tidyverse functions currently don't claim to preserve rownames. So I'm going to close for now, but we may come back to in the future as we improve support for rownames in vctrs.

PR #4553, opened 2019-09-06
Fix README dtplyr link

arcoutte:
small typo in Backends. Url for dtplyr redirected to dbplyr


batpigandme:
Thank you for this fix!

`README.md` is generated by `README.Rmd`. So, as is, your change would get overwritten the next time we edit the source. 

Do you want to add the change to `README.Rmd`? If not, I can push into this pull request (😱 Git terminology can be so confusing)!

hadley:
Thanks!

PR #4557, opened 2019-09-10
Correct dtplyr url

timchenn:
The url is currently pointing to dbplyr.


batpigandme:
Thanks! We've actually already got this correction in #4553, so I'm going to close this in favour of that (and get that one merged). But always appreciate the help.

PR #4581, opened 2019-10-05
use n as default wt in tally

gergness:
Fixes a bug introduced by 4408 - tally needs to differentiate between a missing `wt` and `wt = NULL`.

- `wt` is missing indicates that it should use a variable named `n` as `wt` by default (used by `tally()`)
- `wt = NULL` indicates that it should not use `n` (used by `count()`)

### DIFF
--- a/R/count-tally.R
+++ b/R/count-tally.R
@@ -72,7 +72,7 @@
 #' starwars %>%
 #'   add_count(species) %>%
 #'   filter(n == 1)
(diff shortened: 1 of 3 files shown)


szimmer:
Is this related to the following? https://community.rstudio.com/t/re-tallying-with-tally/46575

gergness:
this should fix #4645

hadley:
Thanks!

PR #4588, opened 2019-10-09
Document top_n() returns unsorted data frame

takapai:
Made it clear in the documentation that top_n doesn't return an ordered data frame, because I found that surprising. Also changed the example code to show a case where the returned results are not ordered.

Modified the top-n.R and generated top_n.Rd with roxygen2.


hadley:
Thanks!

Issue #3329, opened 2018-01-29
Add documentation about what constitutes a "level of grouping."


jarauh:
For example, when grouping by several variables and summarizing, the returned data frame is still grouped, but by one variable less.  Thus, it seems that grouping by several variables introduces several levels of grouping. This is not what I would have expected.  And this means that the grouping depends on the order of the variables in `...`. By the way, when applying `group_by_all/if/at` and `summarize`, the grouping of the result is difficult to control.

Also, the `...`-argument to `ungroup` is not explained.

krlmlr:
Thanks, agreed that the documentation can be improved here. Would you like to submit a PR that updates documentation for `summarise()` and `ungroup()`?

What do you mean by "difficult to control"? `group_by_at(vars(...))` has `select()` semantics:

``` r
library(magrittr)
suppressPackageStartupMessages(library(dplyr))
iris %>% group_by_at(vars(starts_with("Petal"))) %>% groups()
#> [[1]]
#> Petal.Length
#> 
#> [[2]]
#> Petal.Width
iris %>% group_by_at(vars(Petal.Width, starts_with("Petal"))) %>% groups()
#> [[1]]
#> Petal.Width
#> 
#> [[2]]
#> Petal.Length
```

> Created on 2018-01-30 by the [reprex package](http://reprex.tidyverse.org) (v0.1.1.9000).

hadley:
I think `summarise()` needs a new section that describes what happens to grouping variables.

Issue #3229, opened 2017-11-29
sample_frac() performance issue on grouped df 


cturbelin:
Hi !

I found a big difference of performance  using sample_frac() on grouped_df comparing dplyr 0.5.0 and 0.7.x

Running the following example using R/3.4.1 on RHEL 7 x64

version | mean | min | max
-------- | ------ | ---- | -----
dplyr 0.5.0 | 18 | 14 | 108
dplyr 0.7.4 | 2516 | 2384 | 3082

Run on the same machine, with specific user R library
dplyr 0.5.0 is using rlang 0.1.1, Rcpp 0.12.12
dplyr 0.7.4 is using rlang 0.1.4 and Rccp 0.12.13
Others packages are the same

I obtained the same kind of results on Windows (R.3.3 MRO, all the same except dplyr version)

```r
library(microbenchmark)
library(dplyr)

data = expand.grid(med=1:10, yw=1:36, tranche=1:4, id_tel=1:5)

do_calc_sample = function(data) {
        rr = data %>% group_by(yw, id_tel, tranche) %>% sample_frac(size = .6, replace=FALSE)
}

print(microbenchmark(do_calc_sample(data)))

```

The difference exists without grouped df, the same example removing grouping is about 4x slower with 0.7.4.

Best

lionel-:
Thanks, we're working on it. Capture of expressions is now fast in the dev version of rlang and I'm working on making the evaluation of quosures more efficient (which is the problem here). 

`sample_frac.grouped_df()` will also need a bit of refactoring to avoid reconstructing the overscope ngroups times.

hadley:
See also #3193

cturbelin:
Thank you ! I've just seen the pull request

krlmlr:
Closed in #3193.

PR #4602, opened 2019-10-22
Add documentation for grouping layers

jarauh:
This should fix Issue #3329.

### DIFF
--- a/R/group-by.r
+++ b/R/group-by.r
@@ -34,9 +34,11 @@
 #'
 #' @family grouping functions
 #' @param .data a tbl
-#' @param ... Variables to group by. All tbls accept variable names.
+#' @param ... For `group_by`: Variables to group by. All tbls accept variable names.
 #'   Some tbls will accept functions of variables. Duplicated groups
 #'   will be silently dropped.
+#'
+#' For `ungroup`: Further arguments passed to or from other methods.
 #' @param add When `add = FALSE`, the default, `group_by()` will
 #'   override existing groups. To add to the existing groups, use
 #'   `add = TRUE`.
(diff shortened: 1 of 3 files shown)


hadley:
Are you still interested in working on this? Just wanted to check since it's been a while since you submitted it.

jarauh:
Yes, why not.  I could spend a little time on this.

jarauh:
So the next step would be to add a section to the documentation of `summarise`?

hadley:
I just want to check you were still interested before I gave feedback.

R/group-by.r
```diff
@@ -5,6 +5,13 @@
 #' `group_by()` takes an existing tbl and converts it into a grouped tbl
 #' where operations are performed "by group". `ungroup()` removes grouping.
 #'
+#' @section Grouping layers:
```
hadley: I don't think "grouping layers" is quite the right term because I don't think we using it anywhere else. I'd just just "Grouping"

R/group-by.r
```diff
@@ -5,6 +5,13 @@
...
+#'
+#' When grouping by more than one variable, each variable defines its own
```
hadley: It's only `summarise()` that has this behaviour, so I think you can remove the language around "function such as".

I'd also recommend adding a brief explanation: i.e. after a call to summarise, the outermost group will now have only a single row, and so the grouping is no longer useful, and is hence discarded.

hadley:
Fixes #3229

hadley:
I ended up incorporating your ideas myself because it seem like other parts of the doc need more work too. Thanks for your help!

Issue #4660, opened 2019-12-13
Can we get a sort argument in top_n?


martj42:
You know, like there is in `count()`. Would save a few keystrokes.

Peace!

ergodiclife:
Hey @martj42, I made a pull request (#4664) to address this - but I'm not sure if the tidyverse team will accept my proposal. So in the meantime, try this:
```
library(dplyr)

# Add a new, 'count'-like 'sort' argument to 'top_n'
top_n <- function(x, n, wt, sort = FALSE) {
  wt <- enquo(wt)
  if (rlang::quo_is_missing(wt)) {
    vars <- tbl_vars(x)
    wt_name <- vars[length(vars)]
    rlang::inform(glue::glue("Selecting by ", wt_name))
    wt <- sym(wt_name)
  }
  out <- filter(x, dplyr:::top_n_rank({{ n }}, !!wt))
  if (sort) {
    arrange(out, desc(!!wt))
  } else {
    out
  }  
}

# Default
top_n(women, 5, height)
#  height weight
#1     68    146
#2     69    150
#3     70    154
#4     71    159
#5     72    164

# With sort = TRUE 
top_n(women, 5, height, sort = TRUE)
#  height weight
#1     72    164
#2     71    159
#3     70    154
#4     69    150
#5     68    146

```

PR #4664, opened 2019-12-15
Add a new sort argument to address #4660

ergodiclife:
Added a `dplyr::count`-like `sort` argument (`FALSE` by default) to `top_n`.

### DIFF
--- a/R/top-n.R
+++ b/R/top-n.R
@@ -52,7 +52,7 @@
 #'   top_n(1, G)
 #' }
 #' }
-top_n <- function(x, n, wt) {
+top_n <- function(x, n, wt, sort = FALSE) {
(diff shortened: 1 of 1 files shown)


hadley:
@ergodiclife thanks for this! I'm not sure I can merge or not yet, because I want to consider it along with #4494 and #4467. I'm also not sure if you need to change the direction of the sort when `n` is negative?

hadley:
I'm going in a new direction (as described in https://github.com/tidyverse/dplyr/issues/4494#issuecomment-569996736) due to the other problems with `top_n()`. Hopefully the new `slice_min()` and `slice_max()` will help solve your motivating problem, and I do deeply appreciate you taking a stab at a PR.

PR #4687, opened 2020-01-02
New family of slice helpers

hadley:
Motivated by the need to clean up a number of interface issues with `top_n()`, and the discovery that `sample_n()`/`sample_frac()` falls naturally into this new category.

@batpigandme can you please focus on the language explaining the retirement in the news and in the docs?

@lionel- can you please review the overall interface, thinking about consistency with other dplyr and tidyverse conventions?


R/compat-lifecycle.R
```diff
@@ -195,7 +195,7 @@ lifecycle_img <- function(stage, url) {
     maturing = ,
     stable = ,
     questioning = ,
-    archived =
+    retired =
```
lionel-: Probably better to update the whole file including the timestamp so we don't have multiple versions around: https://github.com/r-lib/rlang/blob/master/R/compat-lifecycle.R

R/slice.R
```diff
@@ -0,0 +1,202 @@
...
+#'
+#' # All slice helpers operate per group, silently truncating to the group
+#' # size, so the following code works without error
+#' df %>% group_by(group) %>% slice_head(n = 2)
+#'
+#' # When specifying the proportion of rows to include, non-integer sizes
+#' # are rounded down, so group a gets 0 rows
+#' df %>% group_by(group) %>% slice_head(prop = 0.5)
+slice_head <- function(.data, ..., n, prop) {
+  UseMethod("slice_head")
```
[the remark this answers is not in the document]
hadley: I think they have to be generics in order to support dbplyr, which can't easily support slice (but can support these more specialised helpers)

R/slice.R
```diff
@@ -0,0 +1,202 @@
...
+#' * `slice_head()` and `slice_tail()` select the first or last rows.
+#' * `slice_sample()` randomly selects rows.
+#' * `slice_min()` and `slice_max()` select rows with highest or lowest values
+#'   of a variable.
+#'
+#' If `.data` is a [grouped_df], the operation will be performed on each group,
+#' so that (e.g.) `slice_head(df, n = 5)` will select the first five rows in
+#' each group.
+#'
+#' @param .data A data frame or extension.
```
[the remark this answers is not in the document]
hadley: I _think_ for dplyr, it makes more sense to consistently use `.data` everywhere

R/slice.R
```diff
@@ -0,0 +1,202 @@
...
+#' @param ... Additional arguments passed on to methods.
```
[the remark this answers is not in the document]
hadley: Then why do we even use dots?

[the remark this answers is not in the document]
lionel-: One reason that is relevant to all functions is forcing named arguments. Another reason for S3 generics is that it makes it possible to introduce new arguments or to pass `...` to methods in the future.

R/slice.R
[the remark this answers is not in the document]
lionel-: I think we should choose one rule and stick to it everywhere :)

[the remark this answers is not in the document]
hadley: I think consistency with `slice()` is probably the more important principle here.

R/compat-lifecycle.R
hadley: Good point, will do.

R/slice.R
```diff
@@ -0,0 +1,202 @@
...
+#' @param n,prop Provide either `n`, the number of rows, or `prop`, the
+#'   proportion of rows to select. If `n` is greater than the number of
+#'   rows in the group (or `prop > 1`), it will be silently truncated to the
+#'   group size. If the `prop`ortion of a group size is not an integer, it will
+#'   be rounded down.
+#' @export
```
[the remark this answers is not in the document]
hadley: Tracking in https://github.com/tidyverse/dplyr/issues/4699

NEWS.md
```diff
@@ -1,5 +1,22 @@
...
+  `slice_sample()`. See `?sample_n` for details about why, and for examples
+  converting from old to new usage.
+
+* `top_n()` has been retired in favour of `slice_min()`/`slice_max()`. See
+  `?top_n` for details about why, and how to convert old to new usage (#4494).
+
+* `slice()` gains a new set of helpers:
+
+  * `slice_head()` and `slice_tail()` select the first and last rows, like
+    `head()` and `tail()`, but will work by group.
```
batpigandme: Would it be correct to say _within_ groups? Or maybe just expand "by group, returning a record for each" (or something to that effect)?

R/sample.R
```diff
@@ -1,56 +1,76 @@
 #' Sample n rows from a table
 #'
-#' This is a wrapper around [sample.int()] to make it easy to
-#' select random rows from a table. It currently only works for local
-#' tbls.
+#' @description
+#' \Sexpr[results=rd, stage=render]{dplyr:::lifecycle("retired")}
+#' `sample_n()` and `sample_frac()` have been retired in favour of
+#' [slice_sample()]. While they will not be deprecated in the near future,
+#' retirement means that we will only perform critical bug fixes so we recommend
```
batpigandme:
```suggestion
#' retirement means that we will only perform critical bug fixes, so we recommend
```

R/sample.R
```diff
@@ -1,56 +1,76 @@
...
+#' moving to the newer alternative.
 #'
-#' @param tbl tbl of data.
-#' @param size <[`tidy-select`][dplyr_tidy_select]> For `sample_n()`, the number of rows to select.
+#' These functions were retired because we realised that rather than having
```
batpigandme:
```suggestion
#' These functions were retired because we realised that, rather than having
```

R/sample.R
```diff
@@ -1,56 +1,76 @@
...
+#' two functions it was more convenient to have two mutually exclusive arguments
```
batpigandme:
```suggestion
#' two functions, it was more convenient to have two mutually exclusive arguments
```

batpigandme: This sentence could also be flipped to read ~ "it was more convenient to have two mutually exclusive arguments instead of two functions"

R/sample.R
```diff
@@ -1,56 +1,76 @@
...
+#' to a single function. This also made it to clean up a few other smaller
```
batpigandme:
```suggestion
#' to a single function. This also made it easier to clean up a few other smaller
```

batpigandme: That might not be the right word, but there _is_ a word missing there.

R/slice.R
```diff
@@ -115,3 +115,208 @@ slice.tbl_df <- function(.data, ..., .preserve = FALSE) {
...
+#' # slice_min() and slice_max() may return more rows than requested
+#' # in the presence of ties. Use with_ties = FALSE to suppress
+#' mtcars %>% slice_min(cyl, n = 1)
+#' mtcars %>% slice_min(cyl, n = 1, with_ties = FALSE)
+#'
+#' # slice_sample() allows you to random select with or without replacement
+#' mtcars %>% slice_sample(n = 5)
+#' mtcars %>% slice_sample(n = 5, replace = TRUE)
+#'
+#' # you can optional weight by a variable - this code weights by the
```
batpigandme:
```suggestion
#' # you can optionally weight by a variable - this code weights by the
```

R/slice.R
```diff
@@ -115,3 +115,208 @@ slice.tbl_df <- function(.data, ..., .preserve = FALSE) {
...
+#' df <- tibble(
+#'   group = rep(c("a", "b", "c"), c(1, 2, 4)),
+#'   x = runif(7)
+#' )
+#'
+#' # All slice helpers operate per group, silently truncating to the group
+#' # size, so the following code works without error
+#' df %>% group_by(group) %>% slice_head(n = 2)
+#'
+#' # When specifying the proportion of rows to include, non-integer sizes
```
batpigandme:
```suggestion
#' # When specifying the proportion of rows to include non-integer sizes
```

R/slice.R
```diff
@@ -115,3 +115,208 @@ slice.tbl_df <- function(.data, ..., .preserve = FALSE) {
...
+  idx <- switch(size$type,
+    n =    function(n) seq2(max(n - size$n + 1, 1), n),
+    prop = function(n) seq2(max(n - size$prop * n + 1, 1), n)
+  )
+  slice(.data, idx(dplyr::n()))
+}
+
+#' @export
+#' @rdname slice_head
+#' @param order_by Variable, or function or variables, to order by.
```
batpigandme:
```suggestion
#' @param order_by Variable, or function, or variables to order by.
```

R/top-n.R
```diff
@@ -1,57 +1,43 @@
...
 #'
-#' This is a convenient wrapper that uses [filter()] and
-#' [min_rank()] to select the top or bottom entries in each group,
-#' with respect to `wt`. Note, the resulting data frame will not be
-#' ordered or sorted.
+#' @description
+#' \Sexpr[results=rd, stage=render]{dplyr:::lifecycle("retired")}
+#' `top_n()` has been retired in favour of [slice_min()]/[slice_max()].
+#' While it will not be deprecated in the near future, retirement means
+#' that we will only perform critical bug fixes so we recommend moving to the
```
batpigandme:
```suggestion
#' that we will only perform critical bug fixes, so we recommend moving to the
```

PR #4689, opened 2020-01-03
Better documentation for generic functions

### LINKED ISSUE #4688: Better method documentation
Each generic (over `.data`) should describe the parameter in the same way, and provide a methods section that dynamically lists methods provided by currently loaded packages.

```R
#' @param .data A data frame, data frame extension (e.g. a tibble), or a lazy data frame 
#'   (e.g. from dbplyr or dtplyr).
...
#' @section Methods:
#' This function is a generic, which means that different packages can provide implementations 
#' (methods) for other classes. See the documentation of individual methods for extra arguments 
#' and differences in behaviour. 
#' 
#' Currently loaded packages provide methods for:
#' * dplyr: data.frame, tbl_df
#' * dbplyr: lazy_db
```

The bulleted list in the methods section would be generated at view-time, and would include links to the documentation if available, building off of `generics:::methods_find()`.

hadley:
Fixes #4688


hadley:
Don't forget about set ops.

PR #4712, opened 2020-01-07
Allow modification of grouping_vars

### LINKED ISSUE #4709: mutate() should allow modification of grouping variables
If any grouping variables are modified; regroup at the end

hadley:
Fixes #4709


hadley:
@lionel- is that comment for this PR? I don't see where `drop` applies

lionel-:
In r-lib/vctrs#690 I treated grouped-df created with `drop = FALSE` as standing for a static subclass. I think it'd be useful to introduce (or preserve) restrictions as soon as possible.

hadley:
I don't think we understand the semantics of statically grouped frames enough to know whether or not that should be an error yet. And implementing them will be easier after #4715, so I think we should be able to talk more about them next week.

PR #4720, opened 2020-01-08
Upgrade tidyselect

hadley:


R/select.R
```diff
@@ -1,27 +1,38 @@
...
+#' or `FALSE`) like `is.numeric`, `is.character`, and `is.factor`
+#' to select variables with specific types.
+#'
+#' Selections can be combined using Boolean algebra like:
+#'
+#' * `starts_with("a") & ends_with("x")`: start with "a" and end with "x"
+#' * `starts_with("a") | starts_with("b")`: start with "a" or "b"
+#' * `!starts_with("a")`: doesn't start with "a"
+#'
+#' To remove variable from a selection, use `-`:
```
lionel-:
```suggestion
#' To remove variables from a selection, use `-`:
```

R/select.R
```diff
@@ -99,77 +104,72 @@ rename <- function(.data, ...) {
...
 }
 
 #' @export
-select.data.frame <- function(.data, ...) {
-  # Pass via splicing to avoid matching vars_select() arguments
-  vars <- tidyselect::vars_select(tbl_vars(.data), !!!enquos(...))
-  select_impl(.data, vars)
-}
+select.grouped_df <- function(.data, ...) {
+  idx <- tidyselect::eval_select(expr(c(...)), .data)
```
lionel-: I think this should be `loc` to follow the vctrs convention.

R/select.R
```diff
@@ -99,77 +104,72 @@ rename <- function(.data, ...) {
...
-    groups <- select_impl(groups, group_vars)
+  new_grouped_df(data, groups)
+}
 
-    out <- new_grouped_df(out, groups)
-  }
+# Helpers -----------------------------------------------------------------
 
-  out
+group_idx <- function(data, idx, force_group = FALSE) {
```
lionel-: `group_loc()` following vctrs?

R/select.R
```diff
@@ -36,45 +47,39 @@
...
 #' # Select the grouping variables:
 #' starwars %>% group_by(gender) %>% select(group_cols())
 #'
+#' # Moving variables around --------------------------
+#' # Move `Species` to the front
+#' select(iris, Species, everything())
+#'
+#' # Move `Sepal.Length` to the back: first select everything except
+#' # `Sepal.Length`, then select `Sepal.Length`
+#' select(iris, -Sepal.Length, Sepal.Length)
```
lionel-: Do we still encourage unary `-`? Or should this be:

```r
select(iris, !Sepal.Length, Sepal.Length)
```

hadley: Good catch, thanks

R/select.R
hadley: Oh yeah, thanks

PR #4744, opened 2020-01-12
Restore tibble compatibility

krlmlr:
by reassigning the class attribute after `NextMethod()`.

Builds on top of #4743.

Closes #4739.


hadley:
This definitely doesn't seem correct to me — either `[.tbl_df` should drop all attributes, or it should keep all attributes.

PR #4762, opened 2020-01-16
Implement keep = T in left_join(), right_join()

ianmcook:
@hadley as discussed in #4589, this is a quick first attempt to implement `keep = TRUE` for `left_join()` and `right_join()` in the same way as `full_join()` in 1e76e8a. It also keeps the suffix on the join column from the left table when `keep = TRUE`. No new examples or tests added yet.

I'll use these examples to add tests in a later commit:
```
band_members %>% left_join(band_instruments, by = c("name"), keep = TRUE)
band_members %>% right_join(band_instruments, by = c("name"), keep = TRUE)
band_members %>% full_join(band_instruments, by = c("name"), keep = TRUE)

band_members %>% left_join(band_instruments2, by = c("name" = "artist"), keep = TRUE)
band_members %>% right_join(band_instruments2, by = c("name" = "artist"), keep = TRUE)
band_members %>% full_join(band_instruments2, by = c("name" = "artist"), keep = TRUE)
```

Two questions before I go any further:
- In `right_join()`, when `keep = FALSE` and when the join key columns do not have the same names in the right and left columns, the output really should use the join key column names from the _right_ table, not the left table—because that's where they come from. But this would be a breaking change for existing dplyr code. What are your thoughts about this?
- It would be better to refactor `join_cols()` in a symmetrical way instead of treating the left table as more special than the right table. But that's a bigger endeavor than I have time for right now. Is this something you want to do?

### DIFF
--- a/R/join.r
+++ b/R/join.r
@@ -84,7 +84,7 @@
 #' # Use a named `by` if the join variables have different names
 #' band_members %>% full_join(band_instruments2, by = c("name" = "artist"))
 #' # By default, the join keys from `x` and `y` are coalesced in the output; use
-#' # `keep = TRUE` # to keep the join keys from both `x` and `y`
+#' # `keep = TRUE` to keep the join keys from both `x` and `y`
 #' band_members %>%
 #'   full_join(band_instruments2, by = c("name" = "artist"), keep = TRUE)
 #'
--- a/R/join.r
+++ b/R/join.r
@@ -60,8 +60,9 @@
 #' @param suffix If there are non-joined duplicate variables in `x` and
 #'   `y`, these suffixes will be added to the output to disambiguate them.
 #'   Should be a character vector of length 2.
-#' @param keep Should the join keys from `y` be preserved in the output?
-#'    Only applies to `nest_join()` and `full_join()`.
+#' @param keep Should the join keys from both `x` and `y` be preserved in the
+#'   output? Only applies to `nest_join()`, `left_join()`, `right_join()`, and
+#'   `full_join()`.
 #' @param ... Other parameters passed onto methods.
 #' @param na_matches Should `NA` and `NaN` values match one another?
 #'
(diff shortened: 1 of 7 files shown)


hadley:
I think the answer to both your questions is that joins in dplyr are deliberately not symmetrical — they do their best to preserve the rows and columns of `x`. If I was to start again today, I'm 80% sure that's where I'd end up because of subtle coercion considerations (because coercion can't be 100% symmetric).

Thanks for working on this!

R/join.r
```diff
@@ -82,8 +83,8 @@
 #'
 #' # Use a named `by` if the join variables have different names
 #' band_members %>% full_join(band_instruments2, by = c("name" = "artist"))
-#' # By default only the join keys from `x` are kept; use `keep = TRUE`
-#' # to keep both
+#' # By default, the join keys from `x` and `y` are coalesced in the output; use
+#' # `keep = TRUE` # to keep the join keys from both `x` and `y`
```
hadley:
```suggestion
#' # `keep = TRUE` to keep the join keys from both `x` and `y`
```

R/join.r
```diff
@@ -324,10 +327,9 @@ join_mutate <- function(x, y, by, type,
 
   out <- as_tibble(x_out)
   out <- vec_slice(out, c(rows$x, rep_along(rows$y_extra, NA_integer_)))
-  out[names(x_key)] <- vec_cast(out[names(x_key)], vec_ptype2(x_key, y_key))
 
-  # If we're not keeping all y keys, need to copy over for the new rows
   if (!keep) {
+    out[names(x_key)] <- vec_cast(out[names(x_key)], vec_ptype2(x_key, y_key))
```
hadley: Can you explain why you moved this line? I'm fairly confident that it's correct but I'd like to talk it through just to be sure.

ianmcook: This is the counterpart to the change in involving `x_is_aux` in the definition of `join_cols()` in `join-cols.R`.

In `join-cols.R`, when `keep` is `TRUE`, the `.x` suffix is now added to the names of join columns in `x` that have the same name as join columns in `y`. (Previously the suffix was not added to those columns.)

So now here in `join_mutate()`, when `keep` is `TRUE`, the names of the join columns in `x` will not necessarily match the names of `x_key`, so this line could error in that case. And since when `keep` is `TRUE` we're keeping the separate join key columns from both tables, we don't need to cast to a common type in that case.

ianmcook:
> I think adding a couple of tests would be adequate; I think these options are sufficiently esoteric that they don't need examples.
Ok, thanks, I added some simple tests in `test-join.r`

R/join.r
hadley: Got it, thanks!

hadley:
Thanks @ianmcook!

PR #4773, opened 2020-01-17
Give more control over which variables mutate() should keep

hadley:
This is an extremely quick and hacky implementation of `morph()`, as described in #3721.


hadley:
I think I'll make this a clearly marked experiment argument of `mutate()` that only applies to data frames. If it seems successful (and isn't too hard to implement later) we can broaden it, but for now it seems reasonable to keep it as tightly scoped as possible.

I'll also add tests that use `across()` and `cur_data()` so we don't accidentally break in the future.

hadley:
Actually I think `cur_data()` will have to trigger variable usage, at least until we figure out how to make subsetting lazier.

R/data-mask.R
```diff
@@ -24,12 +24,23 @@ DataMask <- R6Class("DataMask",
...
       }
 
-      binding_fn <- function(index, chunks = resolve_chunks(index)){
-        # chunks is a promise of the list of all chunks for the column
-        # at this index, so resolve_chunks() is only called when
-        # the active binding is touched
-        function() .subset2(chunks, private$current_group)
+      if (track_usage) {
+        private$used <- rep(FALSE, ncol(data))
+        binding_fn <- function(index, chunks = resolve_chunks(index)) {
```
krlmlr: Is it safer to `force(index)` here? I remember https://github.com/tidyverse/ggplot2/pull/2807 caused by missing `force()` when returning a closure, but I haven't found an instance where this implementation breaks.

R/data-mask.R
```diff
@@ -24,12 +24,23 @@ DataMask <- R6Class("DataMask",
...
+          function() {
+            private$used[[index]] <- TRUE
+            .subset2(chunks, private$current_group)
+          }
+        }
+      } else {
+        binding_fn <- function(index, chunks = resolve_chunks(index)){
```
krlmlr: Same here?

hadley:
@mine-cetinkaya-rundel suggested another interesting variation — show the only new variables and the existing variables used to compute them. Maybe `remove = "unused"`

R/data-mask.R
hadley: Given that it works, I think we can leave as is

PR #4774, opened 2020-01-17
Control position of new mutate() columns

### LINKED ISSUE #2047: Allow mutate() to choose the position of new columns
These arguments would specify the position where the new columns are inserted, before or after some column given as index or name.

See also https://github.com/hadley/tibble/issues/99.

hadley:
Proof of concept implementation — I don't think we can make this much simpler without introducing a new dplyr generic that allows you to modify columns and change their order at the same time. I think that's ok though, since `group_by()` (which sometime called `mutate()`) has something of the same flavour.

Will need some refactoring (probably pull out and export `relocate_cols()`) to make this easier for other backends to use.

Fixes #2047.


R/mutate.R
```diff
@@ -142,9 +144,18 @@ mutate <- function(.data, ...) {
 }
 
 #' @export
-mutate.data.frame <- function(.data, ...) {
+mutate.data.frame <- function(.data, ..., .before = NULL, .after = NULL) {
   cols <- mutate_cols(.data, ...)
-  dplyr_col_modify(.data, cols)
+  out <- dplyr_col_modify(.data, cols)
+
+  if (!quo_is_null(enquo(.before)) || !quo_is_null(enquo(.after))) {
```
lionel-: `.before` and `.after` are defused twice, once with `enquo()` and once with `{{`. This will cause repeated side effects via unquoting.

To fix this, the defused arguments should be saved in variables and unquoted later on with `!!`.

R/mutate.R
```diff
@@ -142,9 +144,18 @@ mutate <- function(.data, ...) {
...
+    # Only change the order of new columns
+    new <- setdiff(names(cols), names(data))
+    relocate(out, any_of(new), .before = {{ .before }}, .after = {{ .after }})
```
krlmlr: It looks like we can combine `.before` with `.remove = "used"` if we update `out` before applying the `.remove` logic.

```suggestion
    out <- relocate(out, any_of(new), .before = {{ .before }}, .after = {{ .after }})
    out
```

R/relocate.R
```diff
@@ -40,20 +40,23 @@
...
 
-  if (!missing(.before) && !missing(.after)) {
+  has_before <- !quo_is_null(enquo(.before))
+  has_after <- !quo_is_null(enquo(.after))
+
+  if (has_before && has_after) {
     abort("Must supply only one of `.before` and `.after`")
-  } else if (!missing(.before) && missing(.after)) {
+  } else if (has_before) {
     where <- tidyselect::eval_select(enexpr(.before), .data)
```
krlmlr: Do we need `enquo(.before)` here, and similarly below?

R/mutate.R
```diff
@@ -57,7 +57,9 @@
 #'     if ungrouped).
 #'   * `NULL`, to remove the column.
 #'   * A data frame or tibble, to create multiple columns in the output.
-#'
+#' @param .before,.after <[`tidy-select`][dplyr_tidy_select]> Optionally,
+#'   control where new columns should be appear (the default is to add to the
+#'   left hand side).
```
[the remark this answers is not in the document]
krlmlr: I wonder if `.after = last_col()` should be the (implicit?) default for both `mutate()` and `relocate()`.

Currently this is the implicit default for `mutate()`, but for `relocate()` the implicit default seems to be `.before = 1L`. New users of `relocate()` might expect `mutate()` semantics.

R/relocate.R
hadley: Yup, thanks for spotting

R/mutate.R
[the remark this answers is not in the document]
hadley: I think `relocate()` is sufficiently different from `mutate()` that it shouldn't cause a slash.

R/mutate.R
hadley: Will do — handling missing values + unquoting feels clunky to me at the moment; I think we should think about better tooling at some point.

R/mutate.R
hadley: Oh yeah, makes sense.

hadley:
All comments resolved, but I'd appreciate another look through since this interacts with the new `.keep` argument.

R/mutate.R
```diff
@@ -108,19 +107,20 @@
...
+#' df %>% mutate(z = x + y, .before = 1)
+#' df %>% mutate(z = x + y, .after = x)
+#'
+#' # By default, mutate() keeps all columns from the input data.
+#' # Experimental: You can override with `.keep`
+#' df <- tibble(x = 1, y = 2, a = "a", b = "b")
+#' df %>% mutate(z = x + y, .keep = "all") # the default
+#' df %>% mutate(z = x + y, .keep = "used")
+#' df %>% mutate(z = x + y, .keep = "unused")
+#' df %>% mutate(z = x + y, .keep = "none")
```
krlmlr:
```suggestion
#' df %>% mutate(z = x + y, .keep = "none") # same as transmute()
```

R/relocate.R
```diff
@@ -40,21 +40,26 @@
...
+  } else if (has_before) {
+    where <- tidyselect::eval_select(.before, .data)
     to_move <- c(setdiff(to_move, where), where)
-  } else if (missing(.before) && !missing(.after)) {
-    where <- tidyselect::eval_select(enquo(.after), .data)
+  } else if (has_after) {
+    where <- tidyselect::eval_select(.after, .data)
     to_move <- c(where, setdiff(to_move, where))
   } else {
     where <- 1L
```
krlmlr: I'm still not sure we should use different defaults for `relocate()` and `mutate()` . Should I open a new issue?

hadley: I don't think this is going to cause confusion because the fundamental purposes of the functions are different.

PR #4816, opened 2020-01-31
Proofread and update documentation for *_join functions

### LINKED ISSUE #4787: Revamp `*_join()` docs

davisadamw:
Fixes #4787

### DIFF
--- a/R/join.r
+++ b/R/join.r
@@ -9,7 +9,7 @@
 #' * `right_join()`: includes all rows in `y`.
 #' * `full_join()`: includes all rows in `x` or `y`.
 #'
-#' If there are multiple matches between `x` and `y`, all combination of the
+#' If there are multiple matches between `x` and `y`, all combinations of the
 #' matches are returned.
 #'
 #' @return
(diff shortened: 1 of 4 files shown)


R/join.r
```diff
@@ -9,27 +9,27 @@
 #' * `right_join()`: includes all rows in `y`.
 #' * `full_join()`: includes all rows in `x` or `y`.
 #'
-#' If there are multiple matches between `x` and `y`, all combination of the
+#' If there are multiple matches between rows in `x` and `y`, all combinations of the
```
hadley: I recently realised it's slightly simpler to say: If a row in `x` matches multiple rows in `y`, all the rows in `y` will be returned. Does that make sense?

davisadamw: That is simpler and describes the behavior accurately. I think it's still helpful to make it clear that the number of rows returned from each multiple match is [matches from X] * [matches from Y] that set of match, so maybe something like: If a row in `x` matches multiple rows in `y`, all the rows in `y` will be returned once for each matching row in `x`

hadley: Sounds good!

davisadamw: Great, thanks!

PR #4818, opened 2020-01-31
Update lead + lag documentation

### LINKED ISSUE #4798: Refresh lead() + lag() docs

laurakwiley:
Fixes #4798


R/lead-lag.R
```diff
@@ -1,33 +1,33 @@
...
-#' lag(1:10, 1)
 #' lead(1:10, 1)
+#' lag(1:10, 1)
 #'
 #' x <- runif(5)
 #' cbind(ahead = lead(x), x, behind = lag(x))
 #'
-#' # Use order_by if data not already ordered
+#' # If data are not already ordered, use order_by
 #' df <- data.frame(year = 2000:2005, value = (0:5) ^ 2)
```
krlmlr: Should we use the pipe in the examples? We can use `sample_frac()` for scrambling the data.

R/lead-lag.R
```diff
@@ -1,33 +1,33 @@
...
 #' @examples
 #' lead(1:10, 1)
 #' lead(1:10, 2)
 #'
-#' lag(1:10, 1)
 #' lead(1:10, 1)
+#' lag(1:10, 1)
 #'
 #' x <- runif(5)
 #' cbind(ahead = lead(x), x, behind = lag(x))
```
krlmlr: While we're at it:

```suggestion
#' tibble(ahead = lead(x), x, behind = lag(x))
```

R/lead-lag.R
laurakwiley: Great idea! Will also make consistent with other suggestion and changing to tibble.

hadley:
I made a couple small tweaks — the most important is that I kept the .Rd name the same to avoid breaking existing links.

PR #4823, opened 2020-01-31
Refresh select() docs

### LINKED ISSUE #4793: Refresh `select()` docs

DanielleQuinn:
Fixes #4793 including proofreading, updated examples, improved datasets, etc.

### DIFF
--- a/R/select.R
+++ b/R/select.R
@@ -67,7 +67,7 @@
 #' select(starwars, is.numeric)
 #'
 #' # Optionally, rename individual variables as they are selected,
-#' # in the format "new_name" = "old_name"
+#' # in the format `new_name = old_name`
 #' select(starwars, character_name = name, character_height = height)
 #'
 #' # Use num_range() to select variables with numeric suffixes
--- a/R/select.R
+++ b/R/select.R
@@ -60,15 +60,15 @@
 #' @family single table verbs
 #' @export
 #' @examples
-#' iris <- as_tibble(iris) # Convert to simplify appearance of object when printed
-#' select(iris, starts_with("Petal"))
-#' select(iris, ends_with("Width"))
-#' select(iris, !starts_with("Petal"))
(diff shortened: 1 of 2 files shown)


R/select.R
```diff
@@ -18,38 +18,38 @@
...
+#'   are positions, so expressions like `x:y` can be used to select a range of
 #'   variables.
 #' @return
 #' An object of the same type as `.data`.
 #' * Rows are not affected.
 #' * Output columns are a subset of input columns, potentially with a different
 #'   order. Columns will be renamed if `new_name = old_name` form is used.
 #' * Data frame attributes are preserved.
-#' * Groups are maintained; you can not select off grouping variables.
+#' * Groups are maintained.
```
lionel-: It might be better to keep the redundant part of the sentence, for clarity.

R/select.R
```diff
@@ -60,28 +60,36 @@
...
+#' starwars %>% group_by(gender, eye_color) %>% select(group_cols())
 #'
-#' # Moving variables around --------------------------
+#' # Using select() semantics in across()
+#' starwars %>% summarise(across(cols = height:mass, fns = ~mean(.x, na.rm = TRUE)))
+#'
+#' # Applying tidy eval to select()
+#' # See dplyr::tidyeval for more information
+#' mycol <- c("height", "mass")
+#' starwars %>% select({{mycol}})
```
lionel-: I think this needs to be an example with a function. Using `{{` on objects that are not function arguments works for technical reasons, but is not the intended usage.

Perhaps something like this?

```r
averages <- function(data, vars) {
  data %>%
    select({{ vars }}) %>%
    lapply(mean, na.rm = TRUE)
}
starwars %>% averages(height)
starwars %>% averages(c(height, mass))
```

lionel-: Can you also mention `?dplyr_tidy_eval` for more information please?

lionel-:
Thanks!

PR #4829, opened 2020-02-01
Create group by vignette

### LINKED ISSUE #4779: Grouped data vignette
We need a vignette that broadly explains how grouping works, and shows how grouping and the various verbs interact.

* `use_vignette("grouping.Rmd")`
* Copy in existing content from https://dplyr.tidyverse.org/articles/dplyr.html#grouped-operations (which scheduled for removal)
* Ensure each one table verb has a section with a small example that demonstrates how grouping interacts with its operation
    * un-`select()` grouping vars
    * `arrange()` not [grouped by default](https://github.com/tidyverse/dplyr/issues/3782)
    * show example of filter to remove entire groups, e.g. `filter(df, n() == 1)`
* Should try and work with one dataset (`starwars`?) as much as possible.
* How to access data about the "current" group with `cur_group()` and friends
* Briefly show the underlying data structure accessed with `group_data()`

This is a lot to do in a day, so if you don't get it all done, please submit a PR anyway, as your work will make it much easier for us to finish off.

MikeKSmith:
Fixes #4779

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -97,6 +97,9 @@
 
 ## Grouping
 
+* New `vignette("grouping")` gives more details about how dplyr verbs change
+  when applied to grouped data frames (#4779, @MikeKSmith).
+
 * Grouped data frames now have `names<-`, `[[<-`, `[<-` and `$<-` methods that
   re-generate the underlying grouping. Note that modifying grouping variables 
   in multiple steps (i.e. `df$grp1 <- 1; df$grp2 <- 1`) will be inefficient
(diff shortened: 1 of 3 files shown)


hadley:
Thanks @MikeKSmith! Will be a little while before I get a chance to review because I'm going on vacation.

MikeKSmith:
Enjoy a very well deserved break...

hadley:
I think I've touched just about every line you wrote, but I want to emphasise how useful your contribution was — most of the big ideas are still there (albeit in a different order) and just having some text to "react against" was super useful. Thanks so much for working on this!

PR #4833, opened 2020-02-04
Refresh `bind()` docs #4794 partial refresh

estern95:
As a user, I find the following examples unclear/could use some rethinking:

1)

`#' # Note that for historical reasons, lists containing vectors are
#' # always treated as data frames. Thus their vectors are treated as
#' # columns rather than rows, and their inner names are ignored:
#' ll <- list(
#'   a = c(A = 1, B = 2),
#'   b = c(A = 3, B = 4)
#' )
#' bind_rows(ll)
#'
#' # You can circumvent that behaviour with explicit splicing:
#' bind_rows(!!!ll)`

I am not fully clear on the naming convention in this example `ll`, also it looks like the code has changed and the results of `bind_rows(ll)` and `bind_rows(!!!ll)` return the same result.

2)

This no longer breaks since single length vectors are repeated over the length. I adjusted it in a pull request.

`#' \dontrun{
#' # Rows do need to match when column-binding
#' bind_cols(data.frame(x = 1), data.frame(y = 1:2))
#' }`


hadley:
Thanks @estern95 — that looks great!

Issue #4826, opened 2020-01-31
slice_max and slice_min do not work if variable contains NAs


JaneWall:
whereas the following works

---

When the variable being sorted on contains an NA, slice_max and slice_min give an unintelligent error.  The programs should instead choose the max or min with na removed.

```r
library(nycflights13)
slice_max(flights, arr_delay, n = 3) 

#but this works
 flights %>%
    filter(!is.na(dep_delay)) %>%
    slice_max(dep_delay, n = 3)
```

courtiol:
I agree there could be an argument `with_na` (similar to `with_ties`) in `slice_min/max`, but you can always do:
```
flights %>%
  slice_max(na.omit(arr_delay), n = 3) 
```

wfjvdham:
Using `na.omit()` gives a different result than using the `filter()` as you can see here:

``` r
library(nycflights13)
library(dplyr)
#> 
#> Attaching package: 'dplyr'
#> The following objects are masked from 'package:stats':
#> 
#>     filter, lag
#> The following objects are masked from 'package:base':
#> 
#>     intersect, setdiff, setequal, union

flights %>%
  slice_max(na.omit(dep_delay), n = 3) 
#> # A tibble: 3 x 19
#>    year month   day dep_time sched_dep_time dep_delay arr_time sched_arr_time
#>   <int> <int> <int>    <int>          <int>     <dbl>    <int>          <int>
#> 1  2013     1     9      607            608        -1      741            755
#> 2  2013     6     9     1543           1555       -12     1706           1715
#> 3  2013     1    10     1016           1020        -4     1319           1330
#> # … with 11 more variables: arr_delay <dbl>, carrier <chr>, flight <int>,
#> #   tailnum <chr>, origin <chr>, dest <chr>, air_time <dbl>, distance <dbl>,
#> #   hour <dbl>, minute <dbl>, time_hour <dttm>

flights %>%
  filter(!is.na(dep_delay)) %>%
  slice_max(dep_delay, n = 3)
#> # A tibble: 3 x 19
#>    year month   day dep_time sched_dep_time dep_delay arr_time sched_arr_time
#>   <int> <int> <int>    <int>          <int>     <dbl>    <int>          <int>
#> 1  2013     1     9      641            900      1301     1242           1530
#> 2  2013     6    15     1432           1935      1137     1607           2120
#> 3  2013     1    10     1121           1635      1126     1239           1810
[...]

PR #4857, opened 2020-02-12
Prevent slice_min/max from crashing when there are NA's

wfjvdham:
Removes `NA`'s from the columns.

Closes https://github.com/tidyverse/dplyr/issues/4826

### DIFF
--- a/R/slice.R
+++ b/R/slice.R
@@ -170,8 +170,8 @@ slice_min.data.frame <- function(.data, order_by, ..., n, prop, with_ties = TRUE
   size <- check_slice_size(n, prop)
   if (with_ties) {
     idx <- switch(size$type,
-      n =    function(x, n) head(order(x), sum(min_rank(x) <= size$n, na.rm = TRUE)),
-      prop = function(x, n) head(order(x), sum(min_rank(x) <= size$prop * n, na.rm = TRUE))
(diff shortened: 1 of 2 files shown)


R/slice.R
```diff
@@ -170,8 +170,8 @@ slice_min.data.frame <- function(.data, order_by, ..., n, prop, with_ties = TRUE
   size <- check_slice_size(n, prop)
   if (with_ties) {
     idx <- switch(size$type,
-      n =    function(x, n) head(order(x), sum(min_rank(x) <= size$n)),
-      prop = function(x, n) head(order(x), sum(min_rank(x) <= size$prop * n)),
+      n =    function(x, n) head(order(x), sum(min_rank(x) <= size$n, na.rm = TRUE)),
```
hadley: I think at this point it's probably worth extracting out the repeated code into a function like:

```R
something <- function(x, y) {
  sum(min_rank(x) < y, na.rm = TRUE)
}
```

But I'm not sure what to call it

R/slice.R
```diff
@@ -314,3 +318,7 @@ sample_int <- function(n, size, replace = FALSE, wt = NULL) {
   }
 }
 
+nr_ranks_lesser_or_equal_to <- function(x, y) {
```
hadley: How about `smaller_ranks()`?

hadley:
Thanks @wfjvdham!

Issue #4803, opened 2020-01-25
Request: Inside `dplyr:::tbl_if_vars()`, use `pull()` instead of base column indexing


bschneidr:
This is a very minor request and based on my tests shouldn't produce any noticeable changes in any functionality of `dplyr`. However, it would be [helpful for the `srvyr` package](https://github.com/gergness/srvyr/issues/67) and could potentially  be helpful for other packages which are centered around objects inheriting the `tbl_df` class.

Inside the `tbl_if_vars()` function, there's a line of code which checks the contents of a column using base R column indexing. I'd like to replace that instance of column indexing with a call to `dplyr::pull()`. 

Here's the current code:

https://github.com/tidyverse/dplyr/blob/dbb5acc5af2097afa2b43650d86f2198b5136c75/R/colwise.R#L220

And here's the proposed change (a single line):

```r
    column <- pull(.tbl, tibble_vars[[i]])
```

This produces the same result for regular tibbles, except it also works in the case of special objects such as `tbl_svy`s from the srvyr package where the concept of base column indexing doesn't make sense but the use of `pull()` does. 

If this seems reasonable, I can go ahead and submit a pull request for review from this branch in a fork of the repo:

https://github.com/bschneidr/dplyr/tree/use-pull-in-filter-if-vars

hadley:
A PR would be useful thanks. But please note we’re moving away from these functions in favour of across()

bschneidr:
Thanks for considering this, Hadley, and thank you for the heads-up on the move towards `across()`.

PR #4879, opened 2020-02-22
Inside `dplyr:::tbl_if_vars()`, use `pull()` instead of base column indexing

bschneidr:
This addresses the request in #4803.


hadley:
For future reference, if you include `Fixes #xyz` in a commit message (or the text of the PR), the corresponding issue will be automatically closed when the PR is merged.

PR #4921, opened 2020-03-02
Restore compatibility with CRAN tibble

krlmlr:
Most issues arise because `[<-.data.frame()` doesn't support all corner cases that we do. Merging thisallows us to isolate "our" downstream failures from tibble's.

### DIFF
--- a/R/generics.R
+++ b/R/generics.R
@@ -102,8 +102,32 @@ dplyr_col_modify <- function(data, cols) {
 
 #' @export
 dplyr_col_modify.data.frame <- function(data, cols) {
-  data[names(cols)] <- cols
-  data
+  # Work around corner cases with CRAN tibble:
+  # - Names of zero-length vectors
+  if (length(cols) == 0) return(data)
+
+  # - Assigning a scalar to a zero-row data frame
+  if (nrow(data) == 0) {
+    cols[] <- map(cols, vec_slice, 0)
+  }
+
+  # - Issues with UTF-8 names
+  names(data) <- as_utf8_character(names2(data))
+  names(cols) <- as_utf8_character(names2(cols))
+
+  # - Removal of inner names by [<-.data.frame
+  cols_null <- map_lgl(cols, is.null)
+  data_list <- unclass(data)
+  data_list[names2(cols)[!cols_null]] <- cols[!cols_null]
+  data_list[names(data_list) %in% names2(cols)[cols_null]] <- NULL
+  class(data_list) <- class(data)
+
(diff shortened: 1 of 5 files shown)


R/generics.R
```diff
@@ -102,8 +102,32 @@ dplyr_col_modify <- function(data, cols) {
 
 #' @export
 dplyr_col_modify.data.frame <- function(data, cols) {
-  data[names(cols)] <- cols
-  data
+  # Work around corner cases with CRAN tibble:
```
hadley: Does `vec_slice<-` happen to work here?

krlmlr: No, perhaps because `cols` can be a list?

hadley: Oh of course, `vec_slice()` is the wrong direction.

R/generics.R
```diff
@@ -102,8 +102,31 @@ dplyr_col_modify <- function(data, cols) {
...
+
+  # - Assigning a scalar to a zero-row data frame
+  if (nrow(data) == 0) {
+    cols[] <- map(cols, vec_slice, 0)
+  }
+
+  # - Issues with UTF-8 names
+  names(data) <- as_utf8_character(names2(data))
+
+  # - Removal of inner names by [<-.data.frame
```
hadley: Can you please explain this in a bit more detail?

krlmlr:
``` r
x <- data.frame(a = 1:3)
b <- setNames(2:4, letters[2:4])
b
#> b c d 
#> 2 3 4
x$b <- b
x
#>   a b
#> 1 1 2
#> 2 2 3
#> 3 3 4
x$b
#> [1] 2 3 4
```

<sup>Created on 2020-03-03 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

I haven't found a way to keep the inner names.

R/generics.R
```diff
@@ -102,8 +102,31 @@ dplyr_col_modify <- function(data, cols) {
 
 #' @export
 dplyr_col_modify.data.frame <- function(data, cols) {
-  data[names(cols)] <- cols
-  data
+  # Work around corner cases with CRAN tibble:
```
hadley: This makes me wonder if there are currently a bunch of subtle bugs that we're not testing because we're using `[<-.data.frame` in a bunch of cases. It might be safer to just reimplement the bare minimum of `[<-.tibble` here, rather than attempting to patch up an existing method.

hadley: I'm working on a PR that implements this approach so we can talk about it.

hadley:
See alternative approach in #4937

PR #4923, opened 2020-03-03
New rename_with()

### LINKED ISSUE #4771: Need equivalent of scoped select/rename helpers
Since `across()` doesn't help with this problem. 

Maybe `select()` and `rename()` could gain a `.transform` argument that would be applied to the variable names?

hadley:
Fixes #4771

I think this interface is correct despite having `.fn` before `.col` — most of the time I think you'll be renaming all columns, assuming that the renaming function will handle columns that don't fit the pattern. 

I like the use of `_with` to evoke transformation with a function.


hadley:
@atusy how does this look to you?

atusy:
I upvote `rename_with`, and totally agree `.fn` before `.col` would be useful because

>  renaming function will handle columns that don't fit the pattern

I also like the naming because `rename with function` makes sense as a sentence.

R/rename.R
```diff
@@ -39,3 +48,26 @@ rename.data.frame <- function(.data, ...) {
...
+#'   defaults to all columns.
+rename_with <- function(.data, .fn, .cols = everything(), ...) {
+  UseMethod("rename_with")
+}
+
+
+#' @export
+rename_with.data.frame <- function(.data, .fn, .cols = everything(), ...) {
+  .fn <- as_function(.fn)
+  cols <- tidyselect::eval_select(enexpr(.cols), .data)
```
lionel-: Defusing an argument should always be done with `enquo()`.

R/rename.R
```diff
@@ -39,3 +48,26 @@ rename.data.frame <- function(.data, ...) {
...
+
+  names <- names(.data)
+  names[cols] <- .fn(names[cols], ...)
```
lionel-: Needs name-repair validation as `.fn` might be generating non-unique names, or names conflicting with existing names that were not selected for renaming..

tests/testthat/test-rename.R
```diff
@@ -34,3 +34,18 @@ test_that("can rename with duplicate columns", {
   df <- tibble(x = 1, x = 2, y = 1, .name_repair = "minimal")
   expect_named(df %>% rename(x2 = 2), c("x", "x2", "y"))
 })
+
+# rename_with -------------------------------------------------------------
+
+test_that("can select columns", {
```
lionel-: Needs a test for the renaming case, e.g. `c(foo = cyl)`

R/rename.R
hadley: Except for the `expr(c(...))` which is what always throws me off.

lionel-: This is the common pattern though, defuse your own expression with `expr()`, and the user expression with `enquo()`.

tests/testthat/test-rename.R
hadley: I think this function shouldn't allow renaming: https://github.com/r-lib/tidyselect/issues/178. We can fix after the eventual next release of tidyselect.

R/rename.R
hadley: Good point, I'll fix.

Random thought: should `set_names()` get a `.name_repair` argument?

PR #4932, opened 2020-03-05
Add section on breaking changes

krlmlr:
to NEWS.

I'm aware of the "Lifecycle changes" section. The new text highlights behavior changes that break tests that worked in dplyr 0.8.4.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -2,9 +2,11 @@
 
 ## Breaking changes
 
-* `bind_cols()` no longer forces tibbles.
+* `bind_cols()` no longer converts to a tibble, returns a data frame if the input is a data frame.
 
 * `bind_rows()` and `combine()` use vctrs coercion rules.
+
+    * Combining factor and character creates a character without warning, combining factors creates a factor with levels combined.
 
     * For time columns, the time zone of the first argument is used.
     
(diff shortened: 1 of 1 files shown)


NEWS.md
```diff
@@ -1,5 +1,30 @@
 # dplyr 1.0.0 (in development)
 
+## Breaking changes
+
+* `bind_cols()` no longer forces tibbles.
```
hadley: What do you mean by forces?

NEWS.md
```diff
@@ -1,5 +1,30 @@
...
+
+* `bind_rows()` and `combine()` use vctrs coercion rules.
+
+    * For time columns, the time zone of the first argument is used.
+
```
hadley: Need comment about factor + factor and factor + character here

NEWS.md
```diff
@@ -1,5 +1,30 @@
...
+* `bind_rows()` and `combine()` use vctrs coercion rules.
+
+    * For time columns, the time zone of the first argument is used.
+    
+    * Classed atomic vectors are no longer accepted, `vctrs::vec_is()` is required for all inputs.
+
+* `bind_rows()` and other functions use vctrs name repair, see `?vctrs::vec_as_names`.
+
+* `count()` no longer removes the last level of grouping.
```
hadley: Can you file an issue? I thought `count()` always preserved the previous grouping, so this is a mistake I introduced and we can fix it.

NEWS.md
```diff
@@ -1,5 +1,30 @@
...
+
+* `all.equal.tbl_df()` removed, equality checks for data frames no longer ignore row order or groupings.
```
hadley: Need to comment that this was used by `expect_equal()`; I think it also means that `expect_equal( tibble(x=1), data.frame(x=1))` is now false, but you'd need to check that it was true before.

NEWS.md
krlmlr: This was a false negative -- the test was incorrect, which was revealed by the changes in `all.equal.tbl_df()` .

hadley:
Thanks!

PR #4937, opened 2020-03-05
Use CRAN tibble

hadley:
@krlmlr what do you think are the potential risks of this approach?

@DavisVaughan do you seen any likely performance issues here? (e.g. am I doing the attribute munging performantly)


tests/testthat/test-group-by.r
```diff
@@ -102,8 +102,8 @@ test_that("group_by orders by groups. #242", {
 })
 
 test_that("Can group_by() a POSIXlt", {
-  df <- data.frame(times = 1:5, x = 1:5)
-  df$times <- as.POSIXlt(seq.Date(Sys.Date(), length.out = 5, by = "day"))
+  skip_if_not_installed("tibble", "2.99.99")
```
krlmlr: Do we need to change this? Looks like the old version still might work.

R/generics.R
```diff
@@ -102,8 +102,27 @@ dplyr_col_modify <- function(data, cols) {
...
+  out <- vec_data(data)
+  out[names(cols)] <- cols
+
+  is_null <- map_lgl(out, is.null)
+  out <- out[!is_null]
+
+  # Restore attributes (apart from names)
+  attr <- attributes(data)
+  attr$names <- names(out)
+  attributes(out) <- attr
```
krlmlr: This will be slow and mangle row names. @romainfrancois added C code to tibble to copy over attributes, we could port that code here: https://github.com/tidyverse/tibble/blob/master/src/attributes.c

R/generics.R
```diff
@@ -102,8 +102,27 @@ dplyr_col_modify <- function(data, cols) {
...
+  cols <- lapply(cols, vec_recycle, size = nrow(data))
+
+  names(cols) <- as_utf8_character(names2(cols))
+  names(data) <- as_utf8_character(names2(data))
+
+  out <- vec_data(data)
+  out[names(cols)] <- cols
+
+  is_null <- map_lgl(out, is.null)
+  out <- out[!is_null]
```
krlmlr: I'd really like to avoid iterating over `out`, surprisingly this might be the fastest solution in many cases. Should we do this only if there are `NULL` values in `cols`?

```suggestion
  # Iterating over out will be faster than splitting cols into
  # NULL and non-NULL values in many cases.
  if (any(map_lgl(cols, is.null))) {
    is_null <- map_lgl(out, is.null)
    out <- out[!is_null]
  }
```

tests/testthat/test-group-by.r
hadley: It didn't, so I rewrote the test to more clearly express the intent.

R/generics.R
hadley: We could also do a for loop + `[[`:

```R
for (col in seq_along(cols)) {
  nm <- names(cols)[[i]]
  data[[nm]] <- cols[[i]]
```

krlmlr: That's how tibble is doing it.

R/generics.R
hadley: I hacked around the row names problem for now, and I'll wait to see if @DavisVaughan has vctrs based suggestions for a fix since I'd prefer not to add more C++ code for this.

R/generics.R
```diff
@@ -102,8 +102,28 @@ dplyr_col_modify <- function(data, cols) {
 
 #' @export
 dplyr_col_modify.data.frame <- function(data, cols) {
-  data[names(cols)] <- cols
-  data
+  # Implement from first principles to avoiding edge cases in [.data.frame
+  # and [.tibble in 2.1.3 and earlier
+
+  # Apply tidyverse recycling rules
+  cols <- lapply(cols, vec_recycle, size = nrow(data))
```
DavisVaughan:
```suggestion
  cols <- vec_recycle_common(!!!cols, .size = nrow(data))
```

R/generics.R
```diff
@@ -102,8 +102,28 @@ dplyr_col_modify <- function(data, cols) {
...
+
+  names(cols) <- as_utf8_character(names2(cols))
+  names(data) <- as_utf8_character(names2(data))
+
+  out <- vec_data(data)
+  for (i in seq_along(cols)) {
+    nm <- names(cols)[[i]]
+    out[[nm]] <- cols[[i]]
+  }
```
DavisVaughan:
```suggestion
  out <- vec_data(data)

  nms <- as_utf8_character(names2(cols))
  names(out) <- as_utf8_character(names2(out))

  for (i in seq_along(cols)) {
    nm <- nms[[i]]
    out[[nm]] <- cols[[i]]
  }
```

Mainly this avoids the repeated call to `names(cols)[[i]]` in the loop, but also calls `vec_data()` earlier to avoid data frame dispatch on `names()` and `names<-()`.

R/generics.R
```diff
@@ -102,8 +102,28 @@ dplyr_col_modify <- function(data, cols) {
...
+
+  # Restore attributes (apart from names)
+  attr <- attributes(data)
+  attr$names <- names(out)
+  attr$row.names <- .row_names_info(data, 0L)
+  attributes(out) <- attr
+
+  out
```
DavisVaughan: I think this looks fine from a performance perspective. I'm a bit surprised you can attach compact row names at the R level to the attribute list like that, neat trick!

I might have used `new_data_frame()`, which also avoids materializing row names:

```
  size <- nrow(data)
  
  # Restore attributes (apart from names)
  attr <- attributes(data)
  attr$names <- NULL

  # Pass class to `new_data_frame()` manually
  class <- setdiff(attr$class, "data.frame")
  attr$class <- NULL

  out <- exec(new_data_frame, x = out, n = size, class = class, !!!attr)
```

but I discovered a few bugs in `new_data_frame()` when I did so https://github.com/r-lib/vctrs/issues/883

so what you have looks fine

PR #4942, opened 2020-03-06
n_distinct() can use .data in mutate()

krlmlr:
Closes tidyverse/tibble#721.

### DIFF
--- a/R/distinct.R
+++ b/R/distinct.R
@@ -128,7 +128,7 @@ distinct.data.frame <- function(.data, ..., .keep_all = FALSE) {
 #'
 #' This is a faster and more concise equivalent of `length(unique(x))`
 #'
-#' @param \dots Vectors of values, passed on to [list2()]
+#' @param ... <[`dynamic-dots`][rlang::dyn-dots]> Vectors, passed on to [list2()]
 #' @param na.rm if `TRUE` missing values don't count
 #' @examples
 #' x <- sample(1:10, 1e5, rep = TRUE)
--- a/R/distinct.R
+++ b/R/distinct.R
@@ -128,7 +128,7 @@ distinct.data.frame <- function(.data, ..., .keep_all = FALSE) {
 #'
 #' This is a faster and more concise equivalent of `length(unique(x))`
 #'
-#' @param \dots vectors of values
+#' @param \dots Vectors of values, passed on to [list2()]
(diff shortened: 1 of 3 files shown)


R/distinct.R
```diff
@@ -128,15 +128,15 @@ distinct.data.frame <- function(.data, ..., .keep_all = FALSE) {
 #'
 #' This is a faster and more concise equivalent of `length(unique(x))`
 #'
-#' @param \dots vectors of values
+#' @param \dots Vectors of values, passed on to [list2()]
```
lionel-: Should use dynamic dots tag.

krlmlr:
I checked, no other calls of the form `tibble(...)` .

I'm happy to revise the `tibble()` UI, let's discuss in https://github.com/tidyverse/tibble/issues/721.

lionel-:
> I supposed !!!list2() would no longer be needed after this : tidyverse/tibble@d0707ef
>
> Perhaps we can leave a comment to change then.

Going through `list()` or `list2()` is definitely needed, otherwise `n_distinct()` would inherit the semantics of `tibble()`.

krlmlr:
In other words, do we allow `mutate()` semantics in `n_distinct()`? f8aa4561, not part of this branch yet.

lionel-:
IMO it would only make the function more confusing. However it's probably not worth to spend time thinking about this. Tibble semantics were introduced by error, and it seems better to avoid changing semantics without an actual need or request.

DavisVaughan:
FWIW I think the simplest implementation is:

```r
  data <- list2(...)
  names(data) <- as.character(seq_along(data))
  
  size <- vec_size_common(!!!data)
  data <- vec_recycle_common(!!!data, .size = size)
  
  data <- new_data_frame(data, n = size)
  
  if (isTRUE(na.rm)) {
    data <- vec_slice(data, !vec_equal_na(data))
  }
  
  vec_unique_count(data)
}
```

(This is on top of #4950, which tweaks the `na.rm = TRUE` case)

This:

- Avoids `tibble()` data mask issues (which were also fixed by https://github.com/tidyverse/tibble/pull/725 anyways)
- Does not have `mutate()` semantics

hadley:
@DavisVaughan do you want to alter #4950 to do that? I think that's the simplest way forward.

PR #4944, opened 2020-03-06
Update to latest actions workflow

DavisVaughan:
I just re-ran `usethis::use_github_actions_tidy()`, ensuring that the `_R_CHECK_FORCE_SUGGESTS_: false` variable is kept, which seems like something that was manually turned off.

This should ensure that codecov starts working again


DavisVaughan:
Looks like it works 👍 https://codecov.io/github/tidyverse/dplyr

PR #4947, opened 2020-03-08
Use new_data_frame()

krlmlr:
This is still a lot of extra code I'd rather see elsewhere, ideally implemented in C. Perhaps in `vctrs::new_data_frame()` or in a wrapper.

### DIFF
--- a/R/generics.R
+++ b/R/generics.R
@@ -125,7 +125,12 @@ dplyr_col_modify.data.frame <- function(data, cols) {
     attr$row.names <- .row_names_info(data, 0L)
   }
   attr$class <- setdiff(attr$class, "data.frame")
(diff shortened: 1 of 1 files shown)


hadley:
This seems more complicated that the current code, so I'm not sure what the advantage is?

krlmlr:
We're using `new_data_frame()` now, and we keep row names.

Querying row names is expensive in R < 3.5. With adequate tooling (https://github.com/r-lib/rlang/issues/940) we could do a very fast and concise implementation. Earlier I suggested to tweak `new_data_frame()` so that we can use it directly (https://github.com/r-lib/vctrs/issues/892), this only makes sense if the logic works for all use cases.

krlmlr:
...we could also revert to using `[.tbl_df()` once tibble 3.0.0 is out.

PR #4948, opened 2020-03-08
Fix deprecated function names in src_mysql() and src_postgres()

yutannihilation:
Just happened to find these. Followup of #4941


hadley:
Thanks!

Issue #4983, opened 2020-03-14
count() broken for lazy tables


krlmlr:
``` r
library(dplyr)

dbplyr::memdb_frame(a = 1, b = 2) %>%
  count(a)
#> Warning: `src_sqlite()` is deprecated as of dplyr 1.0.0.
#> Please use `tbl()` directly with a database connection
#> This warning is displayed once every 8 hours.
#> Call `lifecycle::last_warnings()` to see where this warning was generated.
#> Warning: The `add` argument of `group_by()` is deprecated as of dplyr 1.0.0.
#> Please use the `.add` argument instead.
#> This warning is displayed once every 8 hours.
#> Call `lifecycle::last_warnings()` to see where this warning was generated.
#> Error in UseMethod("group_data"): no applicable method for 'group_data' applied to an object of class "c('tbl_SQLiteConnection', 'tbl_dbi', 'tbl_sql', 'tbl_lazy', 'tbl')"
```

<sup>Created on 2020-03-14 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

I can take this on.

hadley:
This has the same root cause as #4984, so I'll take it on.

hadley:
The root cause of the problem is that `group_by()` needs to know the names of unpacked data frame columns so that it works with `across()`. That means it currently has its own partial implementation of `mutate()` that doesn't go through the dispatch needed to work with other backends.

One possible way to work around this would be for `mutate()` to return information about which columns it has added. Maybe it could signal a condition?

krlmlr:
We could also implement something like `build_mutate_spec()` with the invariant that

```r
mutate(...)
```

is the same as

```r
mutate(!!!build_mutate_spec(...))
```

or perhaps

```r
mutate_spec(build_mutate_spec(...))
```

. Useful for other dplyr backends.

PR #4987, opened 2020-03-15
WIP: Tests for add_computed_columns()

krlmlr:
For #4983.


hadley:
This is a bit premature since I'm not sure we want to keep this function.

PR #4993, opened 2020-03-16
Edits to rowwise and colwise vignettes

mine-cetinkaya-rundel:
- Simplify first two examples in rowwise vignette
- Reorder two sections in rowwise vignette to increase complexity as you read down
- Minor grammatical edits / typos

I've also opened #4992, which needs to be addressed in the rowwise vignette.

And finally, I think the Gotchas section in the colwise vignette is still a bit confusing. I think it would be nice to add a "because ..." at the end of "Here `n` becomes `NA` because the standard deviation of 3 is `NA`", something along the lines of "because dplyr first does ... and then does ...", but I wasn't sure about the precise language for what's happening under the hood.

### DIFF
--- a/vignettes/colwise.Rmd
+++ b/vignettes/colwise.Rmd
@@ -281,7 +281,7 @@ There are a few exceptions to this rule:
     df %>% filter(rowAny(across(is.numeric, ~ .x > 0)))
     ```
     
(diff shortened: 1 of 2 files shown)


hadley:
Thanks @mine-cetinkaya-rundel!

PR #5052, opened 2020-03-26
Improve hint for old grouping

### LINKED ISSUE #5027: Corrupt grouped_df data using the old format
I think this error should include a hint, like "Regroup with `group_by()` or strip grouping with `ungroup()`" (including the names of the variables if we can easily extract them)

hadley:
Fixes #5027


hadley:
Thanks!

PR #5061, opened 2020-03-27
Don't count regroups in tests

lionel-:
Fixes tests once r-lib/vctrs#951 is merged.


lionel-:
Now merged.

hadley:
I think we should revert this since we're not going to use `vctrs::vec_cbind()`

PR #5081, opened 2020-04-01
Improve website navbar + article index

hadley:
I can't think of a particularly easy way to share the rendered result, but here's a screenshot. Hopefully that will give you a sense of how each bit of the yaml affects the final page.

Screenshot 2020-04-01 at 7 57 39 AM

### DIFF
--- a/vignettes/two-table.Rmd
+++ b/vignettes/two-table.Rmd
@@ -1,5 +1,11 @@
 ---
 title: "Two-table verbs"
+description: >
+  Most dplyr verbs work with a single data set, but most data analyses involve
+  multiple datasets. This vignette introduces you to the dplyr verbs that 
+  work with more one than data set, and introduces to the mutating joins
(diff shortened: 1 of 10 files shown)


vignettes/two-table.Rmd
```diff
@@ -1,5 +1,11 @@
 ---
 title: "Two-table verbs"
+description: >
+  Most dplyr verbs work with a single data set, but most data analyses involve
+  multiple datasets. This vignette introduces you to the dplyr verbs that 
+  work with more one than data set, and introduces to the mutating joins
+  (which are used primarily to add new columns), filtering joins (which are
+  used to select rows), and the set operations.
```
batpigandme, line 8:
```suggestion
  work with more one than data set, and introduces to the mutating joins, 
  filtering joins, and the set operations.
```

batpigandme: I think the parenthetical make this description a little unwieldy in the articles listing page.

PR #5102, opened 2020-04-13
Minor formatting edits on dplyr vignette

mine-cetinkaya-rundel:
A few minor formatting edits on the dplyr vignette that shows up on the main pkgdown page

- Some inline function names were not formatted as code, those are now fixed
- A few places with two spaces after a period, now using only single space
- Inconsistency between code chunks re: single/multi-line pipe formatting is now made consistent -- no line break if `%>%` used only once


hadley:
Thanks! What do you think of the overall flow?

mine-cetinkaya-rundel:
I like it, I think it reads well. 

I went back and forth between replacing all `foo(df, ...)` with `df %>% foo(...)` but ultimately thought it's best to showcase both options so people aren't surprised by the `foo(df, ...)` usage elsewhere.

hadley:
Thanks!

Issue #5103, opened 2020-04-13
Vignette titles: rowwise and colwise


mine-cetinkaya-rundel:
I think the vignettes called "colwise operations" and "rowwise operations" should be renamed to "Column-wise operations" and "Row-wise operations", respectively. There is a function called `rowwise` but not a function called `colwise`, but the naming and how these are displayed in the vignettes menu (especially the fact that they both start with small letters) makes it seem like there may be a function called `colwise`.

<img width="224" alt="Screenshot 2020-04-13 at 13 22 27" src="https://user-images.githubusercontent.com/5965649/79120104-ee496380-7d89-11ea-955b-93b6de580f59.png">

hadley:
Do you want to do a PR? You just need to update the vignette title in the `.Rmd` (in two places 😬)

PR #5104, opened 2020-04-13
Vignette rename: row-wise and col-wise

mine-cetinkaya-rundel:
- Rename colwise and rowwise vignettes as discussed in #5103 
- Update references to these actions in prose to read "column-wise" and "row-wise" to match titles

### DIFF
--- a/vignettes/colwise.Rmd
+++ b/vignettes/colwise.Rmd
@@ -1,5 +1,5 @@
 ---
-title: "colwise operations"
+title: "Column-wise operations"
 description: >
   Learn how to easily repeat the same operation across multiple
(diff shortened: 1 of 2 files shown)


hadley:
BTW with the new "linked issues" UI in Github, it works better to include `Fixes #5103` in the PR description, not just the individual commits.

mine-cetinkaya-rundel:
Thanks! Is it better to include it in both commit and PR or just PR? (for future reference)

hadley:
I tend to include it in both because I add it in the commit message at the same time I add a bullet to NEWS (which doesn't apply here).

PR #5121, opened 2020-04-16
Adjust tests for waldo

hadley:
WIP since I'm also using it as a lens to improve waldo


hadley:
A swath need adjustments for:

```
`attr(actual, 'tzone')` is a character vector ('')
`attr(expected, 'tzone')` is absent
```

Then the behaviour of `expect_equivalent()` is broken: https://github.com/r-lib/waldo/issues/26

PR #5142, opened 2020-04-23
rows_()

### LINKED ISSUE #4654: Consider new row mutation functions
I've been wondering if dplyr needs a new set of verbs specifically designed for modifying rows identified by a set of keys:

* insert new rows — would error if keys already exist; can specify where to add (how?). Similar to `tibble::add_row()`.

* update values — overrides existing values. (Similar to https://github.com/tidyverse/tidyr/issues/183)

* patch values — like update, but only replaces missing values (Also similar to https://github.com/tidyverse/tidyr/issues/183)

* upsert — update or insert depending on presence/absence of keys

## Basic sketch

```R
df <- tribble(
  ~ x, ~ y,
  1, 2,
  2, NA,
)

# Insert can add multiple rows
df %>% insert(x = 3:4, y = 5)
#>       x     y
#>   <dbl> <dbl>
#> 1     1     2
#> 2     2    NA
#> 3     3     5
#> 4     4     5

# And new columns:
df %>% insert(x = 3, z = 1)
#>       x     y     z
#>   <dbl> <dbl> <dbl>
#> 1     1     2    NA
#> 2     2    NA    NA
#> 3     3    NA     1

#> Error: key variable `x` is missing
df %>% insert(x = 1, y = 5)
#> Error: key `x = 1` already present

# Update modifies existing rows:
df %>% update(x = 1, y = 3)
#>       x     y
#>   <dbl> <dbl>
#> 1     1     3
#> 2     2    NA

# Erroring if there's no match:
df %>% update(x = 3, y = 3)
#> Error: key `x = 3` doesn't exist

# patch() works like update but only changes missing values
df %>% patch(x = 1:2, y = 10)
#>       x     y
#>   <dbl> <dbl>
#> 1     1     3
#> 2     2    10

# upsert() updates or inserts as needed:
df %>% upsert(x = c(1, 3), y = 3)
#>       x     y
#>   <dbl> <dbl>
#> 1     1     3
#> 2     2    NA
#> 3     3     3
```

## Thoughts

* All functions would either take multiple named inputs or a single unnamed data frame.

* I think unlike SQL and `==`, it makes sense for `NA` to match `NA` in keys. That's consistent with `match()`.

* If `.key` were ordered (can we determine that quickly?) `insert()` could insert new rows in right place. Otherwise would be have `.before` and `.after` that would be a single row data frame of key values?

* `.key` would work like `.by`; defaulting to common variables with a message. But unlike joins, would check key set for uniqueness, erroring if not (i.e. no Cartesian products). That means `insert(x, y)` would always yield `nrow(x) + nrow(y)` and `update(x, y)` would always yield `nrow(x)`.

* Should all these functions get a `rows` prefix? Or suffix? `rows_insert()`, `rows_update()`, `rows_patch()`, ...? Does that suggest there's an equivalent set of `col` functions? `cols_insert()` (a single col join, like a vlookup? error if col already exists?), `cols_move()` (#4598), `cols_patch()` (df form of `coalesce()`?)  

## Backends

* Grouping would be ignored (but recomputed if needed)
* Would error for dbplyr as modifying data is out of scope.
* Could modifying in place for dtplyr when `immutable = FALSE`
* dm objects would use predefined primary keys and error if `.key` supplied
* An imaginary mutable googlesheet could perform these operations on the underlying sheet

krlmlr:
Extracted from https://github.com/krlmlr/dm/pull/319.

Closes #4654.


R/rows.R
```diff
@@ -0,0 +1,277 @@
+#' Manipulate individual rows
+#'
+#' @description
+#' \lifecycle{experimental}
+#'
+#' These methods provide a framework for manipulating individual rows
+#' in existing tables, modeled after the SQL operations
+#' `INSERT`, `UPDATE` and `DELETE`.
+#' All operations expect existing and new data to be compatible.
```
hadley, line 9: What does compatible mean?

R/rows.R
```diff
@@ -0,0 +1,277 @@
...
+#' rows_insert
+#'
+#' `rows_insert()` adds new rows.
+#' This operation corresponds to `INSERT` in SQL.
+#' If `by` is non-empty, no two rows with the same values in the key columns
+#' are permitted.
+#' @rdname rows
+#' @export
+rows_insert <- function(x, y, by = NULL, ..., copy = FALSE, inplace = NULL) {
+  check_rows_args(x, y, by)
```
hadley, line 69: Need to use ellipsis to declare what we expect with `...`

R/rows.R
```diff
@@ -0,0 +1,277 @@
...
+#' @export
+rows_truncate.data.frame <- function(x, ..., copy = FALSE, inplace = NULL) {
+  df_inplace(inplace)
+
+  x[integer(), ]
+}
+
+
+
+check_rows_args <- function(x, y, by) {
```
hadley, line 233: Either these checks need to do in the methods, or you'll have to rewrite with an eye to dplyr genericity (i.e. use `tbl_vars()` instead of `colnames()`)

R/rows.R
```diff
@@ -0,0 +1,277 @@
...
+check_by <- function(by, y) {
+  if (!is.null(by) && !is.character(by)) {
+    abort("`by` must be `NULL` or a character vector.")
+  }
+  if (!is.null(names(by)) || any(names(by) != "")) {
+    abort("`by` must not have names.")
+  }
+  bad <- setdiff(by, colnames(y))
+  if (has_length(bad)) {
+    abort(class = "dplyr_rows_extra_column",
```
hadley, line 257: Don't they have to exist in `x` too?

R/rows.R
```diff
@@ -0,0 +1,277 @@
...
+      name = bad
+    )
+  }
+}
+
+check_by <- function(by, y) {
+  if (!is.null(by) && !is.character(by)) {
+    abort("`by` must be `NULL` or a character vector.")
+  }
+  if (!is.null(names(by)) || any(names(by) != "")) {
```
hadley, line 252: Can use `rlang::is_named()` here

R/rows.R
```diff
@@ -0,0 +1,277 @@
...
+    abort(class = "dplyr_rows_extra_column",
+      "All `by` columns must exist in `y`.",
+      name = bad
+    )
+  }
+}
+
+df_key <- function(y, by) {
+  if (is.null(by)) {
+    set_names(1L, names(y)[[1]])
```
hadley, line 266: Why are we naming these?

R/rows.R
```diff
@@ -0,0 +1,277 @@
...
+}
+
+#' @export
+rows_insert.data.frame <- function(x, y, by = NULL, ..., copy = FALSE, inplace = NULL) {
+  y <- auto_copy(x, y, copy = copy)
+  y_key <- df_key(y, by)
+  x_key <- df_key(x, names(y_key))
+  df_inplace(inplace)
+
+  if (has_length(y_key)) {
```
hadley, line 80: Why would `y_key` be absent?

R/rows.R
```diff
@@ -0,0 +1,277 @@
...
+  bad <- setdiff(by, colnames(y))
+  if (has_length(bad)) {
+    abort(class = "dplyr_rows_extra_column",
+      "All `by` columns must exist in `y`.",
+      name = bad
+    )
+  }
+}
+
+df_key <- function(y, by) {
```
hadley, line 264: Maybe these should be `row_df_*`?

R/rows.R
hadley: I think the code would be a decent amount simpler if the keys were character vectors, since then we only need `key` not `x_key` and `y_key`.

hadley: I think the code would be a decent amount simpler if the keys were character vectors, since then we only need `key` not `x_key` and `y_key`.

R/rows.R
```diff
@@ -0,0 +1,277 @@
...
+    set_names(1L, names(y)[[1]])
+  } else {
+    idx <- match(by, names(y))
+    set_names(idx, by)
+  }
+}
+
+df_inplace <- function(inplace) {
+  if (is_true(inplace)) {
+    inform("`rows_...()` ignore `inplace` ignored for data frames.")
```
hadley, line 275:
```suggestion
    inform("`rows_...()` ignore `inplace` argument for data frames.")
```

R/rows.R
```diff
@@ -0,0 +1,277 @@
...
+
+#' rows_insert
+#'
+#' `rows_insert()` adds new rows.
+#' This operation corresponds to `INSERT` in SQL.
+#' If `by` is non-empty, no two rows with the same values in the key columns
+#' are permitted.
+#' @rdname rows
+#' @export
+rows_insert <- function(x, y, by = NULL, ..., copy = FALSE, inplace = NULL) {
```
hadley, line 68: Shouldn't the default for `inplace` be `FALSE`?

R/rows.R
hadley: Also I think this specific branch needs to emit an message saying which keys have been used.

tests/testthat/test-rows.txt
```diff
@@ -0,0 +1,103 @@
...
+Delete and truncate
+===================
+
+> rows_delete(data, tibble(a = 2:4))
+# A tibble: 1 x 3
+      a b         c
+  <int> <chr> <dbl>
+1     1 a       0.5
+
+> rows_delete(data, tibble(a = 2:4, b = "b"))
```
hadley, line 77: Should this message to indicate that `b` is ignored?

tests/testthat/test-rows.R
```diff
@@ -0,0 +1,28 @@
+scoped_options(lifecycle_verbosity = "quiet")
+
+verify_output("test-rows.txt", {
```
hadley, line 3: I think these look good, so it would be useful to turn into individual tests now.

tests/testthat/test-rows.R
```diff
@@ -0,0 +1,28 @@
+scoped_options(lifecycle_verbosity = "quiet")
```
hadley, line 1: Why is this needed?

R/rows.R
hadley: Also I'm reasonably certain that this should be an error.

R/rows.R
```diff
@@ -0,0 +1,277 @@
...
+  idx_new <- idx[new]
+
+  x[idx_existing, names(y)] <- y[!new, ]
+  vctrs::vec_rbind(x, y[new, ])
+}
+
+#' rows_delete
+#'
+#' `rows_delete()` deletes existing rows.
+#' This operation corresponds to `DELETE` in SQL.
```
jennybc, line 192: How is this different from `anti_join()`? (I know the answer is about `by`, but seems worth exploring.)

R/rows.R
```diff
@@ -0,0 +1,277 @@
...
+  }
+
+  vctrs::vec_rbind(x, y)
+}
+
+#' rows_update
+#'
+#' `rows_update()` updates existing rows.
+#' This operation corresponds to `UPDATE` in SQL.
+#' `by` is mandatory and defaults to the first column in `y`.
```
jennybc, line 98: I find this statement (here and all the other instances) confusing because it seems to contradict itself. If `by` is mandatory, why is the signature `by = NULL` and we silently default to the first column of `y`?

Related to my proposal to create a small exported helper for the purposes of making this default transparent.

R/rows.R
```diff
@@ -0,0 +1,277 @@
...
+  df_inplace(inplace)
+
+  idx <- vctrs::vec_match(y[y_key], x[x_key])
+  # FIXME: Check key in x? https://github.com/r-lib/vctrs/issues/1032
+  x[-idx[!is.na(idx)], ]
+}
+
+#' rows_truncate
+#'
+#' `rows_truncate()` removes all rows.
```
jennybc, line 215: At first I thought, "what is this good for?" But then I realized it's good for making a 0-row prototype of a data frame. Maybe worth pointing out.

R/rows.R
```diff
@@ -0,0 +1,277 @@
...
+  y <- auto_copy(x, y, copy = copy)
+  y_key <- df_key(y, by)
+  x_key <- df_key(x, names(y_key))
+  df_inplace(inplace)
+
+  if (has_length(y_key)) {
+    idx <- vctrs::vec_match(y[y_key], x[x_key])
+    bad <- which(!is.na(idx))
+    if (has_length(bad)) {
+      abort(class = "dplyr_rows_insert_duplicate",
```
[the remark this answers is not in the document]
hadley: I think we'll leave better errors for a future release; I want to make sure that the verbs are solid before we invest a bunch in good errors.

R/rows.R
hadley: Also if you were using it on a google sheet (say), it would actually delete rows from the source.

hadley:
I also wonder if we should be checking key validity by default — it would make the functions a little  slower but would guarantee that you learn about problems early.

R/rows.R
```diff
@@ -0,0 +1,277 @@
...
+#' rows_delete(data, tibble(a = 2:4))
+#' rows_delete(data, tibble(a = 2:4, b = "b"))
+#' rows_delete(data, tibble(a = 2:4, b = "b"), by = c("a", "b"))
+#' rows_truncate(data)
+NULL
+
+
+#' rows_insert
+#'
+#' `rows_insert()` adds new rows.
```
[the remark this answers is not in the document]
hadley: We had considered a `sort` argument that would resort after inserting. Maybe we could set that to `TRUE` if `x` was already sorted?

hadley:
I'll take a pass on this; so please hold off on making any more changes until I've merged mine back in.

R/rows.R
krlmlr: I'd argue that the default for `copy` should be `NULL` and be interpreted as `FALSE`. I used `copy = FALSE` for consistency with joins.

It seems that we need `NULL` defaults for the other arguments to easily distinguish missing values, and to be able to tweak the semantics without visible changes in the interface. I don't mind a `first_column()` helper that could be referred to from the docs, I'm not sure we should add it to the interface just yet.

Regarding `inplace` there's two options:

- Throw an error if `inplace = TRUE` is requested but can't be fulfilled (data frames, database view)
- Ignore `inplace = TRUE` on immutable backends

The first option seems "purer" but also more pedantic. I don't have a strong opinion, we need to allow `inplace = TRUE` to fail anyway.

R/rows.R
krlmlr: `by = character()` -- no key defined, duplicates permitted. Do we need to mention specifically in the docs?

R/rows.R
krlmlr: Is this about adding a link to `anti_join()` to establish context?

Some operations already exist with a different name. The goal of this API is updates on backends that support it, via the `inplace` argument. Some duplication with existing functionality seems unavoidable, perhaps we can eventually supersede semi- and anti-joins?

`inplace = FALSE` allows to test drive changes before actually writing to your database/data.table/Google sheet. (For the latter, "dry run" might mean that you save a temporary copy of your sheet, apply the operation there and destroy it when done.) `inplace = TRUE` fails if the data source is immutable, as with data frames.

R/rows.R
krlmlr: Let's use `tbl_vars()` if it's supported better than `colnames()` -- the latter works on databases though.

R/rows.R
krlmlr: We also need to detect `by = c("x", "y" = "z")` . Do we need `is_unnamed()`?

R/rows.R
krlmlr: Yes, that's handled earlier though. I wonder how to phrase this.

R/rows.R
krlmlr: Using numeric indexes is minimally faster. Forgot about the message.

tests/testthat/test-rows.R
krlmlr: I guess it's unrelated, forgot to remove.

tests/testthat/test-rows.R
krlmlr: Let's keep the output test for a little while too.

tests/testthat/test-rows.txt
krlmlr: For dm, is there an easy way to turn off that message, other than `select(y, !!!by)`? I'm fine with muffling if the message has a class.

R/rows.R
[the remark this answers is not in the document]
jennybc: How do you know what it was / should be sorted on? 🤔

R/rows.R
krlmlr: Maybe:

```suggestion
#' `by` is mandatory, if omitted the first column in `y` is used.
```

R/rows.R
jennybc: Ok, yes, the differences are even greater for mutable backends. I hadn't thought about that.

R/rows.R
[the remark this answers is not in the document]
jennybc: I've gotten various feature requests over the years to insert rows into Google Sheets. But I've never formed a clear sense of how someone would want to express their intention about *where* the new row(s) should go. But I know it will come up and, once I can figure out how to elicit something concrete from the user, the Sheets API definitely supports this an insertion. This is why I'm asking.

R/rows.R
jennybc: That's definitely better. I think it probably needs to message to help people remember that the default behaviour for `by` for `row_*()` functions differs from that of the `*_join()` functions.

R/rows.R
```diff
@@ -0,0 +1,245 @@
...
+  UseMethod("rows_insert")
+}
+
+#' @export
+rows_insert.data.frame <- function(x, y, by = NULL, ..., copy = FALSE, in_place = FALSE) {
+  key <- check_key(by, x, y)
+  y <- auto_copy(x, y, copy = copy)
+  df_in_place(in_place)
+
+  check_key_df(x, key, df_name = "x")
```
hadley, line 86: I think it makes sense to enforce this checking — it should generally be quite quick, and if we later decide that the performance is prohibitive, we can add an argument to make checking optional. And either way, it's easy to start strict and relax than the opposite.

R/rows.R
```diff
@@ -0,0 +1,245 @@
...
+  check_key_df(y, key, df_name = "y")
+  idx <- vctrs::vec_match(y[key], x[key])
+  new <- is.na(idx)
+  idx_existing <- idx[!new]
+  idx_new <- idx[new]
+
+  x[idx_existing, names(y)] <- vec_slice(y, !new)
+  vctrs::vec_rbind(x, vec_slice(y, new))
```
hadley, line 165: This isn't quite right because it won't preserve the type of `x`. Maybe we need to coerce `x` and `y` to data frames, and then restore on output?

R/rows.R
```diff
@@ -1,37 +1,50 @@
...
 #' @description
-#' \lifecycle{experimental}
-#'
-#' These methods provide a framework for manipulating individual rows
-#' in existing tables, modeled after the SQL operations
-#' `INSERT`, `UPDATE` and `DELETE`.
-#' All operations expect existing and new data to be compatible.
+#' \Sexpr[results=rd, stage=render]{lifecycle::badge("experimental")}
+#'
+#' These functions provide a framework for modifying rows in table using
```
jennybc, line 6:
```suggestion
#' These functions provide a framework for modifying rows in a table using
```

R/rows.R
```diff
@@ -1,37 +1,50 @@
...
+#' a second table of data. The two tables are matched `by` a set of key
+#' variables whose values must uniquely identify each row. The functions are
+#' inspired by SQL's `INSERT`, `UPDATE` and `DELETE`, and can optionally
```
jennybc, line 9:
```suggestion
#' inspired by SQL's `INSERT`, `UPDATE`, and `DELETE`, and can optionally
```

R/rows.R
```diff
@@ -1,37 +1,50 @@
...
+#' modified `in_place` for selected backends.
```
jennybc, line 10:
```suggestion
#' modify `in_place` for selected backends.
```

R/rows.R
```diff
@@ -1,37 +1,50 @@
...
+#'
+#' * `rows_insert()` adds new rows (like `INSERT`); key values in `y` must
+#'    not occur in `x`.
+#' * `rows_update()` modifies existing rows (like `UPDATE`); key values in
+#'   `y` must occur in `x`.
+#' * `rows_patch()` works like `rows_update()` but only overwrites `NA` values.
+#' * `rows_upsert()` inserts or updates depending on whether or not the the
```
jennybc, line 17:
```suggestion
#' * `rows_upsert()` inserts or updates depending on whether or not the
```

R/rows.R
```diff
@@ -1,37 +1,50 @@
...
+#'   key value in `x` exists in `y`.
```
jennybc, line 18: Is the other way 'round more natural?
```suggestion
#'   key value in `y` already exists in `x`.
```

R/rows.R
```diff
@@ -1,37 +1,50 @@
...
+#'   When `TRUE`, a modified version of `x` is returned invisibly;
+#'   when `FALSE`, a new object representing the resulting changes is returned.
+#' @returns
+#' An object of the same type as `x`. The order of the rows and columns of `x`
+#' is preserved as much as possible. The output has the following properties:
+#'
+#' * `rows_update()` preserves rows as is; `rows_insert()` and `rows_upsert()`
+#'   return all existing rows and potentially new rows; `rows_delete()` returns
+#'   a subset of the rows.
+#' * Columns are not modified.
```
jennybc, line 43:
```suggestion
#' * Columns are not added, removed, or relocated, though the data may be updated.
```

R/rows.R
jennybc: So is `rows_truncate()` staying but is being deliberately de-emphasized in docs? Because it's currently missing from the main bullet list with one item per function.

krlmlr: This is more about providing an entry point for databases, where `TRUNCATE` is a thing. It feels odd to move this generic to dbplyr, even if its usability here is limited.

hadley: I'm not still not convinced it's that important, and I worry about giving people a tool that deletes every row in a database. I think if you really want a truncate, it's reasonable to have to write the SQL yourself. I'd prefer to wait until someone actually needs this before implementing it.

R/rows.R
```diff
@@ -0,0 +1,245 @@
...
+#' @export
+rows_insert.data.frame <- function(x, y, by = NULL, ..., copy = FALSE, in_place = FALSE) {
+  key <- check_key(by, x, y)
+  y <- auto_copy(x, y, copy = copy)
+  df_in_place(in_place)
+
+  check_key_df(x, key, df_name = "x")
+  check_key_df(y, key, df_name = "y")
+
+  idx <- vctrs::vec_match(y[key], x[key])
```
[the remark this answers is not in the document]
krlmlr: How should `key = character()` work for `rows_update()` or `rows_delete()`?

Theory suggests that "no key" means "all columns define the key", i.e. `key = colnames(x)`. This is expensive to check, though.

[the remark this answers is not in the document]
hadley: Right, it's not clear. I think that suggests keys should always be required.

krlmlr:
Done. Adapting `dplyr_reconstruct()` along the way -- we don't need to restore integer row names.

krlmlr:
Resolved conflicts, tests pass now.

hadley:
@krlmlr thanks for all your hard work on this — it's an exciting addition to the dplyr API!

krlmlr:
It's great to have this in dplyr, thanks for your edits and for adding it on such short notice.

Issue #5132, opened 2020-04-20
Preserve free-floating data.frame attributes where possible?


lionel-:
From r-lib/vctrs#703. Probably best tackled with `dplyr_reconstruct()`.

* tidyverse/dplyr#4617
* tidyverse/dplyr#3429

hadley:
Seems like we're already doing this now anyway.

PR #5160, opened 2020-04-29
Simplify `dplyr_reconstruct()`

DavisVaughan:
Closes #5158 

This drastically simplifies `dplyr_reconstruct.data.frame()` by taking an approach similar to `dplyr_col_modify.data.frame()`, which just restores the attributes directly.

I've also pulled over the tests from #5158, but I think the attributes one might have been backwards. I think we care about keeping the attributes from `template`, not `data`.

The job of the data frame method of `dplyr_reconstruct()` is now very simple and well defined. It keeps all attributes of `template` except:

- `names` in case columns were modified
- `row.names` in case rows were modified

There was some code about the `.drop` attribute in the old `dplyr_reconstruct.data.frame()`, but I don't think it belonged there because we have a `grouped_df` method as well. It also wouldn't have worked right because the way to get to the drop attribute is through the group data of the data frame. It's not on the data frame directly.

### DIFF
--- a/R/generics.R
+++ b/R/generics.R
@@ -158,19 +158,12 @@ dplyr_reconstruct <- function(data, template) {
 
 #' @export
 dplyr_reconstruct.data.frame <- function(data, template) {
-  attr_old <- attributes(template)
-  attr_new <- attributes(data)
+  attrs <- attributes(template)
 
-  to_copy <- setdiff(names(attr_old), c("class", "row.names", "names", ".drop"))
-  attr_new[to_copy] <- attr_old[to_copy]
+  attrs$names <- names(data)
+  attrs$row.names <- .row_names_info(data, type = 0L)
 
-  # `new_data_frame()` will add the `"data.frame"` class
-  class <- setdiff(class(template), "data.frame")
-  attr_new[["class"]] <- NULL
-
-  size <- vec_size(data)
-
-  data <- exec(new_data_frame, x = data, n = size, class = class, !!! attr_new)
+  attributes(data) <- attrs
 
   data
 }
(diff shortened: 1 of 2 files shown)


hadley:
I think this also fixes #5132?

DavisVaughan:
That issue links to #4617, which has this example:

```r

df1 <- tibble(
  a = 1:2, 
  b = 2:1
)

df2 <- tibble(
  a = structure(1:2, foo = "bar"),
  c = 2:1
)

left_join(df1, df2, by = "a")$a
#> [1] 1 2
left_join(df2, df1, by = "a")$a
#> [1] 1 2
#> attr(,"foo")
#> [1] "bar"
```

But honestly I sort of feel like this behavior is right. You get the attributes of the first input?

DavisVaughan:
With `dplyr_col_modify()`, `dplyr_row_slice()`, and `dplyr_reconstruct()` already in place, I don't currently see any cases where we _aren't_ preserving free floating attributes. I think we are doing the right thing now?

R/generics.R
```diff
@@ -158,19 +158,12 @@ dplyr_reconstruct <- function(data, template) {
...
 #' @export
 dplyr_reconstruct.data.frame <- function(data, template) {
-  attr_old <- attributes(template)
-  attr_new <- attributes(data)
+  attrs <- attributes(template)
 
-  to_copy <- setdiff(names(attr_old), c("class", "row.names", "names", ".drop"))
-  attr_new[to_copy] <- attr_old[to_copy]
+  attrs$names <- names(data)
+  attrs$row.names <- .row_names_info(data, type = 0L)
```
krlmlr, line 164: I think this would benefit from a lightweight C function that walks the pairlist, replaces in place and appends if necessary.

hadley: Yeah, that might be a useful helper for vctrs or similar. It doesn't need to block this PR.

PR #5167, opened 2020-04-30
Refactor `group_split()` to use `dplyr_row_slice()`

### LINKED ISSUE #5165: Should `nest_by()` and `group_split()` (and friends) preserve subclass, if `dplyr_row_slice()` implemented?
If the package defines `[` and `dplyr_row_slice()` for the subclass, can we expect `nest_by()` and `group_[split/nest]()` to work out of box to return subclass (not tibble) in the partitions, without explicitly defining these methods?
 
``` r
library(tsibble)
dplyr_row_slice(pedestrian, 1:3)
#> # A tsibble: 3 x 5 [1h] <Australia/Melbourne>
#> # Key:       Sensor [1]
#>   Sensor         Date_Time           Date        Time Count
#>   <chr>          <dttm>              <date>     <int> <int>
#> 1 Birrarung Marr 2015-01-01 00:00:00 2015-01-01     0  1630
#> 2 Birrarung Marr 2015-01-01 01:00:00 2015-01-01     1   826
#> 3 Birrarung Marr 2015-01-01 02:00:00 2015-01-01     2   567
nest_by(pedestrian, Sensor)
#> # A tibble: 4 x 2
#> # Rowwise:  Sensor
#>   Sensor                                      data
#>   <chr>                         <list<tbl_df[,4]>>
#> 1 Birrarung Marr                      [14,566 × 4]
#> 2 Bourke Street Mall (North)          [16,414 × 4]
#> 3 QV Market-Elizabeth St (West)       [17,518 × 4]
#> 4 Southern Cross Station              [17,539 × 4]
group_split(pedestrian, Sensor)[[1]]
#> # A tibble: 14,566 x 5
#>    Sensor         Date_Time           Date        Time Count
#>    <chr>          <dttm>              <date>     <int> <int>
#>  1 Birrarung Marr 2015-01-01 00:00:00 2015-01-01     0  1630
#>  2 Birrarung Marr 2015-01-01 01:00:00 2015-01-01     1   826
#>  3 Birrarung Marr 2015-01-01 02:00:00 2015-01-01     2   567
#>  4 Birrarung Marr 2015-01-01 03:00:00 2015-01-01     3   264
#>  5 Birrarung Marr 2015-01-01 04:00:00 2015-01-01     4   139
#>  6 Birrarung Marr 2015-01-01 05:00:00 2015-01-01     5    77
#>  7 Birrarung Marr 2015-01-01 06:00:00 2015-01-01     6    44
#>  8 Birrarung Marr 2015-01-01 07:00:00 2015-01-01     7    56
#>  9 Birrarung Marr 2015-01-01 08:00:00 2015-01-01     8   113
#> 10 Birrarung Marr 2015-01-01 09:00:00 2015-01-01     9   166
#> # … with 14,556 more rows
```

<sup>Created on 2020-04-30 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

Also, I can't get rid of `arrange()` for tsibble,  `dplyr_row_slice()` helps me to remove `filter()` though.

DavisVaughan:
Closes #5165 

`group_split()` has been refactored to accomplish a few goals:

- It now uses `dplyr_row_slice()` to split with. Previously this was `vec_slice()`.

- There was a case where we used `[, i]` rather than 1D `[i]` when `keep = FALSE`. This has been fixed.

- group-splitting a rowwise data frame used to return a list-of rowwise data frames. I believe this is inconsistent, and it should just return a list of tibbles, like with grouped-dfs. The change has been made to make this more consistent.

- Methods for `data.frame`, `grouped_df`, and `rowwise_df` all now share the same implementation

@earowang, this makes `group_split(pedestrian, Sensor)` work since you have a `dplyr_row_slice()` method. You will still need a `group_split.grouped_ts()` method. It should just have to call `NextMethod()` to get the `group_split.grouped_df()` behavior, and then map over the resulting list of tibbles, coercing them to tbl_ts if applicable. I don't think we can do much better here. Adding that method should fix `nest_by()` automatically.

### DIFF
--- a/R/group_split.R
+++ b/R/group_split.R
@@ -112,15 +112,7 @@ group_split.grouped_df <- function(.tbl, ..., keep = TRUE) {
 group_split_impl <- function(data, grouped_data) {
   indices <- group_rows(grouped_data)
 
-  n <- length(indices)
-
-  out <- vector("list", n)
-
-  for (i in seq_len(n)) {
-    index <- indices[[i]]
-    out[[i]] <- dplyr_row_slice(data, index)
-  }
-
+  out <- map(indices, dplyr_row_slice, data = data)
   out <- new_list_of(out, ptype = vec_ptype(data))
 
   out
--- a/R/group_split.R
+++ b/R/group_split.R
@@ -72,11 +72,12 @@ group_split <- function(.tbl, ..., keep = TRUE) {
 
 #' @export
 group_split.data.frame <- function(.tbl, ..., keep = TRUE) {
-  if (dots_n(...)) {
-    group_split(group_by(.tbl, ...), keep = keep)
-  } else {
-    new_list_of(list(.tbl), ptype = .tbl[0L, ])
-  }
+  data <- .tbl
+  grouped_data <- group_by(.tbl, ...)
+
+  data <- group_split_col_slice(data, grouped_data, keep)
+
+  group_split_impl(data, grouped_data)
 }
 
 #' @export
@@ -87,10 +88,13 @@ group_split.rowwise_df <- function(.tbl, ..., keep = TRUE) {
   if (!missing(keep)) {
     warn("keep is ignored in group_split(<rowwise_df>)")
   }
-  new_list_of(
-    map(seq_len(nrow(.tbl)), vec_slice, x = .tbl),
-    ptype = vec_slice(.tbl, 0L)
-  )
+
+  grouped_data <- .tbl
+  data <- as_tibble(.tbl)
+
+  data <- group_split_col_slice(data, grouped_data, keep = TRUE)
+
+  group_split_impl(data, grouped_data)
 }
 
 #' @export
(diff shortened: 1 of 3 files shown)


R/group_split.R
```diff
@@ -72,11 +72,12 @@ group_split <- function(.tbl, ..., keep = TRUE) {
...
 group_split.data.frame <- function(.tbl, ..., keep = TRUE) {
-  if (dots_n(...)) {
-    group_split(group_by(.tbl, ...), keep = keep)
-  } else {
-    new_list_of(list(.tbl), ptype = .tbl[0L, ])
-  }
+  data <- .tbl
+  grouped_data <- group_by(.tbl, ...)
+
+  data <- group_split_col_slice(data, grouped_data, keep)
```
hadley, line 78: I think the function names to more clearly indicate that this is about dropping the grouping variables out of the splits.

hadley: Also, if we're grouping here, can't we just call `group_split.grouped_df()`? And then `group_split_col_slice` can be inlined into `group_split.grouped_df`?

DavisVaughan: > can't we just call group_split.grouped_df()

I don't think so. By design, `group_split.grouped_df()` coerces the grouped-df to tibble and splits that, so you'd end up with a list of tibbles, rather than a list of the tibble subclasses.

---

To me it's about gather two pieces of information: an ungrouped data frame (possibly subclassed) to split, and a grouped data frame holding information about how to split.

With `group_split.data.frame()` you have the ungrouped thing you split on, so you have to generate the grouped data.

With `group_split.grouped_df()` you have the grouped data, but need the ungrouped thing to split on. The best we can do here is make the grouped data into a bare tibble. This is why Earo will need a `group_split.grouped_ts()` method to return a list of tbl_ts.

Once the two pieces of information have been gathered, they share a common implementation in `group_split_impl()`

hadley: Ok, makes sense. Thanks!

PR #5169, opened 2020-04-30
Improvements to dplyr extensibility

### LINKED ISSUE #5161: Fix dplyr_reconstruct genericity
It should be dispatching on `template`. Can also remove `tibble` check now that we no longer rely on tibble methods

hadley:
Incorporates fixes and considerably improves the documentation along the lines suggested by @davisvaughan.

Fixes #5161

### DIFF
--- a/R/generics.R
+++ b/R/generics.R
@@ -1,18 +1,50 @@
-#' Extending dplyr
+#' Extending dplyr with new data frame subclasses
 #'
 #' @description
 #' \Sexpr[results=rd, stage=render]{lifecycle::badge("experimental")}
 #'
 #' These three functions, along with `names<-` and 1d numeric `[`
-#' (i.e. `x[loc]`), provide a minimal interface for extending dplyr to work
-#' with a new subclass of data frame. `dplyr_rows_slice()` and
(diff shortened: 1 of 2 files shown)


R/generics.R
```diff
@@ -1,18 +1,50 @@
...
+#'
+#' * If you have __scalar__ attributes that depend on __columns__, implement a
+#'   `dplyr_reconstruct()` method and a 1d `[` method. For example, if your
+#'   class requires that certain columns be present, your method should return
+#'   a data.frame or tibble when those columns are removed.
+#'
+#' * If your attributes are __vectorised__ over __rows__, implement a
+#'   `dplyr_row_slice()` method. This gives you access to `i` so you can
+#'   modify the row attribute accordingly. You'll also need to think carefully
+#'   about how to recompute the attribute in `dplyr_reconstruct()`, and
```
DavisVaughan, line 37: I think this note about `dplyr_reconstruct()` is a nice addition. I wasn't sure how to provide specific advice on it, but it will definitely need to be considered if you have a row vectorized attribute and something like a left-join is done

PR #5175, opened 2020-05-01
Simplify `group_split()` using `ungroup()`, not `as_tibble()`

DavisVaughan:
Follow up to #5167 

I have since realized that `group_split()` can be implemented in such a way that tsibble likely won't need to supply any methods for it.

The key was to switch a usage of `as_tibble()` for `ungroup()` in `group_split.grouped_df()`. Ungrouping is generic, so when a `grouped_ts` is ungrouped, it can correctly be dropped down to a `tbl_ts`. The previous usage of `as_tibble()` was incorrectly forcing this to be a bare tibble.

So now if tibble subclasses does both of these things, then `group_split()` works out of the box.
- Implement a `group_by.<tbl_subcls>()` method that returns a `<grouped_subcls>`
- Implement a `ungroup.<grouped_subcls>()` method that returns a `<tbl_subcls>`

tsibble does both of these.

---

This also unifies the code a bit more. We still can't make `group_split.data.frame()` go through `group_split.grouped_df()` though. If `gdf <- group_by(df, ...)` has empty dots, we get an ungrouped data frame back and by then immediately calling `group_split(gdf)` we would get stuck in an infinite loop.

---

Here is tsibble working out of the box. This is pinned to the commit before @earowang added her own `group_split.grouped_ts` method. @earowang you probably won't need that anymore after this, unless the behavior of `keep = FALSE` and `[.tbl_ts` is problematic.

<details>

```r
library(tsibble)
# devtools::install_github("tidyverts/tsibble", ref = "41d9e910447d3f8313241396eb3ee554fcede9d0")

pedestrian %>%
  group_split(Sensor) %>%
  .[[1]] %>%
  head(2)
#> # A tsibble: 2 x 5 [1h] <Australia/Melbourne>
#> # Key:       Sensor [1]
#>   Sensor         Date_Time           Date        Time Count
#>   <chr>          <dttm>              <date>     <int> <int>
#> 1 Birrarung Marr 2015-01-01 00:00:00 2015-01-01     0  1630
#> 2 Birrarung Marr 2015-01-01 01:00:00 2015-01-01     1   826

# this happens because of tsibble's `[` method
# a custom `group_split` method would be required to fix it
pedestrian %>%
  group_split(Sensor, keep = FALSE) %>%
  .[[1]] %>%
  head(2)
#> # A tibble: 2 x 4
#>   Date_Time           Date        Time Count
#>   <dttm>              <date>     <int> <int>
#> 1 2015-01-01 00:00:00 2015-01-01     0  1630
#> 2 2015-01-01 01:00:00 2015-01-01     1   826

pedestrian %>%
  group_by(Sensor) %>%
  group_split() %>%
  .[[1]] %>%
  head(2)
#> # A tsibble: 2 x 5 [1h] <Australia/Melbourne>
#> # Key:       Sensor [1]
#>   Sensor         Date_Time           Date        Time Count
#>   <chr>          <dttm>              <date>     <int> <int>
#> 1 Birrarung Marr 2015-01-01 00:00:00 2015-01-01     0  1630
#> 2 Birrarung Marr 2015-01-01 01:00:00 2015-01-01     1   826

pedestrian %>%
  nest_by(Sensor, .keep = TRUE)
#> # A tibble: 4 x 2
#> # Rowwise:  Sensor
#>   Sensor                                      data
#>   <chr>                         <list<tbl_ts[,5]>>
#> 1 Birrarung Marr                      [14,566 × 5]
#> 2 Bourke Street Mall (North)          [16,414 × 5]
#> 3 QV Market-Elizabeth St (West)       [17,518 × 5]
#> 4 Southern Cross Station              [17,539 × 5]
```

</details>

### DIFF
--- a/R/group_split.R
+++ b/R/group_split.R
@@ -72,15 +72,8 @@ group_split <- function(.tbl, ..., keep = TRUE) {
 
 #' @export
 group_split.data.frame <- function(.tbl, ..., keep = TRUE) {
-  data <- .tbl
-  grouped_data <- group_by(.tbl, ...)
-
-  if (!keep) {
-    cols <- group_vars(grouped_data)
-    data <- drop_cols(data, cols)
-  }
-
-  group_split_impl(data, grouped_data)
+  data <- group_by(.tbl, ...)
+  group_split_impl(data, keep = keep)
 }
 
 #' @export
@@ -92,10 +85,7 @@ group_split.rowwise_df <- function(.tbl, ..., keep = TRUE) {
     warn("keep is ignored in group_split(<rowwise_df>)")
   }
 
-  grouped_data <- .tbl
-  data <- as_tibble(.tbl)
-
-  group_split_impl(data, grouped_data)
+  group_split_impl(.tbl, keep = TRUE)
 }
 
 #' @export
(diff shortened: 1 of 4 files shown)


hadley:
Thanks!

earowang:
Thanks Davis. Yep, I removed `group_split()` for `grouped_ts`, and it works straight way.

PR #5187, opened 2020-05-04
Convert coalesece to use vctrs

### LINKED ISSUE #5186: df_patch (via coalesce) is too picky about types
``` r

df <- tibble(a = 1:3, b = letters[c(1:2, NA)], c = 0.5 + 0:2)
df %>% rows_patch(tibble(a = 2:3, b = "B"))
#> Matching, by = "a"
#> # A tibble: 3 x 3
#>       a b         c
#>   <int> <chr> <dbl>
#> 1     1 a       0.5
#> 2     2 b       1.5
#> 3     3 B       2.5
df %>% rows_patch(tibble(a = c(2, 3), b = "B"))
#> Matching, by = "a"
#> Error: Argument 2 must be an integer vector, not a double vector
```

hadley:
Fixes #5186

I'm slightly nervous about changing this at this point in release cycle, but I think it should be an innocuous change because it only increases the set of valid inputs.


DavisVaughan:
I feel like there is also room for a funs `patch(x, ...)` function that would cast to the type of `x` and recycle to the type of `x`. That feels more appropriate in `rows_patch()` to me, but I haven’t thought much about how the current approach with coalesce could break down

hadley:
We'll switch to the better version in funs in 1.1.0.

PR #5206, opened 2020-05-07
Ensure that `filter()` always calls `dplyr_row_slice()`

DavisVaughan:
Same reasoning as #5205


DavisVaughan:
@hadley same thing here

DavisVaughan:
I'm systematically making my way through them as I add tests to rsample

PR #5221, opened 2020-05-11
Update tidyselect docs

lionel-:
* Import upstream documentation.

* Switch to `where()`.


lionel-:
I had to embed the tidyselect documentation after all, because of this issue with inherited sections: https://github.com/r-lib/roxygen2/issues/1097

And because I made a mistake with the overview Rmd file which is not installed in `inst/` as it should. I'll make a note to fix it once these issues are solved.

PR #5236, opened 2020-05-12
Refactor encoding test helpers and use withr::local_locale()

jennybc:
Alternative to #5231. withr does have support for locale change, so I used that + some probing for a non-UTF-8 locale.

### DIFF
--- a/R/data-mask.R
+++ b/R/data-mask.R
@@ -53,7 +53,9 @@ DataMask <- R6Class("DataMask",
 
       promises <- map(seq_len(ncol(data)), function(.x) expr(promise_fn(!!.x)))
 
-      env_bind_lazy(private$bindings, !!!set_names(promises, names_bindings))
+      suppressWarnings(
+        env_bind_lazy(private$bindings, !!!set_names(promises, names_bindings))
+      )
 
       private$mask <- new_data_mask(private$bindings)
       private$mask$.data <- as_data_pronoun(private$mask)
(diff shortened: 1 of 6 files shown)


tests/testthat/helper-encoding.R
```diff
@@ -28,31 +28,41 @@ get_alien_lang_string <- function() {
...
-    }
+non_utf8_encoding <- function(enc = NULL) {
+  if (!l10n_info()$`UTF-8`) {
+    return(Sys.getlocale("LC_CTYPE"))
+  }
+  enc <- enc %||% c(
+    "en_US.ISO8859-1",
+    "en_US.ISO8859-15",
+    "fr_CH.ISO8859-1",
+    "fr_CH.ISO8859-15"
```
jennybc, line 51: We could add more if we see skips on a system where there are non-UTF-8 locales available.

jennybc:
I saw withr is already used elsewhere below `tests/` so I figured it was fair game.

R/data-mask.R
```diff
@@ -53,7 +53,9 @@ DataMask <- R6Class("DataMask",
 
       promises <- map(seq_len(ncol(data)), function(.x) expr(promise_fn(!!.x)))
 
-      env_bind_lazy(private$bindings, !!!set_names(promises, names_bindings))
+      suppressWarnings(
```
lionel-, line 56: This makes me feel anxious.

lionel-: Also it needs a comment about the kind of anticipated warnings. Also if this needs to be fixed here it might need to be fixed upstream.

jennybc: This is from @romainfrancois. I had asked a similar question in slack.

PR #5266, opened 2020-05-23
Use private class for rows errors for now

krlmlr:
CC @lionel-.

### DIFF
--- a/R/rows.R
+++ b/R/rows.R
@@ -91,7 +91,7 @@ rows_insert.data.frame <- function(x, y, by = NULL, ..., copy = FALSE, in_place 
   idx <- vctrs::vec_match(y[key], x[key])
   bad <- which(!is.na(idx))
   if (has_length(bad)) {
-    abort(class = "dplyr_rows_insert_duplicate",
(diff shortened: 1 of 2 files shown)


krlmlr:
Good catch!

hadley:
Per our standard policy, we shouldn't even be throwing classed errors here.

krlmlr:
I feel that's a step backward. Can we adapt the standard policy?

lionel-:
It feels like there really are only two error classes here, `dplyr_error_rows_missing` and `dplyr_error_rows_duplicate`?

krlmlr:
Sure, we can aggregate the error classes.

lionel-:
It's probably best to just remove all the classes for now then, especially since testthat does not understand the `:::` separator yet.

krlmlr:
Let's revisit later.

lionel-:
I think we should either remove the classes or use triple colons.

lionel-:
Or at least use the `dplyr_error_` namespace if we keep them in 1.0.

PR #5277, opened 2020-05-27
Strip attributes before reconstruction

lionel-:
Makes it easier to implement methods and prevents unexpected leaking of irrelevant attributes.

### DIFF
--- a/R/generics.R
+++ b/R/generics.R
@@ -179,6 +179,13 @@ dplyr_col_modify.rowwise_df <- function(data, cols) {
 #' @export
 #' @rdname dplyr_extending
 dplyr_reconstruct <- function(data, template) {
+  # Strip attributes before dispatch to make it easier to implement
+  # methods and prevent unexpected leaking of irrelevant attributes.
+  data <- dplyr_new_data_frame(data)
+  return(dplyr_reconstruct_dispatch(data, template))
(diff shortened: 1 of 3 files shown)


hadley:
What's the motivation for this? I'd prefer not to change this code at this time unless you have a very strong argument.

lionel-:
It's to make it simpler to implement methods, and have more predictable outputs even with implementations that expect bare data frames. Currently we are sloppily sending things like tibbles as `data` argument. It doesn't have any impact on the four packages that implement reconstruct.

lionel-:
Thanks to this PR I have just noticed the `dplyr_reconstruct.sf` method I have implemented yesterday is perhaps incorrect.

After bind-rows, vctrs has already reconstructed the data frame in case of a same-attributes fallback for ptype2 (sf doesn't implement ptype2). This means my reconstruct method gets a full sf data frame as `data` input. Then it calls `st_as_sf()`, which is a generic whose `sf` method is a no-op. In other circumstances, the input might have a different class and the control flow will change by dispatching on `st_as_sf.data.frame()`.

This illustrates why I think we should fix this before too many packages implement `dplyr_reconstruct()`.

lionel-:
Following Slack discussion, merging now and we can revert later if revdeps reveal problems.

PR #5331, opened 2020-06-14
Use `.env$i` in `tbl_if_vars()`

### LINKED ISSUE #5330: Mutate_if breaks when column name is "i"
`mutate_if` is broken (the condition applies to all columns) when a column is named `i`:

``` r
dplyr::mutate_if(tibble::tibble(i = "a", j = 1), is.character, 
    tidyr::replace_na, replace = "")
#> # A tibble: 1 x 2
#>   i     j    
#>   <chr> <chr>
#> 1 a     1
```

<sup>Created on 2020-06-12 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

---

I noticed this for the first time with version 1.0

markfairbanks:
Fixes #5330

### DIFF
--- a/R/colwise.R
+++ b/R/colwise.R
@@ -216,7 +216,7 @@ tbl_if_vars <- function(.tbl, .p, .env, ..., .include_group_vars = FALSE) {
   selected <- new_logical(n)
 
   for (i in seq_len(n)) {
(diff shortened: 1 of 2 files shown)


lionel-:
Awesome, thanks for the report, diagnostic, and fix!

PR #5354, opened 2020-06-23
Use `vec_proxy_order()` in place of `vec_proxy_compare(relax = TRUE)`

DavisVaughan:
Follow up to https://github.com/r-lib/vctrs/pull/1155 which removes the experimental `relax` argument in favor of `vec_proxy_order()`

We get errors without this since `ellipsis::check_dots_empty()` is run on the vctrs side, and having the `relax` arg is an issue. Using `vec_proxy_order()` is also now the way to order list-columns.

Note that this bumps dplyr to use dev vctrs

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -41,7 +41,7 @@ Imports:
     tibble (>= 2.1.3),
     tidyselect (>= 1.1.0),
     utils,
-    vctrs (>= 0.3.1.9000)
+    vctrs (>= 0.3.1)
 Suggests: 
     bench,
     broom,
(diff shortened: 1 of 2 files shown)


hadley:
We want to release dplyr next week, so I don't think we can depend on dev vctrs.

lionel-:
Fixes #5362

hadley:
Can you please make this condition on version of vctrs so we can include in 1.0.1?

DavisVaughan:
Mac devel error is due to the `Non-file package-anchored link(s)` roxygen links issue

PR #5383, opened 2020-07-06
Prefer withCallingHandlers() when rethrowing

krlmlr:
because it's faster than `tryCatch()` and reduces the depth of the call stack.


lionel-:
oops already done in #5322

PR #5395, opened 2020-07-13
Use ! instead of - to negate the selection in docs

### LINKED ISSUE #5392: Getting started vignette should use `!` to negate the selection instead of `-`
dplyr 1.0.0 introduced boolean operators, but it seems they are not advertised. At least, since `-` is no more documented, the vignette should use `!` instead.

c.f. https://tidyselect.r-lib.org/reference/language.html

If this sounds good, I'm happy to create a pull request for this.

yutannihilation:
Fix #5392 

Btw, it seems dplyr still uses roxygen 7.1.0. I think you should use 7.1.1 before the next CRAN submission to avoid crossref warnings on R-devel.

### DIFF
--- a/R/doc-params.R
+++ b/R/doc-params.R
@@ -112,7 +112,7 @@ NULL
 #' *   If you have a character vector of column names, use `all_of()`
 #'     or `any_of()`, depending on whether or not you want unknown variable
 #'     names to cause an error, e.g `select(df, all_of(vars))`,
-#'     `select(df, -any_of(vars))`.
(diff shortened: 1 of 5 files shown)


lionel-:
Thanks!

hadley:
@yutannihilation btw CRAN backed out of that change, so roxygen2 7.1.2 will too.

yutannihilation:
Wow, I didn't know it... Thanks for the information!

PR #5405, opened 2020-07-14
Improve detection of modern order proxy

lionel-:
In the end the next version of vctrs will not include the order proxy. So we detect it from the formal signature of `vec_proxy_compare()` instead of the vctrs version. Edit: Now detecting presence of `vec_proxy_order()` in the vctrs namespace.

### DIFF
--- a/R/zzz.r
+++ b/R/zzz.r
@@ -1,3 +1,13 @@
+
+# FIXME: Temporary util until the API change from
+# https://github.com/r-lib/vctrs/pull/1155 has been on CRAN for long
+# enough to reasonably expect that users have upgraded
+dplyr_proxy_order <- function(x) NULL
+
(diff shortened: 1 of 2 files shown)


R/zzz.r
```diff
@@ -1,3 +1,13 @@
+
+# FIXME: Temporary util until the API change from
+# https://github.com/r-lib/vctrs/pull/1155 has been on CRAN for long
+# enough to reasonably expect that users have upgraded
+dplyr_proxy_order <- function(x) NULL
```
hadley, line 5: I don't know if it's worth it, but you could use `delayedAssign()` instead of assignment on load. The chief advantage is that then all this code would live `arrange.R`

lionel-: Good point, I always forget about it!

PR #5450, opened 2020-07-31
Update inner_join description

MikeJohnPage:
The description for `inner_join` and `full_join` are the same - indicating that all rows from x and y are joined. This is not the case with an `inner_join`. I have updated the description to highlight the fact that an `inner_join` only joins all rows from x and y **where there is a match in the key**.

### DIFF
--- a/R/join.r
+++ b/R/join.r
@@ -4,7 +4,7 @@
 #' The mutating joins add columns from `y` to `x`, matching rows based on the
 #' keys:
 #'
-#' * `inner_join()`: includes all rows in `x` and `y`.
+#' * `inner_join()`: includes all rows in `x` and `y` that match.
 #' * `left_join()`: includes all rows in `x`.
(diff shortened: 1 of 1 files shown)


hadley:
I think the existing doc is technically correct, but extremely confusing 😄. The intent is something like "includes all rows that occur in both `x` and `y`"

MikeJohnPage:
Ah, I see what the docs were trying to portray now. But yes, slightly confusing! Maybe its better to mirror the [dplyr cheat sheet](https://github.com/rstudio/cheatsheets/blob/master/data-transformation.pdf):

- **Left join**: Join matching values from y to x.
- **Right join**: Join matching values from x to y.
- **Inner join**: Join data. Retain only rows with matches.
- **Full join**: Join data. Retains all values, all rows.

I feel the distinctions made for `inner_join` and `full_join` above are much clearer :blush:

PR #5504, opened 2020-08-28
feat: new .keep_last argument

igorkf:
I created this new  argument called `.keep_last` to be able to keep the last duplicated row, because currently the `distinct()` function always keep the first duplicated row.


igorkf:
I did an error.

PR #5607, opened 2020-11-16
Add keep argument to inner_join

### LINKED ISSUE #5581: Add 'keep' argument to inner_join?
The `keep` argument, to preserve join key columns from both the `x` and `y` data frames (when the name differs), is available for all mutate joins except `inner_join`. It looks like it should be straightforward to add in the existing call to `join_mutate()`.

Unlike with left/right/full joins, the paired key columns in an inner join will always be identical so admittedly it's less useful to retain both, but still potentially useful in some applications (e.g. to facilitate subsequent joins, or for later data validation).

``` r

df1 <- tibble(
  id1 = letters[1:3],
  x = 1:3
)

df2 <- tibble(
  id2 = letters[1:2],
  y = 1:2
)

# use 'keep' arg to retain both join key columns (id1 and id2)
left_join(df1, df2, by = c("id1" = "id2"), keep = TRUE)
#> # A tibble: 3 x 4
#>   id1       x id2       y
#>   <chr> <int> <chr> <int>
#> 1 a         1 a         1
#> 2 b         2 b         2
#> 3 c         3 <NA>     NA

# keep arg also avail for full_join, right_join, and nest_join
# but inner_join doesn't have keep arg so can't retain join key from y
inner_join(df1, df2, by = c("id1" = "id2"), keep = TRUE)
#> # A tibble: 2 x 3
#>   id1       x     y
#>   <chr> <int> <int>
#> 1 a         1     1
#> 2 b         2     2
```

<sup>Created on 2020-10-31 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

patrickbarks:
Enables 'keep' argument for `inner_join()`, to preserve join key columns from both the x and y data frames when the name differs. Closes #5581


R/join.r
```diff
@@ -67,8 +67,8 @@
 #'   `y`, these suffixes will be added to the output to disambiguate them.
 #'   Should be a character vector of length 2.
 #' @param keep Should the join keys from both `x` and `y` be preserved in the
-#'   output? Only applies to `nest_join()`, `left_join()`, `right_join()`, and
-#'   `full_join()`.
+#'   output? Only applies to `nest_join()`, `inner_join()`, `left_join()`,
```
hadley, line 70: Does this now apply to all join functions? (or at least functions documented here)

hadley (commented):
Looks great — can you please add a bullet to the top of `NEWS.md`? It should briefly [describe the change](http://style.tidyverse.org/news.html) and end with `(@yourname, #issuenumber)`.

R/join.r
patrickbarks: Yes, all the mutate join functions now have a `keep` arg so I think the "Only applies to..." part can be removed here. The nest join doc inherits the same parameter description. Here too I think the "Only applies to..." part isn't necessary, as it's clear enough that this doc is only for nest_join (which does already have `keep`).

hadley (approved):

hadley:
Thanks!

PR #5610, opened 2020-11-18
Add ... to signature of db_list_tables

MichaelChirico:
We have an extension which needs further arguments. <strike>I'll file another PR so as not to break `dbplyr` shortly.</strike>

Actually I see that as of https://github.com/tidyverse/dbplyr/pull/499, `dbplyr` no longer uses this generic. Will it be deprecated here as well?

### DIFF
--- a/R/dbplyr.R
+++ b/R/dbplyr.R
@@ -40,6 +40,7 @@
 #'  backend to throw an error if a table exists when it shouldn't.
 #' @name backend_dbplyr
 #' @param con A database connection.
+#' @param ... Arguments passed to other methods.
 #' @keywords internal
 NULL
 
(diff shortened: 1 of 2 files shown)


hadley:
@MichaelChirico the deprecation plan is described in https://dbplyr.tidyverse.org/articles/backend-2.html. I think you probably just want to modify the dbplyr generic.

MichaelChirico:
Thanks... closing here. Will come back to this later.

MichaelChirico:
@hadley looking at this again, I'm a bit confused. Did you mean to say to modify the _`dplyr`_ generic?

`db_list_tables` isn't a part of `dbplyr` anymore from what I'm seeing at HEAD...

hadley:
Sorry, I forgot which generics went where — instead of `db_list_tables`, dbplyr now uses `DBI::dbListTables()`. But it looks like that was only ever used as part of `db_write_table()` (https://dbplyr.tidyverse.org/articles/backend-2.html#unused-generics-1) so you might actually want to define a `dbWriteTable()` method. What are you trying to do?

MichaelChirico:
We have a method that takes another argument to `db_list_tables`, but the generic doesn't have `...`, so I was just going to add that to facilitate extensions.

We use `db_list_tables` directly for interactive cases, so `dbWriteTable` feels... indirect. Any thoughts?

PR #5824, opened 2021-03-23
Supply proper caller environments to dplyr masks

lionel-:
Usually does not matter but the caller env will be useful for fixing #5815.

Since they are exported, `group_by_prepare()` and `distinct_prepare()` use `caller_env(2)` as default. Best to supply the caller env explicitly though.


romainfrancois (approved):
Thanks :-)

PR #5835, opened 2021-04-05
Tunnel `across()` arguments for key deparsing

lionel-:
This way we don't need these `.call` arguments. The use of `{{` is optional (and should probably be considered internal for now).

TODO: One test is failing. Need to copy code of `sexp_address()` from rlang into dplyr.

Issues of overcaching: With this PR we now include the quosure environment in the keys. We are still overcaching in some cases (env-expressions in `.cols` and function calls generating `.fns`). Would be nice to fix for consistency and predictability.


romainfrancois (commented):
This looks like a better option than `.call=`. We should measure the cost of this new `key_deparse()` though.

PR #5857, opened 2021-04-20
Expand `if_any()` and `if_all()` expressions

lionel-:
Branched from #5856.

Following Romain's approach for expanding `across()` expressions, this PR expands `filter(if_any(c(x, y), ~ f(.x)))` to `filter(f(x) | f(y))`. 

```r
gdf <- tibble(x = seq_len(1e5), y = x) %>% group_by(x)

bench::mark(
  expr = gdf %>% filter(y %% 10 == 0),
  across = gdf %>% filter(if_any(y, ~ . %% 10 == 0))
)[1:8]

#> # Before
#>   expression      min   median `itr/sec` mem_alloc `gc/sec` n_itr
#>   <bch:expr> <bch:tm> <bch:tm>     <dbl> <bch:byt>    <dbl> <int>
#> 1 expr       234.65ms 242.25ms    4.12      11.4MB     2.75     3
#> 2 across        1.05m    1.05m    0.0158   435.6MB     1.72     1

#> # After
#>   expression     min  median `itr/sec` mem_alloc `gc/sec` n_itr  n_gc
#>   <bch:expr> <bch:t> <bch:t>     <dbl> <bch:byt>    <dbl> <int> <dbl>
#> 1 expr         241ms   246ms      3.97    11.4MB     21.0    20   106
#> 2 across       240ms   246ms      3.95    11.6MB     22.1    20   112
```

In the future we could export the expander. Then `if_any()` could be implemented in backends like dbplyr by calling the expander. Ideally we would do the expansion in the generic, this way the feature comes for free for all backends.

Like in the `across()` expansion, we can't expand when arguments are passed through `...`. I'm starting to think we should embrace a "macro-centric" design and deprecate `...`. This would ensure consistent optimisation in the dplyr backend. If other backends end up using the expanders, then this would also ensure consistent behaviour across backends.

Speaking of macro-centric design, this PR changes the non-expanded versions of `if_any()` and `if_across()` to use the base operators `|` and `&`. This ensures the behaviour is consistent across optimised and non-optimised code paths. This is a small behaviour change because we are no longer using vctrs to coerce the inputs.


romainfrancois (commented):
Looks good. I'm not sure how this will conflict with #5847 but I can pick it up once this pr is merged.

PR #5868, opened 2021-04-28
Add a `.locale` argument to `arrange()` and use radix ordering

### LINKED ISSUE #4962: Performance drop-off for `arrange()`

DavisVaughan:
Closes #4962 
Closes #5090 
Part of #5808 

Requires this vctrs PR https://github.com/r-lib/vctrs/pull/1375

The above vctrs PR switches `vec_order()` to always use radix ordering. This propagates through to dplyr in 3 places:

- `arrange()`
- `group_by()`
- `with_order()`

This PR tackles `arrange()` by providing a new `.locale` argument with 3 possibilities:

- `NULL`, the default, for C locale
- A single string, like `"en_US"`, to use a specific locale, backed by and requiring stringi
- A formula/function to generate the sort key (i.e. `stringi::stri_sort_key()`, or `tolower()`)

This PR is currently focused on the implementation, I've left TODOs for the news bullet and for documentation of locale handling.

The snapshot tests are a little noisy. Dev vctrs requires dev rlang, which tweaked how deparsing with `~` works, changing a few snapshot tests here.

Build errors seem to be from the fact that rlang now exports ellipsis functions, causing some warnings

---

This would be a meaningful breaking change, as the system locale is no longer being respected. However, there are multiple benefits from this change:
- Sorting character vectors in the C locale is much much faster
- Even generating a sort key and then sorting in the C locale is faster than the current approach
- This removes dependence on an environment variable (LC_COLLATE), which we generally advocate in favor of explicitly specifying arguments. In other words, this gets us better consistency across platforms.
- This increases consistency with data.table, who always uses the C locale
- This increases consistency with stringi (and you could make an argument for stringr too)

For English, the biggest change is that uppercase letters now sort before any lowercase letters. Previously it used a natural ordering of `c("a", "A", "b", "B")`.

For other languages, sorting directly in the C locale is often not very meaningful, so they would have to supply a locale identifier.


R/arrange.R
```diff
@@ -68,22 +67,42 @@ arrange <- function(.data, ..., .by_group = FALSE) {
 
 #' @param .by_group If `TRUE`, will sort first by grouping variable. Applies to
 #'   grouped data frames only.
+#' @param .locale The locale to sort character vectors in.
+#'
+#'   - If `NULL`, the default, character vectors are sorted in the C locale.
```
hadley, line 72: Mention that this is the fastest?

tests/testthat/_snaps/across.md
```diff
@@ -25,7 +25,7 @@
 # if_any() and if_all() aborts when predicate mistakingly used in .cols= (#5732)
 
     Code
-      (expect_error(filter(df, if_any(~.x > 5))))
+      (expect_error(filter(df, if_any(~ .x > 5))))
```
hadley, line 28: I wonder what's causing these diffs

hadley (commented):
LGTM — we'll just need to leave this sit until we figure out a unified strategy.

tests/testthat/_snaps/across.md
DavisVaughan: I made @lionel- add a news bullet 🙃  https://github.com/r-lib/rlang/commit/43413afcaab59a6d8f5ca6c4bd81c631a84b95ba

tests/testthat/test-arrange.r
```diff
@@ -79,14 +79,46 @@ test_that("arrange handles S4 classes (#1105)", {
...
+  x <- c("casa", "\u00e1rbol", "zona", "\u00f3rgano")
+  df <- tibble(x = x)
+
+  res <- arrange(df, x, .locale = "es_MX")
+  expect_identical(res$x, x[c(2, 1, 4, 3)])
+})
+
+test_that("arrange validates `.locale`", {
+  df <- tibble()
+  expect_snapshot_error(arrange(df, .locale = 1))
```
lionel-, line 119: I think we now generally use:

```r
expect_snapshot({
  (expect_error(foo()))
  (expect_error(bar(), class = "bar"))
})
```

This way we see the relevant `Code` section in the snapshot next to the output.

R/arrange.R
```diff
@@ -131,24 +153,47 @@ arrange_rows <- function(.data, dots) {
...
+
+    fn <- function(x) {
+      stringi::stri_sort_key(x, locale = locale)
+    }
+
+    return(fn)
+  }
+
+  if (is_formula(locale)) {
+    locale <- as_function(locale)
```
lionel-, line 191:
```suggestion
    return(as_function(locale))
```

tests/testthat/test-arrange.r
DavisVaughan: What is `expect_snapshot_error()` used for then?

lionel-: It was added before we discovered the `expect_snapshot({ (expect_error()) })` pattern.

hadley: I think in most cases it's fine to use:

```R
expect_snapshot(error = TRUE, {
  foo()
  bar()
})
```

I think you only need to use `(expect_error(foo()))` when it's particularly important what the error class is.

lionel-: This will only check that one of `foo()` or `bar()` fails, not both of them, which is a big downside I think.

hadley: Right, but in most cases you just want to see all the error messages together in one place, and the chance of one error failing to fail on CRAN is slim.

R/arrange.R
```diff
@@ -68,22 +67,43 @@ arrange <- function(.data, ..., .by_group = FALSE) {
...
+#'
+#'   - If a single string is supplied, then this will be used as the locale
+#'     identifier to sort with. For example, `"en_US"` will sort with the
+#'     American English locale. This requires the stringi package. Use
+#'     [stringi::stri_locale_list()] to generate a list of possible locale
+#'     identifiers.
+#'
+#'   - If a formula or function of one argument is supplied, it is applied to
+#'     all character columns to transform them into alternate strings that
+#'     sort correctly in the C locale. For example, `~ tolower(.x)` results
```
lionel-, line 83: I think we should probably not showcase this usage too much, and instead focus on locale key generators. It's probably better to apply transformations like `tolower()` in the `arrange()` call itself. This way more users are aware they can apply transformations in `arrange()` calls, compose with `across()`, and then the resulting transformation can easily be used with specific locales.

```r
df <- tibble(x = c(1, 1, 1), y = c("foo", "bar", "FOO"))

arrange(df, x, tolower(y))
arrange(df, x, across(where(is.character), tolower))

# Not useful in that case but another transformation might be
arrange(df, x, across(where(is.character), tolower), .locale = "en_US")
```

DavisVaughan: That sounds fine to me. Should we disallow it altogether?

lionel-: Good idea, it's probably better to be conservative in the UI at first

DavisVaughan:
Updated so that:

- `.locale = NULL` defaults to `"en"` if stringi is available, otherwise it falls back to the C locale with a warning
- `.locale = "C"` is now a valid option

DavisVaughan:
Updated so that:

- We have a new exported `dplyr_locale()` helper. It defaults to `"en"` if stringi is installed, otherwise it falls back to `"C"` with a warning. Users can set `options(tidyverse.locale_collation)` if they want to globally override the default locale.
- `.locale` in `arrange()` now defaults to `dplyr_locale()`
- `.locale` no longer allows `NULL`
- The scoped variants of `arrange()` have gained `.locale` to ease the transition a little

DavisVaughan:
Closing in favor of the cleaner https://github.com/tidyverse/dplyr/pull/5942

Issue #3293, opened 2018-01-08
Feature Request: Port CRAN's fuzzyjoin Funtions to dplyr


jtelleriar:
I would suggest porting some of CRAN's fuzzyjoin package functions to dplyr (With proper Contribution Acknowledgement):

https://cran.r-project.org/web/packages/fuzzyjoin/index.html

Of special interest is "fuzzy_join" function, which allows to do SQL queries such as these:
```
SELECT
    REPLACE(PATH, 'W:\\', '\\\\naswoav.usr.corp.gamesa.es\\WO_AV_DATA\\') AS PATH
  FROM
    dispoin.t_DISPOIN_DIR AS A
    INNER JOIN dispoin.t_scadas AS B
      ON
        A.PATH LIKE CONCAT('%', ScadaCode, '%')
        AND (A.PATH LIKE '%2016%' OR A.PATH LIKE '%2017%')
        -- También se puede usar  A.PATH REGEXP '2016|2017';
        AND A.PATH NOT LIKE '%Data_10M%'
        AND A.PATH NOT LIKE '%rejectAL%'
        AND A.PATH NOT LIKE '%REPROCESSED_FILES%'
        AND A.PATH NOT LIKE '%GA677_201308_AL.csv.s3542017.zip%';
```

krlmlr:
Thanks. Are you suggesting to run the fuzzy join directly on the database?

jtelleriar:
Either in database or in R itself.

However, when you query an SQL database I guess you could always do a cheat like this:
```
t_DISPOIN_DIR <- tbl(con, "t_DISPOIN_DIR")

t_DISPOIN_DIR %>%
sql(
"
SELECT
    REPLACE(PATH, 'W:\\', '\\\\naswoav.usr.corp.gamesa.es\\WO_AV_DATA\\') AS PATH
  FROM
    dispoin.t_DISPOIN_DIR AS A
    INNER JOIN dispoin.t_scadas AS B
      ON
        A.PATH LIKE CONCAT('%', ScadaCode, '%')
        AND (A.PATH LIKE '%2016%' OR A.PATH LIKE '%2017%')
        -- También se puede usar  A.PATH REGEXP '2016|2017';
        AND A.PATH NOT LIKE '%Data_10M%'
        AND A.PATH NOT LIKE '%rejectAL%'
        AND A.PATH NOT LIKE '%REPROCESSED_FILES%'
[...]

PR #5910, opened 2021-06-08
Flexible joins

### LINKED ISSUE #2240: join_by(): Syntax for generic joins
https://github.com/hadley/dplyr/issues/557#issuecomment-53483154 and https://github.com/hadley/dplyr/issues/378#issuecomment-212609049 propose a syntax for generic and rolling joins:

```r
left_join( 
  FundMonths, Returns, 
  join_by(FundID == FundID, yearmonth > gmonth + 3, yearmonth <= gmonth + 15)
)

left_join( 
  events, days,
  join_by(collector_id == collector_id, event_timestamp >= largest(day))
)
```

We can pass this to the SQL engine (and perhaps to data tables) with relatively little work, the main challenge will be to implement this for data frames.

### LINKED ISSUE #5413: Duplicate control of rows in join commands
It would be incredible if you could specify which kind of duplicates in the join variable(s) is(are) acceptable as in [the STATA merge command](https://www.stata.com/manuals13/dmerge.pdf). 

This means having the possibility to decide whether it's ok to have: 
- Duplicates in the join variables of the master data but not in the join variables of the using data (a m:1 join)
- Duplicates in the join variables of the using data but not in the join variables of the master data (a 1:m join)
- Duplicates in both cases (an m:m join)
- No duplicates (a 1:1 join)

I work in education, and some times data-sets were poorly digitized because some students appear to have been enrolled in more than one school (duplicates in the join variable).  When that happens both in the master and using data, those students multiply exponentially in the resulting data-frame, as it happens now with, for example, left_join.  You may not want that. Also, if you are not cautious enough, you may not pay attention to duplicate students, that "ninjaly" add more observations to your results. 

STATA used to have the same issue and only had two commands "merge" or "join" somehow similar to R, and did not let you decide when duplicates are acceptable (i.e duplicate information in the using data when merging students to school characteristics, school id is the join variable) and when it may not be acceptable (i.e duplicate information in the master and using data when merging a student list to a student characteristic, student id is the join variable). At some point, they made that feature feasible and it's great because it allows you to merge and make quality control in one single step. 

Thank you for your time and amazing work. Hope this helps.

### LINKED ISSUE #5661: give `*_join()` more control on duplicated and unmatched rows
As @lionel- nicely proposed in #3967 it would be great to have more control on what the `*_join()` functions do on duplicated or unmatched rows. Proposed arguments:

* `.unmatched = c("keep", "drop", "error")`
* `.duplicates = c("error", "product")`

for example a 1:1 join would then simply be `.unmatched = "error"` and `.duplicates = "error"`.

As the arguments should rather be added to the existing `*_join()` functions the "keep" and "drop" options for `.unmatched` don't really make sense for all joins. Rather, one would have

`left/right/full_join()` -> `.unmatched = c("keep", "error")`
`inner_join()` -> `.unmatched = c("drop", "error")`

### LINKED ISSUE #5914: Allow users to validate the cardinality in joins
One thing I often do in production code after a merge is validating my assumptions on unique keys and relationship between them in both merge inputs (i.e. [cardinality](https://en.wikipedia.org/wiki/Cardinality_%28data_modeling%29) checks). For example, if I have duplicated keys in one data frame and perform an inner join, I end up with more rows that both input tables had, which might come as a surprise and can cause downstream problems if I don't manually assert this. 

In a script, checking these conditions and raising a (helpful) error can take a few lines of code quickly, let alone the verbosity if you have multiple sequential joins. For that reason, I propose to offer the user the option to opt in for cardinality validation as part of the join, very much like Python's [pandas implemented this](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DataFrame.merge.html):

```python
import pandas
pandas.merge(..., validate = 'one_to_one')
```

In {dplyr}, it could be

```r
dplyr::inner_join(..., validate = "1:N")
```

Maybe we could the official abbreviations (not sure they even exist, maybe case insensitive), e.g. `1:n`, `m:n` etc.

If this is out of scope for {dplyr}, maybe it's something for @krlmlr in {dm}?

DavisVaughan:
Closes #5914 
Closes #5661 
Closes #5413 
Closes #2240 (a whopping 5 years old! with 138 thumbs up!)

# Reading order

- NEWS

- `join-by.R`
   - To read `join_by()` docs / examples and to get a general idea of how it works
   - I've tried to error intelligently for many bad user inputs in the parsing code. This probably doesn't need quite as deep of a review.

- `join.R` 
   - To read updated docs
   - Note that all joins go through `join_rows()` now

- `join-rows.R`
   - This basically remaps dplyr arguments to `vec_matches()` arguments and calls `vec_matches()`
   - Many errors are rethrown from vctrs with more applicable error messages

- `join-cols.R`
   - To see how `keep = NULL` is handled
   - `standardise_join_by()` was replaced by `as_join_by()` and `join_by_common()` in `join-by.R`

- Tests
   - I've added quite a few for the `join_by()` parsing checks
   - I've also added snapshot tests for the rethrown errors from `dplyr_matches()`
   - It's worth noting that a number of tests needed an explicit `multiple = "all"` now to silence the new warning

# Summary

This PR implements has two main purposes:

1. It adds `join_by()` for creating a _join specification_ which can be specified as the `by` argument to any join. This allows:
    - Equi join specification with unnamed left and right hand sides, i.e. `join_by(date1 == date2)`
 
    - Specification of non-equi joins like `join_by(id, date1 >= date2)`

    - Specification of rolling joins with `join_by(id, preceding(date1, date2))` and `join_by(id, following(date1, date2))`
    
    - Shortcuts for some complex non-equi joins: `between()`, `overlaps()`, `within()`. 

    - There is a restriction that the left and right hand side of the join conditions have to be symbols or strings, i.e. you can't do `join_by(x + 1 > y)`. It is unclear what the resulting column name should be if you did this, and you can get into order of operation issues, i.e. `!x > y` where `!` has a higher precedence than `>`.

    - To support non-equi joins, `keep` has gain a completely new default value of `NULL`. In non-equi joins like `join_by(sale_date >= commercial_date)`, since the information in the two columns isn't exactly the same, you almost always want to keep both columns. So `keep = NULL` implies `keep = FALSE` for equi conditions and `keep = TRUE` for non-equi conditions. This should be fully backwards compatible.

2. It adds two new "quality control" arguments to most of the join functions that have been requested over the years.
    - `multiple` is a new argument for controlling what happens when a row in `x` matches multiple rows in `y`. It allows `c("all", "first", "last", "warning", "error")`. The default is `NULL`, which uses `"warning"` for equi and rolling joins, where multiple matches are surprising, and `"all"` for cross joins and when there is at least 1 non-equi join condition, where multiple matches are expected. This is a change from the current CRAN dplyr behavior, which never warns.

    - `unmatched` is a new argument for controlling what happens when a row would be dropped because it doesn't have a match. It allows `c("drop", "error")`.

    - Combined with the proposed `enforce()`, this allows for a number of useful checks to ensure that you are doing a 1:1, 1:m, m:1, m:m style join.

# Not implemented

Future enhancements for other PRs (as suggested from comments):
- `join_any()` that would accept arbitrary predicates and would perform a much slower cartesian join + filter using the predicates. It could do it in batches to reduce peak memory. This would require a completely separate engine path for each type of join.

- `join_at()` a tidy select interface for equi joins if you have many columns of the same name to join by. This would have an `as_join_by()` method that translated the result to a `dplyr_join_by` object.

# Examples

<details>
 
``` r
library(dplyr)

set.seed(123)
dates <- as.Date("2019-01-01") + 0:4
needles <- tibble(dates = dates, x = sample(length(dates)))

set.seed(123)
lower <- as.Date("2019-01-01") + sample(6, 5, replace = TRUE)
upper <- lower + sample(2, 5, replace = TRUE)
haystack <- tibble(lower = lower, upper = upper, y = sample(length(lower)))

needles
#> # A tibble: 5 x 2
#>   dates          x
#>   <date>     <int>
#> 1 2019-01-01     3
#> 2 2019-01-02     2
#> 3 2019-01-03     5
#> 4 2019-01-04     4
#> 5 2019-01-05     1
haystack
#> # A tibble: 5 x 3
#>   lower      upper          y
#>   <date>     <date>     <int>
#> 1 2019-01-04 2019-01-06     1
#> 2 2019-01-07 2019-01-08     2
#> 3 2019-01-04 2019-01-05     3
#> 4 2019-01-03 2019-01-05     4
#> 5 2019-01-03 2019-01-05     5

# Non-equi join
# For each row in `needles`, find locations in `haystack` matching the condition
left_join(needles, haystack, by = join_by(dates >= lower, dates <= upper))
#> # A tibble: 12 x 5
#>    dates          x lower      upper          y
#>    <date>     <int> <date>     <date>     <int>
#>  1 2019-01-01     3 NA         NA            NA
#>  2 2019-01-02     2 NA         NA            NA
#>  3 2019-01-03     5 2019-01-03 2019-01-05     4
#>  4 2019-01-03     5 2019-01-03 2019-01-05     5
#>  5 2019-01-04     4 2019-01-04 2019-01-06     1
#>  6 2019-01-04     4 2019-01-04 2019-01-05     3
#>  7 2019-01-04     4 2019-01-03 2019-01-05     4
#>  8 2019-01-04     4 2019-01-03 2019-01-05     5
#>  9 2019-01-05     1 2019-01-04 2019-01-06     1
#> 10 2019-01-05     1 2019-01-04 2019-01-05     3
#> 11 2019-01-05     1 2019-01-03 2019-01-05     4
#> 12 2019-01-05     1 2019-01-03 2019-01-05     5

# If we really cared about including row 2 from `haystack` that wasn't matched by `needles`...
full_join(needles, haystack, by = join_by(dates >= lower, dates <= upper))
#> # A tibble: 13 x 5
#>    dates          x lower      upper          y
#>    <date>     <int> <date>     <date>     <int>
#>  1 2019-01-01     3 NA         NA            NA
#>  2 2019-01-02     2 NA         NA            NA
#>  3 2019-01-03     5 2019-01-03 2019-01-05     4
#>  4 2019-01-03     5 2019-01-03 2019-01-05     5
#>  5 2019-01-04     4 2019-01-03 2019-01-05     4
#>  6 2019-01-04     4 2019-01-03 2019-01-05     5
#>  7 2019-01-04     4 2019-01-04 2019-01-05     3
#>  8 2019-01-04     4 2019-01-04 2019-01-06     1
#>  9 2019-01-05     1 2019-01-03 2019-01-05     4
#> 10 2019-01-05     1 2019-01-03 2019-01-05     5
#> 11 2019-01-05     1 2019-01-04 2019-01-05     3
#> 12 2019-01-05     1 2019-01-04 2019-01-06     1
#> 13 NA            NA 2019-01-07 2019-01-08     2


sales <- tibble(
  sale_id = c("S0", "S1", "S2", "S3", "S4", "S5"),
  sale_date = as.Date(c("2013-2-20", "2014-5-1", "2014-5-1", "2014-6-15", "2014-7-1", "2014-12-31"))
)
commercials <- tibble(
  comm_id = c("C1", "C2", "C3", "C4", "C5"),
  comm_date = as.Date(c("2014-1-1", "2014-4-1", "2014-7-1", "2014-9-15", "2014-9-15"))
)

sales
#> # A tibble: 6 x 2
#>   sale_id sale_date 
#>   <chr>   <date>    
#> 1 S0      2013-02-20
#> 2 S1      2014-05-01
#> 3 S2      2014-05-01
#> 4 S3      2014-06-15
#> 5 S4      2014-07-01
#> 6 S5      2014-12-31
commercials
#> # A tibble: 5 x 2
#>   comm_id comm_date 
#>   <chr>   <date>    
#> 1 C1      2014-01-01
#> 2 C2      2014-04-01
#> 3 C3      2014-07-01
#> 4 C4      2014-09-15
#> 5 C5      2014-09-15

# Rolling join:
# "Find the most recent commercial that aired just before this sale"
left_join(sales, commercials, by = join_by(max(sale_date >= comm_date)))
#> # A tibble: 7 x 4
#>   sale_id sale_date  comm_id comm_date 
#>   <chr>   <date>     <chr>   <date>    
#> 1 S0      2013-02-20 <NA>    NA        
#> 2 S1      2014-05-01 C2      2014-04-01
#> 3 S2      2014-05-01 C2      2014-04-01
#> 4 S3      2014-06-15 C2      2014-04-01
#> 5 S4      2014-07-01 C3      2014-07-01
#> 6 S5      2014-12-31 C4      2014-09-15
#> 7 S5      2014-12-31 C5      2014-09-15

# Notice that rows 6 and 7 have two commercials that aired on the same date.
# Limit that with `multiple`.
left_join(sales[6,], commercials, by = join_by(max(sale_date >= comm_date)), multiple = "first")
#> # A tibble: 1 x 4
#>   sale_id sale_date  comm_id comm_date 
#>   <chr>   <date>     <chr>   <date>    
#> 1 S5      2014-12-31 C4      2014-09-15
left_join(sales[6,], commercials, by = join_by(max(sale_date >= comm_date)), multiple = "last")
#> # A tibble: 1 x 4
#>   sale_id sale_date  comm_id comm_date 
#>   <chr>   <date>     <chr>   <date>    
#> 1 S5      2014-12-31 C5      2014-09-15

# "Find the most recent commercial that aired just before this sale,
# but limit it to within 40 days of the sale"
sales <- mutate(sales, sale_date_lower = sale_date - 40)

left_join(sales, commercials, by = join_by(max(sale_date >= comm_date), sale_date_lower <= comm_date))
#> # A tibble: 6 x 5
#>   sale_id sale_date  sale_date_lower comm_id comm_date 
#>   <chr>   <date>     <date>          <chr>   <date>    
#> 1 S0      2013-02-20 2013-01-11      <NA>    NA        
#> 2 S1      2014-05-01 2014-03-22      C2      2014-04-01
#> 3 S2      2014-05-01 2014-03-22      C2      2014-04-01
#> 4 S3      2014-06-15 2014-05-06      <NA>    NA        
#> 5 S4      2014-07-01 2014-05-22      C3      2014-07-01
#> 6 S5      2014-12-31 2014-11-21      <NA>    NA
```

<\details>


DavisVaughan:
@hadley if you have any initial thoughts about the API here, this is probably a good time for that (see above examples / description). If you have any other wish-list items that immediately come to mind, I'd also welcome them.

tests/testthat/test-join-cols.R
```diff
@@ -1,94 +1,98 @@
...
+  expect_snapshot(error = TRUE, standardise_join_by(NULL, xy, c("a", "b")))
+  expect_snapshot(error = TRUE, standardise_join_by(FALSE, xy, xy))
 
-  expect_snapshot(error = TRUE, join_cols(xy, xy, by = FALSE))
-  expect_snapshot(error = TRUE, join_cols(xy, xy, by = list(1, 2)))
-  expect_snapshot(error = TRUE, join_cols(xy, xy, by = c("x", "x")))
-  expect_snapshot(error = TRUE, join_cols(xy, xy, by = c("x", NA)))
-  expect_snapshot(error = TRUE, join_cols(xy, xy, by = c("aaa", "bbb")))
+  # join vars errors
+  expect_snapshot(error = TRUE, join_cols(xy, xy, by = standardise_join_by(list(1, 2), xy, xy)))
```
hadley, line 90: Might be worth calling `standardise_join_by()` in `join_cols()`? Depending on what you're thinking here, that function might also be called `as_join_by()`? (Assuming you think there's a common underlying class)

hadley (commented):
This looks like exactly what I was imagining for extending the `by` specification 😄 

With this interface in place there are a couple of other join helpers I can imagine:

* `join_any()` which would accept any list of predicates, by first joining a cartesian join (possibly in batches) and then applying the predicates. The performance wouldn't be any better than a cartesian join + a filter, but we since we could do it iteratively, we could reduce peak memory.
* `join_at()` which would use tidyselect to make it easier to join by many common variables

I'm not sure we would actually implement these, but I like having the option.

tests/testthat/test-join-cols.R
DavisVaughan: I've renamed it to `as_join_by()` and refactored a bit, which helped clean up some of the tests.

It needs to live outside `join_cols()`, because I need a join-by object for parts of `join_rows()` as well. It felt wrong to have `join_cols()` create and return the join-by object, since it should be purely focused on the column positions / names.

davidchall:
Please ignore this comment if it's outside scope (or causes more problems than it solves)...

Will this `join_by()` interface be able to accommodate functions in the future? (i.e. #3293) For example,

```r
# geospatial join
inner_join(x, y, by = join_by(~ haversine_distance_km(.x$lat, .x$lon, .y$lat, .y$lon) < 5))

# string distance
inner_join(text, dict, by = join_by(~ stringdist::stringdist(.x$word, .y$word, method = "dl") <= 1))

# IP network membership
inner_join(clients, networks, by = join_by(~ ipaddress::is_within(.x$address, .y$network))
```

This might not be a good first example because this use case is already handled by dedicated geospatial packages, but there are other use cases that require functions in joins. Many SQL flavors support using functions during join clauses ([example](http://sqlfiddle.com/#!9/590676/1/0)).

The [fuzzyjoin](https://github.com/dgrtwo/fuzzyjoin) package provides a generic interface (`fuzzyjoin::fuzzy_join()`) for using custom functions in joins. It uses a Cartesian join, but reduces peak memory by considering each LHS row independently. I'm struggling to imagine how this feature could be compatible with https://github.com/r-lib/vctrs/pull/1387 though.

DavisVaughan:
That would be what Hadley's suggested `join_any()` would do, which would probably come in a separate PR and would be backed by a different implementation that works similar to fuzzyjoin (i.e. cartesian style join that works over batches of `x`).

This is focused on implementing the non-equi binary conditions of `>`, `>=`, `<`, and `<=`, along with max/min rolling joins, as efficiently as possible. It should be at least an order of magnitude faster than the equivalent fuzzyjoin solution, but I understand that there are cases that it won't solve, which is where the slower but more flexible alternative would come in.

davidchall:
Gotcha - thanks a lot for taking the time to explain!

DavisVaughan:
Note: Go back and answer a few popular community questions, like:

- [ ] https://community.rstudio.com/t/tidy-way-to-range-join-tables-on-an-interval-of-dates/7881/2

tests/testthat/_snaps/bind.md
```diff
@@ -2,21 +2,21 @@
...
 
     Code
       df <- bind_cols(list(1, 2))
-    Message <simpleMessage>
+    Message <rlib_message_name_repair>
       New names:
-      * NA -> ...1
-      * NA -> ...2
+      * `` -> `...1`
+      * `` -> `...2`
```
DavisVaughan, line 19: These are just from using dev vctrs

NEWS.md
```diff
@@ -1,5 +1,42 @@
...
+    instead.
+    
+  * `keep` now defaults to `NULL` rather than `FALSE`. `NULL` implies
+    `keep = FALSE` for equi-join conditions, but `keep = TRUE` for non-equi
+    join conditions, since you generally want to preserve both sides of a
+    non-equi join.
+    
+  * `multiple` is a new argument for controlling what happens when a row
+    in `x` matches multiple rows in `y`. For equi joins and rolling joins,
+    where this is usually surprising, this defaults to signaling a `"warning"`,
```
DavisVaughan, line 29: Does this need to be called out more somehow?

R/join-by.R
```diff
@@ -0,0 +1,641 @@
...
+#'   Optionally, a non-equi join condition can be wrapped in `max()` or `min()`
+#'   to specify a rolling join.
+#'
+#'   If a single column name is provided without any join conditions, it is
+#'   interpreted as if that column name was duplicated on each side of `==`,
+#'   i.e. `x` is interpreted as `x == x`.
+#'
+#' @export
+#' @examples
+#' sales <- tibble(
```
DavisVaughan, line 108: It seemed useful to put the rolling / non-equi join examples in the `join_by()` docs rather than cluttering the main join help page

R/join-cols.R
```diff
@@ -1,21 +1,33 @@
...
-  if (!keep) {
+  if (is_null(keep)) {
+    # in x_out, equi key variables need to keep the same name, and non-equi
+    # key variables and aux variables need suffixes for duplicates that appear
+    # in y_out
+    equi <- by$condition == "=="
+    y_aux <- setdiff(y_names, c(by$x[equi], by$y[equi]))
+    x_ignore <- by$x[equi]
+    x_check <- !x_names %in% x_ignore
+    names(x_loc)[x_check] <- add_suffixes(x_names[x_check], c(x_ignore, y_aux), suffix$x)
```
DavisVaughan, line 23: Remember, basically `keep = TRUE` for the non-equi key columns and `keep = FALSE` for the equi key columns. Only the non-equi key columns from `x` may get suffixes.

R/join-cols.R
```diff
@@ -24,58 +36,24 @@ join_cols <- function(x_names, y_names, by = NULL, suffix = c(".x", ".y"), keep
...
-  # key = named location to use for matching
+  # key = named locations to use for matching
   # out = named locations to use in output
   list(
     x = list(key = x_by, out = x_loc),
     y = list(key = y_by, out = y_loc)
   )
 }
 
-standardise_join_by <- function(by, x_names, y_names) {
```
DavisVaughan, line 39: Replaced by `as_join_by()` and `join_by_common()`

R/join-cols.R
```diff
@@ -88,11 +66,19 @@ check_join_vars <- function(vars, names) {
     ))
   }
 
+  if (!is_false(keep)) {
+    # Non-equi columns are allowed to be duplicated when `keep = TRUE/NULL`
+    non_equi <- condition != "=="
+    vars <- c(vars[!non_equi], unique(vars[non_equi]))
```
DavisVaughan, line 72: i.e. `join_by(date >= promo_lower, date <= promo_upper)` duplicates `date` but that's fine

R/join-rows.R
```diff
@@ -1,56 +1,214 @@
...
+  warn_join(
+    message = c(
+      glue("Each row in `x` can match at most 1 row in `y`."),
+      i = glue("Row {i} of `x` matches multiple rows.")
+    ),
+    class = "dplyr_warning_join_matches_multiple"
+  )
+
+  # Cancel `cnd`
+  maybe_restart("muffleWarning")
```
DavisVaughan, line 131: I guess we need this? I'm copying `mutate_cols()` here

R/join.r
```diff
@@ -68,14 +73,42 @@
...
+#'     the data in key columns corresponding to rows that only exist in `y` are
+#'     merged into the key columns from `x`.
 #' @param ... Other parameters passed onto methods.
-#' @param na_matches Should `NA` and `NaN` values match one another?
+#' @param na_matches Should two `NA` or two `NaN` values match?
+#'   - `"na"`, the default, treats two `NA` or two `NaN` values as equal, like
+#'   `%in%`, [match()], and [merge()].
+#'   - `"never"` treats two `NA` or two `NaN` values as different, and will
+#'   never match them together or to any other values. This is similar to joins
+#'   for database sources and to `base::merge(incomparables = NA)`.
```
DavisVaughan, line 88: We used to say `incomparable = FALSE` here, but that's wrong 😱 

```r
> merge(data.frame(x = NA, y = 1), data.frame(x = NA, z = 1), incomparables = FALSE, all = TRUE)
   x y z
1 NA 1 1
> merge(data.frame(x = NA, y = 1), data.frame(x = NA, z = 1), incomparables = NA, all = TRUE)
   x  y  z
1 NA  1 NA
2 NA NA  1
```

R/join.r
```diff
@@ -320,74 +482,135 @@ nest_join.data.frame <- function(x, y, by = NULL, copy = FALSE, keep = FALSE, na
...
   y_key <- set_names(y_in[vars$y$key], names(vars$y$key))
 
-  idx <- switch(type,
-    semi = vec_in(x_key, y_key, na_equal = na_equal),
-    anti = !vec_in(x_key, y_key, na_equal = na_equal)
+  condition <- standardise_join_condition(by)
+  filter <- by$filter
+
+  # We only care about whether or not any matches exist
+  multiple <- "first"
```
DavisVaughan, line 584: A neat usage of `multiple = "first"`. This makes `anti_join()` and `semi_join()` a little slower than it was with `vec_match()`, but now it also works for non-equi joins

_pkgdown.yml
```diff
@@ -37,6 +37,7 @@ reference:
   - left_join
   - nest_join
   - semi_join
+  - join_by
```
DavisVaughan, line 40: I put it in Two Table Verbs to be close to the join functions

tests/testthat/test-join.r
```diff
@@ -23,19 +23,22 @@ test_that("mutating joins preserve row and column order of x", {
 
 test_that("even when column names change", {
   df1 <- data.frame(x = c(1, 1, 2, 3), z = 1:4, a = 1)
-  df2 <- data.frame(z = 1:4, b = 1, x = c(1, 2, 2, 4))
+  df2 <- data.frame(z = 1:3, b = 1, x = c(1, 2, 4))
```
DavisVaughan, line 26: I've tweaked the values in a few tests so they don't require `multiple = "all"`. If I could tell that multiple matches wasn't the main focus of the test, I figured this would be cleaner long term.

R/join-rows.R
```diff
@@ -1,56 +1,214 @@
...
+                          haystack,
+                          ...,
+                          condition = "==",
+                          filter = "none",
+                          missing = "match",
+                          no_match = NA_integer_,
+                          remaining = "drop",
+                          multiple = "all") {
+  withCallingHandlers(
+    vctrs:::vec_matches(
```
DavisVaughan, line 40: I'll remove the `:::` when I merge in the related vctrs PR

NEWS.md
```diff
@@ -1,5 +1,42 @@
...
+  enable more flexible join operations, while also providing tools to perform
+  quality control checks directly in the join call (#5914, #5661, #5413, #2240).
+
+  * A _join specification_ can now be created through `join_by()`. This allows
+    you to specify both the left and right hand side of a join using unquoted
+    column names, such as `join_by(sale_date == commercial_date)`. Join
+    specifications can be supplied to any `*_join()` function as the `by`
+    argument.
+    
+  * Join specifications also allow for more flexible joins, such as non-equi
```
hadley, line 13: I think this bullet needs to include a full list of all the expressions that are supported

R/join-by.R
```diff
@@ -0,0 +1,641 @@
+#' Join specifications
+#'
+#' @description
+#' `join_by()` constructs a specification to join by using a small domain
```
hadley, line 4:
```suggestion
#' `join_by()` constructs a specification that describes how to join two tables using a small domain
```

R/join-by.R
```diff
@@ -0,0 +1,641 @@
...
+#' specific language. The result can be supplied as the `by` argument to any
+#' of the join functions (such as [left_join()]).
+#'
+#' `join_by()` is constructed from a set of expressions, with each expression
+#' being separated by a comma. Each expression should reference a column from
```
hadley, line 9: Maybe

```
#' `join_by()` is constructed from a comma-separated set of expressions of the form
#' `column_from_x OP column_from_y`,  where `OP` is one of `==`, ... as described below
```

R/join-by.R
```diff
@@ -0,0 +1,641 @@
...
+#' the left-hand table to match to the right-hand table, with some operator
+#' specifying how to match them together. Multiple expressions are combined
+#' using an "and" operation.
+#'
+#' ## Equi Joins:
```
hadley, line 14:
```suggestion
#' ## Equi joins:
```

We use sentence case for headings

R/join-by.R
```diff
@@ -0,0 +1,641 @@
...
+#'
+#' Equi joins match on equality, and are the most common type of join. To
```
hadley, line 16: @batpigandme equi-join or equi join?

R/join-by.R
hadley: Hmmm, or maybe it would be better to start with a bulleted list of all the options and then say that more details follow?

R/join-by.R
```diff
@@ -0,0 +1,641 @@
...
+#'   capacity. Equivalent to `x_lower <= y_upper, x_upper >= y_lower`.
+#'
+#' Internally, arguments are matched by position and should not be named.
+#'
+#' These conditions assume that the ranges are specified logically, i.e.
+#' `x_lower` should always be less than or equal to `x_upper`.
+#'
+#' ## Column Referencing:
+#'
+#' When specifying join conditions, the default assumes that column names on
```
hadley, line 72:
```suggestion
#' When specifying join conditions, `join_by` assumes that column names on
```

R/join-by.R
```diff
@@ -0,0 +1,641 @@
...
+#' the left-hand side of the condition refer to the left-hand table, and names
+#' on the right-hand side of the condition refer to the right-hand table.
```
hadley, line 74:
```suggestion
#' the left-hand side of the condition refer to the left-hand table (`x`), and names
#' on the right-hand side of the condition refer to the right-hand table (`y`).
```

R/join-by.R
```diff
@@ -0,0 +1,641 @@
...
+#'
+#' - `overlaps(x_lower, x_upper, y_lower, y_upper)`
+#'
+#'   Matches when `[x_lower, x_upper]` overlaps `[y_lower, y_upper]` in any
+#'   capacity. Equivalent to `x_lower <= y_upper, x_upper >= y_lower`.
+#'
+#' Internally, arguments are matched by position and should not be named.
+#'
+#' These conditions assume that the ranges are specified logically, i.e.
+#' `x_lower` should always be less than or equal to `x_upper`.
```
hadley, line 68:
```suggestion
#' These conditions assume that the ranges are well-formed, i.e.
#' `x_lower <= `x_upper`.
```

hadley (commented):
Made it up to the end of `join-cols.R`. I'll tackle the rest later

R/join-by.R
DavisVaughan: Keeping in mind we also have non-equi join and rolling join

R/join-rows.R
```diff
@@ -1,56 +1,214 @@
...
-      x_name <- sub(rx, "", cnd$x_arg)
-      y_name <- sub(rx, "", cnd$y_arg)
-
-      abort(c(
-        glue("Can't join on `x${x_name}` x `y${y_name}` because of incompatible types."),
-        i = glue("`x${x_name}` is of type <{x_type}>>.", x_type = vec_ptype_full(cnd$x)),
-        i = glue("`y${y_name}` is of type <{y_type}>>.", y_type = vec_ptype_full(cnd$y))
-      ))
-    }
+  missing <- standardise_join_missing(type, na_matches, unmatched)
```
hadley, line 11: I think we typically call these `check_`. These functions are a little different to our usual check functions which are more about throwing errors, but I think they're close enough to share a prefix.

R/join-rows.R
hadley: Could you comment a bit general on this rethrowing of errors feels as an API? Like does it seem pretty natural or is it annoying?

R/join-rows.R
```diff
@@ -1,56 +1,214 @@
...
+    vctrs_warning_matches_multiple = rethrow_warning_join_matches_multiple
+  )
+}
+
+rethrow_error_join_incompatible_type <- function(cnd) {
+  rx <- "^[^$]+[$]"
+  x_name <- sub(rx, "", cnd$x_arg)
+  y_name <- sub(rx, "", cnd$y_arg)
+
+  stop_join(
```
hadley, line 66: I think it makes sense to not use the `parent` argument of `abort()` here, but could you explain why?

R/join-rows.R
```diff
@@ -1,56 +1,214 @@
...
+    vctrs_error_matches_nothing = rethrow_error_join_matches_nothing,
+    vctrs_error_matches_missing = rethrow_error_join_matches_missing,
+    vctrs_error_matches_remaining = rethrow_error_join_matches_remaining,
+    vctrs_error_matches_multiple = rethrow_error_join_matches_multiple,
+    vctrs_warning_matches_multiple = rethrow_warning_join_matches_multiple
+  )
+}
+
+rethrow_error_join_incompatible_type <- function(cnd) {
+  rx <- "^[^$]+[$]"
```
hadley, line 62: It's a bit unfortunate we have to do regexp munging here. Can you explain why?

R/join-rows.R
```diff
@@ -1,56 +1,214 @@
...
+  abort(message = message, class = c(class, "dplyr_error"), ...)
+}
+warn_dplyr <- function(message = NULL, class = NULL, ...) {
+  warn(message = message, class = c(class, "dplyr_warning"), ...)
+}
+
+standardise_join_missing <- function(type, na_matches, unmatched) {
+  if (na_matches == "na") {
+    return("match")
+  }
```
hadley, line 151: I think these would be maybe easier to reason about if they all had the form:

```R
if ( ) { 

} else if () { 

} else if () {

}
```

Currently they're a slightly unusual mix of if/else with early returns.

R/join.r
```diff
@@ -9,8 +9,9 @@
 #' * `right_join()`: includes all rows in `y`.
 #' * `full_join()`: includes all rows in `x` or `y`.
 #'
-#' If a row in `x` matches multiple rows in `y`, all the rows in `y` will be returned
-#' once for each matching row in `x`.
+#' By default, if a row in `x` matches multiple rows in `y`, all of the matching
+#' rows in `y` will be returned. This can be adjusted using the `multiple`
```
hadley, line 13: Mention the warning? (Or wait until we're more clear about the default)

R/join.r
```diff
@@ -42,23 +43,27 @@
 #' @param x,y A pair of data frames, data frame extensions (e.g. a tibble), or
 #'   lazy data frames (e.g. from dbplyr or dtplyr). See *Methods*, below, for
 #'   more details.
-#' @param by A character vector of variables to join by.
+#' @param by A character vector of variables to join by, or a join specification
```
hadley, line 46:
```suggestion
#' @param by A character vector of variables to join by or a join specification
```

R/join.r
```diff
@@ -42,23 +43,27 @@
...
+#'   created through [join_by()].
```
hadley, line 47:
```suggestion
#'   created with [join_by()].
```

R/join.r
```diff
@@ -42,23 +43,27 @@
...
-#'   variables in common across `x` and `y`. A message lists the variables so that you
-#'   can check they're correct; suppress the message by supplying `by` explicitly.
+#'   variables in common across `x` and `y`. A message lists the variables so
+#'   that you can check they're correct; suppress the message by supplying `by`
+#'   explicitly.
 #'
-#'   To join by different variables on `x` and `y`, use a named vector.
-#'   For example, `by = c("a" = "b")` will match `x$a` to `y$b`.
+#'   To join on different variables between `x` and `y`, use a named vector. For
+#'   example, `by = c("a" = "b")` will match `x$a` to `y$b`.
```
hadley, line 55: Maybe include `join_by()` equivalent? ``Or use `join_by(a == b)` ``

I feel like people might prefer that.

R/join.r
```diff
@@ -68,14 +73,42 @@
 #'   Should be a character vector of length 2.
 #' @param keep Should the join keys from both `x` and `y` be preserved in the
 #'   output?
+#'   - If `NULL`, the default, equi join conditions retain only the keys from
+#'     `x`, while non-equi join conditions retain the keys from both inputs.
```
hadley, line 77: Remove "conditions"?

R/join.r
hadley: Ooops!

R/join.r
```diff
@@ -68,14 +73,42 @@
...
-#' @param na_matches Should `NA` and `NaN` values match one another?
+#' @param na_matches Should two `NA` or two `NaN` values match?
+#'   - `"na"`, the default, treats two `NA` or two `NaN` values as equal, like
+#'   `%in%`, [match()], and [merge()].
+#'   - `"never"` treats two `NA` or two `NaN` values as different, and will
+#'   never match them together or to any other values. This is similar to joins
+#'   for database sources and to `base::merge(incomparables = NA)`.
+#' @param multiple Handling of rows in `x` with multiple matches in `y`.
+#'   For each row of `x`:
+#'   - `"all"` returns every match detected in `y`.
```
hadley, line 91: Mention `This is the same behaviour as SQL`?

R/join.r
```diff
@@ -300,9 +445,26 @@ nest_join.data.frame <- function(x, y, by = NULL, copy = FALSE, keep = FALSE, na
...
-  y_split <- vec_group_loc(y_key)
-  matches <- vec_match(x_key, y_split$key)
-  y_loc <- y_split$loc[matches]
+  condition <- standardise_join_condition(by)
+  filter <- by$filter
+
+  type <- "nest"
+
+  # `nest_join()` currently assumes `na_matches = "na"`
+  na_matches <- "na"
```
hadley, line 454: Just pass in the class to `join_rows()`?  Probably worth adding `na_matches` as an argument while you're in here.

R/join.r
hadley: Nice! I hadn't even thought about the intersection of semi/anti and non-equi joins.

_pkgdown.yml
hadley: :+1:

tests/testthat/test-join.r
hadley: Makes sense to me.

hadley (approved):
Overall, looks great and I think this is good to merge once you've read through my smaller suggestions, and we're ready to start on dplyr 1.1.

R/join-rows.R
DavisVaughan: There were a few things that tripped me up:

- I started with `tryCatch()` but when I got to the warning handler I had to switch to `withCallingHandlers()`. I only figured that out by looking at `mutate_cols()`

- For the warning handler, I had no idea I had to do `maybe_restart("muffleWarning")` to cancel the warning and I'm still not entirely sure what that does / why I have to do it.

It seems like `rlang::with_handlers()` should have been where I started, and I should have wrapped the warning handler in `rlang::calling()`?

I really like the ability to customize the error messages to be join specific. The `vec_matches()` error messages are good, but they don't make sense in this context.

I think that this is overkill for most scenarios, but for dplyr it is a pretty useful pattern. `vec_matches()` also has `needles_arg` and `haystack_arg` arguments, but I still prefer the pattern used here in this case.

I think pulling out the handlers into their own named functions is the cleanest way to do this. Inlining all the handlers gets pretty messy.

R/join-by.R
```diff
@@ -0,0 +1,641 @@
...
+    x = glue("Expression {i} contains {err_expr(expr)}.")
+  ))
+}
+
+parse_join_by_dollar <- function(expr, i) {
+  n <- length(expr)
+
+  if (n != 3L) {
+    abort(c(
+      "Expressions containing `$` must have length 3.",
```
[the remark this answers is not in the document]
DavisVaughan: I think I tried this, but something in `expect_snapshot()` can't handle this even though it runs interactively. It seems sufficiently rare that I'm not too worried about it

```r
> join_by(`$`(1, 2, 3) == b)
Error in `parse_join_by_dollar()` at dplyr/R/join-by.R:369:4: 
Expressions containing `$` must have length 3.
x Expression 1 contains `1$2`, which is length 4.

> expect_snapshot(error = TRUE, join_by(`$`(1, 2, 3) == b))
Code
  join_by(1$2 == b)
Error <simpleError>
  <text>:1:11: unexpected numeric constant
  1: join_by(1$2
                ^
```

R/join-by.R
```diff
@@ -0,0 +1,641 @@
...
+  n <- length(expr)
+
+  if (n != 3L) {
+    abort(c(
+      "Expressions containing a binary operation must have length 3.",
+      x = glue("Expression {i} contains {err_expr(expr)}, which is length {n}.")
+    ))
+  }
+
+  condition <- as_string(expr[[1]])
```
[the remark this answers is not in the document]
DavisVaughan: Both `parse_join_by_expr()` and `parse_join_by_filter()` check for that ahead of time and error in a way that is more informative than I could have done from within `parse_join_by_binary()`

R/join-by.R
```diff
@@ -0,0 +1,641 @@
...
+  top <- as_string(expr[[1]])
+
+  if (is_true(top == "$")) {
+    abort(c(
+      "When specifying a single column name, `$` cannot be used.",
+      x = glue("Expression {i} is {err_expr(expr)}.")
+    ))
+  }
+
+  if (is_true(top == "between")) {
```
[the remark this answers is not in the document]
DavisVaughan:
```r
> rlang::is_true
function (x) 
{
    identical(x, TRUE)
}
```

I'm probably way to conservative here though.

And yes I think a switch would make sense, I'll update

R/join-by.R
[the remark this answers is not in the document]
hadley: Oh yeah, because `expect_snapshot()` has to generate `code` by deparsing the expression. And something clearly fails there.

R/join-rows.R
DavisVaughan: I don't think it is really checking anything. It is more about remapping the combination of `{type, na_matches, unmatched}` to a new variable, `missing`.

Maybe `compute_` or `remap_`?

R/join-rows.R
DavisVaughan: I think it has to do with whether or not the context of the parent message would be useful for your users. Like, here we have some high level procedure that failed, and the parent message details the precise location and reason for the failure which the high level procedure has no way of knowing ahead of time

```r
file <- "http://foo.bar/baz"

withCallingHandlers(
  download.file(file, destfile = tempfile()),
  error = function(err) {
    msg <- sprintf("Can't analyze file `%s`", file)
    rlang::abort(msg, parent = err)
  }
)
#> Error: 
#>   Can't analyze file `http://foo.bar/baz`
#> Caused by error in `download.file()`: 
#>   cannot open URL 'http://foo.bar/baz'
```

But for these rethrown errors we know exactly what went wrong (because of the error class name) and we are essentially restating the error message in a completely different context, so including the vctrs message too would be confusing at worst and sort of useless at best:

``` r
library(dplyr)

df <- data.frame(x = c(1, 1))

left_join(df, df, "x", multiple = "error")
#> Error: Each row in `x` can match at most 1 row in `y`.
#> ℹ Row 1 of `x` matches multiple rows.

vctrs:::vec_matches(df, df, multiple = "error")
#> Error: Each element can match at most 1 observation.
#> x The element at location 1 has multiple matches.
```

hadley: Makes sense!

R/join-rows.R
hadley: Ok, leave as is then.

R/join-rows.R
DavisVaughan: That was some legacy code that we don't actually need. We can pass in `needles_arg = "x"` and `haystack_arg = "y"` directly so we don't need this.

R/join.r
DavisVaughan: Good idea. I'll mention the warning and how to silence it. The top level Description section seems like a good place for that

```
#' By default, if a row in `x` matches multiple rows in `y`, all of the matching
#' rows in `y` will be returned. If this occurs in an equi join or a rolling
#' join, a warning will be thrown stating that multiple matches have been
#' detected since this is usually surprising. If multiple matches are expected
#' in these cases, silence this warning by explicitly setting `multiple =
#' "all"`.
```

R/join.r
DavisVaughan: I sort of do mean conditions. Like if you mix equality conditions and inequality conditions together, the equality ones retain keys from x and the inequality ones retain keys from both

``` r
library(dplyr)

segments <- tibble(
  segment_id = 1:4,
  chromosome = c("chr1", "chr2", "chr2", "chr1"),
  start = c(140, 210, 380, 230),
  end = c(150, 240, 415, 280)
)

reference <- tibble(
  reference_id = 1:4,
  chromosome = c("chr1", "chr1", "chr2", "chr2"),
  start = c(100, 200, 300, 400),
  end = c(150, 250, 399, 450)
)

# chromosome is the equi join condition (keep = FALSE)
# x$start, y$start, y$end are the non equi join conditions (keep = TRUE)
by <- join_by(chromosome, between(start, start, end))

full_join(segments, reference, by)
#> # A tibble: 5 × 7
#>   segment_id chromosome start.x end.x reference_id start.y end.y
#>        <int> <chr>        <dbl> <dbl>        <int>   <dbl> <dbl>
#> 1          1 chr1           140   150            1     100   150
#> 2          2 chr2           210   240           NA      NA    NA
#> 3          3 chr2           380   415            3     300   399
#> 4          4 chr1           230   280            2     200   250
#> 5         NA chr2            NA    NA            4     400   450
```

Does this sound better?

```
#'   - If `NULL`, the default, conditions joining on equality retain only the
#'     keys from `x`, while conditions joining on an inequality retain the keys
#'     from both inputs.
```

R/join.r
DavisVaughan: Good idea, added `na_matches` to `nest_join()` since it was the only join that didn't have that argument

PR #5942, opened 2021-07-01
Add a `.locale` argument to `arrange()` and implement `dplyr_locale()`

### LINKED ISSUE #4962: Performance drop-off for `arrange()`

DavisVaughan:
Supersedes #5868 (as this is a cleaner git history)
Closes #4962
Closes #5090
Part of #5808

Requires dev vctrs.

This PR adds a `.locale` argument to `arrange()`, with 3 possible options:
- `dplyr_locale()`, the default. See below.
- A single stringi locale identifier, like `"en_US"` or `"fr"`. This requires stringi.
- `"C"`, for sorting in the C locale. This does not require stringi.

`dplyr_locale()` is a new exported helper and returns a string representing the default locale to use. It has the following properties:
- If stringi is installed, use `"en"`.
- If stringi is not installed, use `"C"` with a warning. The warning can be silenced by installing stringi or specifying `.locale = "C"` explicitly.
- If the `dplyr.locale` option is set to a stringi locale identifier or `"C"`, that overrides the above default behavior.

This is a breaking change, as we no longer respect the `LC_COLLATE` option.


tests/testthat/_snaps/bind.md
```diff
@@ -2,21 +2,21 @@
...
 
     Code
       df <- bind_cols(list(1, 2))
-    Message <simpleMessage>
+    Message <rlib_message_name_repair>
       New names:
-      * NA -> ...1
-      * NA -> ...2
+      * `` -> `...1`
+      * `` -> `...2`
```
DavisVaughan, line 19: These come from updating to dev vctrs, which uses `rlang::names_inform_repair()` now

DavisVaughan:
@hadley @romainfrancois @lionel- this is finally ready to merge. Some of you looked at #5868, but it has been awhile since then so I'd appreciate a quick review here and final signoff that we are good to merge these changes!

NEWS.md
```diff
@@ -1,5 +1,19 @@
 # dplyr (development version)
 
+* `arrange()` now uses a new process for sorting character vectors. In
+   particular, a much faster radix-sort based algorithm has been implemented,
+   which is heavily inspired by data.table's `forder()`. Additionally, a new
+   argument, `.locale`, now controls the locale that character vectors are
+   ordered in. `.locale` defaults to `dplyr_locale()`, which uses the American
```
hadley, line 7: I think this needs a crisp explanation that the behaviour has changed, and should probably link back to the tidyup for a fuller explanation/justification.

R/arrange.R
```diff
@@ -68,22 +64,40 @@ arrange <- function(.data, ..., .by_group = FALSE) {
...
+#'   - Defaults to [dplyr_locale()], which uses the American English locale
+#'     if the stringi package is installed and the global default locale
+#'     has not been altered. See the help page for [dplyr_locale()] for the
+#'     exact details.
+#'
+#'   - If a single string is supplied, then this will be used as the locale
+#'     identifier to sort with. For example, `"fr"` will sort with the French
+#'     locale. This requires the stringi package. Use
+#'     [stringi::stri_locale_list()] to generate a list of possible locale
+#'     identifiers.
```
hadley, line 78:
```suggestion
#'   - If a single string is supplied, then this will be used as the sorting locale
#'     For example, `"fr"` will sort with the French locale. This requires the stringi 
#'     package. Use [stringi::stri_locale_list()] to generate a list of possible locale
#'     identifiers.
```

Is there a precise name for these sorts of identifiers?

R/locale.R
```diff
@@ -0,0 +1,70 @@
...
+#' `dplyr_locale()` returns a single string representing the default locale used
+#' by dplyr when ordering character vectors. It is used as the default value of
+#' `.locale` in [arrange()].
+#'
+#' ## Default locale
+#'
+#' - If stringi >= 1.5.3 is installed, the default locale is set to American
+#'   English, represented by the locale identifier `"en"`.
+#'
+#' - If stringi >= 1.5.3 is not installed, the default locale falls back to the
```
hadley, line 13:
```suggestion
#' - If stringi is not installed or is older than 1.5.3, the default locale falls back to the
```

hadley (approved):
Approved with the caveat that we don't want to merge until after the 1.0.8 release.

R/arrange.R
DavisVaughan: I don't think so. The ICU documentation calls them "locale IDs" but doesn't get more specific than that when referring to the whole locale string, so that is what I went off of.

For `"en_US"`, it refers to the `"en"` part as a _language code_, and the `"US"` part as a _country code_

https://unicode-org.github.io/icu/userguide/locale/#the-locale-concept

NEWS.md
DavisVaughan: I took a second pass in 3e5b6aabdc66c19a01b3b7fcb85389366a2afb76

DavisVaughan:
Superseded by #6263

PR #5993, opened 2021-08-25
Require dev rlang

DavisVaughan:
Closes r-lib/rlang#1254

This was coming with https://github.com/tidyverse/dplyr/pull/5942, which is close to being ready to merge, but doing this in its own PR is much cleaner since there are a few changes:

- A few snapshot tests are changed since now there is a space in anonymous functions between `~` and the body
- The `check_dots_empty()` error comes from rlang now and is a little different
- In two places, we relied on `abort()` auto-naming the non-header bullets. It no longer does this, so I've added explicit `x` names on these bullets. See also https://github.com/r-lib/rlang/issues/1254

@lionel- I imagine that rlang will probably be sent to CRAN before dplyr, so there will be a short period where the two places we weren't explicitly naming the bullets will now emit unbulleted messages, but they are fairly rare errors so I think that is okay (grouping by a nonexistent variable and having an error with the implicit mutate() inside an arrange() call)

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -37,7 +37,7 @@ Imports:
     magrittr (>= 1.5),
     methods,
     R6,
-    rlang (>= 0.4.10),
+    rlang (>= 0.4.11.9001),
     tibble (>= 2.1.3),
     tidyselect (>= 1.1.0),
     utils,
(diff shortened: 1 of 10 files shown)


romainfrancois (approved):
Thanks :-)

PR #6000, opened 2021-09-05
Update storms data to 2020

steveharoz:
* The `storms` data only went up to 2015, so I updated the data through 2020.

* **More easy to update in the future**: The code for generating the storms dataset depended on a file that was not available and used deprecated functions. These issues have been fixed. As long as the data source URL is updated, it should be simple to rerun and update this dataset annually.

* I added some example code for the `storms` man page (see graph below)

* Clarified the units and meaning of the diameter columns

* Possible breaking change: I changed `ts_diameter` and `hu_diameter` to `tropicalstorm_force_diameter` and `hurricane_force_diameter`. They weren't used in any examples, and barely used in other github repositories, so impact should be very minimal.

* Fixes issue #5899 and @romainfrancois's [comment](https://github.com/tidyverse/dplyr/issues/5899#issuecomment-851288418).

image


steveharoz:
Hmm... using ggplot in the example causes the checks to fail. Anyone have suggestions? It'd be a shame to drop the example, but maybe it's necessary?

romainfrancois (approved):

erikmeesters:
Hi guys, I'm a coral reef ecologist working in an EU project on Caribbean coral reefs and very interested in these scripts and figures, however, I'm not very familiar with github. I can code R, but I'm not an expert like you guys (more old school I fear). Can I use your code and examples? I've downloaded the latest hurdat text file and used your script to make it into a normal data frame. Seems to work all fine. It's a great help. Thanks for this. Erik

steveharoz:
Hi @erikmeesters. Of course! All the code here is freely available via the MIT license. 

For future reference, go to the root of the repository (https://github.com/tidyverse/dplyr/), and open the `LICENSE.md` file.

PR #6141, opened 2021-12-25
Don't reexport deprecated trunc_mat()

krlmlr:
It doesn't seem to be imported anywhere from dplyr, only from tibble: https://github.com/search?q=org%3Acran+trunc_mat+filename%3ANAMESPACE&type=Code&ref=advsearch&l=&l=


romainfrancois (approved):
LGTM, Thanks @krlmlr

PR #6164, opened 2022-01-25
Preserve custom factor metadata in group_by(..., .drop=FALSE)

gergness:
Before this PR, the metadata on a factor variable beyond the standard base constructs (the levels attribute, and the "factor" and possibly "ordered" classes) is lost. 

I'm not sure how much this is worth going into, but I was hoping to extend the behavior of factors, especially the `.drop` behavior in https://github.com/gergness/srvyr/issues/134. This is admittedly niche, but it feels in the spirit of recent work in `vctrs` that allows extending base vector classes. Because of the recent advances in preserving metadata, it was surprising to me that this case doesn't.

Before the PR:
``` r

df <- tibble(
  x = structure(1L, levels = "x", extra_info = "xyz", class = c("custom", "factor"))
)
nodrop <- df %>% group_by(x, .drop = FALSE) %>% group_data()

class(nodrop$x)
#> [1] "factor"
attributes(nodrop$x)
#> $levels
#> [1] "x"
#> 
#> $class
#> [1] "factor"
```

After this PR:
``` r

df <- tibble(
  x = structure(1L, levels = "x", extra_info = "xyz", class = c("custom", "factor"))
)
nodrop <- df %>% group_by(x, .drop = FALSE) %>% group_data()

class(nodrop$x)
#> [1] "custom" "factor"
attributes(nodrop$x)
#> $levels
#> [1] "x"
#> 
#> $extra_info
#> [1] "xyz"
#> 
#> $class
#> [1] "custom" "factor"
```

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,7 @@
 # dplyr (development version)
 
+* `group_by(..., .drop = FALSE)` preserves all metadata on a factor including for subclasses (@gergness, #6164).
+
 * The deprecated `trunc_mat()` is no longer reexported from dplyr, no CRAN packages import it from dplyr (#6141).
 
 * dplyr now uses `rlang::check_installed()` to prompt you whether to install
--- a/R/grouped-df.r
+++ b/R/grouped-df.r
@@ -77,11 +77,7 @@ compute_groups <- function(data, vars, drop = FALSE) {
     # Make the new keys from the old keys and the new_indices
     new_keys <- pmap(list(old_keys, new_indices, uniques), function(key, index, unique) {
       if (is.factor(key)) {
-        if (is.ordered(key)) {
-          new_ordered(index, levels = levels(key))
-        } else {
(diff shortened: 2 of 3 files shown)


DavisVaughan:
@gergness I appreciate what you are trying to do, but we are a bit uncomfortable committing to supporting this feature:

- `tidyr::complete()` and `tidyr::expand()` also don't support factor subclasses

- Just reattaching the attributes back onto the class after it has been "sliced" won't work in all cases. In some cases the attributes might need to be "updated" based on the new data that is there (i.e. if you have a vectorized attribute, similar to `names`) and there is no correct way to do that here

- We aren't sure that base R factors/ordered factors are even something that should be subclassed at all. Typically a class needs to be designed with subclassing in mind from the start, and we aren't convinced base R has done that.

Again, we really appreciate the PR and hope you will continue to contribute in the future!

PR #6203, opened 2022-03-02
Implement `conflict` and `unmatched` arguments for the `rows_*()` family

### LINKED ISSUE #5699: `rows_patch(x, y, ...)` shows unhelpful error message when `nrow(y) > nrow(x)`
When I try to patch a data frame (`x` in the below example) using another data frame with more rows (`y`), an error is thrown with an unhelpful message. It took some time to figure out that the row number difference was the source of the error. `rlang::last_error()` and `rlang::last_trace()` didn't help.

Is there any reason a data frame *should not* be patched by a data frame with more rows? I could be missing some theory here. If there *is* a good reason why this should not be allowed, can I add that to the documentation for `rows_patch()` and friends?

Example:

``` r

x = data.frame(id = c(1, 2, 3), foo = c("a", "b", NA))
y = data.frame(id = c(1, 3, 5, 7), foo = c("a", "c", "e", "g"))
dplyr::rows_patch(x, y, by = "id")
#> Error: Attempting to patch missing rows.
```

<sup>Created on 2021-01-20 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>
<details>
<summary>
Session info
</summary>

``` r
sessionInfo()
#> R version 4.0.3 (2020-10-10)
#> Platform: x86_64-w64-mingw32/x64 (64-bit)
#> Running under: Windows 10 x64 (build 19041)
#> 
#> Matrix products: default
#> 
#> locale:
#> [1] LC_COLLATE=English_United States.1252 
#> [2] LC_CTYPE=English_United States.1252   
#> [3] LC_MONETARY=English_United States.1252
#> [4] LC_NUMERIC=C                          
#> [5] LC_TIME=English_United States.1252    
#> 
#> attached base packages:
#> [1] stats     graphics  grDevices utils     datasets  methods   base     
#> 
#> other attached packages:
#> [1] dplyr_1.0.3
#> 
#> loaded via a namespace (and not attached):
#>  [1] knitr_1.30       magrittr_2.0.1   tidyselect_1.1.0 R6_2.5.0        
#>  [5] rlang_0.4.10     stringr_1.4.0    highr_0.8        tools_4.0.3     
#>  [9] xfun_0.20        DBI_1.1.1        htmltools_0.5.1  ellipsis_0.3.1  
#> [13] yaml_2.2.1       digest_0.6.27    assertthat_0.2.1 tibble_3.0.5    
#> [17] lifecycle_0.2.0  crayon_1.3.4     purrr_0.3.4      vctrs_0.3.6     
#> [21] glue_1.4.2       evaluate_0.14    rmarkdown_2.6    stringi_1.5.3   
#> [25] compiler_4.0.3   pillar_1.4.7     generics_0.1.0   pkgconfig_2.0.3
```

</details>

Expected output:

``` r
data.frame(id = c(1, 2, 3), foo = c("a", "b", "c"))
#>   id foo
#> 1  1   a
#> 2  2   b
#> 3  3   c
```

### LINKED ISSUE #5984: rows_patch fails when y has key values that don't occur in x
I tried to use `row_patch` to fill in missing values in a data frame.  It failed by design, because it didn't meet the requirement for the function: "key values in y must occur in x".  However, I was able to successfully execute the row patch by using the `match` function from base R (see below).  

This isn't a typical bug report because it is clearly stated that this is a limitation of the function, but I wanted to point out that it can be done using a different function from base R.  Since `row_patch` is currently an experimental function in `dplyr`, I wanted to provide feedback to hopefully improve it.  `row_patch` is much cleaner and I would prefer to use it if possible.

It would also be nice if `row_patch` could preserve other variables in y that don't occur in x.  `match` can handle this as well.

```r
D1 <- data.frame(
  id=seq(1,3),
  x=c("cow",NA,"sheep"))

D2 <- data.frame(
  id=seq(1,4),
  x=c("cow","turtle","parrot","frog"))

#rows_patch fails
D1 %>%
  rows_patch(D2, by = "id")

#match works
na <- is.na(D1$x)

D1$x[na] <- D2$x[match(D1$id[na],D2$id)]
```

DavisVaughan:
Follow-up to #6199 and continuation of the plan from here https://github.com/tidyverse/dplyr/pull/5588#issuecomment-1048373544

Closes #5984 
Closes #5699 

This PR adds a new `conflict` argument to `rows_insert()`, `rows_update()`, `rows_patch()`, and `rows_delete()`, inspired by the `ON CONFLICT *` command from SQL's `INSERT`.

- For `rows_insert()`, a conflict arises if there is a key in `y` that _already_ exists in `x`. By default, it `"error"`s, but you can now also `"ignore"` the rows in `y` where this occurs, which is like `ON CONFLICT DO NOTHING`.

- For `rows_update()`, `rows_patch()`, and `rows_delete()`, a conflict arises if there is a key in `y` that _doesn't_ exist in `x` (because the behavior is then ambiguous). Again, by default it `"error"`s, but you can also `"ignore"` those rows in `y`.

---

A few more notes:

- `conflict` is a bit of a non-standard name for the update/patch/delete behavior, but I think it works well enough. The alternative was to use `conflict` for `rows_insert()` and come up with some other argument name for these 3. Or use `matched` for insert and `unmatched` for update/patch/delete. I am open to suggestions if anyone feels strongly against the current scheme.

- In https://github.com/tidyverse/dplyr/pull/5588#issuecomment-1025599455, @mgirlich mentioned the possibility of `rows_insert(conflict = "insert")` to insert the `y` rows anyways even if there are conflicts, but this felt too similar to `bind_rows()` to me, so I left it out (i.e. it would always insert the `y` rows, no matter what, which is basically just binding the rows together after validating that the `by` columns exist in both inputs). I think we could consider adding it later if people ask for it.

- I've used a default of `"error"` because that is what the current behavior is, and it feels reasonable to opt in to this more lax behavior.

### DIFF
--- a/R/rows.R
+++ b/R/rows.R
@@ -103,7 +103,7 @@ rows_insert <- function(x,
                         y,
                         by = NULL,
                         ...,
-                        conflict = "error",
+                        conflict = c("error", "ignore"),
                         copy = FALSE,
                         in_place = FALSE) {
   lifecycle::signal_stage("experimental", "rows_insert()")
@@ -115,7 +115,7 @@ rows_insert.data.frame <- function(x,
                                    y,
                                    by = NULL,
                                    ...,
-                                   conflict = "error",
+                                   conflict = c("error", "ignore"),
                                    copy = FALSE,
                                    in_place = FALSE) {
   check_dots_empty()
@@ -145,7 +145,7 @@ rows_update <- function(x,
                         y,
                         by = NULL,
                         ...,
-                        conflict = "error",
+                        conflict = c("error", "ignore"),
                         copy = FALSE,
                         in_place = FALSE) {
   lifecycle::signal_stage("experimental", "rows_update()")
@@ -157,7 +157,7 @@ rows_update.data.frame <- function(x,
                                    y,
                                    by = NULL,
                                    ...,
-                                   conflict = "error",
+                                   conflict = c("error", "ignore"),
                                    copy = FALSE,
                                    in_place = FALSE) {
   check_dots_empty()
@@ -198,7 +198,7 @@ rows_patch <- function(x,
                        y,
                        by = NULL,
                        ...,
-                       conflict = "error",
+                       conflict = c("error", "ignore"),
                        copy = FALSE,
                        in_place = FALSE) {
   lifecycle::signal_stage("experimental", "rows_patch()")
@@ -210,7 +210,7 @@ rows_patch.data.frame <- function(x,
                                   y,
                                   by = NULL,
                                   ...,
-                                  conflict = "error",
+                                  conflict = c("error", "ignore"),
                                   copy = FALSE,
                                   in_place = FALSE) {
   check_dots_empty()
@@ -305,7 +305,7 @@ rows_delete <- function(x,
                         y,
                         by = NULL,
                         ...,
-                        conflict = "error",
+                        conflict = c("error", "ignore"),
                         copy = FALSE,
                         in_place = FALSE) {
   lifecycle::signal_stage("experimental", "rows_delete()")
@@ -317,7 +317,7 @@ rows_delete.data.frame <- function(x,
                                    y,
                                    by = NULL,
                                    ...,
-                                   conflict = "error",
+                                   conflict = c("error", "ignore"),
                                    copy = FALSE,
                                    in_place = FALSE) {
   check_dots_empty()
--- a/R/rows.R
+++ b/R/rows.R
@@ -461,8 +461,9 @@ rows_check_y_conflict <- function(x_key,
       rows_matched <- err_locs(rows_matched)
 
       message <- c(
-        "`y` must contain keys that don't exist in `x`.",
-        i = glue("The following rows in `y` have keys that already exist in `x`: {rows_matched}.")
+        "`y` can't contain keys that already exist in `x`.",
+        i = glue("The following rows in `y` have keys that already exist in `x`: {rows_matched}."),
+        i = "Use `conflict = \"ignore\"` if you want to ignore these `y` rows."
       )
 
       abort(message, call = error_call)
@@ -495,7 +496,8 @@ rows_check_y_unmatched <- function(x_key,
 
       message <- c(
         "`y` must contain keys that already exist in `x`.",
-        i = glue("The following rows in `y` have keys that don't exist in `x`: {rows_unmatched}.")
+        i = glue("The following rows in `y` have keys that don't exist in `x`: {rows_unmatched}."),
+        i = "Use `unmatched = \"ignore\"` if you want to ignore these `y` rows."
       )
 
       abort(message, call = error_call)
--- a/R/rows.R
+++ b/R/rows.R
@@ -9,15 +9,15 @@
 #' by SQL's `INSERT`, `UPDATE`, and `DELETE`, and can optionally modify
 #' `in_place` for selected backends.
 #'
-#' * `rows_insert()` adds new rows (like `INSERT`). Key values in `y` must
-#'    not occur in `x`.
-#' * `rows_update()` modifies existing rows (like `UPDATE`). Key values in
-#'   `y` must occur in `x`, and key values in `y` must be unique.
+#' * `rows_insert()` adds new rows (like `INSERT`). By default, key values in
+#'   `y` must not exist in `x`.
+#' * `rows_update()` modifies existing rows (like `UPDATE`). Key values in `y`
+#'   must be unique, and, by default, key values in `y` must exist in `x`.
 #' * `rows_patch()` works like `rows_update()` but only overwrites `NA` values.
 #' * `rows_upsert()` inserts or updates depending on whether or not the
 #'   key value in `y` already exists in `x`. Key values in `y` must be unique.
-#' * `rows_delete()` deletes rows (like `DELETE`). Key values in `y` must
-#'   exist in `x`.
+#' * `rows_delete()` deletes rows (like `DELETE`). By default, key values in `y`
+#'   must exist in `x`.
 #'
 #' @inheritParams left_join
 #' @param x,y A pair of data frames or data frame extensions (e.g. a tibble).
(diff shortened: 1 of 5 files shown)


R/rows.R
```diff
@@ -112,6 +145,7 @@ rows_update <- function(x,
                         y,
                         by = NULL,
                         ...,
+                        conflict = "error",
```
hadley, line 148:
```suggestion
                        conflict = c("error", "ignore"),
```

hadley (approved):
Seems reasonable to me. I just skimmed the code.

mgirlich:
Looks great to me and I'm looking forward to use the new power of the `rows_*()` functions! 🥳 

For databases it probably makes sense to allow `rows_insert(by = character())`.
So, it might make sense to allow `rows_insert(by = character())` for local tables as well. What do you think?

Off topic question:
I have some weird and unnecessary line breaks in my error messages. Is this a `cli` issue? And is this a known issue?
image

DavisVaughan:
`by = character()` would mean "always insert these rows, there are no keys to check"? Which is exactly `bind_rows()` right? (I realize we don't have a `bind_rows()` for dbplyr, just confirming). I guess for databases you just want a way to insert arbitrary rows?

It would be a little weird to only allow `by = character()` for 1 of the `rows_*()` functions, but they don't have too much meaning otherwise, i.e. `rows_update(by = character())`, `rows_patch(by = character())`, and `rows_delete(by = character())` would either be no-ops or errors based on the value of `conflict` that is set, I guess?

I'm not sure about this feature yet, but I think we can merge this PR and think about it in another PR

DavisVaughan:
@mgirlich I've been told the aggressive line wrapping is due to https://github.com/r-lib/cli/issues/288, where we aren't sure if there is an error inspector on the right side of the console, which would take up some of the width if it is there

mgirlich:
> `by = character()` would mean "always insert these rows, there are no keys to check"? Which is exactly `bind_rows()` right? (I realize we don't have a `bind_rows()` for dbplyr, just confirming).

Yes.

> I guess for databases you just want a way to insert arbitrary rows?

Yes.

> It would be a little weird to only allow `by = character()` for 1 of the `rows_*()` functions, but they don't have too much meaning otherwise, i.e. `rows_update(by = character())`, `rows_patch(by = character())`, and `rows_delete(by = character())` would either be no-ops or errors based on the value of `conflict` that is set, I guess?

I agree that it is a bit weird. But it also feels weird to me that I can do most things with the `rows_*()` functions but then suddenly have to use `bind_rows()` (and for databases even something different).

Maybe one way out of this is to add a function `rows_append()`, which would not have the argument `by`. This would probably also be a bit safer and communicate the intent a bit better.

> I'm not sure about this feature yet, but I think we can merge this PR and think about it in another PR

Sounds good.

R/rows.R
```diff
@@ -34,6 +34,20 @@
 #'
 #'   When `TRUE`, a modified version of `x` is returned invisibly;
 #'   when `FALSE`, a new object representing the resulting changes is returned.
+#' @param conflict How should keys in `y` that conflict with keys in `x` be
+#'   handled?
+#'
+#'   For `rows_insert()`, a conflict arises if there is a key in `y` that
+#'   already exists in `x`.
+#'
+#'   For `rows_patch()`, `rows_delete()`, and `rows_update()`, a conflict arises
```
mgirlich, line 43:
```suggestion
#'   For `rows_update()`, `rows_patch()`, and `rows_delete()`, a conflict arises
```

mgirlich (approved):

DavisVaughan:
I reread this PR recently and it took me awhile to re-remember what `conflict` means for the non-`rows_insert()` cases.

Since it seems somewhat confusing, I might just go ahead and split it into two arguments. `conflict` for `rows_insert()` and `nonexistent` for `rows_update/patch/delete()`. i.e. "handle the case where you are attempting to modify a nonexistent key in `x`"

krlmlr:
I agree that `conflict` and `nonexistent` should be distinct arguments. Should we then change to `conflict = "drop"` and change `conflict = "ignore"` to insert duplicated rows? This will be useful for databases where supporting `conflict = "drop"` is difficult.

https://github.com/tidyverse/dbplyr/pull/748

krlmlr:
Added to 33cabfe6531a54c59ac79ab08c833ff2cb2e5823, feel free to adapt.

mgirlich:
> I agree that `conflict` and `nonexistent` should be distinct arguments. Should we then change to `conflict = "drop"` and change `conflict = "ignore"` to allow duplicated rows? This will be useful for databases where supporting `conflict = "error"` is difficult.
> 
> [tidyverse/dbplyr#748](https://github.com/tidyverse/dbplyr/pull/748)

`conflict = "ignore"` would then be the same as `bind_rows()` resp. my proposed `rows_append()`? It feels weird to specify `by` and have `conflict = "ignore"`.

krlmlr:
Yes, `conflict = "ignore", in_place = FALSE` should do `union()`, and with `in_place = TRUE` it should plainly insert rows without any extra checks. Only `conflict = "drop"` would insert fewer rows in case of duplicates.

`by` will also be provided by `dm_rows_insert()`, and could be implicit if keys are stored as a property of the target table object. Supporting `by = c()` is another alternative, but for dm it would be easier if we just supported `conflict = "ignore"` .

krlmlr:
Are you planning to make use of my PR?

krlmlr:
I updated it, just in case.

DavisVaughan:
@krlmlr I'm definitely making use of it but will just pull pieces of it into a separate commit here and give you credit. Thanks for drafting it for us!

DavisVaughan:
Final set of changes:

- `rows_insert()` got `conflict = c("error", "ignore")`
- `rows_update/patch/delete()` got `unmatched = c("error", "ignore")`
- Will add `rows_append()` in a follow-up PR

We went with `unmatched` instead of `nonexistent` since it seemed a little clearer and "matching" is probably a little more familiar to R users

We kept `conflict` for `rows_insert()` because that maps 1:1 to the SQL `ON CONFLICT` command

PR #6210, opened 2022-03-07
update bind.r documentation

ahcyip:
Name of .id argument was a bit ambiguous - clarified to state that it defines the name of the column


ahcyip:
I see that the description of the .id argument in the map_dfr description was updated here
https://github.com/tidyverse/purrr/commit/553242e52aee2dafb1e08f703ea87098c3a691b1
https://purrr.tidyverse.org/reference/map.html
and it is much clearer.

I am going to propose to revise the dplyr::bind_rows documentation to be consistent.

DavisVaughan:
Thanks @ahcyip!

PR #6219, opened 2022-03-23
Slice min/max ignores NA's when with_ties is set to FALSE

wfjvdham:
When the `with_ties` argument was set to `FALSE` `NA`s where not ignored anymore. This PR fixes that. Problem was mentioned here: https://github.com/tidyverse/dplyr/issues/4826#issuecomment-1064040917


hadley:
Let's discuss this first: #6347. I'm not sure the current behaviour for `with_ties = TRUE` is correct.

PR #6221, opened 2022-03-24
Update to rlang 1.0.2 snapshots

DavisVaughan:
- Updated two snapshots related to internal errors now mentioning the package name

- Require rlang >= 1.0.2 as this isn't much to ask of the user, and ensures that we are all developing with a version of rlang that gives consistent snapshots

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -40,7 +40,7 @@ Imports:
     tibble (>= 2.1.3),
     tidyselect (>= 1.1.1),
     utils,
-    vctrs (>= 0.3.5), 
+    vctrs (>= 0.4.1), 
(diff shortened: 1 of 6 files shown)


krlmlr:
vctrs 0.4.0 happened in the meantime, if we update snapshots there, wording changes to the worse in a few places due to https://github.com/r-lib/vctrs/issues/1543 .

romainfrancois (approved):

DavisVaughan:
We are going to fix the linked vctrs issue and do a patch vctrs release in around a week or less.

I'll hold off on this until that is on CRAN and then I'll bump the vctrs version and regenerate all snapshots with the newest rlang and vctrs versions.

Issue #6049, opened 2021-10-11
Feature request: Allow `lag/lead` to use length-X `n`


pnacht:
`lag` and `lead` currently mandate that the lag value `n` be a length-1 integer value. Is there a specific reason for this?

I was able to write the following function which allows for length-X vectors in `n`:

``` r
myLag <- function(x, n = 1, default = NA) {
    i <- pmax(seq_along(x) - n, 0)
    x <- c(default, x)
    x[i + 1]
}

myLag(10:15)
#> [1] NA 10 11 12 13 14
myLag(10:15, 0:5)
#> [1] 10 10 10 10 10 10
myLag(10:15, 5:0)
#> [1] NA NA NA 11 13 15
myLag(10:15, 1:2)
#> [1] NA NA 11 11 13 13
myLag(10:15, 2:1)
#> [1] NA 10 10 12 12 14
```

<sup>Created on 2021-10-11 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

Obviously, it lacks some of the checks in `dplyr::lag`, but those would be straightforward to include. It also doesn't include the `order_by` argument and implementation, but at a glance that also seems to be a simple drop-in.

With this implementation, if `x`'s length is not a multiple of `n`'s, a warning is thrown and the output is usually not what was expected, but that could be handled by the standard dplyr error of `"n must be of length {length(x)} or 1, not {length(n)}"` (though that'd make some of the examples above not work, but I'm fine with that).

But most importantly, is there a reason why `lag` currently requires `n` to be length-1?

hadley:
@romin-ebrahimi I think the most useful place to start would be a minimal patch to the existing `lead()` and `lag()` functions.  I think the basic idea is something like this:

```R
lag <- function(x, n = 1L, default = NA, order_by = NULL, ...) {
  if (!is.null(order_by)) {
    return(with_order(order_by, lag, x, n = n, default = default))
  }

  if (inherits(x, "ts")) {
    msg <- "`x` must be a vector, not a ts object, do you want `stats::lag()`?"
    abort(msg)
  }

  if (!is.numeric(n) || any(n < 0)) {
    msg <- glue("`n` must be a vector of positive integers.")
    abort(msg)
  }

  if (vec_size(default) != 1L) {
    msg <- glue("`default` must be size 1, not size {vec_size(default)}")
    abort(msg)
  }

  inputs <- fix_call(vec_cast_common(default = default, x = x))

[...]

PR #6235, opened 2022-04-18
lead and lag accept vector of positive integers

romin-ebrahimi:
This closes out #6049 by modifying the lead and lag functions to allow for an input vector of positive integers.

### DIFF
--- a/R/lead-lag.R
+++ b/R/lead-lag.R
@@ -54,7 +54,7 @@ lag <- function(x, n = 1L, default = NA, order_by = NULL, ...) {
     abort(msg)
   }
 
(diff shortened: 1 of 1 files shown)


romin-ebrahimi:
I have no idea why the unit test in testthat/test-window.R is failing. The tests pass locally. It fails with the following: `Error in seq_len(ifelse(i >= xlen, NA_integer_, i)): argument must be coercible to non-negative integer`. Any ideas on how to fix the unit test failure is greatly appreciated.

DavisVaughan:
Thanks for the PR, but see https://github.com/tidyverse/dplyr/issues/6049#issuecomment-1185796702

PR #6238, opened 2022-04-23
Feature Request: Integrate `across()` into count/tally

darentsai:
The argument `wt` in count/tally is designed for performing weighted counts. It also can be used to simplify the process of summation. E.g., `df %>% count(a, wt = b)` is equivalent to `df %>% group_by(a) %>% summarise(n = sum(b))` under the hood. Currently it takes one variable at most as input, but in some cases we need to sum multiple variables at the same time. So I think it might be convenient to make it a general version to pass more than one columns into it. To achieve it, this PR integrates `across()` into count/tally and transforms the argument type of `wt` from ___\<data-masking\>___ to ___\<tidy-select\>___. The following is some examples.

In common use without setting `wt`:

```r
iris %>%
  count(Species)

#>      Species  n
#> 1     setosa 50
#> 2 versicolor 50
#> 3  virginica 50
```

```r
iris %>%
  count(Species, wt = starts_with("Sepal"))

#>      Species Sepal.Length Sepal.Width
#> 1     setosa        250.3       171.4
#> 2 versicolor        296.8       138.5
#> 3  virginica        329.4       148.7
```

Now the arg `name` can take a glue specification that describes how to name the output columns.

```r
iris %>%
  count(Species, wt = starts_with("Sepal"), name = "{.col}_Sum")

#>      Species Sepal.Length_Sum Sepal.Width_Sum
#> 1     setosa            250.3           171.4
#> 2 versicolor            296.8           138.5
#> 3  virginica            329.4           148.7
```

A bigger difference is that the original `add_count()` with one variable passed to `wt` will create an additional column named `n` to indicate the sum of that variable. However, this PR defaults to create total columns with the same column names and hence overwrite previous columns.

```r
iris %>%
  add_count(Species, wt = starts_with("Sepal"))

#>     Sepal.Length Sepal.Width Petal.Length Petal.Width    Species
#> 1          250.3       171.4          1.4         0.2     setosa
#> 2          250.3       171.4          1.4         0.2     setosa
#> ...
#> 51         296.8       138.5          4.7         1.4 versicolor
#> 52         296.8       138.5          4.5         1.5 versicolor
#> ...
#> 101        329.4       148.7          6.0         2.5  virginica
#> 102        329.4       148.7          5.1         1.9  virginica
```

If you want to keep the former columns, you need to make different names by specifying `name`.

```r
iris %>%
  add_count(Species, wt = starts_with("Sepal"),
            name = "{gsub('[a-z.]', '', .col)}_Sum")

#>     Sepal.Length Sepal.Width Petal.Length Petal.Width    Species     SL_Sum   SW_Sum
#> 1            5.1         3.5          1.4         0.2     setosa      250.3    171.4
#> 2            4.9         3.0          1.4         0.2     setosa      250.3    171.4
#> 3            4.7         3.2          1.3         0.2     setosa      250.3    171.4
```

### DIFF
--- a/R/count-tally.R
+++ b/R/count-tally.R
@@ -93,10 +93,14 @@ tally <- function(x, wt = NULL, sort = FALSE, name = NULL) {
 #' @export
 tally.data.frame <- function(x, wt = NULL, sort = FALSE, name = NULL) {
   n <- tally_n(x, {{ wt }})
-  name <- check_name(name, group_vars(x))
 
   local_options(dplyr.summarise.inform = FALSE)
-  out <- summarise(x, !!name := !!n)
+  out <- if (is_call(n, "n", n = 0)) {
+    name <- check_name(name, group_vars(x))
+    summarise(x, !!name := !!n)
+  } else {
+    summarise(x, !!n)
+  }
 
   if (sort) {
     arrange(out, desc(!!sym(name)))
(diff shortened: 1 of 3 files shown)


hadley:
I already find `count()` to be uncomfortably complex and so I'd need convincing before making it more complex. The best way to start that process is with an issue.

PR #6239, opened 2022-04-24
Remove existing continuous benchmark infrastructure

DavisVaughan:
It hasn't worked for quite a while. We can come back to this in the future, but it is probably best to completely remove this for now, and then use this PR as a reference if we come back to it.

This should mean that we finally get a green check mark again on the `main` builds.


DavisVaughan:
This seems fairly harmless and I'd really like to see ✅ on the main branch, so I'm going to merge. Feel free to ping me later if you want to revert some of this!

PR #6245, opened 2022-04-26
Quality of life improvements to `select.R` docs

DavisVaughan:
I have two complaints about the current docs:

- Running `devtools::check()` in the console will alter global options, and won't set them back to their original values. This is particularly annoying for `crayon.enabled`.
- Running `devtools::check()` in the console will almost always change `select.Rd` because the Rmd chunks in `select.R` that roxygen2 runs will print out tibbles, and the number of columns printed out depends on the current `width` of the RStudio console.

I fixed the first problem by adding a teardown chunk at the end of the `select()` docs that resets the old options. I fixed the second problem by adding `width = 80` as one of those options.

@hadley can you think of any other way to do this? I tried things like `rlang::local_options()` with and without setting `.frame = knitr::knit_global()` but didn't have any luck. I don't know of a way to "locally" alter these options and force the documentation chunks to respect the changes.


DavisVaughan:
I had a new idea, see the latest commit. It is based on `knit_child()` from https://bookdown.org/yihui/rmarkdown-cookbook/child-document.html, which gives us something we can wrap in `with_options()`. This feels ok?

hadley:
Coincidentally I just did this in roxygen2: https://github.com/r-lib/roxygen2/commit/ddd0c076870abb3f93536829fc9467e95941e8d9

hadley (approved):
Seems fine to me. I didn't know we were doing this sort of manual example section.

PR #6247, opened 2022-04-26
Update license files with `use_mit_license()`

DavisVaughan:
@hadley there isn't anything else required to change copyright holders, right?


hadley (approved):
Nope, that's good — we're not actually changing the copyright holders here, we're just doing a better job of documenting who they are.

PR #6248, opened 2022-04-26
Update group_split.R documentation to use group_by() before group_keys()

delabj:
Noticed that the current documentation for group_split() uses the depreciated use of `...` in an example with group_keys(). Updated to reflect the practice of calling group_by() first.


hadley:
Thanks!

PR #6253, opened 2022-05-01
Faster dplyr_quosures()

krlmlr:
Most of the time is spent in `enquos()`, there is now only one call to that function.

### DIFF
--- a/R/across.R
+++ b/R/across.R
@@ -391,13 +391,21 @@ new_dplyr_quosure <- function(quo, ...) {
 dplyr_quosures <- function(...) {
   quosures <- enquos(..., .ignore_empty = "all")
   names_given <- names2(quosures)
-  names_auto  <- names(enquos(..., .named = TRUE, .ignore_empty = "all"))
 
   for (i in seq_along(quosures)) {
-    quosures[[i]] <- new_dplyr_quosure(quosures[[i]],
-      name_given = names_given[i],
-      name_auto = names_auto[i],
-      is_named = names_given[i] != "",
+    quosure <- quosures[[i]]
+    name_given <- names_given[[i]]
+    is_named <- (name_given != "")
+    if (is_named) {
+      name_auto <- name_given
+    } else {
+      name_auto <- as_label(quosure)
+    }
+
+    quosures[[i]] <- new_dplyr_quosure(quosure,
+      name_given = name_given,
+      name_auto = name_auto,
+      is_named = is_named,
       index = i
     )
   }


romainfrancois (approved):

R/across.R
```diff
@@ -391,13 +391,21 @@ new_dplyr_quosure <- function(quo, ...) {
...
-  names_auto  <- names(enquos(..., .named = TRUE, .ignore_empty = "all"))
 
   for (i in seq_along(quosures)) {
-    quosures[[i]] <- new_dplyr_quosure(quosures[[i]],
-      name_given = names_given[i],
-      name_auto = names_auto[i],
-      is_named = names_given[i] != "",
+    quosure <- quosures[[i]]
+    name_given <- names_given[[i]]
+    is_named <- (name_given != "")
```
lionel-, line 398:
```suggestion
    is_named <- name_given != ""
```

lionel- (commented):
Since you're not defusing named arguments you can use `quos()` instead of `enquos()` to avoid the overhead of `endots()`.

R/across.R
krlmlr: Why make code less clear and explicit?

krlmlr:
Thanks, I don't understand "not defusing named arguments". This is called directly from `mutate()`? In what setting would `enquos()` be more appropriate?

krlmlr:
All tests pass, the speed gains are dramatic:

``` r
library(rlang)

dplyr_quosures <- function(...) {
  quosures <- enquos(..., .ignore_empty = "all")
  names_given <- names2(quosures)
  names_auto <- names(enquos(..., .named = TRUE, .ignore_empty = "all"))

  for (i in seq_along(quosures)) {
    quosures[[i]] <- dplyr:::new_dplyr_quosure(quosures[[i]],
      name_given = names_given[i],
      name_auto = names_auto[i],
      is_named = names_given[i] != "",
      index = i
    )
  }
  quosures
}

dplyr_quosures_new <- function(...) {
  quosures <- quos(..., .ignore_empty = "all")
  names_given <- names2(quosures)

  for (i in seq_along(quosures)) {
    quosure <- quosures[[i]]
    name_given <- names_given[[i]]
    is_named <- (name_given != "")
    if (is_named) {
      name_auto <- name_given
    } else {
      name_auto <- as_label(quosure)
    }

    quosures[[i]] <- dplyr:::new_dplyr_quosure(quosure,
      name_given = name_given,
      name_auto = name_auto,
      is_named = is_named,
      index = i
    )
  }
  quosures
}

bench::mark(dplyr_quosures(a = 1), dplyr_quosures_new(a = 1))
#> # A tibble: 2 × 6
#>   expression                     min   median `itr/sec` mem_alloc `gc/sec`
#>   <bch:expr>                <bch:tm> <bch:tm>     <dbl> <bch:byt>    <dbl>
#> 1 dplyr_quosures(a = 1)      77.98µs  83.07µs    11617.    9.04MB     61.2
#> 2 dplyr_quosures_new(a = 1)   7.34µs   8.28µs   114491.   64.36KB     57.3
```

<sup>Created on 2022-05-04 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

R/across.R
lionel-: I've never seen parentheses used like that in assignments in our packages. I thought this was a typo from refactoring.

lionel-:
> Thanks, I don't understand "not defusing named arguments".

Like `function(..., foo, bar) enquos(..., foo, bar)`.

R/across.R
```diff
@@ -389,15 +389,23 @@ new_dplyr_quosure <- function(quo, ...) {
 }
 
 dplyr_quosures <- function(...) {
-  quosures <- enquos(..., .ignore_empty = "all")
+  quosures <- quos(..., .ignore_empty = "all")
```
DavisVaughan, line 392: It would probably be good to have a comment here saying why we are using `quos()` rather than `enquos()`. If I saw this then I would immediately have an urge to change it to `enquos()`

krlmlr: Good idea, done.

lionel-:
The first version of rlang didn't have `enquos()`. We added it because it's less confusing for users. Also easier to use the extra functionality like `.named = `.

The next minor version of rlang will be focused on tidyeval and `enquos()` should be as fast as `quos()`. It's probably best not to spend too much time on this.

PR #6263, opened 2022-05-10
Add a `.locale` argument to `arrange()` and implement `dplyr_locale()`

DavisVaughan:
Supersedes and Closes #5942 (once again. It just got too out of date to be useful.)
Closes https://github.com/tidyverse/dplyr/issues/4962
Closes https://github.com/tidyverse/dplyr/pull/5090
Part of https://github.com/tidyverse/dplyr/pull/5808

This PR adds a `.locale` argument to `arrange()`, with 3 possible options:
- `dplyr_locale()`, the default. See below.
- A single stringi locale identifier, like `"en_US"` or `"fr"`. This requires stringi.
- `"C"`, for sorting in the C locale. This does not require stringi.

`dplyr_locale()` is a new exported helper and returns a string representing the default locale to use. It has the following properties:
- It defaults to `"C"`
- If the `dplyr.locale` option is set to a stringi locale identifier or `"C"`, that overrides the above default behavior.

This is a breaking change, as we no longer respect the `LC_COLLATE` option.

- [x] Run revdeps one more time after we decide that this is The Way - `"1bfa0724-c563-48de-9068-51476e19cf82"` (Update - no changes to worse in revdeps)


DavisVaughan:
There was only 1 revdep that was affected by this change (somehow). It was {dexter}, which relied on the assumption that both `df1` and `df2` in the example below would have the same ordering:

```r
df1 <- arrange(df, id)

df2 <- df1 %>%
  group_by(id) %>%
  summarize()
```

This previously "worked" because `arrange()` and `group_by()` both used the system locale when internally sorting the values (group by sorts the keys). We now recommend that you explicitly `arrange()` after the `summarize()` if you need this behavior.

PR in https://github.com/dexter-psychometrics/dexter/pull/6

PR #6269, opened 2022-05-13
Advise on how to silence the `multiple = "warning"` message

DavisVaughan:
By pushing people towards explicit usage of `multiple = "all"`


hsbadr:
Could this also be controlled with an option (e.g., `options(dplyr.multiple = "all")`)? I understand that this can generate unexpected results if the user isn't warned about multiple matches, but it defaults to `NULL` (users need opt in a global option). The good thing about the option is to avoid verbose setting of `multiple = "all"` to silence the warnings, which is the old behavior, while being able to explicitly change it in specific calls. So, the `dplyr.multiple` option defaults to `NULL` and would only be used if the argument `multiple` is missing.

DavisVaughan:
We will do a pre-release blog post to get people to try these new features out before we send them to CRAN. If lots of people complain about this then we will consider it, but I'd rather people just switch to explicitly setting `multiple` instead

iangow:
A whole load of code I have is now throwing warnings because of this change. I think the old default (no warning) was preferable. Often one writes `inner_join(y, x)` and now has to flip code to `inner_join(x, y)` (not always easy if `y` is coming at the end of a series of pipes) or add `multiple = "true"` to avoid warnings.

"I'd rather people just switch to explicitly setting `multiple` instead" seems quite dogmatic about how people use `inner_join`, especially as it's introducing an additional difference between `dplyr` and SQL (additionally relevant given `dbplyr`), where `inner_join(y, x)` and `inner_join(x, y)` are equivalent.

DavisVaughan:
We are planning on improving on this in https://github.com/tidyverse/dplyr/issues/6731

iangow:
> We are planning on improving on this in #6731
> 
> I'm optimistic that most people were doing 1-to-many joins and we were inadvertently warning on those.
> 
> After that is fixed it will only automatically warn on many-to-many joins.

I looked at "R for Data Science" and it seems that this change would allow the exposition there to be simplified. (For example, either [this](https://r4ds.hadley.nz/joins.html#allow-multiple-rows) would need to be a different example with many-to-many or could be simply omitted.)

PR #6059, opened 2021-10-25
Avoid reexporting deprecated APIs

krlmlr:
For `tbl_sum()` and `type_sum()`, users can reexport from pillar or from tibble. The other functions should not be used at all.

This needs a revdepcheck.

Closes #6058.

Issue #6262, opened 2022-05-09
dplyr 1.1.0 revdep tracker


DavisVaughan:

hadley:
I'll take these ones:


## `na_if()` vctrs update #6329
- [ ] circumplex: Doing `dplyr::na_if(., "NaN")` on a double vector, I assume to replace `NaN` with `NA`?
- [ ] forestmangr: Doing `dplyr::na_if(., 0)` where `.` is a tibble, I guess expecting each column to get na-if'd?
- [ ] hlaR: Expecting a character `y` to be compatible with integer `x`
- [ ] huito: "`y` <character> to match type of `x` <data.frame>" with `na_if()`, same as forestmangr?
- [ ] itraxR: Looks like CRAN version had a problem that was fixed here in dev version. Related to `na_if()` on a data frame https://github.com/tombishop1/itraxR/commit/5de1717f60dffc134e20fed25344db0dacf37f91
- [ ] JumpeR: Another "Can't convert `y` <double> to match type of `x` <data.frame>"
- [ ] naniar: Another "Can't convert `y` <double> to match type of `x` <tbl_df>" in the `replace-with-na.Rmd` vignette
- [ ] skater: Looks like calling `na_if(<chr>, 0)` in a vignette on column `dadid`
- [ ] suddengains: Looks like `na_if(<dbl>, "-Inf")` which is weird because `-Inf == -Inf` does return `TRUE`
- [ ] SwimmeR: Another: "Can't convert `y` <character> to match type of `x` <data.frame>."

DavisVaughan:
## Join `check_dots_unused()` (our fault) #6599 
- [ ] disk.frame `anti_join(.x, .y, by = by, copy = copy, ..., overwrite = overwrite)`
- [ ] table.express `left_join.ExprBuilder(roll =  #TRUE)`
- [x] tidytree: https://github.com/YuLab-SMU/tidytree/pull/29 this one had a bug too

lionel-:
## `nth()`, `first()`, `last()` vctrs update #6331
- [x] confoundr: "Can't convert `default` <character> to match type of `x` <double>", `first(x, default = "NA")` with double `x` - https://github.com/jwjackson/confoundr/pull/3
- [x] divseg: Does `dplyr::first(dplyr::cur_data())` expecting to get the first col of the data frame? Should use `pull()`. - https://github.com/christopherkenny/divseg/pull/1
[...]

PR #6277, opened 2022-05-24
Remove `data_frame_()` reexport

DavisVaughan:
Part of #6059 

Job ID `"268ad56c-38d4-4a9f-81e4-f241efeafab6"`


DavisVaughan:
Only affected rabhit, and I've already sent in an issue about this. They used `@importFrom dplyr data_frame_` but never actually used it so it should be easy. Tracking in #6262

PR #6281, opened 2022-05-24
Remove `tbl_sum()` and `type_sum()` reexports

DavisVaughan:
Part of #6059


DavisVaughan:
Based on the following failure report I think we can probably do `tbl_sum()` and go ahead and notify the two package authors, but otherwise I think `type_sum()` is going to require a long time to deprecate.

## tbl_sum

- groupr
- tabxplor

## type_sum

Most of the tidymodels / spark ones are probably the same underlying issue. Fixing it in one of two places probably fixes them all. Still a lot of failure though.

Tidymodels related
- baguette
- dials
- discrim
- rules
- stacks
- usemodels
- tune

Spark related
- sparkavro
- sparkbq
- sparkhail
- sparklyr
- sparklyr.flint
- sparklyr.nested
- sparktf
- sparkxgb

Possibly unique failures
- apache.sedona
- casen
- catalog
- DMwR2
- egor
- finnts
- geospark
- graphframes
- nevada
- pguIMP
- rsparkling
- shinyML
- srvyr
- text

DavisVaughan:
Going with https://github.com/tidyverse/dplyr/pull/6284 instead, which only removes `tbl_sum()`. We can revisit the `type_sum()` removal after tidymodels and sparklyr have all updated to use the pillar version.

PR #6225, opened 2022-04-01
More informative error for `case_when()` RHS type and class mismatches

eutwt:
closes #6206 

Current
``` r
dplyr::case_when(c(TRUE, FALSE) ~ c(3, 4), TRUE ~ c('a', 'b'))
#> Error in names(message) <- `*vtmp*`: 'names' attribute [1] must be the same length as the vector [0]
```

<sup>Created on 2022-03-31 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

New
``` r
dplyr::case_when(c(TRUE, FALSE) ~ c(3, 4), TRUE ~ c('a', 'b'))
#> Error in `dplyr::case_when()`:
#> ! `c("a", "b")` must be a double vector, not a character vector.
```

<sup>Created on 2022-03-31 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

I didn't update snapshots because the line-wrapping in my error messages happens earlier than in the snapshot, and I'm not sure why.

PR #6286, opened 2022-06-01
Rewrite `case_when()` using `vec_case_when()`

### LINKED ISSUE #5106: Ensure case_when() and friends use vctrs
So that you know longer need to know about `NA_character_` etc.

### LINKED ISSUE #6206: More informative error for `case_when` failing due to not matching NA
It seems that since recently mutate `case_when` calls are more strictly evaluated with regard to the variable type (or it always was as strict and some other R or package update, now produces integer instead of double values). 
That's fine, but I think, ideally the error message would be more informative than `'names' attribute [1] must be the same length as the vector [0]`, especially for this borderline case where (depending on the data) the same code might produce either integers or doubles.

``` r
library(dplyr)
#> 
#> Attache Paket: 'dplyr'
#> Die folgenden Objekte sind maskiert von 'package:stats':
#> 
#>     filter, lag
#> Die folgenden Objekte sind maskiert von 'package:base':
#> 
library(magrittr)

tibble(a = c("yes", "no"), 
       b = as.integer(c(0,2))) %>% 
  mutate(b = case_when(a == "yes" ~ NA_real_, 
                       TRUE ~ b))
#> Error in `mutate()`:
#> ! Problem while computing `b = case_when(a == "yes" ~ NA_real_, TRUE ~
#>   b)`.
#> Caused by error in `names(message) <- `*vtmp*``:
#> ! Attribut 'names' [1] muss dieselbe Länge haben wie der Vektor [0]
```

<sup>Created on 2022-03-04 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

<details style="margin-bottom:10px;">
<summary>
Session info
</summary>

``` r
sessioninfo::session_info()
#> - Session info ---------------------------------------------------------------
#>  setting  value
#>  version  R version 4.1.2 (2021-11-01)
#>  os       Windows 10 x64 (build 19044)
#>  system   x86_64, mingw32
#>  ui       RTerm
#>  language (EN)
#>  collate  German_Germany.1252
#>  ctype    German_Germany.1252
#>  tz       Europe/Berlin
#>  date     2022-03-04
#>  pandoc   2.14.0.3 @ C:/Program Files/RStudio/bin/pandoc/ (via rmarkdown)
#> 
#> - Packages -------------------------------------------------------------------
#>  package     * version date (UTC) lib source
#>  assertthat    0.2.1   2019-03-21 [1] CRAN (R 4.1.1)
#>  backports     1.4.1   2021-12-13 [1] CRAN (R 4.1.2)
#>  cli           3.2.0   2022-02-14 [1] CRAN (R 4.1.2)
#>  crayon        1.5.0   2022-02-14 [1] CRAN (R 4.1.2)
#>  DBI           1.1.2   2021-12-20 [1] CRAN (R 4.1.2)
#>  digest        0.6.29  2021-12-01 [1] CRAN (R 4.1.2)
#>  dplyr       * 1.0.8   2022-02-08 [1] CRAN (R 4.1.2)
#>  ellipsis      0.3.2   2021-04-29 [1] CRAN (R 4.1.1)
#>  evaluate      0.15    2022-02-18 [1] CRAN (R 4.1.2)
#>  fansi         1.0.2   2022-01-14 [1] CRAN (R 4.1.2)
#>  fastmap       1.1.0   2021-01-25 [1] CRAN (R 4.1.1)
#>  fs            1.5.2   2021-12-08 [1] CRAN (R 4.1.2)
#>  generics      0.1.2   2022-01-31 [1] CRAN (R 4.1.2)
#>  glue          1.6.2   2022-02-24 [1] CRAN (R 4.1.2)
#>  highr         0.9     2021-04-16 [1] CRAN (R 4.1.1)
#>  htmltools     0.5.2   2021-08-25 [1] CRAN (R 4.1.1)
#>  knitr         1.37    2021-12-16 [1] CRAN (R 4.1.2)
#>  lifecycle     1.0.1   2021-09-24 [1] CRAN (R 4.1.2)
#>  magrittr    * 2.0.2   2022-01-26 [1] CRAN (R 4.1.2)
#>  pillar        1.7.0   2022-02-01 [1] CRAN (R 4.1.2)
#>  pkgconfig     2.0.3   2019-09-22 [1] CRAN (R 4.1.1)
#>  purrr         0.3.4   2020-04-17 [1] CRAN (R 4.1.1)
#>  R.cache       0.15.0  2021-04-30 [1] CRAN (R 4.1.2)
#>  R.methodsS3   1.8.1   2020-08-26 [1] CRAN (R 4.1.1)
#>  R.oo          1.24.0  2020-08-26 [1] CRAN (R 4.1.1)
#>  R.utils       2.11.0  2021-09-26 [1] CRAN (R 4.1.2)
#>  R6            2.5.1   2021-08-19 [1] CRAN (R 4.1.1)
#>  reprex        2.0.1   2021-08-05 [1] CRAN (R 4.1.1)
#>  rlang         1.0.1   2022-02-03 [1] CRAN (R 4.1.2)
#>  rmarkdown     2.12    2022-03-02 [1] CRAN (R 4.1.2)
#>  rstudioapi    0.13    2020-11-12 [1] CRAN (R 4.1.1)
#>  sessioninfo   1.2.2   2021-12-06 [1] CRAN (R 4.1.2)
#>  stringi       1.7.6   2021-11-29 [1] CRAN (R 4.1.2)
#>  stringr       1.4.0   2019-02-10 [1] CRAN (R 4.1.1)
#>  styler        1.6.2   2021-09-23 [1] CRAN (R 4.1.2)
#>  tibble        3.1.6   2021-11-07 [1] CRAN (R 4.1.2)
#>  tidyselect    1.1.2   2022-02-21 [1] CRAN (R 4.1.2)
#>  utf8          1.2.2   2021-07-24 [1] CRAN (R 4.1.1)
#>  vctrs         0.3.8   2021-04-29 [1] CRAN (R 4.1.1)
#>  withr         2.5.0   2022-03-03 [1] CRAN (R 4.1.2)
#>  xfun          0.30    2022-03-02 [1] CRAN (R 4.1.2)
#>  yaml          2.3.5   2022-02-21 [1] CRAN (R 4.1.2)
#> 
#>  [1] C:/Users/ga27jar/Documents/R/win-library/4.1
#>  [2] C:/Program Files/R/R-4.1.2/library
#> 
#> ------------------------------------------------------------------------------
```

</details>

### LINKED ISSUE #6261: `case_when()` error path isn't working correctly
``` r
library(dplyr)

# check_type
case_when(TRUE ~ 1, TRUE ~ "x")
#> Error in names(message) <- `*vtmp*`: 'names' attribute [1] must be the same length as the vector [0]

# check_class
case_when(TRUE ~ 1, TRUE ~ structure(1, class = "foo"))
#> Error in names(message) <- `*vtmp*`: 'names' attribute [1] must be the same length as the vector [0]
```

These are supposed to report on invalid types and classes, respectively.

This happens because we are missing an argument here in `replace_with()`. That `NULL` is probably intended to be passed on to `reason`, but it is instead being matched to `name` and that doesn't work. A character name is required for that error path to run correctly.
https://github.com/tidyverse/dplyr/blob/1cff4b9ce7f404955d8a5764b5708eee5972e0e3/R/case_when.R#L174

DavisVaughan:
Closes #6261 
Closes #6206 
Closes #5106
Closes #6145 
Closes #6225 

The intention is to move `vec_case_when()` to vctrs and rewrite in C, but getting the semantics right in R is more important for now.

`vec_case_when()` will also be used to back `if_else()` and `coalesce()` for sure. I think it could also back `na_if()`. It could additionally back `replace_when()` if we want to implement that.

---

`case_when()` has gained a new interface. Formulas are no longer used, instead you pass pairs of `condition/value` inputs. There is also an explicit `.default` argument now, and new `.ptype` and `.size` arguments.

The formula interface still works for now, I'm optimistic that it is 100% backwards compatible. We don't do anything to actively dissuade people from using the formula interface right now. However, you can't use any of the new arguments with the old interface.

---

Open question:

Should we have a `.missing` argument in `case_when()` and `vec_case_when()`? This would:
- Match `if_else()`
- Be easier than supplying `is.na(x), "value"` for most cases
- Provide a way to handle the case where the missing value pops up in the computation rather than in `x` itself (see below)

We decided that `.default` should ONLY handle the case where all `condition`s are `FALSE`. Meaning it doesn't handle the case where no conditions are `TRUE` and at least one of those conditions is `NA` - the `NA` now gets propagated through. This is generally what people want, see this example:

``` r

x <- c(1, 2, NA, 4, 5)

# Confusing because `NA` isn't propagated.
# The `NA` gets assigned to `"high"` which feels wrong.
case_when(
  x <= 2 ~ "low",
  x <= 4 ~ "med",
  TRUE ~ "high"
)
#> [1] "low"  "low"  "high" "med"  "high"

# New interface propagates `NA` - good!
case_when(
  x <= 2, "low",
  x <= 4, "med",
  .default = "high"
)
#> [1] "low"  "low"  NA     "med"  "high"

# Handle `NA` easily because we knew where it came from
case_when(
  x <= 2, "low",
  x <= 4, "med",
  is.na(x), "unknown",
  .default = "high"
)
#> [1] "low"     "low"     "unknown" "med"     "high"
```

The above was the motivating case for this change. But it is a little more frustrating if the `NA`s occur in the computation of the condition rather than in `x` itself. I still think the `.default` behavior is right, the problem is that it is hard to explicitly handle the `NA` cases now. A `.missing` argument would allow us to easily explicitly handle the computed `NA`s.

``` r

x <- c(-1, 0, 1)

# Generated `NA` is treated like `FALSE`.
# `TRUE ~` applies to everything that is left, but does that really make sense?
case_when(
  sqrt(x) > 0 ~ "big",
  TRUE ~ "little"
)
#> Warning in sqrt(x): NaNs produced
#> [1] "little" "little" "big"

# Propagates `NA`, which seems reasonable
case_when(
  sqrt(x) > 0, "big",
  .default = "little"
)
#> Warning in sqrt(x): NaNs produced
#> [1] NA       "little" "big"

# But there is no easy way to "handle" the `NA` because it happens
# in the computation, not in `x`
case_when(
  sqrt(x) > 0, "big",
  is.na(x), "missing",
  .default = "little"
)
#> Warning in sqrt(x): NaNs produced
#> [1] NA       "little" "big"

# Proposal:
# case_when(
#   sqrt(x) > 0, "big",
#   .default = "little",
#   .missing = "missing"
# )
```

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,55 @@
 # dplyr (development version)
 
+* `case_when()` has a new interface and has been rewritten to utilize vctrs. The
+  existing formula interface will continue to work for awhile, but we now feel
+  that it is suboptimal and encourage you to switch to the new interface
+  (#5106).
+  
+  Where you used to use:
+  
+  ```
+  case_when(
+    condition1 ~ value1,
+    condition2 ~ value2,
+    TRUE ~ default
+  )
+  ```
+  
+  You now supply pairs of inputs (with no `~`) and an explicit `.default`
+  argument:
+  
+  ```
+  case_when(
+    condition1, value1,
+    condition2, value2,
+    .default = default
+  )
+  ```
+  
+  The rewrite with vctrs includes a number of additional benefits:
+  
+  - The types of the values no longer have to match exactly. For example, the
+    following no longer requires you to use `NA_character_` instead of just
+    `NA`.
+  
+    ```
+    case_when(
+      x %in% c("little", "small"), "one"
+      x %in% c("big", "large"), "two",
+      x %in% c("missing", "unknown"), NA
+    )
+    ```
+    
+  - `case_when()` now supports a larger variety of value types. For example,
+    you can use a data frame to create multiple columns at once.
+    
+  - There are new `.ptype` and `.size` arguments which allow you to enforce
+    a particular output type and size. This allows you to construct a completely
+    type and size stable call to `case_when()`.
+  
+  - The error thrown when types or lengths were incorrect has now been
+    improved (#6261, #6206).
+
 * `tbl_sum()` is no longer reexported from tibble (#6284).
 
 * `slice_sample()` now gives a more informative error when `replace = FALSE` and
--- a/NEWS.md
+++ b/NEWS.md
@@ -26,6 +26,11 @@
   )
   ```
   
+  Additionally, unlike `TRUE ~ default`, the `.default` argument isn't used for
+  cases where an `NA` should have been propagated through. To handle those
+  cases, there is a new `.missing` argument. This should replace most uses
+  of `is.na(x) ~ value`.
+  
   The rewrite with vctrs includes a number of additional benefits:
   
   - The types of the values no longer have to match exactly. For example, the
--- a/R/case_when.R
+++ b/R/case_when.R
@@ -138,6 +138,17 @@
 #'     )
 #'   )
 #'
+#' # You can use data frames to create multiple columns at once.
+#' # The data frames will automatically be unpacked.
+#' starwars %>%
+#'   select(name:mass, gender, species) %>%
+#'   mutate(
+#'     case_when(
+#'       height > 200 | mass > 200, tibble(type = "large", vehicle = "car"),
+#'       species == "Droid", tibble(type = "robot", vehicle = "none"),
+#'       .default = tibble(type = "other", vehicle = "unknown")
+#'     )
+#'   )
 #'
 #' # `case_when()` is not a tidy eval function. If you'd like to reuse
 #' # the same patterns, extract the `case_when()` call in a normal
--- a/R/case_when.R
+++ b/R/case_when.R
@@ -1,104 +1,140 @@
-#' A general vectorised if
+#' A general vectorised if-else
 #'
-#' This function allows you to vectorise multiple [if_else()]
-#' statements. It is an R equivalent of the SQL `CASE WHEN` statement.
-#' If no cases match, `NA` is returned.
+#' This function allows you to vectorise multiple [if_else()] statements. It is
+#' an R equivalent of the SQL `CASE WHEN` statement. If no cases match, a
+#' missing value is returned unless a `.default` is supplied.
 #'
-#' @param ... <[`dynamic-dots`][rlang::dyn-dots]> A sequence of two-sided formulas. The left hand side (LHS)
-#'   determines which values match this case. The right hand side (RHS)
-#'   provides the replacement value.
+#' @param ... <[`dynamic-dots`][rlang::dyn-dots]>
 #'
-#'   The LHS must evaluate to a logical vector. The RHS does not need to be
-#'   logical, but all RHSs must evaluate to the same type of vector.
+#'   Pairs of inputs supplied like
+#'   `condition1, value1, condition2, value2, ...`. The `condition`s determine
+#'   which values match this case, the `value`s provide the value to use for
+#'   this case.
 #'
-#'   Both LHS and RHS may have the same length of either 1 or `n`. The
-#'   value of `n` must be consistent across all cases. The case of
-#'   `n == 0` is treated as a variant of `n != 1`.
+#'   Each `condition` input must evaluate to a logical vector. All `condition`
+#'   inputs must be the same length.
 #'
-#'   `NULL` inputs are ignored.
+#'   The `value` inputs will be coerced to their common type. All `value`
+#'   inputs must be length 1 or the same length as the `condition`s.
+#'
+#'   An `NA` in a `condition` will result in a missing value in that location
+#'   of the output unless another `condition` evaluates to `TRUE` for that
+#'   location.
+#'
+#'   If the `...` are named, those names will be utilized in any error messages.
+#'
+#' @param .default The default value used when all `condition`s return `FALSE`
+#'   for a particular location. `.default` must be length 1 or the same length
+#'   as the `condition`s. `.default` participates in the computation of the
+#'   common type alongside the `value` inputs.
+#'
+#'   If `NULL`, the default, a missing value will be placed in the result.
+#'
+#' @param .ptype An optional prototype declaring the desired output type. If
+#'   supplied, this overrides the common type of the `value` inputs.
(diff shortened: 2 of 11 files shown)


markfairbanks:
I think the proposal of a `.missing` arg is a good way to go since `.default` won't handle `NA`s. In the case of using one vector the use case isn't quite as obvious, but in the case of 2+ vectors you would have to keep note of all vectors used in a condition and create an `is.na()` condition for each one.

``` r
devtools::load_all(".")
#> ℹ Loading dplyr

x <- c(1, 2, NA, 4, 5)
y <- c(1, 2, 3, NA, 5)
z <- c(1, 2, 3, 4, NA)

case_when(
  x <= 2 & y <= 2, "low",
  x <= 4 & y <= 4, "med",
  x == 5 & z == 5, "different",
  is.na(x) | is.na(y) | is.na(z), "unknown",
  .default = "high"
)
#> [1] "low"     "low"     "unknown" "unknown" "unknown"
```

DavisVaughan:
This also comes up when `%in%` is used, which I think is a fairly common use case of `case_when()`.

``` r
x <- c("a", "b", NA, "e")

vec_case_when(
  x %in% c("a", "e", "i", "o", "u"), "vowel",
  .default = "consonant",
  .missing = "missing"
)
#> [1] "vowel"     "consonant" "consonant" "vowel"
```

Maybe the original design where `.default` applies to both `FALSE` and `NA` was correct? So you are forced to deal with the missing values yourself, however they might come up, rather than relying on `.missing` which might not always work as you expect (because sometimes your `NA`s can become `FALSE` when the condition is computed). The main reason we didn't like the original design is because when missing values do propagate through, you get weird results like this, which the new version does handle better.

```r
x <- c(1, 2, NA)

dplyr::case_when(
  x %% 2 == 0  ~ "odd",
  TRUE ~ "even"
)
#> [1] "even" "odd"  "even"

vec_case_when(
  x %% 2 == 0, "odd",
  .default = "even"
)
#> [1] "even" "odd"  NA
```

DavisVaughan:
Superseded by #6300

PR #6291, opened 2022-06-08
[vignettes/base.Rmd] Fix function arguments order

rivizan:
When comparing to `select(df, starts_with("x"))`, the grepl function had inverted arguments.


hadley:
Thanks!

PR #6018, opened 2021-09-16
Update `group_by()` algorithm to utilize `vec_locate_sorted_groups()`

DavisVaughan:
Probably merge after #5942 when we are working on dplyr 1.1.0

Closes #5808 
Closes #4406

PR #6297, opened 2022-06-14
Update `group_by()` algorithm to utilize `vec_locate_sorted_groups()`

### LINKED ISSUE #4406: group_by performance: potential for easy and substantial improvement

DavisVaughan:
Supersedes #6018, too out of date
Closes #4406 

Linked tidyup, https://github.com/tidyverse/tidyups/blob/main/006-dplyr-group-by-ordering.md

Main changes:
- `group_by()` now internally computes and orders groups with `vec_locate_sorted_groups()`, which is often much faster than the previous approach

- `group_by()` now internally sorts character vector group columns in the C locale rather than the system locale. This affects the ordering of the result of functions that use the group data, like `summarise()` and `group_split()`. Note that this is consistent with the new default of `arrange()`, which also uses the C locale.

- A temporary boolean global option, ~`dplyr.legacy_group_by_locale`~ `dplyr.legacy_locale`, has been added that allows users to revert back to respecting the system locale. This is mainly for users that are on a deadline and need it to "just work" like it did before. It is better to explicitly call `arrange(.locale =)` after `summarise()` instead. I tried to make it clear that this option will be removed in the future, and gave an example of what to do instead.
  - We will update `arrange()` to respect this as well.

`"8965bce6-c05d-498f-a9f3-edc8d8f49740"`

### DIFF
--- a/R/group-by.r
+++ b/R/group-by.r
@@ -38,9 +38,8 @@
 #'
 #' @section Ordering:
 #' Currently, `group_by()` internally orders the groups in ascending order. This
-#' should be seen as an implementation detail and is not a behavior that should
-#' be relied upon. This results in ordered output from functions that aggregate
-#' groups, such as [summarise()].
+#' results in ordered output from functions that aggregate groups, such as
+#' [summarise()].
 #'
 #' When used as grouping columns, character vectors are ordered in the C locale
 #' for performance and reproducibility across R sessions. If the resulting
--- a/R/group-by.r
+++ b/R/group-by.r
@@ -59,10 +59,10 @@
 #'
 #' Prior to dplyr 1.1.0, character vector grouping columns were ordered in the
 #' system locale. If you need to temporarily revert to this behavior, you can
-#' set the global option `dplyr.legacy_group_by_locale` to `TRUE`, but this
-#' should be used sparingly and you should expect this option to be removed in a
-#' future version of dplyr. It is better to update existing code to explicitly
-#' call `arrange()` instead.
+#' set the global option `dplyr.legacy_locale` to `TRUE`, but this should be
+#' used sparingly and you should expect this option to be removed in a future
+#' version of dplyr. It is better to update existing code to explicitly call
+#' `arrange()` instead.
 #'
 #' @export
 #' @examples
--- a/R/group-by.r
+++ b/R/group-by.r
@@ -35,6 +35,36 @@
 #'
 #' * `group_by()`: \Sexpr[stage=render,results=rd]{dplyr:::methods_rd("group_by")}.
 #' * `ungroup()`: \Sexpr[stage=render,results=rd]{dplyr:::methods_rd("ungroup")}.
+#'
+#' @section Ordering:
+#' Currently, `group_by()` internally orders the groups in ascending order. This
+#' should be seen as an implementation detail and is not a behavior that should
+#' be relied upon. This results in ordered output from functions that aggregate
+#' groups, such as [summarise()].
+#'
+#' When used as grouping columns, character vectors are ordered in the C locale
+#' for performance and reproducibility across R sessions. If the resulting
+#' ordering of your grouped operation matters and is dependent on the locale,
+#' you should follow up the grouped operation with an explicit call to
+#' [arrange()] and set the `.locale` argument. For example:
+#'
+#' ```
+#' data %>%
+#'   group_by(chr) %>%
+#'   summarise(avg = mean(x)) %>%
+#'   arrange(chr, .locale = "en")
+#' ```
+#'
+#' This is often useful as a preliminary step before generating content intended
+#' for humans, such as an HTML table.
+#'
+#' Prior to dplyr 1.1.0, character vector grouping columns were ordered in the
+#' system locale. If you need to temporarily revert to this behavior, you can
+#' set the global option `dplyr.legacy_group_by_locale` to `TRUE`, but this
+#' should be used sparingly and you should expect this option to be removed in a
+#' future version of dplyr. It is better to update existing code to explicitly
+#' call `arrange()` instead.
+#'
 #' @export
 #' @examples
 #' by_cyl <- mtcars %>% group_by(cyl)
--- a/R/grouped-df.r
+++ b/R/grouped-df.r
@@ -306,11 +306,7 @@ dplyr_locate_sorted_groups <- function(x) {
   out
 }
 dplyr_legacy_group_by_locale <- function() {
-  out <- peek_option("dplyr.legacy_group_by_locale")
-
-  if (is_null(out)) {
-    out <- FALSE
-  }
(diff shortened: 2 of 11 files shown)


tests/testthat/test-grouped-df.r
```diff
@@ -189,3 +189,32 @@ test_that("helper gives meaningful error messages", {
...
+  on_mac <- identical(tolower(Sys.info()[["sysname"]]), "darwin")
+  skip_if_not(on_mac, message = "Not on Mac. Unsure if we can use 'en_US' locale.")
+
+  local_options(dplyr.legacy_group_by_locale = TRUE)
+  withr::local_collate("en_US")
+
+  df <- tibble(x = c("a", "A", "Z", "b"))
+  result <- compute_groups(df, "x")
+  expect_identical(result$x, c("a", "A", "b", "Z"))
+})
```
DavisVaughan, line 215: This seemed like the best I could do to have a test that used a non-C locale so we could ensure the global option was working.
- On Windows it isn't called `"en_US"` except in R >=4.2.0 with the new UTF-8 support
- On Linux there might not be extensive locale support

R/group-by.r
```diff
@@ -35,6 +35,36 @@
 #'
 #' * `group_by()`: \Sexpr[stage=render,results=rd]{dplyr:::methods_rd("group_by")}.
 #' * `ungroup()`: \Sexpr[stage=render,results=rd]{dplyr:::methods_rd("ungroup")}.
+#'
+#' @section Ordering:
+#' Currently, `group_by()` internally orders the groups in ascending order. This
+#' should be seen as an implementation detail and is not a behavior that should
```
hadley, line 41: I don't think there's much point in saying this, as no matter what we say, people will rely on it accidentally.

R/group-by.r
```diff
@@ -35,6 +35,36 @@
...
+#'   summarise(avg = mean(x)) %>%
+#'   arrange(chr, .locale = "en")
+#' ```
+#'
+#' This is often useful as a preliminary step before generating content intended
+#' for humans, such as an HTML table.
+#'
+#' Prior to dplyr 1.1.0, character vector grouping columns were ordered in the
+#' system locale. If you need to temporarily revert to this behavior, you can
+#' set the global option `dplyr.legacy_group_by_locale` to `TRUE`, but this
```
hadley, line 63: Could we make a single option that brings back the legacy behaviour of `arrange()`?

R/grouped-df.r
```diff
@@ -294,11 +294,33 @@ expand_groups <- function(old_groups, positions, nr) {
...
+  }
+
+  out
+}
+dplyr_legacy_group_by_locale <- function() {
+  out <- peek_option("dplyr.legacy_group_by_locale")
+
+  if (is_null(out)) {
+    out <- FALSE
+  }
```
hadley, line 313:
```suggestion
  out <- getOption("dplyr.legacy_group_by_locale", FALSE)
```

?

tests/testthat/test-grouped-df.r
hadley: I think you can just skip on windows. I'm pretty sure it should be safe to assume that linux has a `en_US` locale.

R/group-by.r
DavisVaughan: We could have these two global options

```
# Picked up in `dplyr_locale()` and used by `arrange()` as the default stringi locale
dplyr.locale = NULL / string

# If `TRUE`:
# - `arrange()` would always use `vec_order_base()` and would ignore `.locale`
# - `group_by()` would use `vec_order_base()`
dplyr.legacy_locale = NULL / bool
```

DavisVaughan: I updated this pr to respect `dplyr.legacy_locale` in `group_by()`. `arrange()` will take a little work so I'll do that in a follow up PR

PR #6301, opened 2022-06-18
Error if a column is provided twice in `...` of `relocate`

### LINKED ISSUE #6209: `relocate.data.frame()` shouldn't rely on `vec_unique()` retaining names
Currently, `relocate()` relies on the fact that `vec_unique()` keeps names
https://github.com/tidyverse/dplyr/blob/8abb54b60e40ef7c619156a12b14872cb2eb7989/R/relocate.R#L83-L85

We think we should actually drop names from the result of `vec_unique()`, since the returned names are reliant on the fact that only the _first_ encountered duplicate value is returned (this is also consistent with `unique()`). See https://github.com/r-lib/vctrs/issues/1442 and https://github.com/r-lib/vctrs/pull/1532 where we tried to implement this, but found that dplyr relies on this behavior currently.

---

One easy way to fix this is to use `vec_unique_loc()` to get the locations, and then `vec_slice()`, which will retain names.

That should work in the immediate term but it may be worth looking at the code to also see why we currently get names off the unique values, and if there is something else we should be doing instead, since that _feels_ unreliable.

eutwt:
closes #6209 

``` r
#> Warning: package 'dplyr' was built under R version 4.1.2
df <- tibble(a = 1, b = 2)

#### current behavior
df %>% 
  relocate(new = b, new2 = b)
#> # A tibble: 1 × 2
#>     new     a
#>   <dbl> <dbl>
#> 1     2     1

#### this PR
devtools::load_all("~/Documents/GitHub/dplyr_tv")
#> ℹ Loading dplyr
df %>% 
  relocate(new = b, new2 = b)
#> Error in `relocate()`:
#> ! Named arguments must be unique, but `b` is supplied twice

## only errors if duplicated field is named
df %>% 
  relocate(b, b)
#> # A tibble: 1 × 2
#>       b     a
#>   <dbl> <dbl>
#> 1     2     1
```

<sup>Created on 2022-07-18 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

### DIFF
--- a/R/relocate.R
+++ b/R/relocate.R
@@ -52,6 +52,10 @@ relocate <- function(.data, ..., .before = NULL, .after = NULL) {
 #' @export
 relocate.data.frame <- function(.data, ..., .before = NULL, .after = NULL) {
   to_move <- tidyselect::eval_select(expr(c(...)), .data)
+  if (anyDuplicated(to_move)) {
+    duplicated_col <- names(.data)[which.max(duplicated(to_move))]
+    abort(glue("Columns to move must be unique, but `{duplicated_col}` is supplied twice"))
+  }
 
   .before <- enquo(.before)
   .after <- enquo(.after)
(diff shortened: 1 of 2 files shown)


hadley:
@DavisVaughan can you review this please?

Issue #6309, opened 2022-06-27
Should pass error calls instead of using `fix_call()`


lionel-:
vctrs and tidyselect have been updated to support error calls. We should generally use this instead of `fix_call()`. This should improve backtraces as the trace suffix from the bottom call up to `call = <frame>` will be hidden by default.

Also tidyselect now creates chained errors, and the `fix_call()` in `select()` doesn't interact well with these. The parent error is missing:

```r
f <- function() g()
g <- function() h()
h <- function() abort("foo")
dplyr::select(mtcars, f())
#> ! Problem while evaluating `f()`.
```

hadley:
Looks like we still need `error_call`/`call` arguments for `vec_rbind()`, and `vec_cbind()`.

Issue #6308, opened 2022-06-26
Errors in mask methods carry wrong calls


lionel-:
CRAN rlang doesn't display calls of the form `foo$fn()`. Dev rlang now does, which reveals a few call problems in the error snapshots, for instance:

```r
tibble() %>% summarise(stop("{"))
#> Error in `summarise()`:
#> ! Problem while computing `..1 = stop("{")`.
#> Caused by error in `mask$eval_all_summarise()`:
#> ! {
```

```r
tbl <- tibble(x = 1:2, y = 1:2)
tbl %>%
  group_by(x) %>%
  mutate(y = NULL, a = sum(y))
#> Error in `mutate()`:
#> ! Problem while computing `a = sum(y)`.
#> ℹ The error occurred in group 1: x = 1.
#> Caused by error in `mask$eval_all_mutate()`:
#> ! object 'y' not found
```

PR #6310, opened 2022-06-27
Use chained error created by tidyselect

lionel-:
Part of #6309

Uses chained errors now created by tidyselect:

```r
f <- function() g()
g <- function() h()
h <- function() abort("foo")

dplyr::select(mtcars, f())
#> Error in `dplyr::select()`:
#> ! Problem while evaluating `f()`.
#> Caused by error in `h()`:
#> ! foo
```

Errors within the selection DSL are borrowed instead of chained, as before:

```r
dplyr::select(mtcars, foobar)
#> Error in `dplyr::select()`:
#> ! Can't subset columns that don't exist.
#> ✖ Column `foobar` doesn't exist.
```

@DavisVaughan @hadley Could you take a look at the wording of these chained errors please? If they look good to you, I'll send tidyselect to CRAN.

(I expect R CMD check to fail as the snapshots in the main branch still need to be updated for the latest rlang, see #6308.)

### DIFF
--- a/R/select.R
+++ b/R/select.R
@@ -66,9 +66,10 @@ select.list <- function(.data, ...) {
 select.data.frame <- function(.data, ...) {
   error_call <- dplyr_error_call()
 
-  loc <- tidyselect_fix_call(
-    tidyselect::eval_select(expr(c(...)), .data),
-    call = error_call
+  loc <- tidyselect::eval_select(
+    expr(c(...)),
+    data = .data,
+    error_call = error_call
   )
   loc <- ensure_group_vars(loc, .data, notify = TRUE)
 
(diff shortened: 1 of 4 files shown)


hadley (approved):
LGTM

R/select.R
```diff
@@ -66,9 +66,10 @@ select.list <- function(.data, ...) {
 select.data.frame <- function(.data, ...) {
   error_call <- dplyr_error_call()
 
-  loc <- tidyselect_fix_call(
```
DavisVaughan, line 69: Can you please remove `tidyselect_fix_call()`?

DavisVaughan (approved):
LGTM, can we switch the main branch to rlang 1.0.3 and tidyselect 1.1.3 in a different PR and merge that in, then rebase this on that?

hadley:
Can we finish this off and get it in?

PR #6312, opened 2022-06-28
Tweak join_by() docs

hadley:
This eliminates the very long description in favour of a new "join types" section.


DavisVaughan (approved):
Looks great! I merge main into this, which should fix the build failures

PR #6320, opened 2022-07-05
Update the storms dataset

### LINKED ISSUE #6319: Update Storms dataset
* 2020 and 2021 data is available
* Some storms are missing. Probably time to redo (and simplify!) the parser.
  * 2020:  Bertha, Dolly, Laura, Delta
  * 2021: Grace

Updated data file:
https://www.nhc.noaa.gov/data/hurdat/hurdat2-1851-2021-041922.txt

I've been maintaining it, so I'll have a look.

steveharoz:
(closes #6319)

1. A bug in the code to reformat the data was causing some storms to be dropped. That's been fixed.
2. Data for 2021 storms has been added.
3. I've added data for earlier storms (1852-1974).

Point 3 might be worth discussing. Whoever originally added the dataset to dplyr dropped storms before 1975. I've been doing the same since I've been updating it, but I haven't seen a clear rationale. Considerations for adding the early data:
* PRO: More data
* CON: Bigger data file 42k -> 88k
* CON: The early data may be less useful. E.g., many storms from the 1800s have only a single data point.
* PRO: This is supposed to be an educational dataset. Filtering out the less useful data is a simple realistic exercise for learning dplyr.


mydatacz:
Hi Steve!

Quick suggestion:  to add the other status codes to the line below - if someone does want to keep the filter commented out, they won’t have to go back and clean the status for those.

Thanks again!

status = factor(recode(status, "HU" = "hurricane", "TS" = "tropical storm", "TD" = "tropical depression"))
EX – extratropical
SD – subtropical depression
SS – subtropical storm
LO – low 
WV – tropical wave
DB – disturbance

steveharoz:
Trying to figure out what's going on with storm categorization. Bug in my parser? Or inconsistency in NOAA's categorization?

```r
> storms %>% 
+     filter(category > 0, !(status %in% c("hurricane", "EX"))) %>% 
+     select(name, year, month, day, hour, lat, long, status, wind, pressure)
# A tibble: 6 × 10
  name      year month   day  hour   lat  long status          wind pressure
  <chr>    <dbl> <dbl> <int> <dbl> <dbl> <dbl> <fct>          <int>    <int>
1 AL091968  1968     9    20    12  35.5 -49.5 SS                75      976
2 AL091968  1968     9    21    12  39.6 -44.7 SS                65      982
3 AL181979  1979    10    24    18  40.5 -62   SS                65      985
4 EMILY     2005     7    20    18  25   -98.7 tropical storm    70      975
5 DORIAN    2019     9     7    18  42.8 -64.6 LO                80      954
6 DORIAN    2019     9     8     0  45.2 -62.9 LO                80      956
```

mydatacz:
I think it may not be either issues.  I think the data is likely correct.

Subtropical, Extratropical, Lows and Disturbances can all have high wind intensity, but that doesn't mean they are hurricanes.  A storm needs to be determined to be a tropical cyclone before it can rise to the level of a hurricane (based on wind speed/intensity).  Definitions of types of storms here:  https://www.nhc.noaa.gov/aboutgloss.shtml

https://www.nhc.noaa.gov/data/hurdat/hurdat2-format-atl-1851-2021.pdf

HU (Spaces 20-21, before 4th comma) – Status of system. Options are:
TD – Tropical cyclone of tropical depression intensity (< 34 knots)
TS – Tropical cyclone of tropical storm intensity (34-63 knots)
HU – Tropical cyclone of hurricane intensity (> 64 knots)
EX – Extratropical cyclone (of any intensity)
SD – Subtropical cyclone of subtropical depression intensity (< 34 knots)
SS – Subtropical cyclone of subtropical storm intensity (> 34 knots)
LO – A low that is neither a tropical cyclone, a subtropical cyclone, nor an extratropical cyclone (of any intensity) WV – Tropical Wave (of any intensity)
DB – Disturbance (of any intensity)

So if a storm does not meet the requirements to be classified as a tropical cyclone, regardless of wind speed, it will never have a status of hurricane.

For example, you've probably experienced wind conditions higher than 34 - 47 knots, which is a gale, unless the wind was associated with a storm that was already determined to be a tropical cyclone (based on additional criteria aside from wind).  

Or you may have been in a winter Nor'easter (extratropical storm that can have winds over 65 knots but isn't a hurricane).  I don't know the measurements behind it, but reading the definition and the link below it appears there are multiple characteristics/metrics used to determine if a storm is a tropical cyclone.  

https://www.weather.gov/source/zhu/ZHU_Training_Page/tropical_stuff/sub_extra_tropical/subtropical.htm

Category > 0 had me stumped also at first, but then I realized the categories are based on a wind scale.  Non-tropical cyclone storms also get assigned categories based on wind in the data.  At first glance they appear to coincide with the Saffir-Simpson Wind scale, but I don't know that for sure.  Having followed weather reports closely as a sailor, though, I've never heard NOAA refer to a category 1 gale or category 1 nor'easter, so I think NOAA only uses categories to describe hurricanes.  https://www.nhc.noaa.gov/aboutsshws.php

luis.df <- storms %>%
  filter(name == "Luis") %>%
  select(year, name, category, status, wind)

47 | 1995 | Luis | 2 | hurricane | 95
48 | 1995 | Luis | 2 | hurricane | 90

53 | 1995 | Luis | 2 | EX | 85
54 | 1995 | Luis | 2 | EX | 95
55 | 1995 | Luis | 3 | EX | 105
56 | 1995 | Luis | 2 | EX | 90
57 | 1995 | Luis | 1 | EX | 75
58 | 1995 | Luis | 0 | EX | 60

Hurricane Luis, for example, was an extratropical storm at some point, with wind categories 1, 2, and 3, but the other characteristic of the storm at that time did not meet the criteria for a tropical cyclone anymore, even though the winds were often higher that when it was a tropical cyclone.

So I guess the take-away would be not to filter the data based on category thinking you are only going to get hurricanes.

mydatacz:
Actually, I just realized when looking at the original hurdat2 format file linked above, it seems category is not a column coming from NOAA.  Since it is a column added in/calculated as part of the dyplr file, maybe only add the category for the tropical depressions, tropical storms and hurricanes, and leave the other storm types with an NA?

mydatacz:
Actually, I just realized when looking at the original hurdat2 format file linked above, it seems category is not a column coming from NOAA. Since it is a column added in/calculated as part of the dyplr file, maybe only add the category for the tropical depressions, tropical storms and hurricanes, and leave the other storm types with an NA?  Although I'm not sure of what the purpose of the -1 and 0 category are for tropical depressions and tropical storms, since the Saffir-Simpson Wind scale doesn't start until 1 for hurricanes.  If you decide to use category for only hurricanes, all the other storm status' could be 0.

mydatacz:
Actually, I just realized when looking at the original hurdat2 format file linked above, it seems category is not a column coming from NOAA. Since it is a column added in/calculated as part of the dyplr file, maybe only add the category for the tropical depressions, tropical storms and hurricanes, and leave the other storm types with an NA? I'm not sure of what the purpose of the -1 and 0 category are for tropical depressions and tropical storms, since the Saffir-Simpson Wind scale doesn't start until 1 for hurricanes. If you decide to use category for only hurricanes, all the other storm status categories could be 0.

steveharoz:
Yes, category is calculated from windspeed. I've made that a bit more clear in docs and set it to NA for everything that's not a hurricane.

R/data-storms.R
```diff
@@ -1,33 +1,41 @@
...
 #' @examples
 #'
-#' # show a plot of the storm paths
+#' # show a plot of the storm paths in 1975 or later
 #' if (requireNamespace("ggplot2", quietly = TRUE)) {
 #'   library(ggplot2)
-#'   ggplot(storms) +
+#'   storms %>%
+#'     filter(year >= 1975) %>%
+#'   ggplot() +
```
hadley, line 38:
```suggestion
#'     ggplot() +
```

hadley: And maybe put the `aes()` in the `ggplot()` call?

data-raw/storms.R
```diff
@@ -88,35 +102,67 @@ storms <- storm_dataframes %>%
...
+  )) %>%
+  relocate(category, .after = status)
 
+# drop storms without at least one record that is a tropical depression or higher
+storms <- storms %>%
+  group_by(year, name) %>%
+  mutate(is_depression_or_higher = any(status %in% c("hurricane", "tropical storm", "tropical depression"))) %>%
+  ungroup() %>%
+  filter(is_depression_or_higher) %>%
+  select(-is_depression_or_higher)
```
hadley, line 153:
```suggestion
  filter(any(status %in% c("hurricane", "tropical storm", "tropical depression"))) %>%
  ungroup()
```

?

NEWS.md
```diff
@@ -1,5 +1,6 @@
 # dplyr (development version)
 
+* `storms` has been updated to include 2021 data, data prior to 1975, and some missing storms that were omitted due to an error.
```
```suggestion
* `storms` has been updated to include 2021 data, data prior to 1975, and some missing storms that were omitted due to an error (#6320).
```

NEWS.md
```diff
@@ -1,5 +1,6 @@
...
 * `coalesce()` now more fully embraces the principles of vctrs (#6265).
```
DavisVaughan, line 4: When we merge this in, remember to add a space between this NEWS bullet and the next one

hadley:
@steveharoz do you want to finish this off?

steveharoz:
@hadley Yeah. I'll finish it later this week.

R/data-storms.R
steveharoz: The aes on its own line is just my one-thing-per-line code style. Feel free to change it though. Commit coming soon...

R/data-storms.R
```diff
@@ -1,33 +1,41 @@
 #' Storm tracks data
 #'
-#' This data is a subset of the NOAA Atlantic hurricane database best track
-#' data, \url{https://www.nhc.noaa.gov/data/#hurdat}. The data includes the
-#' positions and attributes of storms from 1975-2020, measured every six hours
-#' during the lifetime of a storm.
+#' This dataset is the NOAA Atlantic hurricane database best track data, \url{https://www.nhc.noaa.gov/data/#hurdat}.
+#' The data includes the positions and attributes of storms from 1852-2021.
+#' The modern storms are measured every six hours during the lifetime of the storm.
```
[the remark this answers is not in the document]
steveharoz: done. see latest commit.

R/data-storms.R
```diff
@@ -1,33 +1,41 @@
...
-#' \item{category}{Saffir-Simpson storm category (estimated from wind speed.
-#' -1 = Tropical Depression, 0 = Tropical Storm)}
+#' \item{category}{Saffir-Simpson hurricane category calculated from wind speed. Only applies to hurricanes.
+#'   \itemize{
+#'     \item 64+ knots = Category 1
+#'     \item 83+ knots = Category 2
+#'     \item 96+ knots = Category 3
+#'     \item 113+ knots = Category 4
+#'     \item 137+ knots = Category 5
+#'   }
```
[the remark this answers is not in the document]
steveharoz: done. see latest commit.

R/data-storms.R
```diff
@@ -1,33 +1,41 @@
...
 #' \item{pressure}{Air pressure at the storm's center (in millibars)}
-#' \item{tropicalstorm_force_diameter}{Diameter (in nautical miles) of the area experiencing tropical storm strength winds (34 knots or above)}
-#' \item{hurricane_force_diameter}{Diameter (in nautical miles) of the area experiencing hurricane strength winds (64 knots or above)}
+#' \item{tropicalstorm_force_diameter}{Diameter (in nautical miles) of the area experiencing tropical storm strength winds (34 knots or above). Only available starting in 2004.}
+#' \item{hurricane_force_diameter}{Diameter (in nautical miles) of the area experiencing hurricane strength winds (64 knots or above). Only available starting in 2004.}
 #' }
 #' @examples
 #'
-#' # show a plot of the storm paths
+#' # show a plot of the storm paths in 1975 or later
```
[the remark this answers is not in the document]
steveharoz: The facets get too squished in the figure if too many years are included. Is there a way to make the figure bigger in the docs? https://dplyr.tidyverse.org/reference/storms.html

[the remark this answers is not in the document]
steveharoz: Added a little comment explaining that in the example

hadley:
Thanks for the update! I think the last question to resolve is whether it's worth while to include the rows prior to 1975 — I'm worried that this has a high likelihood of breaking existing graphics for little additional gain. I think it's probably safer to not include the historical data here.

steveharoz:
@hadley Yeah, I see the benefit of only having the clean and more complete data.

hadley:
Thanks! I did a couple more docs tweaks because I realised that this is the perfect place to use inline R code.

steveharoz:
Good call on the inline R!

PR #6327, opened 2022-07-11
Bring back legacy `arrange()` behavior with `dplyr.legacy_locale`

DavisVaughan:
Follow up to #6297 and #6263 

So the general idea is:
- In dplyr 1.1.0, the default behavior of both `group_by()` and `arrange()` sort in the C locale
- You can revert to legacy behavior, where both sort with the system locale, with the option `dplyr.legacy_locale = TRUE`
- You can override the `arrange()` locale with an explicit `.locale` argument, ~or with `dplyr.locale = "fr"`, but we prefer the former~
- You cannot override the C locale usage in `group_by()` (unless you revert to legacy behavior)

A few other notes about this PR:
- Removed `dplyr_locale()`, which is no longer useful
- Default `.locale` to `NULL` instead of `dplyr_locale()`
- If multiple locale options are set, hierarchy of precedence is `.locale`, ~`dplyr.locale`~, `dplyr.legacy_locale`

---

One question: Do we even need `dplyr.locale` anymore? Is being able to revert to the legacy behavior good enough? If we think `dplyr.locale` would stick around permanently, then it makes sense to keep it too, since I would like `dplyr.legacy_locale` to phase out. `dplyr.locale` might be useful as a permanent escape hatch for the case where you are calling something that calls `arrange()` for you and doesn't expose `.locale`. (Update: We removed `dplyr.locale`)


R/arrange.R
```diff
@@ -40,9 +40,9 @@
 #'   grouped data frames only.
 #' @param .locale The locale to sort character vectors in.
 #'
-#'   - Defaults to [dplyr_locale()], which uses the `"C"` locale unless this is
-#'     explicitly overriden. See the help page for [dplyr_locale()] for the
-#'     exact details.
+#'   - If `NULL`, the default, then the `"C"` locale will be used unless this is
```
hadley, line 43: Should we allow `"legacy"` here too?

hadley (commented):
I agree that we can drop `dplyr.locale` — I don't think we need some escape hatch to control arrange done elsewhere because in most cases, you can just re-arrange, and it's not likely to be that expensive (because you're probably rearranging for display, so there's not going to be that much data).

R/arrange.R
DavisVaughan: Since the only way to get legacy behavior in `group_by()` is to set the `dplyr.legacy_locale` global option, I think I'd like to keep that as the only way to get legacy behavior in `arrange()` too. I think it should be "all or nothing" legacy behavior that gets applied to both.

R/arrange.R
```diff
@@ -106,16 +106,6 @@ arrange_rows <- function(data,
...
 
-  chr_proxy_collate <- locale_to_chr_proxy_collate(
-    locale = locale,
-    error_call = error_call
-  )
-
-  if (length(dots) == 0L) {
-    out <- seq_len(nrow(data))
-    return(out)
-  }
```
DavisVaughan, line 117: Removed this because we no longer need an early exit when there aren't any `...` supplied. This was problematic before because of the `paste0()` recycling issue and some problems with `order()` with no inputs, but I've nicely worked around those now.

DavisVaughan:
Okay I have removed support for `dplyr.locale` and refreshed the docs and tests to reflect that.

Also updated my original comment in this PR to reflect the new state of how `arrange()` and `group_by()` work in 1.1.0

R/arrange.R
```diff
@@ -210,11 +214,60 @@ locale_to_chr_proxy_collate <- function(locale,
...
+dplyr_order_legacy <- function(data, direction) {
+  if (ncol(data) == 0L) {
+    # Work around `order(!!!list())` returning `NULL`
+    return(seq_len(nrow(data)))
+  }
+
+  proxies <- map2(data, direction, dplyr_proxy_order_legacy)
+  proxies <- unname(proxies)
+
+  inject(order(!!!proxies, decreasing = FALSE, na.last = TRUE))
```
hadley, line 237:
```suggestion
  inject(order(!!!proxies))
```

Seeing the explicit values makes me think you're overriding the defaults

NEWS.md
```diff
@@ -139,11 +139,28 @@
...
   follow up a `group_by()` with functions that use the group data, such as
-  `summarise()` or `group_split()` (#4406, #6297).
-  
-  See the Ordering section of `?group_by()` for more information. For a full
-  explanation of this change, refer to this
-  [tidyup](https://github.com/tidyverse/tidyups/blob/main/006-dplyr-group-by-ordering.md).
+  `summarise()` or `group_split()` (#4406, #6297). See the Ordering section of 
+  `?group_by()` for more information. 
+
+* `arrange()` now uses a faster algorithm for sorting character vectors, which
```
hadley, line 145: I tweaked the NEWS bullets a bit. Please give them another read through to make sure I correctly captured the changes.

NEWS.md
```diff
@@ -139,11 +139,28 @@
...
+  is heavily inspired by data.table's `forder()`. Additionally, the default
+  locale for sorting character vectors is now the C locale, which is a breaking
+  change from the previous behavior that utilized the system locale. The new
+  `.locale` argument can be used to adjust this to, say, the American English
+  locale, which is an optional feature that requires the stringi package. This
+  change improves reproducibility across R sessions and operating systems
+
+* `arrange()` and `group_by()` now both default to using the C locale when
+  ordering or grouping character vectors. This brings _substantial_ performance
```
hadley, line 154: We'll need to include some benchmarks in the blog post.

I wonder if we should say that this makes dplyr consistent with data.table?

R/arrange.R
```diff
@@ -40,9 +40,9 @@
 #'   grouped data frames only.
 #' @param .locale The locale to sort character vectors in.
 #'
-#'   - Defaults to [dplyr_locale()], which uses the `"C"` locale unless this is
-#'     explicitly overridden. See the help page for [dplyr_locale()] for the
-#'     exact details.
+#'   - If `NULL`, the default, then the `"C"` locale will be used unless this is
+#'     explicitly overridden. See the [locale][dplyr-locale] help page for the
+#'     exact details of how to override the default.
```
hadley, line 45:
```suggestion
#'   - If `NULL`, the default, uses the `"C"` locale unless the `dplyr.legacy_locale` escape
#'     hatch is activated. See below and [dplyr-locale] for more details.
```

R/arrange.R
```diff
@@ -176,8 +166,22 @@ arrange_rows <- function(data,
   })
 
   directions <- directions[quo_names %in% names(data)]
+
+  if (is.null(locale) && dplyr_legacy_locale()) {
```
hadley, line 170: I think you could simplify the logic by doing this earlier — if the legacy situation, we currently jump through a bunch of hoops to remove `desc()` calls before eventually adding them back in again. But maybe it's not worth it because of the transmute gymnastics?

hadley (approved):

R/arrange.R
DavisVaughan: I don't think we add them back in?

I think it has to happen here, both paths do the `transmute()` bits and both paths need to strip out `desc()`. 

Unless im missing something?

NEWS.md
DavisVaughan: I took another pass through the NEWS, simplifying the individual group-by and arrange bullets, since the combined bullet does a better job of talking about the C locale parts

I mentioned dplyr/data.table consistency in the 3rd bullet as part of a list of reasons why we made this change

R/arrange.R
DavisVaughan: Oh you mean at the end of `dplyr_proxy_order_legacy()`. CRAN dplyr does it this way too, so I'm inclined to keep it. I think figuring out the transmute stuff in combination with moving this code is going to be more challenging than its worth

hadley: Ok, sounds good.

PR #6328, opened 2022-07-12
Implement `case_match()` and `vec_case_match()`

DavisVaughan:
`case_match()` is a variant of `case_when()` that takes a primary input, `.x`, and then a series of formulas where the LHSs of each formula are values to match against `.x` rather than logical vectors. The LHSs get turned into logical conditions by `vec_in()`, and then the results are passed on to `vec_case_when()`.

It technically closes https://github.com/tidyverse/funs/issues/60

This would function as a direct successor to `recode()`, which is already questioning and has an awkward interface for anything except character vectors (and even there it can be odd).

``` r
char_vec <- sample(c("a", "b", "c"), 10, replace = TRUE)

recode(char_vec, a = "Apple", b = "Banana")
#>  [1] "Banana" "Banana" "c"      "Banana" "c"      "Banana"
#>  [7] "Apple"  "Banana" "c"      "Banana"

case_match(
  char_vec,
  "a" ~ "Apple",
  "b" ~ "Banana",
  .default = char_vec
)
#>  [1] "Banana" "Banana" "c"      "Banana" "c"      "Banana"
#>  [7] "Apple"  "Banana" "c"      "Banana"


recode(char_vec, a = "Apple", b = "Banana", .default = NA_character_)
#>  [1] "Banana" "Banana" NA       "Banana" NA       "Banana"
#>  [7] "Apple"  "Banana" NA       "Banana"

case_match(
  char_vec,
  "a" ~ "Apple",
  "b" ~ "Banana"
)
#>  [1] "Banana" "Banana" NA       "Banana" NA       "Banana"
#>  [7] "Apple"  "Banana" NA       "Banana"


# `case_match()` is more general and works elegantly 
# with more than just character
num_vec <- c(1:4, NA)

recode(num_vec, `1` = "o", `2` = "e", `3` = "o", `4` = "e", .missing = "m")
#> [1] "o" "e" "o" "e" "m"

case_match(
  num_vec,
  c(1, 3) ~ "o",
  c(2, 4) ~ "e",
  NA ~ "m"
)
#> [1] "o" "e" "o" "e" "m"


# More of a programmatic usage
level_key <- c(a = "apple", b = "banana", c = "carrot")
recode(char_vec, !!!level_key)
#>  [1] "banana" "banana" "carrot" "banana" "carrot" "banana"
#>  [7] "apple"  "banana" "carrot" "banana"

vec_case_match(
  needles = char_vec,
  haystacks = as.list(names(level_key)),
  values = as.list(level_key),
  default = char_vec
)
#>  [1] "banana" "banana" "carrot" "banana" "carrot" "banana"
#>  [7] "apple"  "banana" "carrot" "banana"
```

I still think a `replace_match()` would be useful here, like:

```r
# type stable replacement wrapper around case_match()
replace_match <- function(.x, ...) {
  ptype <- vec_ptype(.x)
  ptype <- vec_ptype_finalise(ptype)
  case_match(.x = .x, ..., .default = .x, .ptype = ptype)
}

# very close to compactness of recode()
replace_match(
  char_vec,
  "a" ~ "Apple",
  "b" ~ "Banana"
)

# instead of 
case_match(
  char_vec,
  "a" ~ "Apple",
  "b" ~ "Banana",
  .default = char_vec
)
```

`replace_match()` could also be used instead of a match-like version of `na_if()`

```r
x <- c("a", "NA", "NaN", "no")
replace_match(x, c("NA", "NaN", "no") ~ NA)
```

In forcats, we could have `fct_case_match()` as a successor to `recode_factor()`, but its interface would probably be the other way around, like:

```r
fct_case_match(
  .x,
  odd = c(1, 3),
  even = c(2, 4),
  ordered = FALSE
)

fct_case_when(
  odd = .x %in% c(1, 3),
  even = .x %in% c(2, 4),
  ordered = FALSE
)
```


DavisVaughan:
A better name for this might be `case_switch()`. i.e. it is a vectorized switch statement.

It just has the nice property of being able to collapse cases with the same right-hand sides into one line

```r
case_switch(
  num_vec,
  1 ~ "o",
  3 ~ "o",
  2 ~ "e",
  4 ~ "e",
  NA ~ "m"
)

case_switch(
  num_vec,
  c(1, 3) ~ "o",
  c(2, 4) ~ "e",
  NA ~ "m"
)
```

NEWS.md
```diff
@@ -1,5 +1,10 @@
 # dplyr (development version)
 
+* New `case_match()` function that is a "vectorised switch" variant of
```
hadley, line 3: You're staying with `case_match()` instead of `case_switch()`?

R/case-match.R
```diff
@@ -0,0 +1,131 @@
+#' A general vectorised switch
+#'
+#' This function is a variant of [case_when()] that vectorises the idea behind
```
hadley, line 3: Maybe take some of the intro text from `recode()`? I feel like this doesn't yet do a good job of describing what this function does.

R/case-match.R
```diff
@@ -0,0 +1,131 @@
...
+#'   If `NULL`, the default, a missing value will be used.
+#'
+#' @param .ptype An optional prototype declaring the desired output type. If
+#'   supplied, this overrides the common type of the RHS inputs.
+#'
+#' @return
+#' A vector with the same size as `.x` and the same type as the common type of
+#' the RHS inputs in `...`.
+#'
+#' @seealso [case_when()]
```
hadley, line 41: I wonder if it's worth including an example like this:

```R
case_match(letters[1:3], "a" ~ "x", "b" ~ "y", .ptype = factor(levels = c("y", "x")))
```

R/case-match.R
```diff
@@ -0,0 +1,131 @@
...
+#' case_match(
+#'   x,
+#'   "a" ~ 1,
+#'   "b" ~ 2,
+#'   "c" ~ 3,
+#'   "d" ~ 4,
+#'   NA ~ 0,
+#'   .default = 100
+#' )
+#'
```
hadley, line 68: I'd add an example in the opposite direction (integer -> character) as well

R/case-match.R
```diff
@@ -0,0 +1,131 @@
...
+#' # Input values can be grouped into the same expression to map them to the
+#' # same output value
+#' case_match(
+#'   x,
+#'   c("a", "b") ~ "low",
+#'   c("c", "d", "e") ~ "high"
+#' )
+#'
+#' # `.default` is allowed to be vectorized, and you can supply a `.ptype` to
```
hadley, line 77: I don't think this example is going to be useful most readers. Maybe just show the use of `.default = x` to replace selected values and leave everything else as is?

```R
case_match(x, "a" ~ "Apple", .default = x)
```

Also why do you need `.ptype` here? To avoid accidentally forcing integers to doubles?

R/case-match.R
```diff
@@ -0,0 +1,131 @@
...
+#' # force a particular output type. Combining these features together allows
+#' # you to create a type stable "replace match" helper.
+#' replace_match <- function(.x, ...) {
```
hadley, line 80: Are you sure we don't want to provide this function in dplyr?

R/case-match.R
```diff
@@ -0,0 +1,131 @@
...
+#' that matches using values rather than logical expressions. If no cases match,
+#' a missing value is returned unless a `.default` is supplied.
+#'
+#' @param .x A vector to match against.
+#'
+#' @param ... <[`dynamic-dots`][rlang::dyn-dots]> A sequence of two-sided
+#'   formulas. The left hand side (LHS) determines which values match this case.
+#'   The right hand side (RHS) provides the replacement value.
+#'
+#'   The LHS inputs must evaluate to vectors with the same type as `.x`. There
```
hadley, line 14: Do we need to say anything about this case `case_match(x, "a" ~ 1, "a" ~ 2)` ?

R/case-match.R
hadley: And _maybe_ an example with tibble output? 

```R
case_match(x, 
  "a" ~ tibble(x = 3), 
  "b" ~ data.frame(x = 1, y = 2)
)
```

R/case_when.R
```diff
@@ -200,30 +207,23 @@ case_when_formula_evaluate <- function(args,
 
   for (i in seq_len(n_args)) {
     pair <- quos_pairs[[i]]
-    conditions[[i]] <- eval_tidy(pair$lhs, env = default_env)
-    values[[i]] <- eval_tidy(pair$rhs, env = default_env)
+    lhs[[i]] <- eval_tidy(pair$lhs, env = default_env)
```
hadley, line 210: Should we use error wrapping here?

_pkgdown.yml
```diff
@@ -72,6 +72,7 @@ reference:
   contents:
   - between
   - case_when
+  - case_match
```
hadley, line 75:
```suggestion
  - case_match
  - case_when
```

Based on the alphabet 😛

NEWS.md
DavisVaughan: Yea I tried out `case_switch()` but had a few reasons for preferring `case_match()` in the end:

- I seem to use "match" very frequently in the docs and the test descriptions, making me think that is the better name

- The LHSs can handle multiple values, which isn't very switch like but is match like

- The vctrs implementation has the signature `vec_case_match(needles, haystacks, values)`, which makes a lot of sense. But `vec_case_switch()` just wouldn't align the same way because needles/haystacks don't make as much sense there.

- If we made a forcats version of this that produced a factor, we'd probably flip the order of the inputs and use named arguments where the name represents the output level and the values represent the values to map to that level, like `fct_case_match(x, lvl1 = c(val1, val2), lvl2 = val3)`. Calling this `fct_case_switch()` would be fairly misleading because you aren't switching off the names.

R/case-match.R
DavisVaughan: I'm hoping that forcats provides something like this instead:

```r
fct_case_match(
  letters[1:4], 
  x = "a",
  y = c("b", "d"),
  .default = NA,
  .ordered = TRUE
)
```

which would be elegant enough that I wouldn't want to mention a less optimal interface here

R/case-match.R
DavisVaughan: > Also why do you need .ptype here?

Yea, to be type stable on `x`

R/case-match.R
DavisVaughan: I haven't been able to decide if we should or not because I think it would imply we also need `replace_when()`

DavisVaughan: Having `replace_when()` and `replace_match()` would be nice alternatives to `replace()` though, especially it we don't do `mutate(.when = )` right now

R/case_when.R
DavisVaughan: Do you have an example of why it would be useful? Since this is just evaluating the LHS/RHS of the formula, I feel like the only error we could hit would be:

``` r
dplyr::case_when(foo ~ 1)
#> Error in eval_tidy(pair$lhs, env = default_env): object 'foo' not found
```

But I'm not sure we can improve on that anyways

NEWS.md
hadley: Ok, sounds good!

tests/testthat/test-case-match.R
```diff
@@ -0,0 +1,47 @@
+test_that("LHS can match multiple values", {
```
hadley, line 1: I added a few tests of basic properties here

hadley (approved):
Need check/better error message for `NULL` lhs and rhs:

```R
case_match(letters, "x" ~ NULL)
case_match(letters, NULL ~ "x")
```

Otherwise, looks great!

Do we want to superseded `recode()` in this PR or a separate one?

R/case_when.R
DavisVaughan: I've updated to throw a chained error like this

DavisVaughan:
> Do we want to superseded recode() in this PR or a separate one?

I'll leave that for another PR, I want to get this one in

PR #6329, opened 2022-07-12
Rewrite `na_if()` using vctrs

DavisVaughan:
We talked about making `na_if()` work in a match-like way, like `na_if(x, c("bad", "NA"))`, but I no longer think we should do that. While that might be more useful, I think:
- It would break existing code where people are currently supplying vector `y` values and expect `==` to be used, which does seem to come up on a quick GitHub scan
- It would no longer match `NULLIF()` from SQL, which it is advertised as mimicking. The Microsoft SQL docs state that `NULLIF(e1, e2)` should be treated as a "searched CASE expression", meaning that the following should be exactly equivalent

  ```
  NULLIF(e1, e2)

  CASE  
    WHEN e1 = e2 THEN NULL  
    ELSE e1  
  END 
  ```

  https://docs.microsoft.com/en-us/sql/t-sql/language-elements/nullif-transact-sql?view=sql-server-ver16#remarks

If people want a match-like `na_if()`, we could add `replace_match(x, c("bad", "NA") ~ NA)` as part of #6328 

---

Other notes on what changes from this update:
- Now casts `y` to the type of `x`, with the intention of making it very clear that this is type stable on `x`
- Now uses `vec_equal()` for comparison
- Now uses `vec_assign()` for assignment

`"9c7db457-e0e3-4d62-8d92-1a8af03dae11"`

### DIFF
--- a/R/na_if.R
+++ b/R/na_if.R
@@ -1,18 +1,23 @@
-#' Convert values to NA
+#' Convert values to `NA`
 #'
-#' This is a translation of the SQL command `NULLIF`. It is useful
-#' if you want to convert an annoying value to `NA`.
+#' This is a translation of the SQL command `NULLIF`. It is useful if you want
+#' to convert an annoying value to `NA`.
 #'
 #' @param x Vector to modify
-#' @param y Value to replace with NA
+#' @param y Value or vector to compare against. When `x` and `y` are equal, the
+#'   value in `x` will be replaced with `NA`.
+#'
+#'   `y` is cast to the type of `x` before comparison.
+#'
+#'   `y` is [recycled][vctrs::vector_recycling_rules] to the size of `x` before
+#'   comparison. This means that `y` can be a vector with the same size as `x`,
+#'   but most of the time this will be a single value.
 #' @return A modified version of `x` that replaces any values that
-#'   are equal to `y` with NA.
+#'   are equal to `y` with `NA`.
 #' @seealso [coalesce()] to replace missing values with a specified
 #'   value.
 #'
 #'   [tidyr::replace_na()] to replace `NA` with a value.
-#'
-#'   [recode()] to more generally replace values.
 #' @export
 #' @examples
 #' na_if(1:5, 5:1)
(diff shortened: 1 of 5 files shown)


DavisVaughan:
These changes broke 10 revdeps (see https://github.com/tidyverse/dplyr/issues/6262), but all of them are off-label usage of `na_if()`:
- 6 are related to doing something like `na_if(<tibble>, 0)` to replace all instances of 0 across any column of the tibble with `NA`. This accidentally worked because `<tibble> == 0` returns a logical matrix that `[<-` knows how to deal with. But I consider this off-label usage
- 2 are doing `dplyr::na_if(., "NaN")` or `dplyr::na_if(., "-Inf")`, i.e. some weird conversion from double to character to perform the comparison
- 4 are expecting a double `x` and character `y` (or vice versa) to be compatible, sometimes in very weird ways like: `dplyr::na_if(., "NaN")` or `dplyr::na_if(., "-Inf")`, where they seem to be trying to convert `.` to character before making the comparison

R/na_if.R
```diff
@@ -24,19 +29,25 @@
...
 #' starwars %>%
 #'    mutate(across(where(is.character), ~na_if(., "unknown")))
 na_if <- function(x, y) {
-  check_length(y, x, "`y`", glue("same as `x`"))
+  # Type and size stable on `x`
+  y <- vec_cast(x = y, to = x, x_arg = "y", to_arg = "x")
+  y <- vec_recycle(y, size = vec_size(x), x_arg = "y")
+
+  na <- vec_init(x)
+  where <- vec_equal(x, y)
```
hadley, line 48: Should this be `na_equal = TRUE`?

DavisVaughan: Interesting, I think the _only_ time that would matter is making `na_if(x, NaN)` replace `NaN` values in `x` with `NA_real_`, but that in itself might be useful enough to make that change (since `na_if(x, NA)` is essentially a no-op and doesn't make sense no matter what `na_equal` is)

DavisVaughan: I ended up switching to `na_equal = TRUE`, it feels useful with no downsides that I can think of

hadley (approved):

PR #6331, opened 2022-07-13
Rewrite `nth()`, `first()`, and `last()` using vctrs

DavisVaughan:
Revival of https://github.com/tidyverse/dplyr/pull/4636
Technically closes https://github.com/tidyverse/funs/issues/35

I am now quite confident that `first()` and friends should use `[` / `vec_slice()` rather than `[[` / `vec_get()`. I think the invariant for these functions should be that they return an object with:
- Size 1
- The same type as `x`

This guarantees they will be useful with `summarise()`, which is good because `?summarise` already mentions them as "Useful Functions"

This change mainly affects lists, which now return a list of size 1 rather than the actual element of the list. But I think that change is correct - i.e. it feels weird for "first" to return something that could be size >1. It also affects data frames, which previously returned a single column but now return a single row (which again, feels more consistent).

I'll check revdeps to ensure this doesn't cause a big uproar, but I think these are relatively uncommon edge cases 
`"252be158-3a4a-4e70-a233-4a9a0a858637"`


DavisVaughan:
This one might actually be too extreme. It seems like a lot of people use `first()` and `last()` in a programmatic way outside of data frames. In particular, a number of people use it on lists and expect it to extract the first/last element of that list.

14 broken revdeps:
- 7 due to `first()` or `last()` on a list, expecting to get list element
- 5 due to `first()` or `last()` on a data frame, expecting to get a column. Should use `pull()` instead.
- 1 due to stricter casting rules (can't cast character to double)
- 1 unknown

---

We might be able to make it use `vec_slice2()`-like behavior instead. i.e.:
- For most types (including lists), uses `[[`
- For data frames or matrices, selects a single row (and drops any row names)

It depends on how strongly we feel about the invariants of these functions (i.e. should they always return an object the same type as `x` with size 1?)

hadley:
Worth incorporating #6257?

DavisVaughan:
I thought about it but I think it'll be easier to me to keep it straight if I do that in a separate PR

This one is ready for review now though

R/nth-value.R
```diff
@@ -1,25 +1,37 @@
...
-#'   If a double is supplied, it will be silently truncated.
-#' @param order_by An optional vector used to determine the order
-#' @param default A default value to use if the position does not exist in
-#'   the input. This is guessed by default for base vectors, where a
-#'   missing value of the appropriate type is returned, and for lists, where
-#'   a `NULL` is return.
+#' @return
+#' If `x` is a list, a single element from that list.
+#'
+#' Otherwise, a vector the same type as `x` with size 1.
```
hadley, line 33:
```suggestion
#' If `x` is a list, a single element from that list.
#' Otherwise, a size-1 vector the same type as `x`.
```

R/nth-value.R
```diff
@@ -31,62 +43,113 @@
...
 #'
 #' last(x)
-#' # Second argument provides optional ordering
-#' last(x, y)
+#' # `order_by` provides optional ordering
+#' last(x, order_by = y)
 #'
-#' # These functions always return a single value
+#' # With empty atomic vectors, these functions still return a single value
 #' first(integer())
```
hadley, line 64: Move to below the OOB example?

hadley (approved):

PR #6339, opened 2022-07-19
Remove `tbl_sql` methods for `count()` and `tally()`

### LINKED ISSUE #6338: Remove method `count.tbl_sql()`
dplyr defines a method `count.tbl_sql()`. dbplyr defines a method `count.tbl_lazy()`. Now, the dplyr method wins due to the classes of database tables. I guess it would make sense to simply remove the method from dplyr (same issue for `tally.tbl_sql()`).

DavisVaughan:
Closes #6338 
Closes https://github.com/tidyverse/dbplyr/issues/940

See https://github.com/tidyverse/dplyr/issues/6338#issuecomment-1189423451 for the full diagnosis

@mgirlich could you please confirm that this is working as you expect? I think I've gotten it right, but you know more about dbplyr. You might want to add a test that is similar to the one here on the dbplyr side as well.

`"c5253d35-98ae-4a66-a3b7-78e93d321fd3"`

### DIFF
--- a/R/count-tally.R
+++ b/R/count-tally.R
@@ -76,13 +76,10 @@ count.data.frame <- function(x, ..., wt = NULL, sort = FALSE, name = NULL, .drop
   out <- tally(out, wt = !!enquo(wt), sort = sort, name = name)
 
   # Ensure grouping is transient
-  if (is.data.frame(x)) {
-    out <- dplyr_reconstruct(out, x)
-  }
+  out <- dplyr_reconstruct(out, x)
+
   out
 }
-
-count.tbl_sql <- count.data.frame
 
 #' @export
 #' @rdname count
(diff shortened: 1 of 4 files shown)


R/count-tally.R
```diff
@@ -76,14 +76,11 @@ count.data.frame <- function(x, ..., wt = NULL, sort = FALSE, name = NULL, .drop
   out <- tally(out, wt = !!enquo(wt), sort = sort, name = name)
 
   # Ensure grouping is transient
-  if (is.data.frame(x)) {
-    out <- dplyr_reconstruct(out, x)
-  }
+  out <- dplyr_reconstruct(out, x)
```
DavisVaughan, line 79: As mentioned here https://github.com/tidyverse/dplyr/issues/6338#issuecomment-1189423451, this `if (is.data.frame(x))` was only added in https://github.com/tidyverse/dplyr/pull/5114 to add support for database backends and can be removed now. We now unconditionally reconstruct to preserve the original grouping as much as possible.

mgirlich:
Thanks, this works as expected 👍

PR #6340, opened 2022-07-19
Retain only the last duplicate when a column is renamed multiple times in `relocate()`

### LINKED ISSUE #6209: `relocate.data.frame()` shouldn't rely on `vec_unique()` retaining names

DavisVaughan:
Closes #6301 
Closes #6209 

Alternative approach vs #6301. I actually don't think we should error here. I think we should take the _last_ name change when a column is renamed multiple times while it is being moved.

CRAN behavior takes the _first_ change, which I feel is wrong:

``` r
library(dplyr)

df <- tibble(a = 1)

relocate(df, x = a, y = a)
#> # A tibble: 1 × 1
#>       x
#>   <dbl>
#> 1     1
```

This PR takes the _last_ name change, which is more consistent with `rename()`:

``` r
library(dplyr)

df <- tibble(a = 1)

relocate(df, x = a, y = a)
#> # A tibble: 1 × 1
#>       y
#>   <dbl>
#> 1     1

rename(df, x = a, y = a)
#> # A tibble: 1 × 1
#>       y
#>   <dbl>
#> 1     1
```

I think this generally makes a little more sense too. We could error here, but I feel like this is a little more forgiving and better matches the spirit of dplyr while still being consistent with existing functions like `rename()`.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,9 @@
 # dplyr (development version)
 
+* `relocate()` now retains the last name change when a single column is renamed
+  multiple times while it is being moved. This better matches the behavior of
+  `rename()` (#6209, with help from @eutwt).
+
 * `na_if()` has been rewritten to utilize vctrs. This comes with the following
   improvements (#6329):
 
--- a/R/relocate.R
+++ b/R/relocate.R
@@ -13,7 +13,8 @@
 #' properties:
 #'
 #' * Rows are not affected.
-#' * The same columns appear in the output, but (usually) in a different place.
+#' * The same columns appear in the output, but (usually) in a different place
+#'   and possibly renamed.
 #' * Data frame attributes are preserved.
 #' * Groups are not affected.
 #' @section Methods:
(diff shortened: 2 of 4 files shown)


DavisVaughan:
@eutwt what do you think about this as an alternative to your PR? I think you are exactly right that only `to_move` would ever contain duplicates, so this tries to handle them up front as quickly as possible, so we don't have to worry about duplicates in the rest of the code.

hadley (approved):

PR #6342, opened 2022-07-20
Factor out `eval_relocate()` and use `dplyr_col_select()` in `relocate()`

### LINKED ISSUE #5829: relocate.data.frame: factor out or make default method
Looking at [relocate.data.frame](https://github.com/tidyverse/dplyr/blob/614389a391504ee17708aca2cce3d92b123ac449/R/relocate.R#L53-L90) in https://github.com/apache/arrow/pull/9798, it seems that the logic there is pretty generic and does not actually require a data.frame. As long as `.data` has a `dim()` method (for `ncol()` on L81), a `[<-` method (L84, to apply the reordering), and a `names()` method (for `tidyselect::eval_select()`), this function should work. 

In the short run, it seems to make sense for us to copy and adapt the `relocate.data.frame` method for Arrow objects rather than try to use the (unexported) dplyr data.frame method off-label. But for the future, we were wondering if there was an opportunity to factor this code differently to allow reuse by non-data.frame backends. Ideas that occurred to us:

* Factor out the core logic and perhaps move to `tidyselect`, as in `vars_relocate` alongside `vars_select` et al. Maybe something like `function(.data, ..., .before = NULL, .after = NULL)` that would return an integer vector of selected/reordered column indices.
* Just rename `relocate.data.frame` to `relocate.default` and then this should just work for any `.data` that has those three methods implemented.

If you think this is worthwhile (or see a better alternative), we're happy to submit a PR.

cc @ianmcook @romainfrancois

### LINKED ISSUE #6167: Select and relocate work inconsistenly on an empty dataframe
I expect that trying to `relocate` a non-existing column with `any_of` would return the original dataframe. Even if the dataframe is empty. The function `select` seems to work more consistently with `any_of`.

My dplyr version is 1.0.7.

``` r
library(tidyverse)
tibble(a=1:2) %>% relocate(any_of("b"))
#> # A tibble: 2 × 1
#>       a
#>   <int>
#> 1     1
#> 2     2
tibble() %>% relocate(any_of("b"))
#> Error: Can't subset columns that don't exist.
#> x Location 1 doesn't exist.
#> ℹ There are only 0 columns.
tibble() %>% select(any_of("b"))
#> # A tibble: 0 × 0
```

<sup>Created on 2022-01-28 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

### LINKED ISSUE #6341: `relocate()` drops attributes attached to bare data frames
Since this is mainly just using `x[i]` it should be using `dplyr_col_select()` internally, which patches the `[` method of base R data frames and data.table objects to retain attributes

``` r
library(dplyr)

df <- vctrs::data_frame(x = 1, y = 2)
attr(df, "foo") <- "bar"

out <- relocate(df, y, .before = x)

attr(out, "foo")
#> NULL
```

Since `mutate()` also uses `relocate()`, that should fix the same kind of issue that comes from `mutate()` + `.before`

``` r
library(dplyr)

df <- vctrs::data_frame(x = 1, y = 2)
attr(df, "foo") <- "bar"

out <- mutate(df, z = 3, .before = x)

attr(out, "foo")
#> NULL
```

DavisVaughan:
Closes #5829 (but we'd still need to move this to tidyselect)
Closes #6167 
Closes #6341 

This PR pulls `eval_relocate()` out of `relocate.data.frame()`. The eventual goal would be to move this to tidyselect for other backends to use it (#5829) but this is a good first step.

The `before` and `after` arguments of `eval_relocate()` take defused input, or `NULL`, which feels right given that `expr` works basically the same way.

`eval_relocate()` returns a named integer vector the same size as `ncol(data)` (important invariant) that describes how to rearrange the columns of `data`. The names are normally the same as the original data, but renaming can happen in `expr` so it is possible for them to be different.

I fixed a few edge case bugs in the implementation of `eval_relocate()`. Mainly around 0-col data frames and what happens when you supply `.before` or `.after` but they result in an empty selection.

I've also started using `dplyr_col_select()` in `relocate.data.frame()` rather than just `.data[pos]`. This solves #6341 and probably a few other consistency bugs since `dplyr_col_select()` has a few extra sanity checks in it.

### DIFF
--- a/R/relocate.R
+++ b/R/relocate.R
@@ -52,49 +52,103 @@ relocate <- function(.data, ..., .before = NULL, .after = NULL) {
 
 #' @export
 relocate.data.frame <- function(.data, ..., .before = NULL, .after = NULL) {
-  to_move <- tidyselect::eval_select(expr(c(...)), .data)
+  loc <- eval_relocate(
+    expr = expr(c(...)),
+    data = .data,
+    before = enquo(.before),
+    after = enquo(.after),
+    before_arg = ".before",
+    after_arg = ".after"
+  )
 
-  # Enforce the invariant that `ncol(.data) == ncol(relocate(.data, ...))` by
+  dplyr_col_select(.data, loc, names = names(loc))
+}
+
+eval_relocate <- function(expr,
+                          data,
+                          ...,
+                          before = NULL,
+                          after = NULL,
+                          before_arg = "before",
+                          after_arg = "after",
+                          env = caller_env(),
+                          error_call = caller_env()) {
+  # `eval_relocate()` returns a named integer vector of size `ncol(data)`
+  # describing how to rearrange `data`. Each location in the range
+  # `seq2(1L, ncol(data))` is represented once. The names are the new names to
+  # assign to those columns. They are typically the same as the original names,
+  # but `expr` does allow for renaming.
+
+  check_dots_empty0(...)
+
+  sel <- tidyselect::eval_select(
+    expr = expr,
+    data = data,
+    env = env,
+    error_call = error_call
+  )
+
+  # Enforce the invariant that relocating can't change the number of columns by
   # retaining only the last instance of a column that is renamed multiple times
-  # while it is being moved
   # TODO: https://github.com/r-lib/vctrs/issues/1442
-  # `to_move <- vec_unique(to_move, which = "last")`
-  loc_last <- which(!duplicated(to_move, fromLast = TRUE))
-  to_move <- vec_slice(to_move, loc_last)
+  # `sel <- vec_unique(sel, which = "last")`
+  loc_last <- which(!duplicated(sel, fromLast = TRUE))
+  sel <- vec_slice(sel, loc_last)
 
-  .before <- enquo(.before)
-  .after <- enquo(.after)
-  has_before <- !quo_is_null(.before)
-  has_after <- !quo_is_null(.after)
+  n <- length(data)
+
+  before <- as_quosure(before, env = env)
+  after <- as_quosure(after, env = env)
+
+  has_before <- !quo_is_null(before)
+  has_after <- !quo_is_null(after)
 
   if (has_before && has_after) {
-    abort("Must supply only one of `.before` and `.after`.")
-  } else if (has_before) {
-    where <- min(unname(tidyselect::eval_select(.before, .data)))
-    if (!where %in% to_move) {
-      to_move <- c(to_move, where)
+    message <- glue("Can't supply both `{before_arg}` and `{after_arg}`.")
+    abort(message, call = error_call)
+  }
+
+  if (has_before) {
+    # TODO: Use `allow_rename = FALSE`. https://github.com/r-lib/tidyselect/issues/225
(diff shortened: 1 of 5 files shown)


DavisVaughan:
@lionel- would you mind looking at this since it would probably end up in tidyselect (https://github.com/r-lib/tidyselect/issues/232)? I think it is ok if it is not 100% equivalent to what you'd implement for `eval_relocate()`, but I think this gets pretty close.

R/relocate.R
```diff
@@ -52,49 +52,103 @@ relocate <- function(.data, ..., .before = NULL, .after = NULL) {
...
+  loc_last <- which(!duplicated(sel, fromLast = TRUE))
+  sel <- vec_slice(sel, loc_last)
 
-  .before <- enquo(.before)
-  .after <- enquo(.after)
-  has_before <- !quo_is_null(.before)
-  has_after <- !quo_is_null(.after)
+  n <- length(data)
+
+  before <- as_quosure(before, env = env)
```
lionel-, line 100: I'm a bit on the fence about whether it'd be better to use `enquo()` for these arguments. But it seems better to be consistent and only take strict arguments in an `eval_` function, so I think this is a good solution.

R/relocate.R
```diff
@@ -52,49 +52,103 @@ relocate <- function(.data, ..., .before = NULL, .after = NULL) {
...
+  has_before <- !quo_is_null(before)
+  has_after <- !quo_is_null(after)
 
   if (has_before && has_after) {
-    abort("Must supply only one of `.before` and `.after`.")
-  } else if (has_before) {
-    where <- min(unname(tidyselect::eval_select(.before, .data)))
-    if (!where %in% to_move) {
-      to_move <- c(to_move, where)
+    message <- glue("Can't supply both `{before_arg}` and `{after_arg}`.")
```
lionel-, line 107: Use `cli_abort()` with `.arg` style classes?

R/relocate.R
```diff
@@ -52,49 +52,103 @@ relocate <- function(.data, ..., .before = NULL, .after = NULL) {
...
+    abort(message, call = error_call)
+  }
+
+  if (has_before) {
+    # TODO: Use `allow_rename = FALSE`. https://github.com/r-lib/tidyselect/issues/225
```
lionel-, line 112: This is fixed in the dev version which I plan to release right before the dplyr release.

lionel- (approved):

R/relocate.R
hadley: I don't think we use `cli_abort()` in dplyr yet?

R/relocate.R
DavisVaughan: Yea I was trying to be consistent with `expr` here, since `eval_select()` captures `expr` in the same way with this `as_quosure()` idea.

I started with `enquo()` but it felt really strange. This felt better to me

R/relocate.R
DavisVaughan: But this bug still exists :/ https://github.com/r-lib/tidyselect/issues/221

lionel-: I've added this to my list of issues to fix for the dplyr release.

PR #6343, opened 2022-07-21
Prevent mask methods from being mentioned in error calls

### LINKED ISSUE #6308: Errors in mask methods carry wrong calls

lionel-:
Closes #6308.

This was a bit tricky because the `call` for these top-level errors is created by R itself, before calling `.handleSimpleError()`. To fix these, I wrapped the evaluation routine in a function called `eval()` which rlang ignores by default.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,8 @@
 # dplyr (development version)
 
+* Fixed an issue with latest rlang that caused internal tools (such as
+  `mask$eval_all_summarise()`) to be mentioned in error messages (#6308).
+
 * `distinct()` now retains attributes of bare data frames (#6318).
 
 * dplyr no longer provides `count()` and `tally()` methods for `tbl_sql`.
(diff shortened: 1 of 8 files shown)


hadley (approved):

lionel-:
I wouldn't say it is hacky. I see it as an escape hatch that we can use for special cases like this.

PR #6346, opened 2022-07-21
Respect requested variable order in distinct()

### LINKED ISSUE #6156: distinct() doesn't respect user-specified column order
I'm again experiencing behavior that was reported as https://github.com/tidyverse/dplyr/issues/3195, then closed years ago.  I'll use the same reprex.  Running dplyr 1.0.7.

```
starwars %>% distinct(skin_color, hair_color)

#> # A tibble: 50 x 2
#>    hair_color    skin_color 
#>    <chr>         <chr>      
#>  1 blond         fair       
#>  2 <NA>          gold       
#>  3 <NA>          white, blue
#>  4 none          white      
#>  5 brown         light      
#>  6 brown, grey   light      
#>  7 <NA>          white, red 
#>  8 black         light      
#>  9 auburn, white fair       
#> 10 auburn, grey  fair       
#> # ... with 40 more rows
Created on 2022-01-14 by the reprex package (v2.0.1)
```

hadley:
Fixes #6156

### DIFF
--- a/R/distinct.R
+++ b/R/distinct.R
@@ -107,7 +107,8 @@ distinct_prepare <- function(.data,
   }
 
   # Always include grouping variables preserving input order
-  out_vars <- intersect(names(.data), c(distinct_vars, group_vars))
+  new_vars <- c(setdiff(group_vars, distinct_vars), distinct_vars)
+  out_vars <- intersect(new_vars, names(.data))
 
   if (.keep_all) {
     keep <- seq_along(.data)
--- a/R/distinct.R
+++ b/R/distinct.R
@@ -106,17 +106,18 @@ distinct_prepare <- function(.data,
     abort(bullets, call = error_call)
   }
 
+  # Only keep unique vars
+  distinct_vars <- unique(distinct_vars)
   # Missing grouping variables are added to the front
   new_vars <- c(setdiff(group_vars, distinct_vars), distinct_vars)
-  out_vars <- intersect(new_vars, names(.data))
 
   if (.keep_all) {
     keep <- seq_along(.data)
   } else {
-    keep <- out_vars
+    keep <- new_vars
   }
 
-  list(data = .data, vars = out_vars, keep = keep)
+  list(data = .data, vars = new_vars, keep = keep)
 }
 
 #' @export
--- a/R/distinct.R
+++ b/R/distinct.R
@@ -106,7 +106,7 @@ distinct_prepare <- function(.data,
     abort(bullets, call = error_call)
   }
 
-  # Always include grouping variables preserving input order
+  # Missing grouping variables are added to the front
   new_vars <- c(setdiff(group_vars, distinct_vars), distinct_vars)
   out_vars <- intersect(new_vars, names(.data))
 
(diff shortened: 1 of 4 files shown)


R/distinct.R
```diff
@@ -107,7 +107,8 @@ distinct_prepare <- function(.data,
   }
 
   # Always include grouping variables preserving input order
```
DavisVaughan, line 109:
```suggestion
  # Missing grouping variables are added to the front
```

This seems like a clearer way to talk about what this is doing, right?

I thought about suggesting you do:

```r
new_vars <- unique(c(group_vars, distinct_vars))
```

but after further review I figured out that wasn't right because you want grouping variables that also exist in `distinct_vars` to respect the order they are specified in `distinct_vars`. I think an updated comment would have gotten me there faster.

R/distinct.R
```diff
@@ -107,7 +107,8 @@ distinct_prepare <- function(.data,
...
-  out_vars <- intersect(names(.data), c(distinct_vars, group_vars))
+  new_vars <- c(setdiff(group_vars, distinct_vars), distinct_vars)
+  out_vars <- intersect(new_vars, names(.data))
```
DavisVaughan, line 111: I was very curious about what this `intersect()` does. I removed it and this test broke:

```r
test_that("distinct doesn't duplicate columns", {
  df <- tibble(a = 1:3, b = 4:6)

  expect_named(df %>% distinct(a, a), "a")
  expect_named(df %>% group_by(a) %>% distinct(a), "a")
})
```

which I think implies that we were using `intersect()` for its "feature" of removing duplicates

``` r
intersect(c(1, 1), 1)
#> [1] 1
```

So probably `unique(new_vars)` is enough and more correct there?

If you think about it, both `distinct_vars` and `group_vars` should always contain cols in `.data`. We explicitly check that this is the case for `distinct_vars` (thats the `missing_vars` bit). Maybe we should change to `unique(new_vars)` and then add an explicit check that `group_vars` are in `names(.data)` (since it is user input) and error if that isn't true.

tests/testthat/test-distinct.R
```diff
@@ -102,9 +102,15 @@ test_that("distinct handles 0 columns edge case (#2954)", {
   expect_equal(nrow(distinct(tibble())), 0L)
 })
 
-test_that("distinct preserves order of the input variables (#3195)",{
+test_that("distinct respects order of the specified variables (#3195)",{
```
DavisVaughan, line 105:
```suggestion
test_that("distinct respects order of the specified variables (#3195, #6156)",{
```

Maybe worth referencing both issues

DavisVaughan (approved):

R/distinct.R
hadley: I don't think we need to check `group_vars()` because it's always computed from the input data. But I agree that `unique()` is clearer than `intersect()`

DavisVaughan: I only mentioned that we check it because `distinct_prepare()` is exported (for whatever reason) and `group_vars` is an argument, so who knows what someone might pass in. Internally I don't think we will ever have an issue.

It is also fine to not check it and just assume that only advanced developers will ever use this, and they should know what they are doing

PR #6349, opened 2022-07-21
Deprecate `all_equal()`

### LINKED ISSUE #6324: Deprecate `all_equal()`
It's been questioning for awhile, and there doesn't seem to be much usage (https://cs.github.com/?scopeName=All+repos&scope=&q=all_equal+language%3AR)

hadley:
Fixes #6324

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,8 @@
 # dplyr (development version)
 
+* `all_equal()` is formally deprecated. We've advise against it for
+  some time, and we explicitly recommend you use something else (#6324).
+
 * `slice()` helpers again produce output equivalent to `slice(.data, 0)` when
   the `n` or `prop` argument is 0, fixing a bug introduced in the previous
   version (@eutwt, #6184).
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,6 +1,6 @@
 # dplyr (development version)
 
-* `all_equal()` is formally deprecated. We've advise against it for
+* `all_equal()` is formally deprecated. We've advised against it for
   some time, and we explicitly recommend you use something else (#6324).
 
 * `slice()` helpers again produce output equivalent to `slice(.data, 0)` when
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,7 +1,8 @@
 # dplyr (development version)
 
 * `all_equal()` is formally deprecated. We've advised against it for
-  some time, and we explicitly recommend you use something else (#6324).
+  some time, and we explicitly recommend you use `all.equal()`,
+  manually reordering the rows and columns (#6324).
 
 * `slice()` helpers again produce output equivalent to `slice(.data, 0)` when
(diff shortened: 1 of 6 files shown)


NEWS.md
```diff
@@ -1,5 +1,8 @@
 # dplyr (development version)
 
+* `all_equal()` is formally deprecated. We've advise against it for
+  some time, and we explicitly recommend you use something else (#6324).
```
DavisVaughan, line 4:
```suggestion
* `all_equal()` is formally deprecated. We've advised against it for
  some time, and we explicitly recommend you use something else (#6324).
```

DavisVaughan: Do you want to mention why we don't like it here? (Like you do in the help file?)

R/all-equal.r
```diff
@@ -20,20 +22,23 @@
 #' scramble <- function(x) x[sample(nrow(x)), sample(ncol(x))]
 #'
 #' # By default, ordering of rows and columns ignored
```
DavisVaughan, line 24:
```suggestion
#' # `all_equal()` ignored row and column ordering by default,
#' # but we now feel that that makes it too easy to make mistakes
```

R/all-equal.r
```diff
@@ -20,20 +22,23 @@
...
-#' # By default all_equal is sensitive to variable differences
-#' df1 <- data.frame(x = "a", stringsAsFactors = FALSE)
-#' df2 <- data.frame(x = factor("a"))
-#' all_equal(df1, df2)
-#' # But you can request dplyr convert similar types
-#' all_equal(df1, df2, convert = TRUE)
+#' # Instead, be explicit about the row and column ordering
+#' all.equal(
+#'   mtcars,
+#'   mtcars[rownames(mtcars), names(mtcars)]
```
DavisVaughan, line 31:
```suggestion
#'   mtcars2[rownames(mtcars), names(mtcars)]
```

?

tests/testthat/test-all-equal.r
```diff
@@ -1,3 +1,7 @@
+test_that("all_equal is deprecated", {
+  expect_snapshot(all_equal(mtcars, mtcars))
```
DavisVaughan, line 2:
```suggestion
  local_options(lifecycle_verbosity = "warning")
  expect_snapshot(all_equal(mtcars, mtcars))
```

Just in case, for whatever reason, `all_equal()` gets called somewhere else in the global session

DavisVaughan (approved):

tests/testthat/test-all-equal.r
hadley: `local_reproducible_output()` already does that automatically.

R/all-equal.r
hadley: Ooops

tests/testthat/test-all-equal.r
DavisVaughan: Neat! Didn't know that

PR #6350, opened 2022-07-21
Mark mutate() args + row_* as no longer experimental

### LINKED ISSUE #6110: Review experimental badges
For dplyr 1.1.0

hadley:
I think these are unquestionably useful. Other experimental functions that I left as is are the `dplyr_*` generics, `group_map()`/`group_walk()`/`group_nest()`/`group_split()`/`group_trim()`, `with_groups()`, and `nest_by()`. I don't see any reason that these would go away, but it feels like we still want to reserve the right to play around with them still.

Fixes #6110

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,10 @@
 # dplyr (development version)
 
+* The `.keep`, `.before`, and `.after` arguments to `mutate()` 
+  are no longer experimental.
+  
+* The `rows_` family of functions are no longer experimental.
+
 * `slice()` helpers again produce output equivalent to `slice(.data, 0)` when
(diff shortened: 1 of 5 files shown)


NEWS.md
```diff
@@ -1,5 +1,10 @@
 # dplyr (development version)
 
+* The `.keep`, `.before`, and `.after` arguments to `mutate()` 
+  are no longer experimental.
+  
+* The `rows_` family of functions are no longer experimental.
```
DavisVaughan, line 6:
```suggestion
* The `rows_*()` family of functions are no longer experimental.
```

?

DavisVaughan (approved):
Looks great. Agree about keeping everything else experimental

PR #6351, opened 2022-07-21
Silently cap `slice_sample()` slice size at `n()` when `replace = FALSE`

### LINKED ISSUE #6185: `slice_sample()` errors if `n` bigger than number of rows
In the man page for `slice` functions, the description for the argument `n` states:

> If n is greater than the number of rows in the group (or prop > 1), the result will be silently truncated to the group size. If the proportion of a group size does not yield an integer number of rows, the absolute value of prop*nrow(.data) is rounded down.

The output of `slice_sample` used to be the same `data.frame` (with different ordering) if `n` is higher than the number of rows, but it is now returning an error.

``` r
library(dplyr)
#> 
#> Attaching package: 'dplyr'
#> 
#>     filter, lag
#> 

tmp <- data.frame(x = c(rep('a',6), rep('b',4)), y = 1:10)
print(tmp)
#>    x  y
#> 1  a  1
#> 2  a  2
#> 3  a  3
#> 4  a  4
#> 5  a  5
#> 6  a  6
#> 7  b  7
#> 8  b  8
#> 9  b  9
#> 10 b 10

tmp %>% slice_sample(n = 15)
#> Error in `slice_sample()`:
#> ! Problem while computing indices.
#> Caused by error in `sample.int()`:
#> ! cannot take a sample larger than the population when 'replace = FALSE'

tmp %>% group_by(x) %>% slice_sample(n = 15)
#> Error in `slice_sample()`:
#> ! Problem while computing indices.
#> ℹ The error occurred in group 1: x = "a".
#> Caused by error in `sample.int()`:
#> ! cannot take a sample larger than the population when 'replace = FALSE'
```

<sup>Created on 2022-02-11 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

eutwt:
closes #6185 

This PR silently caps slice size at `n()` for `slice_sample(..., replace = FALSE)` e.g. 
``` r
data.frame(a = 1) %>% 
  slice_sample(n = 2)
#>   a
#> 1 1
```
The timeline on this, as I understand it:

- In #5962, when adding support for negative `n` and `prop` in slice functions, I moved calculation of "slice size" to one function for all slice helpers. This function (silently) bounded all sizes between `0` and `n()`. I didn't realize, though, that  `slice_sample(..., replace = TRUE)`, allowed users to produce groups with more than `n()` rows, and so in doing this I changed that behavior.
- In revdeps before releasing 1.0.8 it [was found](https://github.com/Nowosad/motif/issues/66) that this change broke a package, and so #6172 reverted to the old behavior for `slice_sample(..., replace = TRUE)` and size > `n()`. The PR also made two other changes
  - For `slice_sample(..., replace = FALSE)`,  supplying a slice size is larger than `n()` is not allowed (causes error)
  - For `slice_sample()` in general, supplying negative `n` or `prop` is not allowed (causes error)
- Issue #6185 was opened requesting that the first of the two "additional" changes above be reverted

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,10 @@
 # dplyr (development version)
 
+* `slice_sample()` returns a data frame or group with the same number of rows as 
+  the input when `replace = FALSE` and `n` is larger than the number of rows or 
+  `prop` is larger than 1. This reverts a change made in 1.0.8, returning to the 
+  behavior of 1.0.7 (#6185)
+
 * `slice()` helpers again produce output equivalent to `slice(.data, 0)` when
   the `n` or `prop` argument is 0, fixing a bug introduced in the previous
   version (@eutwt, #6184).
--- a/R/slice.R
+++ b/R/slice.R
@@ -452,17 +452,7 @@ get_slice_size <- function(n, prop, allow_negative = TRUE, error_call = caller_e
 
 sample_int <- function(n, size, replace = FALSE, wt = NULL, call = caller_env()) {
   if (!replace && n < size) {
-    header <- paste0(
-      "Can't sample without replacement using a size that is larger than ",
-      "the number of rows in the data."
-    )
-    message <- c(
-      header,
-      i = glue("{size} rows were requested in the sample."),
-      i = glue("{n} rows are present in the data."),
-      i = "Set `replace = TRUE` to sample with replacement."
-    )
-    abort(message, call = call)
+    size <- n
   }
 
   if (size == 0L) {
--- a/tests/testthat/_snaps/slice.md
+++ b/tests/testthat/_snaps/slice.md
@@ -12,52 +12,6 @@
       <error/rlang_error>
       Error in `slice_sample()`:
       ! `prop` must be positive.
-    Code
-      (expect_error(df %>% slice_sample(n = 4, replace = FALSE)))
-    Output
-      <error/rlang_error>
-      Error in `slice_sample()`:
-      ! Problem while computing indices.
-      Caused by error:
-      ! Can't sample without replacement using a size that is larger than the number of rows in the data.
-      i 4 rows were requested in the sample.
-      i 1 rows are present in the data.
-      i Set `replace = TRUE` to sample with replacement.
-    Code
-      (expect_error(gdf %>% slice_sample(n = 4, replace = FALSE)))
-    Output
-      <error/rlang_error>
-      Error in `slice_sample()`:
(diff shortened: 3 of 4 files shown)


hadley:
Well that was simpler than I expected 😆 Do you mind briefly also summarising the course of events in a news bullet?

NEWS.md
```diff
@@ -1,5 +1,10 @@
 # dplyr (development version)
 
+* `slice_sample()` returns a data frame or group with the same number of rows as 
+  the input when `replace = FALSE` and `n` is larger than the number of rows or 
+  `prop` is larger than 1. This reverts a change made in 1.0.8, returning to the 
+  behavior of 1.0.7 (#6185)
+
```
DavisVaughan, line 7: Mainly a question for @Hadley:

So, this goes directly against how `sample()` works (and I feel that the `sample()` behavior is reasonable) but we are ok with that?

``` r
sample(1:5, size = 6, replace = FALSE)
#> Error in sample.int(length(x), size, replace, prob): cannot take a sample larger than the population when 'replace = FALSE'
```

DavisVaughan: I personally think it is fine for `slice_sample()` to be the "odd one out" here because `sample()` is also different from `head()` and `tail()` 

``` r
head(1:5, n = 6)
#> [1] 1 2 3 4 5
tail(1:5, n = 6)
#> [1] 1 2 3 4 5
```

DavisVaughan: I guess this can come in handy with grouped data, as described here:
https://github.com/tidyverse/dplyr/issues/6185#issuecomment-1191704069

It is really a question of how strict we want to be. Should `slice_sample(df, n = 10)` have an invariant that it returns a data frame of size `10` (or `10 * n_groups` in the grouped case)? I guess since it is mainly for interactive use, we don't have to be quite as strict as `sample()` (which does have this invariant).

I could go either way on this one

eutwt: I think it's reasonable for `sample(x, bign)` to error because the user can just write `sample(x, min(bign, length(x)))` instead. But `slice_sample(n = min(bign, n()))` isn't possible. On the other hand, you can always just use `slice(slice(sample(n(), min(bign, n()))))`

hadley: Yeah, think of `slice_head(df, n = 5)` — it returns `min(group_size, 5)` for each group. And that's the behaviour we want for `slice_sample()` too.

PR #6352, opened 2022-07-21
Collection of documentation tweaks

### LINKED ISSUE #5943: Doubt about `percent_rank()`
As I see the code of the function: 
https://github.com/tidyverse/dplyr/blob/04454209ea069939d3335c43846c85c725547a89/R/rank.R#L104-L106
How the percentile rank is calculated is only the percentage of values that are less than the value under analysis, and not the percentage of values that are less than or equal to the value of interest, which is the definition of the percentile rank.

``` r
x <- c(1, 1, 2, 2, 5, 100)

# Current work
dplyr::percent_rank(x)
#> [1] 0.0 0.0 0.4 0.4 0.8 1.0

# Expected work
rank(x, ties.method = "max")/sum(!is.na(x))
#> [1] 0.3333333 0.3333333 0.6666667 0.6666667 0.8333333 1.0000000
```

<sup>Created on 2021-07-01 by the [reprex package](https://reprex.tidyverse.org) (v2.0.0)</sup>

### LINKED ISSUE #5958: Inaccurate documentation for `name` argument of `count()`
The documentation for `count()` says of the argument `name`:

> If omitted, it will default to n. If there's already a column called n, it will error, and require you to specify the name.

But when I try calling `count` on a variable named `n`, that's not what happens - it succeeds with a warning:

```
mtcars %>% group_by(cyl, am) %>% summarise(n = n(), .groups = "drop") %>% count(cyl, n)
Storing counts in `nn`, as `n` already present in input
i Use `name = "new_name"` to pick a new name.
# A tibble: 6 x 3
    cyl     n    nn
  <dbl> <int> <int>
1     4     3     1
2     4     8     1
3     6     3     1
4     6     4     1
5     8     2     1
6     8    12     1
```

This is trivial, I think, and I like the current behavior of `count()` warning instead of erroring - I'm just noting the documentation mismatch.

### LINKED ISSUE #6017: Break up verb label into smaller categories
I'd suggest "rows", "columns" and "groups". And joins could become "tables"

And we should reorganise the reference index along the same lines.

See WIP 2e of R4DS for examples: https://r4ds.hadley.nz/data-transform.html#dplyr-basics

### LINKED ISSUE #6205: Clearer across() documentation for multiple .fns
I've found the documentation for across() about the multiple function case more oblique than perhaps it could be, but I'm not familiar enough with all of its applications to be sure I'll get it right if I directly submit a change via pull. This applies to both the Roxygen comment on R/across.R and to a lesser degree the colwise vignette, vignettes/colwise.Rmd

The current text in the Roxygen comment is
>Functions to apply to each of the selected columns. Possible values are:
>    - NULL, to returns the columns untransformed.
>    - A function, e.g. mean.
>    - A purrr-style lambda, e.g. ~ mean(.x, na.rm = TRUE)
>    - A list of functions/lambdas, e.g. list(mean = mean, n_miss = ~ sum(is.na(.x))

and in the vignette,

> The second argument, .fns, is a function or list of functions to apply to each column. This can also be a purrr style formula (or list of formulas) like ~ .x / 2. (This argument is optional, and you can omit it if you just want to get the underlying data; you’ll see that technique used in vignette("rowwise").)
> ...
> ### Multiple functions
>
> You can transform each variable with more than one function by supplying a named list of functions or lambda functions in the second argument:

The intuition that I'm only seeing referred to indirectly is that the provided functions *fan out* into multiple outputs, i.e. if you provided n columns and m functions you'd end up with n x m outputs. The Roxygen intro also refers to transformation in the singular, "across() makes it easy to apply the same transformation to multiple columns...", so it feels like the multiple-output case is de-emphasised. There are other possible interpretations for "Functions to apply to each of the selected columns", e.g. maybe the functions are applied in order, to a single output. (This is of course very achievable through other methods like `purrr::compose` or `. %>% ...` pipes or purrr lambdas, and not necessary for `across` to do, but it's not ruled out by the description.)

It's also not made entirely clear that an _unnamed_ list of functions `c(...)` is probably unusual, i.e. the reason a _named_ list is used is because the name determines the output name (depending on `.names`) - someone might use an unnamed list in an ad-hoc to test a range of transformations, but named lists match the intention behind the design of the function better. The vignette is clearer about this, and while Roxygen uses a named list as an example but leaves the choice unaddressed, i.e. it could just be a code style cue to make a lambda more readable. (I've primarily used across with `mutate`, so the rationale in the `summarise`-style example wasn't immediately obvious.)

My suggestion would be to emphasise that a separate output is produced for each function, e.g. a maybe clumsy alternative wording would be
> Functions to apply to each of the selected columns, where each transformation produces a separate output column based on the column and list names.

Again, I don't know for sure if this conflicts with any current or planned uses for the function. As far as I can tell across more or less enforces this as a constraint, i.e. requires unique names for each output (although of course transform functions can return NULL in order to _not_ produce a new output column. _Optionally_ producing output for a list of functions opens up some interesting variations I guess that vary a little from the n x m account.)

Mostly my sense was that the intention or purpose behind the argument was underspecified in the documentation, making it harder to understand the behaviour unless you already know the behaviour. Partly this could be a consequence of across being a fairly flexible tool so its 'meaning' can look a little different in different uses, but I think we can nudge the docs towards giving a quicker birds-eye view. Hopefully you can see where I'm coming from.

### LINKED ISSUE #6233: Improve ranking function documentation
* Don't just define in terms of `rank()`.
* Explain the consequences of using `min_rank()` vs `dense_rank()` vs `row_number()` for filtering (i.e. what's the max and min number of rows you might get)

### LINKED ISSUE #6326: Base R functions misinterpreted as methods in documentation
Hello, 
In the documentation of generic functions, there is an automatic detection of the methods available for other classes defined in the loaded packages. This is great but :
- the `select()` documentation mentions `list` as a class for which a `select` method exists, and points to `utils::select.list()`.
- the `slice()` documentation mentions `index` as a class for which a `slice` method exists, and points to `base::slice.index()`.

Don't you think they should be removed, by making an exception in the `methods_rd()` function for instance? 

Thanks in advance

Félicien

hadley:
Fixes #6017. Fixes #5958. Fixes #6233. Fixes #5943. Fixes #6326. Fixes #6205.


R/rank.R
```diff
@@ -1,64 +1,91 @@
...
+#'   `rank(ties.method = "first")`.
 #'
-#' * `min_rank()`: equivalent to `rank(ties.method = "min")`
+#' * `min_rank()` gives every tie the same (smallest) value so that
+#'   `c(10, 20, 20, 30)` gets ranks `c(1, 2, 2, 4)`. It's the way ranking usually
+#'   works in spots and is equivalent to `rank(ties.method = "min")`.
 #'
-#' * `dense_rank()`: like `min_rank()`, but with no gaps between
-#'   ranks
+#' * `dense_rank()` works like `min_rank()`, but doesn't leave any no gaps,
```
eutwt, line 15: Extra word ("no") here

hadley: Thanks!

R/rank.R
```diff
@@ -1,64 +1,91 @@
...
+}
+
+#' Bucket a numeric vector into `n` groups
+#'
+#' `ntile()` is a sort of very rough rank, which breaks the input vector into
+#' `n` buckets. If `length(x)` is not an integer multiple of `n`, the size of
+#' the buckets will differ by up to one, with larger buckets coming first.
+#'
+#' @inheritParams row_number
 #' @param n number of groups to split up into.
```
DavisVaughan, line 79:
```suggestion
#' @param n Number of groups to bucket into
```

R/rank.R
```diff
@@ -1,64 +1,91 @@
...
 #' @export
-#' @rdname ranking
+#' @family ranking functions
+#' @examples
+#' x <- c(5, 1, 3, 2, 2, NA)
+#' ntile(x, 2)
+#' ntile(x, 4)
+#'
+#' # If the bucket sizes are uneven, the larger buckets n come first
```
DavisVaughan, line 87:
```suggestion
#' # If the bucket sizes are uneven, the larger buckets come first
```

R/rank.R
```diff
@@ -89,24 +118,39 @@ ntile <- function(x = row_number(), n) {
...
+#'
+#' @description
+#' These two ranking functions implement two slightly different ways to
+#' compute a percentile. For each `x_i` in `x`:
+#'
+#' * `cume_dist(x)` counts the total number of values less than
+#'   or equal to `x_i`, and divides it by the number of observations.
+#'
+#' * `percent_rank(x)` counts the total number of values less than
+#'   `x_i`, and divides it by the number of observations minus 1.
```
DavisVaughan, line 131: Nice, very clear definitions

R/rank.R
```diff
@@ -89,24 +118,39 @@ ntile <- function(x = row_number(), n) {
...
 }
 
+#' Proportional ranking functions
+#'
+#' @description
+#' These two ranking functions implement two slightly different ways to
+#' compute a percentile. For each `x_i` in `x`:
+#'
+#' * `cume_dist(x)` counts the total number of values less than
+#'   or equal to `x_i`, and divides it by the number of observations.
```
DavisVaughan, line 128:
```suggestion
#'   or equal to `x_i`, and divides it by the number of non-missing observations.
```

Maybe something about the fact that "number of observations" doesn't include missing values here and in `percent_rank()`'s definition? Or maybe this is a separate note that goes under these bullet points?

R/rank.R
```diff
@@ -89,24 +118,39 @@ ntile <- function(x = row_number(), n) {
...
-
-#' @export
-#' @rdname ranking
+#' @examples
+#' x <- c(5, 1, 3, 2, 2)
+#'
+#' cume_dist(x)
+#' percent_rank(x)
+#'
+#' # you can understand what's going on by computing it by hand
```
DavisVaughan, line 143:
```suggestion
#' # You can understand what's going on by computing it by hand
```

R/rank.R
```diff
@@ -89,24 +118,39 @@ ntile <- function(x = row_number(), n) {
...
+#' sapply(x, function(xi) sum(x <= xi))
+#' sapply(x, function(xi) sum(x < xi)  / (length(x) - 1))
+#' # the real computations are a little more complex in order to
```
DavisVaughan, line 146:
```suggestion
#' # The real computations are a little more complex in order to
```

R/rank.R
```diff
@@ -89,24 +118,39 @@ ntile <- function(x = row_number(), n) {
...
-#' @export
-#' @rdname ranking
+#' @examples
+#' x <- c(5, 1, 3, 2, 2)
+#'
+#' cume_dist(x)
+#' percent_rank(x)
+#'
+#' # you can understand what's going on by computing it by hand
+#' sapply(x, function(xi) sum(x <= xi))
```
DavisVaughan, line 144:
```suggestion
#' sapply(x, function(xi) sum(x <= xi) / length(x))
```

DavisVaughan: `sapply()`, the horror!

R/select.R
```diff
@@ -1,4 +1,4 @@
-#' Subset columns using their names and types
+#' Keep or drop columns using their names and types
 #'
 #' @description
 #'
```
DavisVaughan, line 4: I think this part of the description section below:

> You can also use predicate functions like is.numeric to select variables based on their properties.

Should mention that you have to combine that with `where()` to use this feature. This was probably written when we allowed `select(data, is.numeric)`, but we no longer allow that

I think I've seen a question about this before

R/rank.R
```diff
@@ -1,64 +1,91 @@
...
-#' @rdname ranking
+#' @family ranking functions
+#' @examples
+#' x <- c(5, 1, 3, 2, 2, NA)
+#' ntile(x, 2)
+#' ntile(x, 4)
+#'
+#' # If the bucket sizes are uneven, the larger buckets n come first
+#' ntile(1:8, 3)
 ntile <- function(x = row_number(), n) {
```
DavisVaughan, line 89: Note to document that `ntile()` can put the same value into different buckets (here the two 10s). i.e. it priorities keeping the size of the buckets the same over keeping repeats of the same value together.

```r
library(dplyr)
​
tibble(
  x = c(1, 5, 1, 10, 3, 4, rep(10, 7))
) %>%
  mutate(r = ntile(x, n = 2)) %>%
  print(n = Inf)
#> # A tibble: 13 × 2
#>        x     r
#>    <dbl> <int>
#>  1     1     1
#>  2     5     1
#>  3     1     1
#>  4    10     1
#>  5     3     1
#>  6     4     1
#>  7    10     1
#>  8    10     2
#>  9    10     2
#> 10    10     2
#> 11    10     2
#> 12    10     2
#> 13    10     2
```

R/sets.r
```diff
@@ -1,11 +1,16 @@
 #' Set operations
 #'
-#' These functions override the set functions provided in base to make them
-#' generic so that efficient versions for data frames and other tables can be
-#' provided. The default methods call the base versions. Beware that
-#' `intersect()`, `union()` and `setdiff()` remove duplicates.
+#' @description
+#' Perform set operations using the rows of a data frame or tibble.
```
DavisVaughan, line 4:
```suggestion
#' Perform set operations using the rows of a data frame.
```

Feels like that goes without saying

R/sets.r
```diff
@@ -1,11 +1,16 @@
...
+#' `intersect()`, `union()`, `setdiff()`, and `setequal()` override the base
+#' functions of the same name in order to make them generic. The existing
+#' behaviour for vectors is preserved by providing default methods that call
+#' the base functions.
 #'
-#' @param x,y objects to perform set function on (ignoring order)
+#' Note that `intersect()`, `union()` and `setdiff()` remove duplicates
+#' in `x` and `y`.
+#'
```
DavisVaughan, line 12: Since `union_all()` is new, I'd like to see a description of what it is here, and how it is different from `union()`

R/sets.r
```diff
@@ -21,21 +26,27 @@
 #' union_all(first, second)
 #' setequal(mtcars, mtcars[32:1, ])
 #'
-#' # Handling of duplicates:
-#' a <- data.frame(column = c(1:10, 10))
-#' b <- data.frame(column = c(1:5, 5))
+#' # Note these functions also remove pre-existing duplicates in x or y:
```
DavisVaughan, line 29:
```suggestion
#' # Note the following 3 functions also remove pre-existing duplicates in `x` or `y`:
```

I thought "these functions" referred to all of the functions on this page, which of course isn't true

R/sets.r
```diff
@@ -21,21 +26,27 @@
...
+#' @usage intersect(x, y, ...)
+#' @importFrom generics intersect
+#' @export intersect
+NULL
+
+#' @name setops
+#' @aliases union
+#' @usage union(x, y, ...)
+#' @importFrom generics union
+#' @export union
```
DavisVaughan, line 49: Screen Shot 2022-08-01 at 4 33 56 PM

Hmm, doing `?intersect` and then clicking on the `Set operations` page for `generics` doesn't seem to link to the dplyr methods page

This doesn't work with the CRAN version of dplyr either

_pkgdown.yml
```diff
@@ -13,86 +13,99 @@ home:
...
 
 - title: Vector functions
+  desc: >
+    Unlike other dplyr functions, these functions work on individual vectors,
+    not data frames.
   contents:
   - across
-  - c_across
   - between
+  - c_across
```
DavisVaughan, line 74: Do you want `across()` and `c_across()` mentioned both here and in their special `Multiple columns` section?

_pkgdown.yml
```diff
@@ -13,86 +13,99 @@ home:
...
   - nest_join
   - semi_join
   - join_by
+  - rows
 
-- title: Grouping
+- subtitle: Multiple columns
+  desc: >
+    Pair these functions with `mutate()`, `summarise()`, `filter()`, and
+    `group_by()` to operate on multiple columns simultaneously
```
DavisVaughan, line 62:
```suggestion
    `group_by()` to operate on multiple columns simultaneously.
```

Since you do it elsewhere

_pkgdown.yml
```diff
@@ -13,86 +13,99 @@ home:
...
   - nth
-  - ranking
+  - ntile
+  - order_by
+  - percent_rank
   - recode
+  - row_number
 
-- title: Data
+- title: Build in datasets
```
DavisVaughan, line 92:
```suggestion
- title: Built in datasets
```

_pkgdown.yml
```diff
@@ -13,86 +13,99 @@ home:
...
   - between
+  - c_across
   - case_when
   - coalesce
   - cumany
   - desc
   - if_else
   - lead
-  - order_by
   - "n"
```
DavisVaughan, line 81: The context dependent expressions that come from including `n()` here somehow feel like they dont belong in this "Vector functions" section

DavisVaughan (approved):
Looks great, just a bunch of small tweaks

R/sets.r
hadley: Ooops, I should describe them all.

_pkgdown.yml
hadley: Oops no

_pkgdown.yml
hadley: Agreed, moved to grouping since that's where they mostly live. Could also consider moving `cur_column()` into `across()` or its own file.

R/doc-methods.R
```diff
@@ -31,6 +31,14 @@ methods_generic <- function(x) {
   # Don't link to self
   info$topic[info$topic == x] <- NA
 
+  # Remove spurious matches in base packages like slice.list or slice.index
+  base_packages <- c(
+    "base", "compiler", "datasets", "graphics", "grDevices", "grid",
+    "methods", "parallel", "splines", "stats", "stats4", "tcltk",
+    "tools", "utils"
+  )
+  info <- info[!info$package %in% base_packages]
```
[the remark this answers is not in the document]
hadley: There was a bug in `methods_generic()`; I've fixed it and they now all look good to me.

Issue #6057, opened 2021-10-19
Shouldn't setequal ignore duplicates? 


juangomezduaso:
I thought that being in the group of set operations, _setequal()_ would deem things like, say,  _mtcars_ and _bind_rows(mtcars, mtcars)_ as equal, but it doesn´t. 
i suppose this is the intended behavior, but in that case: Souldn´t this be explained somewhere in the documentation? My apologies if it already is, as I was unable to find it. 
Taking into account that Rbase behavior in vectors ignores duplicates (and so  _bind_rows(mtcars,mtcars)[[1]]_ and  _mtcars[[1]]_ are equal) I think this clarification would be useful.

DavisVaughan:
Minimal reprex showing the difference with base R

``` r
setequal(1, c(1, 1))
#> [1] TRUE

dplyr::setequal(data.frame(x = 1), data.frame(x = c(1, 1)))
#> [1] FALSE
```

Seems reasonable that we should be consistent and drop duplicates before checking

PR #6354, opened 2022-07-22
`setequal()` improvements

### LINKED ISSUE #6114: setdiff vs. setequal inconsitency when columns have different types
It seems like `setequal` compares column types while `setdiff` doesn't.

```r
df1 <- data.frame(x=1)
df2 <- df1

# df1$x and df2$x are both of type `num`
setdiff(df1, df2) # --> will output 0-rowed data.frame
setequal(df1, df2) # --> will output TRUE

df2$x %<>% as.integer 

# df1$x is of type `num` while df2$x is of type `int`
setdiff(df1, df2) # --> will output 0-rowed data.frame
setequal(df1, df2) # --> will output FALSE
```

I can see how this could actually be the intended behavior. 
However, in that case a hint in the docs would be great.

hadley:
I don't love that there's so much variation in where the casting happens in these functions, but this seems like the simplest fix. `convert = TRUE` here eventually calls `vec_ptype2(x_i, y_i)` so it's still using vctrs rules.

Fixes #6114.


hadley:
I also noticed #6057, and fixing that leads to a cleaner implementation, I think.

DavisVaughan:
Closes #6057

R/sets.r
```diff
@@ -67,44 +67,51 @@ generics::setequal
...
+  out <- vec_unique(vec_rbind(x, y))
   reconstruct_set(out, x)
 }
 
 #' @export
 union_all.data.frame <- function(x, y, ...) {
   check_dots_empty()
-  out <- bind_rows(x, y)
+
+  out <- vec_rbind(x, y)
```
DavisVaughan, line 89: Should this call `check_compatible(x, y)` like `union()`? i.e. to me `union() == vec_unique(union_all())`?

R/sets.r
```diff
@@ -67,44 +67,51 @@ generics::setequal
...
   check_dots_empty()
   check_compatible(x, y)
-  cast <- vec_cast_common(x, y)
-  new_x <- cast[[1L]]
-  new_y <- cast[[2L]]
-  out <- vec_unique(vec_slice(new_x, !vec_in(new_x, new_y)))
+
+  cast <- vec_cast_common(x = x, y = y)
+  out <- vec_unique(vec_slice(cast$x, !vec_in(cast$x, cast$y)))
   reconstruct_set(out, x)
```
DavisVaughan, line 100: Is this `dplyr_reconstruct()`? It certainly looks like it if you compare to `dplyr_reconstruct.grouped_df`

R/sets.r
```diff
@@ -67,44 +67,51 @@ generics::setequal
...
 }
 
 #' @export
 setequal.data.frame <- function(x, y, ...) {
   check_dots_empty()
-  isTRUE(equal_data_frame(x, y))
+  if (!is.data.frame(y)) {
+    abort("`y` must be a data frame. ", call = error_call)
```
DavisVaughan, line 107:
```suggestion
    abort("`y` must be a data frame.")
```

No error call there, plus an extra space. Maybe add a test for this?

hadley:
I don't fully understand the logic behind "linked issues". You'd think it'd be the same as the issues that get closed when the PR is merged but it's not (I already had `Closes #6057` in the commit message).

tests/testthat/test-sets.R
```diff
@@ -90,6 +93,32 @@ test_that("set equality", {
...
+
+test_that("setequal equality ignores duplicated rows", {
+  df1 <- tibble(x = 1)
+  df2 <- df1[c(1, 1, 1), ]
+
+  expect_true(setequal(df1, df2))
+  expect_true(setequal(df2, df1))
+})
+
+test_that("setequal uses coercion rules", {
```
DavisVaughan, line 112:
```suggestion
test_that("setequal uses coercion rules (#6114)", {
```

tests/testthat/test-sets.R
```diff
@@ -90,6 +93,32 @@ test_that("set equality", {
...
 
+test_that("setequal ignores column and row order", {
+  df1 <- tibble(x = 1:2, y = 3:4)
+  df2 <- df1[2:1, 2:1]
+
+  expect_true(setequal(df1, df2))
+  expect_true(setequal(df1, df2))
+})
+
+test_that("setequal equality ignores duplicated rows", {
```
DavisVaughan, line 104: There is an existing (but odd) test above called `"set operations remove duplicates"` if you wanted to add to that

DavisVaughan, line 104:
```suggestion
test_that("setequal equality ignores duplicated rows (#6057)", {
```

R/sets.r
```diff
@@ -67,44 +67,51 @@ generics::setequal
...
+
   cast <- vec_cast_common(x = x, y = y)
-  new_x <- cast[[1L]]
-  new_y <- cast[[2L]]
-  out <- vec_unique(vec_slice(new_x, vec_in(new_x, new_y)))
+  out <- vec_unique(vec_slice(cast$x, vec_in(cast$x, cast$y)))
   reconstruct_set(out, x)
 }
 
 #' @export
```
DavisVaughan, line 76: Random note that the help docs for these could benefit from these methods getting a `#' @rdname setops` tag

If you do `?union_all` the help page only has Usage for `union_all()` even though the help page is for all the methods

DavisVaughan (approved):

R/sets.r
hadley: See my other docs PR

R/sets.r
hadley: Yeah, definitely

R/sets.r
hadley: Yeah, good catch. Must've be written before `dplyr_reconstruct` existed

R/sets.r
hadley: Oops yeah.

tests/testthat/test-sets.R
hadley: I decided to put all the setequal tests together since it's a bit different. I'll do a little more test cleanup.

PR #6355, opened 2022-07-22
Remove `fix_call()`

### LINKED ISSUE #6309: Should pass error calls instead of using `fix_call()`

hadley:
Fixes #6309


tests/testthat/_snaps/colwise-mutate.md
```diff
@@ -12,7 +12,7 @@
       (expect_error(mutate_at(tibble(), "test", ~1)))
     Output
       <error/vctrs_error_subscript_oob>
-      Error in `mutate_at()`:
+      Error in `tbl_at_vars()`:
```
hadley, line 15: I think regressions in superseded functions are fine.

R/across.R
```diff
@@ -297,18 +297,20 @@ across_setup <- function(cols,
   }
   across_cols <- mask$across_cols()
 
-  vars <- fix_call(
-    tidyselect::eval_select(cols, data = across_cols),
-    call = call(across_if_fn)
+  vars <- tidyselect::eval_select(
+    cols,
+    data = across_cols,
+    error_call = call(across_if_fn)
```
lionel-, line 303: If possible it's better to pass a frame environment, so that rlang can hide the (very) heavy tidyselect stack from user backtraces.

R/bind.r
```diff
@@ -139,7 +139,7 @@ bind_rows <- function(..., .id = NULL) {
   if (is.null(.id)) {
     names(dots) <- NULL
   }
-  out <- fix_call(vec_rbind(!!!dots, .names_to = .id))
+  out <- vec_rbind(!!!dots, .names_to = .id)
```
lionel-, line 142: Should we keep the `fix_call()` until `vec_rbind()` and `vec_cbind()` gain an `error_call` argument? Now that we've framed them as backend functions, it makes sense to me to add these arguments. I used to think only input-checking functions should in theory have `call` arguments.

lionel- (approved):
Are we releasing vctrs before dplyr? If so we could add `call` arguments to `vec_rbind()` and `vec_cbind()` to fully finish this.

R/across.R
hadley: I'm just preserving the existing code without understanding it; do you see an obvious way to get the right frame environment?

R/bind.r
hadley: I was thinking we'd keep this PR open until vctrs had the extension points we need.

R/slice.R
```diff
@@ -316,11 +316,7 @@ slice_eval <- function(mask, dots, error_call = caller_env()) {
     }
 
     index <<- 0L
-
-    fix_call(
-      vec_c(!!!out),
-      call = NULL
-    )
+    vec_c(!!!out)
```
DavisVaughan, line 319: I imagine this error call isn't great if we have a type failure here somehow? I'm not sure what path this goes through to get here

DavisVaughan (approved):

R/slice.R
hadley: I'll pick this up in #6386

PR #6356, opened 2022-07-23
Add `percentile_rank()` [WIP]

jmbarbone:
Wanted to get some feedback on viability before making further revisions (cleaner messages, documentation [pending changes from [ac1b789](https://github.com/tidyverse/dplyr/commit/ac1b7890e3fd407ac34b5f1ab6e04ef06044622f)], testing, NEWS).

This is related to #5943 where it seems that a separate `percentile_rank()` (contrast `percent_rank()`) may be desired.  This implements the [formula](https://en.wikipedia.org/wiki/Percentile_rank):

```
PR = (CF - (0.5 * F)) / N
```


hadley:
@jmbarbone thanks for working on this! Unfortunately I don't think we want to add another function here. We discussed it internally and are worried that adding a third option that's only slightly different just makes things more confusing for folks.

PR #6364, opened 2022-07-25
Use consistent definiton of desc in arrange()

### LINKED ISSUE #6231: Re-exporting desc() breaks in arrange
I have a package that re-exports functions from various places so I can easily use them in some internal packages without having to deal with the separate imports. I can easily work around this issue by not doing that, but the bug surprised me and was invisible without direct testing, so I wanted to report it in case you want to deal with it.

Basically, if I reexport `desc()`, and I call it by namespacing, it doesn't do anything in `arrange()`.

```r
# Install my example package that reexports desc.
# devtools::install_github("jonthegeek/mydesc")
df <- data.frame(x = c("first", "second"))
dplyr::arrange(df, dplyr::desc(x))$x[[1]]
#> [1] "second"
dplyr::arrange(df, mydesc::desc(x))$x[[1]]
#> [1] "first"
```

It appears to be an issue with row 98 of arrange.R, but I'm not sure how/if it should be fixed.

hadley:
Fixes #6231

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,8 @@
 # dplyr (development version)
 
+* `df |> arrange(mydesc::desc(x))` works correctly when the mydesc re-exports
+   `dplyr::desc()` (#6231).
+
 * `union_all()`, like `union()`, now requires that data frames be compatible:
   i.e. they have the same columns, and the columns have compatible types.
 
--- a/R/arrange.R
+++ b/R/arrange.R
@@ -116,14 +116,15 @@ arrange_rows <- function(data,
     return(out)
   }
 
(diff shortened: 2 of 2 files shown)


hadley:
I don't see an obvious way to test this without adding a dependency to an artificial package, but I think it's pretty obvious what was going wrong here, and the simple reprex now works:

``` r

df <- data.frame(x = 1:2)
arrange(df, mydesc::desc(x))$x
#> [1] 2 1
```

<sup>Created on 2022-07-25 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

PR #6368, opened 2022-07-26
Automatically unlist() lists of length 1 vectors

### LINKED ISSUE #6302: `rowwise()` + `mutate()` doesn't respect rowwise-ness when copying a list-col
I don't think this is expected. I would have expected the first to return an `x` column containing `c(1, 2, 3, 4)` as a double vector.

``` r
library(dplyr)

df <- tibble(x = list(1, 2, 3, 4))
df <- rowwise(df)

# `x` isn't accessed an element at a time
mutate(df, y = x)
#> # A tibble: 4 × 2
#> # Rowwise: 
#>   x         y        
#>   <list>    <list>   
#> 1 <dbl [1]> <dbl [1]>
#> 2 <dbl [1]> <dbl [1]>
#> 3 <dbl [1]> <dbl [1]>
#> 4 <dbl [1]> <dbl [1]>

# `x` is accessed an element at a time
mutate(df, y = x + 1)
#> # A tibble: 4 × 2
#> # Rowwise: 
#>   x             y
#>   <list>    <dbl>
#> 1 <dbl [1]>     2
#> 2 <dbl [1]>     3
#> 3 <dbl [1]>     4
#> 4 <dbl [1]>     5
```

I think something is off with this section. We can't just use the `.data[[name]]` result here, we need to unchop it (checking that all elements are size 1 first)

https://github.com/tidyverse/dplyr/blob/55dfc1c5b49eae3f1b144ebe7d9fa2c3d79b1835/R/mutate.R#L323-L336

hadley:
Fixes #6302

### DIFF
--- a/R/mutate.R
+++ b/R/mutate.R
@@ -336,7 +336,8 @@ mutate_cols <- function(.data, dots, caller_env, error_call = caller_env()) {
                 dplyr_error_data = list(result_size = sizes[group], expected_size = 1)
               )
             }
-            result <- unlist(result, recursive = FALSE)
+            result_ptype <- attr(result, "ptype", exact = TRUE)
+            result <- vec_unchop(result, ptype = result_ptype)
           }
         } else if (!quo_is_symbolic(quo) && !is.null(quo_get_expr(quo))) {
           # constant, we still need both `result` and `chunks`
--- a/R/mutate.R
+++ b/R/mutate.R
@@ -336,6 +336,7 @@ mutate_cols <- function(.data, dots, caller_env, error_call = caller_env()) {
                 dplyr_error_data = list(result_size = sizes[group], expected_size = 1)
               )
             }
+            result <- unlist(result, recursive = FALSE)
           }
         } else if (!quo_is_symbolic(quo) && !is.null(quo_get_expr(quo))) {
           # constant, we still need both `result` and `chunks`
--- a/tests/testthat/test-mutate.r
+++ b/tests/testthat/test-mutate.r
@@ -230,6 +230,16 @@ test_that("rowwise mutate gives expected results (#1381)", {
   expect_equal(res$y, c(NA, 2, 3))
 })
 
+test_that("rowwise mutate un-lists existing length-1 list-columns", {
+  df <- rowwise(tibble(x = as.list(1:3)))
+  out <- mutate(df, y = x)
+  expect_equal(out$y, 1:3)
+
+  df <- rowwise(tibble(x = list(1, 2:3)))
+  expect_snapshot(mutate(df, y = x), error = TRUE)
+})
+
+
 test_that("grouped mutate does not drop grouping attributes (#1020)", {
   d <- data.frame(subject = c("Jack", "Jill"), id = c(2, 1)) %>% group_by(subject)
   a1 <- names(attributes(d))
--- a/tests/testthat/test-mutate.r
+++ b/tests/testthat/test-mutate.r
@@ -230,11 +230,28 @@ test_that("rowwise mutate gives expected results (#1381)", {
   expect_equal(res$y, c(NA, 2, 3))
 })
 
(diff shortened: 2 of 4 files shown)


tests/testthat/test-mutate.r
```diff
@@ -230,6 +230,16 @@ test_that("rowwise mutate gives expected results (#1381)", {
   expect_equal(res$y, c(NA, 2, 3))
 })
 
+test_that("rowwise mutate un-lists existing length-1 list-columns", {
```
DavisVaughan, line 233:
```suggestion
test_that("rowwise mutate un-lists existing length-1 list-columns (#6302)", {
```

tests/testthat/test-mutate.r
```diff
@@ -230,6 +230,16 @@ test_that("rowwise mutate gives expected results (#1381)", {
...
+  df <- rowwise(tibble(x = as.list(1:3)))
+  out <- mutate(df, y = x)
+  expect_equal(out$y, 1:3)
+
+  df <- rowwise(tibble(x = list(1, 2:3)))
+  expect_snapshot(mutate(df, y = x), error = TRUE)
+})
+
```
DavisVaughan, line 241: Some additional tests that I think are important

``` r
# Different code path from original problem
# (because `z` is based on a "new" column, `y`)
df <- tibble(x = 1:4)
df <- rowwise(df)
out <- mutate(df, y = list(1), z = y)
expect_identical(out$z, rep(1, times = 4))

# Proof that `vec_unchop()` was needed over `unlist()`
df <- tibble(x = list(tibble(a = 1), tibble(a = 2)))
df <- rowwise(df)
out <- mutate(df, y = x)
expect_identical(out$y, tibble(a = c(1, 2)))

# Preserves list-of type when evaluating symbolic column expression
df <- tibble(x = list_of(.ptype = character()))
df <- rowwise(df)
out <- mutate(df, y = x)
expect_identical(out$y, character())

# Needs #6369 too. Should be `y = NULL`, so no `y` column in result.
df <- tibble(x = list())
df <- rowwise(df)
out <- mutate(df, y = x)
expect_named(out, "x")
#> Error: Names of `out` ('x', 'y') don't match 'x'
```

R/mutate.R
```diff
@@ -336,6 +336,7 @@ mutate_cols <- function(.data, dots, caller_env, error_call = caller_env()) {
                 dplyr_error_data = list(result_size = sizes[group], expected_size = 1)
               )
             }
+            result <- unlist(result, recursive = FALSE)
```
DavisVaughan, line 339:
```suggestion
            result_ptype <- attr(result, "ptype", exact = TRUE)
            result <- vec_unchop(result, ptype = result_ptype)
```

DavisVaughan (changes requested):
I think `vec_unchop()` works better than `unlist()` here. Suggested a few tests to go along with that change, plus one that would be useful to add if you merge #6369 first

R/mutate.R
```diff
@@ -336,6 +336,8 @@ mutate_cols <- function(.data, dots, caller_env, error_call = caller_env()) {
                 dplyr_error_data = list(result_size = sizes[group], expected_size = 1)
               )
             }
+            result_ptype <- attr(result, "ptype", exact = TRUE)
+            result <- vec_unchop(result, ptype = result_ptype)
```
hadley, line 340: Do you think we should change the default to `ptype = attr(x, "ptype")`?

DavisVaughan: I'm leaning towards no because we have tried extremely hard to keep the implementation/details of `list_of()` separated from any vctrs internals. In theory that would be admitting that `vec_unchop()` "knows" about the list-of class and I'm resistant to that without an extremely good reason to do so

DavisVaughan (approved):
LGTM. Reminder to add the test that also needs #6369 (depending on which is merged first)

PR #6369, opened 2022-07-26
Better handling of zero-row rowwise mutates()

### LINKED ISSUE #6303: `rowwise()` size zero data frame edge case
There is a weird edge case that occurs with size zero data frames and `rowwise()`. In these cases, a list-col is made up of just `list()`, meaning there aren't any elements in there to extract and use in the expression.

Practically, what ends up happening is that this `Rf_length(column) > 0` branch returns `FALSE`, so the list-column gets chopped, resulting in `list(list())`
https://github.com/tidyverse/dplyr/blob/55dfc1c5b49eae3f1b144ebe7d9fa2c3d79b1835/src/chop.cpp#L20-L24

That results in weird bugs, like this one:

``` r
library(dplyr)

df <- tibble(x = list(1, 2, 3, 4))
df <- rowwise(df)

# Looks good
mutate(df, y = x + 1)
#> # A tibble: 4 × 2
#> # Rowwise: 
#>   x             y
#>   <list>    <dbl>
#> 1 <dbl [1]>     2
#> 2 <dbl [1]>     3
#> 3 <dbl [1]>     4
#> 4 <dbl [1]>     5

# Now it is impossible to know type is in `x`
df <- df[0,]
df
#> # A tibble: 0 × 1
#> # Rowwise: 
#> # … with 1 variable: x <list>
df$x
#> list()

# In practice this ends up doing `x + 1` -> `list() + 1`, which is theoretically
# very different from the `x + 1` above
mutate(df, y = x + 1)
#> Error in `mutate()`:
#> ! Problem while computing `y = x + 1`.
#> Caused by error in `x + 1`:
#> ! non-numeric argument to binary operator
```

I think that `Rf_length(column) > 0` was added to help with this case, but I think it may have made things more confusing and less consistent, because you aren't accessing the "elements" of the list column, you are working on the whole list-col itself

hadley:
Fixes #6303

### DIFF
--- a/R/mutate.R
+++ b/R/mutate.R
@@ -338,6 +338,7 @@ mutate_cols <- function(.data, dots, caller_env, error_call = caller_env()) {
             }
             result_ptype <- attr(result, "ptype", exact = TRUE)
             if (length(result) == 0 && is.null(result_ptype)) {
+              # i.e. `vec_ptype_finalise(unspecified())` (#6369)
               result <- logical()
             } else {
               result <- vec_unchop(result, ptype = result_ptype)
--- a/R/mutate.R
+++ b/R/mutate.R
@@ -337,7 +337,11 @@ mutate_cols <- function(.data, dots, caller_env, error_call = caller_env()) {
               )
             }
             result_ptype <- attr(result, "ptype", exact = TRUE)
-            result <- vec_unchop(result, ptype = result_ptype)
+            if (length(result) == 0 && is.null(result_ptype)) {
+              result <- logical()
+            } else {
+              result <- vec_unchop(result, ptype = result_ptype)
+            }
           }
         } else if (!quo_is_symbolic(quo) && !is.null(quo_get_expr(quo))) {
           # constant, we still need both `result` and `chunks`
--- a/src/chop.cpp
+++ b/src/chop.cpp
@@ -24,8 +24,6 @@ void dplyr_lazy_vec_chop_grouped(SEXP chops_env, SEXP rows, SEXP data, bool roww
         column = PROTECT(Rf_allocVector(VECSXP, 1));
         if (ptype != R_NilValue) {
           SET_VECTOR_ELT(column, 0, ptype);
-        } else {
-          SET_VECTOR_ELT(column, 0, Rf_allocVector(VECSXP, 0));
         }
         SET_PRCODE(prom, column);
         UNPROTECT(2);
--- a/src/chop.cpp
+++ b/src/chop.cpp
@@ -17,11 +17,25 @@ void dplyr_lazy_vec_chop_grouped(SEXP chops_env, SEXP rows, SEXP data, bool roww
     SEXP prom = PROTECT(Rf_allocSExp(PROMSXP));
     SET_PRENV(prom, R_EmptyEnv);
     SEXP column = p_data[i];
-    if (rowwise && vctrs::vec_is_list(column) && Rf_length(column) > 0) {
-      SET_PRCODE(prom, column);
+
+    if (rowwise && vctrs::vec_is_list(column)) {
+      if (Rf_length(column) == 0) {
+        SEXP ptype = PROTECT(Rf_getAttrib(column, Rf_install("ptype")));
+        column = PROTECT(Rf_allocVector(VECSXP, 1));
+        if (ptype != R_NilValue) {
+          SET_VECTOR_ELT(column, 0, ptype);
+        } else {
+          SET_VECTOR_ELT(column, 0, Rf_allocVector(VECSXP, 0));
+        }
+        SET_PRCODE(prom, column);
+        UNPROTECT(2);
+      } else {
+        SET_PRCODE(prom, column);
+      }
     } else {
       SET_PRCODE(prom, Rf_lang3(dplyr::functions::vec_chop, column, rows));
     }
+
     SET_PRVALUE(prom, R_UnboundValue);
 
     Rf_defineVar(rlang::str_as_symbol(p_names[i]), prom, chops_env);
--- a/src/chop.cpp
+++ b/src/chop.cpp
@@ -24,6 +24,8 @@ void dplyr_lazy_vec_chop_grouped(SEXP chops_env, SEXP rows, SEXP data, bool roww
         column = PROTECT(Rf_allocVector(VECSXP, 1));
         if (ptype != R_NilValue) {
           SET_VECTOR_ELT(column, 0, ptype);
+        } else {
+          SET_VECTOR_ELT(column, 0, Rf_allocVector(LGLSXP, 1));
         }
         SET_PRCODE(prom, column);
         UNPROTECT(2);
--- a/src/chop.cpp
+++ b/src/chop.cpp
@@ -25,6 +25,7 @@ void dplyr_lazy_vec_chop_grouped(SEXP chops_env, SEXP rows, SEXP data, bool roww
         if (ptype != R_NilValue) {
           SET_VECTOR_ELT(column, 0, ptype);
         } else {
+          // i.e. `vec_ptype_finalise(unspecified())` (#6369)
           SET_VECTOR_ELT(column, 0, Rf_allocVector(LGLSXP, 1));
         }
         SET_PRCODE(prom, column);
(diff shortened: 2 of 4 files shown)


src/chop.cpp
```diff
@@ -17,11 +17,25 @@ void dplyr_lazy_vec_chop_grouped(SEXP chops_env, SEXP rows, SEXP data, bool roww
...
+
+    if (rowwise && vctrs::vec_is_list(column)) {
+      if (Rf_length(column) == 0) {
+        SEXP ptype = PROTECT(Rf_getAttrib(column, Rf_install("ptype")));
+        column = PROTECT(Rf_allocVector(VECSXP, 1));
+        if (ptype != R_NilValue) {
+          SET_VECTOR_ELT(column, 0, ptype);
+        } else {
+          SET_VECTOR_ELT(column, 0, Rf_allocVector(VECSXP, 0));
+        }
```
DavisVaughan, line 29: I dont think you need this branch. I think the main point of https://github.com/tidyverse/dplyr/issues/6303#issuecomment-1159754285 was to convince us that `column` should evaluate to `list(NULL)` here rather than `list(list())` (and you automatically get `NULL` as the list element from `Rf_allocVector(VECSXP, 1)`)

tests/testthat/test-mutate.r
```diff
@@ -199,13 +199,17 @@ test_that("mutate works on zero-row grouped data frame (#596)", {
...
+  out <- mutate(rf, x2 = x)
+  expect_s3_class(out, "rowwise_df")
+  expect_identical(out$x2, double())
 
+  # preserving classes of list-cols where possible
+  rf <- rowwise(tibble(x = list(), y = vctrs::list_of(.ptype = character())))
+  out <- mutate(rf, x2 = identity(x), y2 = identity(y))
+  expect_equal(out$x2, list())
+  expect_equal(out$y2, character())
+})
```
DavisVaughan, line 212: Would like to see a test that ensures that `x` comes through as `NULL`. Maybe the first 2 of the 3 tests here?

If you do add a test, could the test also link to `(#6303)`?

``` r
df <- tibble(x = list())
df <- rowwise(df)

fn <- function(x) {
  stopifnot(is.null(x))
  TRUE
}

# Prove that `x` is seen as `NULL`
out <- mutate(df, y = fn(x))
expect_identical(out$y, logical())

# This ends up doing `y = NULL`, which I think is correct actually
out <- mutate(df, y = identity(x))
expect_named(out, "x")

# This one fails, but I think that is #6302 (it should also do `y = NULL`)
out <- mutate(df, y = x)
expect_named(out, "x")
#> Error: Names of `out` ('x', 'y') don't match 'x'
```

DavisVaughan (changes requested):

hadley:
New test for `rowwise mutate un-lists existing size-1 list-columns (#6302)` from #6368:

```R
  # Needs #6369 too. Should be `y = NULL`, so no `y` column in result.
  df <- tibble(x = list())
  df <- rowwise(df)
  out <- mutate(df, y = x)
  expect_named(out, "x")
  #> Error: Names of `out` ('x', 'y') don't match 'x'
```

src/chop.cpp
hadley: Oh yeah, I should have commented about that. If I remove that branch, then I get:

```R
rf <- rowwise(tibble(x = list(), y = vctrs::list_of(.ptype = character())))
mutate(rf, x2 = identity(x), y2 = identity(y))
#> # A tibble: 0 × 3
#> # Rowwise: 
#> # … with 3 variables: x <list>, y <list<chr>>, y2 <chr>
```

i.e. `x2` is absent from the output. I assume that's because it gets unpacked into `NULL` which removes the column.

hadley: Oh, but based on your comment below, you think that's correct

hadley: I don't have any strong feelings here; so if you think dropping the column is the correct behaviour I'm happy to go with it. I don't think it's likely to affect much code in the wild.

DavisVaughan: I have a new proposal! In `tidyr::unchop()` in these cases we use `unspecified(0L)`. So we could use that here as well - that would avoid dropping the column (which does feel a little wrong) but still conveys the "unknown"ness of it.

``` r
library(tidyr)

df <- tibble(x = list())
df2 <- unchop(df, x)
df2$x
#> <unspecified> [0]

df <- tibble(x = vctrs::list_of(.ptype = integer()))
df2 <- unchop(df, x)
df2$x
#> integer(0)
```

If you wanted to keep `unspecified()` as a purely internal type here, you could also return `logical()` instead, which is the "finalized" ptype of unspecified.

``` r
library(vctrs)

unspecified()
#> <unspecified> [0]

vec_ptype_finalise(unspecified())
#> logical(0)
```

I'd be fine with either one, I think. I think both of these are better solutions than `NULL` because:
- The column isn't dropped
- They are more consistent with tidyr `unchop()`, because this is effectively an unchopping operation, as seen by the usage of `vec_unchop()` in #6368

DavisVaughan:
With the `unspecified()` or `logical()` changes, I expect these tests to pass:

```r
  df <- tibble(x = list())
  df <- rowwise(df)

  # Symbol evaluation path
  out <- mutate(df, y = x)
  expect_named(out, c("x", "y"))
  expect_identical(out$y, logical()) # or unspecified()

  # Expression evaluation path
  out <- mutate(df, y = identity(x))
  expect_named(out, c("x", "y"))
  expect_identical(out$y, logical()) # or unspecified()
```

DavisVaughan:
This PR and #6368 entangle in a fairly complex way with this example:

```r
  df <- tibble(x = list())
  df <- rowwise(df)
  out <- mutate(df, y = x)
```

When you think of both PRs together, the mutate-cols loop will start out these these two variables:

```r
result <- .data[[name]] # list()
chunks <- mask$resolve(name) # list(logical()) assuming we use logical() as the fill value here
```

The #6368 change means we then end up with:
https://github.com/tidyverse/dplyr/pull/6368/files#diff-9b0f256680edfcc73a444f94fd9d40fdc00c66105a61d2a3d03331a5938f09e8R340

```r
result <- vec_unchop(result) # NULL
```

Because we have a `NULL` `result`, we end up finalizing `result` from the `chunks` instead
https://github.com/tidyverse/dplyr/pull/6368/files#diff-9b0f256680edfcc73a444f94fd9d40fdc00c66105a61d2a3d03331a5938f09e8R380

```r
result <- chunks[[1L]] # logical()
```

This works, but I feel like it kind of works "by luck" and I'd rather `result` be correctly finalized at the `vec_unchop()` step instead. I think the problem is that the C++ code here and the `vec_unchop()` code added in #6368 need to be completely in sync for this to work exactly right.

So I think the `vec_unchop()` code in #6368 needs to set `result` to `logical()` when `identical(result, list()) && is.null(result_ptype)` is `TRUE` (rather than calling `vec_unchop()`)

hadley:
I think I've now incorporated all your suggestions. LMK if you think I've missed any cases in the test.

R/mutate.R
```diff
@@ -337,7 +337,11 @@ mutate_cols <- function(.data, dots, caller_env, error_call = caller_env()) {
               )
             }
             result_ptype <- attr(result, "ptype", exact = TRUE)
-            result <- vec_unchop(result, ptype = result_ptype)
+            if (length(result) == 0 && is.null(result_ptype)) {
+              result <- logical()
```
DavisVaughan, line 341:
```suggestion
              # i.e. `vec_ptype_finalize(unspecified())` (#6369)
              result <- logical()
```

src/chop.cpp
```diff
@@ -17,11 +17,25 @@ void dplyr_lazy_vec_chop_grouped(SEXP chops_env, SEXP rows, SEXP data, bool roww
...
-      SET_PRCODE(prom, column);
+
+    if (rowwise && vctrs::vec_is_list(column)) {
+      if (Rf_length(column) == 0) {
+        SEXP ptype = PROTECT(Rf_getAttrib(column, Rf_install("ptype")));
+        column = PROTECT(Rf_allocVector(VECSXP, 1));
+        if (ptype != R_NilValue) {
+          SET_VECTOR_ELT(column, 0, ptype);
+        } else {
+          SET_VECTOR_ELT(column, 0, Rf_allocVector(LGLSXP, 1));
```
DavisVaughan, line 28:
```suggestion
          // i.e. `vec_ptype_finalize(unspecified())` (#6369)
          SET_VECTOR_ELT(column, 0, Rf_allocVector(LGLSXP, 1));
```

DavisVaughan (approved):
Looks great, thanks!

Added a few inline comments just because this one was so difficult to get right. Hopefully it (along with the tests) helps us to keep these in sync in the future

PR #6371, opened 2022-07-26
Preserve type of `y` in `nest_join()`

### LINKED ISSUE #6295: `nested_join()` to preserve the class of the nested data?
In the example below, if I nest `dt2`, the `rowwise_df` class is preserved in each list column element, I'm wondering if `nest_join()` should have a similar behavior? 

``` r
# set up `dt1` with an id column and some random numbers
set.seed(123)
dt1 <- tibble::tibble(
  id = c(1,2,3),
  a = rnorm(n = 3)
)

# set up `dt2` as a rowwise dataframe with `id2` nested inside `id1` column 
set.seed(123)
dt2 <- tibble::tibble(
  id = rep(c(1,2,3), each = 3),
  id2  = rep(c("a", "b", "c"), 3),
  b = rnorm(n = 9)
) |>
  dplyr::rowwise()

# nesting `dt2` gives each element in data as a `rowwise_df`
dt2 |> tidyr::nest(data = -id)
#> # A tibble: 3 × 2
#>      id data                
#>   <dbl> <list>              
#> 1     1 <rowwise_df [3 × 2]>
#> 2     2 <rowwise_df [3 × 2]>
#> 3     3 <rowwise_df [3 × 2]>

# but `nest_join()` doesnt not preserve the `rowwwise_df` class 
dt1 |> dplyr::nest_join(dt2, by = "id")
#> # A tibble: 3 × 3
#>      id      a dt2             
#>   <dbl>  <dbl> <list>          
#> 1     1 -0.560 <tibble [3 × 2]>
#> 2     2 -0.230 <tibble [3 × 2]>
#> 3     3  1.56  <tibble [3 × 2]>
```

<sup>Created on 2022-06-13 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

hadley:
Fixes #6295


R/join.r
```diff
@@ -513,7 +513,7 @@ nest_join.data.frame <- function(x,
   # changing the key vars because of the cast
   new_cols <- vec_cast(out[names(x_key)], vec_ptype2(x_key, y_key))
 
-  y_out <- set_names(y_in[vars$y$out], names(vars$y$out))
+  y_out <- set_names(y[vars$y$out], names(vars$y$out))
```
DavisVaughan, line 516: I think the main problem with this is that we are typically _very_ careful to only operate on bare tibbles in the join functions, since many `[` methods are broken (for our purposes). The most efficient thing to do is probably:

```r
  y_out <- set_names(y_in[vars$y$out], names(vars$y$out))
  y_out <- map(y_loc, vec_slice, x = y_out)
  y_out <- map(y_out, dplyr_reconstruct, template = y)
  new_cols[[name_var]] <- y_out
```

Then you can also add this test to ensure that attributes are retained on both `x` and `y` when they are bare data frames

```r
df1 <- data.frame(x = c(1, 2), y = c(3, 4))
df2 <- data.frame(x = c(1, 2), z = c(3, 4))

attr(df1, "foo") <- 1
attr(df2, "foo") <- 2

out <- nest_join(df1, df2, by = "x")

expect_identical(attr(out, "foo"), 1)
expect_identical(attr(out$df2[[1]], "foo"), 2)
expect_identical(attr(out$df2[[2]], "foo"), 2)
```

R/join.r
```diff
@@ -513,7 +513,7 @@ nest_join.data.frame <- function(x,
   # changing the key vars because of the cast
```
DavisVaughan, line 513: I think this comment that starts with `Currently, this regroups too often` is outdated?

With this new approach, when both `x` and `y` are grouped data frames we regroup:
- One time for `x`
- `nrow(out)` times for `y`

tests/testthat/test-join.r
```diff
@@ -261,6 +261,15 @@ test_that("nest_join returns list of tibbles (#3570)",{
   expect_s3_class(out$df2[[1]], "tbl_df")
 })
 
+test_that("nest_join respects types of y",{
```
DavisVaughan, line 264:
```suggestion
test_that("nest_join respects types of y (#6295)",{
```

DavisVaughan (changes requested):

R/join.r
hadley: Ah yeah, makes sense.

R/join.r
hadley: Removed the comment and moved explanation to test.

DavisVaughan (approved):

PR #6377, opened 2022-07-30
Check possible values of keep

### LINKED ISSUE #6375: Dev dplyr error when `keep` is not logical
Currently the error isn't informative

``` r
df <- data.frame(a = 1)
dplyr::inner_join(df, df, by = 'a', keep = 1)
#> Error in `x_key[merge]`:
#> ! Must subset columns with a valid subscript vector.
#> ✖ Subscript `merge` has the wrong type `function`.
#> ℹ It must be logical, numeric, or character.
```

<sup>Created on 2022-07-27 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

hadley:
Fixes #6375

### DIFF
--- a/R/join.r
+++ b/R/join.r
@@ -465,10 +465,18 @@ nest_join.data.frame <- function(x,
                                  na_matches = c("na", "never"),
                                  multiple = NULL,
                                  unmatched = "drop") {
+
+  check_keep(keep)
   na_matches <- check_na_matches(na_matches)
   unmatched <- check_unmatched(unmatched)
 
-  name_var <- name %||% as_label(enexpr(y))
+  if (is.null(name)) {
+    name_var <- as_label(enexpr(y))
+  } else if (is_string(name)) {
+    name_var <- name
+  } else {
+    abort("`name` must be a string.")
+  }
 
   x_names <- tbl_vars(x)
   y_names <- tbl_vars(y)
@@ -722,3 +730,4 @@ check_keep <- function(keep, error_call = caller_env()) {
   if (!is_bool(keep) && !is.null(keep)) {
     abort("`keep` must be `TRUE`, `FALSE`, or `NULL`.", call = error_call)
   }
+}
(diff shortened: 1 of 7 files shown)


R/join.r
```diff
@@ -536,6 +536,9 @@ join_mutate <- function(x,
   check_dots_empty0(...)
 
   na_matches <- check_na_matches(na_matches, error_call = error_call)
+  if (!identical(keep, TRUE) && !identical(keep, FALSE) && !identical(keep, NULL)) {
```
eutwt, line 539: Slightly shorter with rlang functions: `!is_true(keep) && !is_false(keep) && !is_null(keep)`

hadley: Good catch, thanks!

R/join.r
```diff
@@ -536,6 +536,9 @@ join_mutate <- function(x,
   check_dots_empty0(...)
 
   na_matches <- check_na_matches(na_matches, error_call = error_call)
+  if (!is_true(keep) && !is_false(keep) && !is.null(keep)) {
+    abort("`keep` must be `TRUE`, `FALSE`, or `NULL`", call = error_call)
+  }
```
DavisVaughan, line 541: A few notes:

- `nest_join()` also uses `keep` but doesn't go through `join_mutate()`, so this check needs to be there too
- You can use `is_bool()` from rlang as another shortcut here
- Added the full stop `.` to the end of the sentence
- It probably makes sense to extract out a helper like `check_unmatched()` and `check_na_matches()`, since you have to use this check in 2 places now.

So maybe:

```r
check_keep <- function(keep, ..., error_call = caller_env()) {
  check_dots_empty0(...)
  
  if (!is_bool(keep) && !is.null(keep)) {
    abort("`keep` must be `TRUE`, `FALSE`, or `NULL`.", call = error_call)
  }
}
```

Plus maybe a `nest_join()` test

DavisVaughan (changes requested):

R/join.r
hadley: I'm not a fan of the `...` pattern for internal functions like this, so I didn't implement it. But I went with the other suggestions and generally improved the error reporting across the join family.

PR #6378, opened 2022-07-30
Move slice checks to generic

### LINKED ISSUE #6361: slice helper error when no `n` or `prop`
For the slice helper functions, would it make sense to move `check_slice_dots()` from the data.frame methods to the generics? My thought is that dtplyr and dbplyr can rely on it to prevent malformed calls and provide the right error. Right now when `n` is not provided, dtplyr has an error message that's not as friendly, and dbplyr has different behavior. 

``` r
library(rlang)
library(glue)
#> Warning: package 'glue' was built under R version 4.1.2
df <- data.frame(a = 1:10)

#> Warning: package 'dplyr' was built under R version 4.1.2

df %>% 
  slice_sample(1)
#> Error in `slice_sample()`:
#> ! `n` must be explicitly named.
#> ℹ Did you mean `slice_sample(n = 1)`?

library(dtplyr)
#> Warning: package 'dtplyr' was built under R version 4.1.2

df %>% 
  lazy_dt() %>% 
  slice_sample(1)
#> Error in `slice_sample()`:
#> ! `...` must be empty.
#> ✖ Problematic argument:
#> • ..1 = 1
#> ℹ Did you forget to name an argument?

library(dbplyr, warn.conflicts = FALSE)
#> Warning: package 'dbplyr' was built under R version 4.1.2

con <- DBI::dbConnect(RSQLite::SQLite(), ":memory:")
copy_to(con, df)
df2 <- tbl(con, "df")

df2 %>% 
  slice_sample(1)
#> # Source:   SQL [1 x 1]
#> # Database: sqlite 3.38.5 [:memory:]
#>       a
#>   <int>
#> 1     6
```

<sup>Created on 2022-07-24 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

hadley:
@eutwt is this what you were thinking?

@DavisVaughan this narrows the interface of possible methods, but I think it should (a) not affect any existing code and (b) narrows the interface in a generally helpful way. I _think_ this is ok because the `slice_` generics already have relatively narrow scope.

Fixes #6361


eutwt:
> @eutwt is this what you were thinking?
 
Yep, that's what I had in mind. 

> (a) not affect any existing code and (b) narrows the interface in a generally helpful way

It can affect existing user code because right now dbplyr doesn't require explicitly naming arguments, e.g.  `slice_min(df, a, 1)` gives a result with no error. But I think it's helpful because that's probably unintentional in dbplyr and would eventually be adjusted anyway, so better to have the check in one place.

R/slice.R
```diff
@@ -205,14 +207,14 @@ slice_min.data.frame <- function(.data, order_by, ..., n, prop, with_ties = TRUE
 #' @export
 #' @rdname slice
 slice_max <- function(.data, order_by, ..., n, prop, with_ties = TRUE) {
```
DavisVaughan, line 209: Maybe add the same validation as `replace` in `slice_sample()` for `with_ties` in `slice_min()` and `slice_max()`? i.e. it must be a bool

R/slice.R
```diff
@@ -395,7 +397,7 @@ check_slice_dots <- function(..., n, prop, error_call = caller_env()) {
     dots <- enquos(...)
 
     if (length(dots) == 1L && names2(dots)[1] == "") {
-      slice_call <- error_call$.Generic
+      slice_call <- caller_call()[[1]]
```
DavisVaughan, line 400:
```suggestion
      slice_call <- frame_call(frame = error_call)[[1]]
```

Is this _slightly_ more correct? Probably doesn't matter much, but we do have `error_call` as an argument here so in theory the direct caller env isn't always the right one to get the `slice_call` from

DavisVaughan (approved):
LGTM, I can't think of anything that this would break

I guess it does now _require_ that `n` and `prop` be arguments of any methods here, but I think we'd want that anyways.

hadley:
Should merge after #6384. Need to add test for `with_ties()`

R/slice.R
hadley: Added by hand, thanks!

hadley:
Includes validation of all boolean arguments.

PR #6379, opened 2022-07-31
Improve `nest_join()` docs

### LINKED ISSUE #6232: `nest_join()` doesn't act on its `keep` argument
See below with comparison to `left_join()` behaviour

``` r
df1 <- tibble(col1 = 1:2, key = letters[1:2])
df2 <- tibble(col2 = 3:4, key2 = letters[2:3])

nest_join(df1, df2, by = c(key = "key2"), keep = TRUE)
#> # A tibble: 2 × 3
#>    col1 key   df2             
#>   <int> <chr> <list>          
#> 1     1 a     <tibble [0 × 2]>
#> 2     2 b     <tibble [1 × 2]>

left_join(df1, df2, by = c(key = "key2"), keep = TRUE)
#> # A tibble: 2 × 4
#>    col1 key    col2 key2 
#>   <int> <chr> <int> <chr>
#> 1     1 a        NA <NA> 
#> 2     2 b         3 b
```

<sup>Created on 2022-04-12 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

hadley:
And change default value of `multiple`; I'm pretty sure this is what `nest_join()` should do, since it doesn't have the Cartesian product problem of the other joins.

Fixes #6232


R/join.r
```diff
@@ -411,25 +411,28 @@ anti_join.data.frame <- function(x, y, by = NULL, copy = FALSE, ..., na_matches
...
-#'   more details.
-#' @param name The name of the list column nesting joins create.
-#'   If `NULL` the name of `y` is used.
+#' @param name The name of the list-column created by the join. If `NULL`,
+#'   the default, the name of `y` is used.
+#' @param keep Should the new list-column contain join keys? The default
+#'   will preserve the join keys for non-equi-joins.
+#' @return
+#' A modified copy of `x` that contains an new column called `name`. The
+#' column contains a list of objects the same type as `y`.
```
DavisVaughan, line 435: Would the bulleted list form that is used in the other join docs be useful here? It would let you state:

- The return type is the same type as `x`.
- The output has exactly the same number of rows as `x`. (This one feels particularly useful to state)
- The output has 1 more column than `x`, a list column named `name` containing data frames the same type as `y`.
- The output groups are taken from `x`.
- When `keep = FALSE`, output columns included in `by` are coerced to their common type between `x` and `y` (so it isn't always _exactly_ the same type as `x`, but it is close).

Screen Shot 2022-08-08 at 6 00 19 PM

R/join.r
```diff
@@ -411,25 +411,28 @@ anti_join.data.frame <- function(x, y, by = NULL, copy = FALSE, ..., na_matches
...
+#' * [left_join()] is a `nest_join()` plus `tidyr::unnest(.drop = FALSE)`.
+#' * [semi_join()] is a `nest_join()` plus a `filter()` where you check
 #'   that every element of data has at least one row,
-#' * `anti_join()` is a `nest_join()` plus a `filter()` where you check every
+#' * [anti_join()] is a `nest_join()` plus a `filter()` where you check every
 #'   element has zero rows.
 #'
-#' @param x,y A pair of data frames, data frame extensions (e.g. a tibble), or
-#'   lazy data frames (e.g. from dbplyr or dtplyr). See *Methods*, below, for
-#'   more details.
```
DavisVaughan, line 430: Benefits of the roxygen2 update!

DavisVaughan (approved):
Looks great with the one comment about temporarily reverting the `multiple` change

R/join.r
```diff
@@ -463,7 +471,7 @@ nest_join.data.frame <- function(x,
                                  name = NULL,
                                  ...,
                                  na_matches = c("na", "never"),
-                                 multiple = NULL,
+                                 multiple = "all",
```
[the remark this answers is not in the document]
hadley: K, will do.

R/join.r
hadley: I like it. I did some polishing based on what you suggested, and also made some minor tweaks to the other join docs.

R/join.r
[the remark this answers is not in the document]
hadley: Could also describe what the different values of `multiple` mean here, in the same way that we define how the different joins work.

PR #6381, opened 2022-07-31
Refactor `n_distinct()`; implement `consecutive_id()`

### LINKED ISSUE #5184: adjacency grouping as a helper function
This has been discussed in various issues and forums, notably:

- https://github.com/tidyverse/dplyr/issues/237, suggesting `group_by_adj()` (suggesting pre-ordered adjacency);
- https://github.com/tidyverse/dplyr/issues/1534;
- https://github.com/tidyverse/dplyr/issues/1661; and
- https://github.com/tidyverse/dplyr/issues/4445 (most recent)

One key difference is that some of those discussions focused on a new `group_by_*` verb, where I suggest having just a helper function is sufficient to scratch the itch sufficiently. Lacking safeguards and sanity checks, it could be as straight-forward as:

```r
rle_group <- function(...) {
  yy <- rle(paste(...))
  rep(seq_along(yy$lengths), yy$lengths)
}

df <- data.frame(x = c(0, 0, 1, 0), y = c(2, 2, 2, 2))
df %>%
  group_by(grp = rle_group(x, y))
# # A tibble: 4 x 3
# # Groups:   grp [3]
#       x     y   grp
#   <dbl> <dbl> <int>
# 1     0     2     1
# 2     0     2     1
# 3     1     2     2
# 4     0     2     3
```

The under-inspiring `Groups: grp` may not be as informative as [Romain's previous suggestion](https://github.com/tidyverse/dplyr/issues/237#issuecomment-35085223) of

```lang-r
> df <- data.frame( x = 1, a = rep(c(1,2), each = 4), b = rep( letters[1:2], each = 2), stringsAsFactors = TRUE )
> gdf <- grouped_df_adj_impl( df, list( quote(x), quote(a), quote(b) ), FALSE )
> summarise(gdf, n = n() )
Source: local data frame [4 x 4]
Groups: x, a

  x a b n
1 1 1 a 2
2 1 1 b 2
3 1 2 a 2
4 1 2 b 2
```

(where the `Groups: x, a` is more useful), it still enables the grouping.

This comes up frequently enough on StackOverflow that I think its utility is justified. I think the implementation could be innocuous enough to not impact other verbs.

The use of `paste` enables "one or more" columns, and happens to side-step `rle`'s avoidance of `factor`. It does add overhead, so a more performant variant might be supported (that does not omit `factor`s). I'd think a variant of [Hadley's initial suggestion](https://github.com/tidyverse/dplyr/issues/237#issuecomment-34117419) (Rcpp-based) could be adapted to handle one-or-more.

I'd be willing to start on a PR if maintainers support the concept.

### LINKED ISSUE #6348: Add more useful examples to `n_distinct()`
Like to show how takes joint distinctness when given multiple columns

```r
library(dplyr)

x <- c(1, 1, 2, 2, 2)
y <- c(2, 2, 1, 2, 2)

# Values 1 and 2 are distinct
n_distinct(x)
#> [1] 2

# Pairs (1, 2), (2, 1), and (2, 2) are distinct
n_distinct(x, y)
#> [1] 3
```

hadley:
Fixes #6348. Fixes #5184.


R/consecutive-id.R
```diff
@@ -0,0 +1,24 @@
+#' Generate a unique identifier for consecutive combinations
+#'
+#' `consecutive_id()` generates a unique identifier that increments every time
+#' a variable (or combination of variables) changes. Inspired by
+#' `data.table::rle()`.
```
DavisVaughan, line 5:
```suggestion
#' `data.table::rleid()`.
```

R/n_distinct.R
```diff
@@ -0,0 +1,42 @@
...
+#' # (2, NA) is dropped, leaving 2 distinct combinations
+#' n_distinct(x, y, na.rm = TRUE)
+#'
+#' # Also works with data frames
+#' n_distinct(data.frame(x, y))
+n_distinct <- function(..., na.rm = FALSE) {
+  check_dots_unnamed()
+  data <- vctrs::data_frame(..., .name_repair = "minimal")
+
+  if (isTRUE(na.rm)) {
```
DavisVaughan, line 36: Maybe add a check using `is_bool(na.rm)` and abort if not bool?

R/n_distinct.R
```diff
@@ -0,0 +1,42 @@
+#' Count unique combinations
+#'
+#' `n_distinct()` counts the number of unique/distinct combinations in a set
+#' of one or more vectors. It's a faster and more concise equivalent to
+#' `nrow(unique(data.frame(...)))`.
+#'
+#' @param ... Unnamed vectors. If multiple vectors, supplied they should have
+#'   the same length. This function does technically respect the recycling
```
DavisVaughan, line 8:
```suggestion
#' @param ... Unnamed vectors. If multiple vectors are supplied, then they should
#'   have the same length. This function does technically respect the recycling
```

R/n_distinct.R
```diff
@@ -0,0 +1,42 @@
...
+#'   rules, but since a constant vector won't affect the unique values,
+#'   this is mostly of theoretical interest.
```
DavisVaughan, line 10: I personally feel that this should not be part of the parameter documentation, because we generally don't want users to even know this exists.

I would probably put it in `@details`? And I'd also probably do `respect the [recycling rules][vctrs::vector_recycling_rules]`

Up to you

R/n_distinct.R
```diff
@@ -0,0 +1,42 @@
...
+#' @param na.rm If `TRUE`, exclude missing values from the count.
+#' @returns A single number..
```
DavisVaughan, line 12:
```suggestion
#' @returns A single number.
```

R/n_distinct.R
```diff
@@ -0,0 +1,42 @@
...
+#' # Pairs (1, 3), (2, 3), and (2, NA) are distinct
+#' n_distinct(x, y)
+#'
+#' # (2, NA) is dropped, leaving 2 distinct combinations
+#' n_distinct(x, y, na.rm = TRUE)
+#'
+#' # Also works with data frames
+#' n_distinct(data.frame(x, y))
+n_distinct <- function(..., na.rm = FALSE) {
+  check_dots_unnamed()
```
DavisVaughan, line 33: The only potential issue I could see with this is if people have been doing `n_distinct(!!!df)`. People seem to do this with `coalesce()` fairly often, but I did a GitHub search and it doesn't seem like they do it with `n_distinct()` very much (probably because most people don't know you can supply >1 vectors).

R/n_distinct.R
```diff
@@ -0,0 +1,42 @@
...
+  data <- vctrs::data_frame(..., .name_repair = "minimal")
```
DavisVaughan, line 34: For `n_distinct()`, there is one edge case difference that comes from using `data_frame()` without an `unpack = FALSE` argument.

``` r
library(vctrs)
library(purrr)

a <- 1
b <- data_frame(x = 2, y = NA)

# CRAN `n_distinct()` (no unpacking)
# The df-col doesn't look "missing" because it is only partially missing,
# so it doesn't get dropped
data <- data_frame(a = a, b = b)
data
#>   a b.x b.y
#> 1 1   2  NA
vec_slice(data, !reduce(map(data, vec_equal_na), `|`))
#>   a b.x b.y
#> 1 1   2  NA

# PR `n_distinct()` (unpacking)
# The df-col is flattened, so the `y` column looks "missing" and the row
# is dropped
data <- data_frame(a, b, .name_repair = "minimal")
data
#>     x  y
#> 1 1 2 NA
vec_slice(data, !reduce(map(data, vec_equal_na), `|`))
#> [1]   x y
```

I'd argue that the new behavior is more correct and more consistent with the vctrs principle of an _incomplete_ row (the current behavior doesn't align well with the vctrs definitions of missingness or incompleteness). I think we should say that `n_distinct(na.rm = TRUE)` drops _incomplete_ rows - i.e. when an `NA` value appears _anywhere_ in the row then that is enough to not count it.

In that case, I'd suggest we simplify the missingness computation to this, which is also much simpler:

```r
if (na.rm) {
  # Drop row if any missing values are present
  complete <- vec_detect_complete(data)
  data <- vec_slice(data, complete)
}
```

The other nice thing here is that `vec_detect_complete()` works the same whether or not `unpack = FALSE` is set

DavisVaughan: This is probably a useful edge case test to add

```r
a <- 1
b <- data_frame(x = 2, y = NA)
n_distinct(a, b, na.rm = TRUE) == 0L
```

tests/testthat/test-consecutive-id.R
```diff
@@ -0,0 +1,20 @@
+test_that("works with simple vectors", {
+  expect_equal(consecutive_id(c(1, 1, 2, 1, 2)), c(1, 1, 2, 3, 4))
+})
+
+test_that("handles data frames", {
+  df <- tibble(x = c(1, 1, 1), y = c(1, 2, 1))
+  expect_equal(consecutive_id(df), c(1, 2, 3))
```
DavisVaughan, line 7:
```suggestion
  df <- tibble(x = c(1, 1, 1, 1), y = c(1, 2, 2, 1))
  expect_equal(consecutive_id(df), c(1, 2, 2, 3))
```

This seems slightly more informative for this function? i.e. it now tests:
- Identical consecutive rows get the same id (now tested with my suggestion)
- A new row always increments the id (was already tested)

DavisVaughan (approved):

R/consecutive-id.R
```diff
@@ -0,0 +1,24 @@
...
+#'
+#' @inheritParams n_distinct
+#' @export
```
DavisVaughan, line 8: Document return value?

R/n_distinct.R
hadley: That'd be easiest with `check_bool()` from #6378 which is waiting on #6384, so I'll just leave as is for now.

R/n_distinct.R
DavisVaughan: I'm fine with its removal, having tests for it (for us to look at later) is enough for me!

R/n_distinct.R
hadley: Yeah, I like this idea.

PR #6383, opened 2022-08-02
Add `na.rm` argument to `slice_min()` and `slice_max()`

### LINKED ISSUE #6177: NA handling in slice_min / slice_max
`slice_min` and `slice_max` drop rows where the value in the `order_by` column is `NA`, returning fewer than the requested number of rows if `n` is large enough.

It's unclear to me whether this is a bug or just intended behavior that requires documenting.

```r
library(tibble)
library(dplyr)

x <- tibble(a = 1:3, b = c(10, NA, 12))
slice_min(x, b, n = 3)
```

eutwt:
closes #6177

PR behavior below. For this example all the results are the same with `slice_max` instead of `slice_min` (`NA` always last).

This is the behavior I'd want, but it seems possible there could be a decent chunk of code out there relying on a default of na.rm = TRUE. 

``` r

df <- tibble(x = c(1, NA, NA))
df %>% slice_min(x, n = 2)
#> # A tibble: 3 × 1
#>       x
#>   <dbl>
#> 1     1
#> 2    NA
#> 3    NA
df %>% slice_min(x, n = 2, with_ties = FALSE)
#> # A tibble: 2 × 1
#>       x
#>   <dbl>
#> 1     1
#> 2    NA
df %>% slice_min(x, n = 2, na.rm = TRUE)
#> # A tibble: 1 × 1
#>       x
#>   <dbl>
#> 1     1
```

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,10 @@
 # dplyr (development version)
 
+* The default output of `slice_min()` and `slice_max()` no longer excludes
+  rows where the `order_by` variable has a missing value. A new argument `na.rm` 
+  is added, defaulting to `FALSE` and allowing rows with a missing value to be 
+  excluded by setting `na.rm = TRUE` (#6177).
+
 * `arrange()` now correctly ignores `NULL` inputs (#6193).
 
 * `*_join()` now error if you supply them with additional arguments that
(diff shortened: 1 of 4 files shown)


eutwt:
Closing in favor of #6384, which is better if only for not reinventing features of `vec_rank`

PR #6385, opened 2022-08-03
Implement symdiff()

### LINKED ISSUE #4811: Add a symmetric difference set function
I often have the need to find elements (rows) that two vectors (data frames) *don't* have in common. Finding these elements using setops is straightforward, but I wonder if it would be ok to add the symmetric difference (aka. union \ intersection) function to save a few key strokes. Below is my proposed solution. I can make the PR if we agree it makes sense to add this.

```r
symdiff <- function(x, y) {
  setdiff(union(x, y), intersect(x, y))
}
```

hadley:
Fixes #4811


NEWS.md
```diff
@@ -1,8 +1,6 @@
 # dplyr (development version)
 
-* Passing `...` to `across()` is now deprecated because the evaluation timing of
-  `...` is ambiguous. Now instead of (e.g.) `across(a:b, mean, na.rm = TRUE)`
-  you should write `across(a:b, ~ mean(.x, na.rm = TRUE))` (#6073).
+* New `symdiff()` function computes the symmetric difference (#4811).
```
DavisVaughan, line 3: Dropped a bullet?

R/sets.r
```diff
@@ -7,41 +7,47 @@
 #' * `union(x, y)` finds all rows in either `x` or `y`, excluding duplicates.
 #' * `union_all(x, y)` finds all rows in either `x` or `y`, including duplicates.
 #' * `setdiff(x, y)` finds all rows in `x` that aren't in `y`.
+#' * `symdiff(x, y)` computes the symetric difference, i.e. all rows in
+#'    `x` that aren't in `y` and all rows in `y` that aren't in `x`.
```
DavisVaughan, line 11:
```suggestion
#' * `symdiff(x, y)` computes the symmetric difference, i.e. all rows in
#'    `x` that aren't in `y` and all rows in `y` that aren't in `x`.
```

R/sets.r
```diff
@@ -7,41 +7,47 @@
...
 #' * `setequal(x, y)` returns `TRUE` if `x` and `y` contain the same rows
 #'   (ignoring order).
 #'
-#' Note that `intersect()`, `union()` and `setdiff()` remove duplicates
-#' in `x` and `y`.
+#' Note that `intersect()`, `union()`, `setdiff()` and `symdiff()` remove
```
DavisVaughan, line 15:
```suggestion
#' Note that `intersect()`, `union()`, `setdiff()`, and `symdiff()` remove
```

DavisVaughan (approved):

PR #6393, opened 2022-08-09
Use `$pick()` when `across(.fns = NULL)` is specified

### LINKED ISSUE #6264: Printing error involving `group_nest`, `rowwise` , and `across`
I came upon the following error when attempting to debug some code.  I think some combination of group_nest/rowwise/across is constructing a tibble that breaks the tbl formatter.

``` r
suppressWarnings(library(tidyverse))
tibble(x = 1:2, y = 1) %>% 
  group_nest(y) %>%
  rowwise() %>% 
  summarise(str = print(across()))
#> Error in `summarise()`:
#> ! Problem while computing `str = print(across())`.
#> i The error occurred in row 1.
#> Caused by error in `[<-.data.frame`:
#> ! replacement element 2 is a matrix/data frame of 2 rows, need 1
```

DavisVaughan:
Closes #6380 (alternative)
Closes #6264

@hadley I think I like this approach more?

- It seems a lot simpler to me
- It avoids a new argument to `current_cols()` (which I think shouldn't ever alter the cols if possible)
- It matches the spirit of what an official `pick()` function would do. It would be semantically the same as `across(.fns = NULL)`, so it is nice to see that we internally use `$pick()` for the `across(.fns = NULL)` code path.

I think this further supports the need for `pick()`. It would allow us to "purify" `across()` a bit more with rowwise data frames. Once we deprecate `.fns = NULL`, `across()` would _always_ work directly on list column elements VS `pick()` which would _always_ preserve the full list-column as-is

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,8 @@
 # dplyr (development version)
 
+* `across()` used without functions inside a rowwise-data frame no longer
+   generates an invalid data frame (#6264).
+
 * New `consecutive_id()` for creating groups based on contiguous runs of the
   same values, like `data.table::rleid()` (#1534).
 
--- a/R/across.R
+++ b/R/across.R
@@ -170,11 +170,9 @@ across <- function(.cols = everything(), .fns = NULL, ..., .names = NULL) {
   names <- setup$names
 
   mask <- peek_mask()
-  data <- mask$current_cols(vars)
 
   if (is.null(fns)) {
-    nrow <- length(mask$current_rows())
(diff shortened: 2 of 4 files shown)


hadley (approved):
Yeah, this looks a lot cleaner to me.

You might also want to remove the same old test that I did at https://github.com/tidyverse/dplyr/pull/6380/files#diff-45e563b486cc3a0bec1b439f59112c55a9ef8688d3ae465ce8fda8ac2c7c576fL514

PR #6401, opened 2022-08-13
Separate docs for bind_rows() and bind_cols()

### LINKED ISSUE #5063: Consider bind_cols() alternatives

hadley:
Fixes #5063


R/bind.r
```diff
@@ -150,8 +91,33 @@ bind_rows <- function(..., .id = NULL) {
...
+#'
+#' @description
+#' Bind any number of data frames by column, making a wider result.
+#' This is similar to `do.call(cbind, dfs)`.
+#'
+#' Where possible prefer using a [join][left_join] to combine multiple
+#' data frames. `bind_cols()` binds the rows in order in which they appear
+#' so it is easy to create meaningless results without realising it.
+#'
+#' @inheritParams bind_rows
```
DavisVaughan, line 104: No need for this, and roxygen2 gives us a nice warning about it now

R/bind.r
```diff
@@ -1,90 +1,31 @@
...
+#' similar to `do.call(rbind, dfs)`, but the output will contain all columns
+#' that appear in any of the inputs.
+#'
+#' @param ... Data frames to combine. Each argument can either be a data frame,
+#'   a list that could be a data frame, or a list of data frames. Columns are
+#'   matched by name, and any missing columns will be filled with `NA`.
+#' @param .id The name of an optional identifier column. Provide a string to
+#'   create an output column that identifies each input. The column will use
+#'   names if available, otherwise it will use positions.
+#' @returns A data frame the same type as the first element of `...`.
```
DavisVaughan, line 13:
```suggestion
#' @return A data frame the same type as the first element of `...`.
```

I didn't know a `@returns` alias existed. Use whichever you prefer I guess, just calling it out in case it was an accident

R/bind.r
```diff
@@ -150,8 +91,33 @@ bind_rows <- function(..., .id = NULL) {
...
+#'
+#' @inheritParams bind_rows
+#' @param ... Data frames to combine. Each argument can either be a data frame,
+#'   a list that could be a data frame, or a list of data frames.
+#'   Inputs are [recycled][vctrs::vector_recycling_rules] to the same length,
+#'   then matched by position.
+#' @inherit bind_rows return
+#' @param .name_repair One of `"unique"`, `"universal"`, or
+#'   `"check_unique"`. See [vctrs::vec_as_names()] for the meaning of these
+#'   options.
```
DavisVaughan, line 112:
```suggestion
#' @param .name_repair One of `"unique"`, `"universal"`, or
#'   `"check_unique"`. See [vctrs::vec_as_names()] for the meaning of these
#'   options.
#' @inherit bind_rows return
```

To keep params together

R/bind.r
```diff
@@ -150,8 +91,33 @@ bind_rows <- function(..., .id = NULL) {
...
+#' Where possible prefer using a [join][left_join] to combine multiple
+#' data frames. `bind_cols()` binds the rows in order in which they appear
+#' so it is easy to create meaningless results without realising it.
+#'
+#' @inheritParams bind_rows
+#' @param ... Data frames to combine. Each argument can either be a data frame,
+#'   a list that could be a data frame, or a list of data frames.
+#'   Inputs are [recycled][vctrs::vector_recycling_rules] to the same length,
+#'   then matched by position.
+#' @inherit bind_rows return
```
DavisVaughan, line 109: If we really don't like `bind_cols()` that much anymore, it might be nice to isolate it completely by not referencing `bind_rows()` docs at all. i.e. use an explicit `@return` section here even if it is just a copy/paste from bind-rows right now.

Not sure how you feel about it

DavisVaughan (approved):

R/bind.r
hadley: I think we prefer `@returns` generally, since it reads a little better.

R/bind.r
hadley: Fine with me.

R/bind.r
hadley: Done by hand

PR #6405, opened 2022-08-16
Allow negative n/prop in slice_sample()

### LINKED ISSUE #6402: slice_*: argument n and prop can be negative for all but slice_sample?
It seems that arguments `n` and `prop` are allowed to be negative (doc: _If a negative value of n or prop is provided, ...._) for all `sample_*` function except for `slice_sample`? Is that desired (yet undocumented?) behaviour, or an omission?

``` r
dplyr::slice_sample(iris, n=-1)
#> Error in `dplyr::slice_sample()`:
#> ! `n` must be positive.
``

### PR DESCRIPTION
**Pull request by hadley**
Fixes #6402

This seems reasonable to me, and already in alignment with what the docs suggest — negative `prop` and `n` drop rows.

### CODE REVIEW

File: R/slice.R [hunk 1]
```diff
@@ -444,24 +444,20 @@ check_bool <- function(x, arg = caller_arg(x), call = caller_env()) {
...
   slice_input <- check_slice_n_prop(n, prop, error_call = error_call)
 
   if (slice_input$type == "n") {
     if (slice_input$n >= 0) {
       function(n) floor(slice_input$n)
-    } else if (allow_negative) {
-      function(n) ceiling(n + slice_input$n)
     } else {
-      abort("`n` must be positive.", call = error_call)
+      function(n) ceiling(n + slice_input$n)
```
eutwt, line 454: If you allow negative n/prop in slice_sample, I think it should truncate at 0 like the other slice_ functions. 

``` r
library(dplyr, warn = FALSE)
df <- data.frame(x = 1)

df %>% 
  slice_head(n = -2)
#> [1] x

df %>% 
  slice_sample(n = -2)
#> Error in `slice_sample()`:
#> ! Problem while computing indices.
#> Caused by error in `sample.int()`:
#> ! invalid 'size' argument
```

<sup>Created on 2022-08-16 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1.9000)</sup>

hadley: Hmmmm, this makes me wonder if `slice_min()` and `slice_max()` are handling negative values correctly. 

I feel like maybe it's better to centralise the `n` + keep vs drop in this function?

eutwt: Yeah, get_slice_size originally did the truncating itself. But when support for negative n/prop in slice_sample was dropped, that piece of logic was removed since it was no longer necessary (for the other slice_ functions, because they use `slice` the "bad" indices are already dropped)

eutwt: Keep in mind though, that slice_ has one exception to the slice_size bounds, which is that `slice_sample` with `replace = TRUE` can create a larger-than-original group size.

hadley: Ah that's true. So `sample_int` probably just needs to handle negative values specifically, and I need to make sure each function is tested individually for its support of negative values.

R/slice.R
```diff
@@ -444,24 +444,20 @@ check_bool <- function(x, arg = caller_arg(x), call = caller_env()) {
...
 
-get_slice_size <- function(n, prop, allow_negative = TRUE, error_call = caller_env()) {
+get_slice_size <- function(n, prop, error_call = caller_env()) {
   slice_input <- check_slice_n_prop(n, prop, error_call = error_call)
 
   if (slice_input$type == "n") {
     if (slice_input$n >= 0) {
       function(n) floor(slice_input$n)
-    } else if (allow_negative) {
-      function(n) ceiling(n + slice_input$n)
```
DavisVaughan, line 454: Random: Does the fact that we have floor/ceiling here imply that `n` is allowed to be fractional? 🤢

R/slice.R
```diff
@@ -444,24 +444,20 @@ check_bool <- function(x, arg = caller_arg(x), call = caller_env()) {
...
+      function(n) ceiling(n + slice_input$n)
     }
   } else if (slice_input$type == "prop") {
     if (slice_input$prop >= 0) {
       function(n) floor(slice_input$prop * n)
-    } else if (allow_negative) {
-      function(n) ceiling(n + slice_input$prop * n)
     } else {
-      abort("`prop` must be positive.", call = error_call)
+      function(n) ceiling(n + slice_input$prop * n)
```
DavisVaughan, line 460:
```suggestion
      function(n) max(ceiling(n + slice_input$prop * n), 0L)
```

I think you could "fix" all of the problems by creating an invariant that `get_slice_size()` always returns a positive size. You'd apply this trick to `ceiling(n + slice_input$n)` too. I think that `slice_min/max/head/tail()` are already working right with OOB negative sizes, but this would be a centralized guarantee that they wouldn't have to even worry about it.

Then I think `sample_int()` is really just this, because it should handle the zero size case already

```r
sample_int <- function(n, size, replace = FALSE, wt = NULL, call = caller_env()) {
  if (!replace && n < size) {
    size <- n
  }
  sample.int(n, size, prob = wt, replace = replace)
}
```

DavisVaughan: In theory you'd have a symmetrical `min(floor(slice_input$prop * n), n)` clamp on the positive branches, which would allow you to remove this kind of check in `slice_head()` and `slice_tail()`

```r
if (to > n) {
  to <- n
}
```

and this in `sample_int()`

```r
if (!replace && n < size) {
  size <- n
}
```

You'd just have 1 special case for the `slice_sample(replace = TRUE)` case to _not_ clamp the positive branch (because, i.e., `slice_sample(mtcars, n = 100, replace = TRUE)` makes sense)

DavisVaughan: Maybe this should be a separate PR, but I do think it would dramatically simplify things

DavisVaughan (approved):

R/slice.R
hadley: Yeah, I think that's right, now that I understand what negative values actually do.

DavisVaughan (approved):

R/slice.R
```diff
@@ -423,48 +415,57 @@ check_slice_n_prop <- function(n, prop, error_call = caller_env()) {
     list(type = "n", n = 1L)
   } else if (!missing(n) && missing(prop)) {
     n <- check_constant(n, "n", error_call = error_call)
-    if (!is.numeric(n) || length(n) != 1 || is.na(n)) {
-      abort("`n` must be a single number.", call = error_call)
+    if (!is_integerish(n, n = 1) || is.na(n)) {
```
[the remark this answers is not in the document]
DavisVaughan: I think you can also just use `check_number()`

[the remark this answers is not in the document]
hadley: `Inf` is explicitly allowed in the tests 😬

R/slice.R
```diff
@@ -423,48 +415,57 @@ check_slice_n_prop <- function(n, prop, error_call = caller_env()) {
...
+      abort(
+        glue("`n` must be a round number, not {obj_type_friendly(n)}."),
+        call = error_call
+      )
     }
     list(type = "n", n = n)
   } else if (!missing(prop) && missing(n)) {
     prop <- check_constant(prop, "prop", error_call = error_call)
     if (!is.numeric(prop) || length(prop) != 1 || is.na(prop)) {
```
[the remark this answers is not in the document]
hadley: Prop isn't supposed to be an integer 😉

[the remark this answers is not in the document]
DavisVaughan: oh duh

PR #6406, opened 2022-08-16
Improve peek_mask() errors

### LINKED ISSUE #6403: Using `across()` inside dplyr's tidy selection verbs causes misleading error message
I'm aware that using `across()` inside dplyr's tidy selection verbs such as `select()` etc. doesn't make much sense, but it throws a misleading error message.

> Error in `context_peek()`:
> ! `across()` must only be used inside dplyr verbs.

Probably a message saying "must only be used inside data masking verbs" is more appropriate. But maybe there is a more beginner friendly formulation to express that.

Unfortunately (due to my current environment setup) I'm not on the latest dplyr version and cannot test if this error has been resolved already. However, I couldn't find an existing issue addressing this specific error although it popped up in a different context in #5333.

Below is one example for `select()`, but we get the same message for all major dplyr verbs that use tidy selection (`rename`, `relocate`, `pull`):

``` r
library(dplyr)

mtcars %>% 
  select(across(starts_with("m")))
#> Error in `context_peek()`:
#> ! `across()` must only be used inside dplyr verbs.
```
<sup>Created on 2022-08-16 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

hadley:
* Capture call rather than function name in `peek_mask()` and `context_peek()`
* Require `location` in `context_peek()`
* Make `peek_mask()` location more specific
* Require mask in `across_setup()` and `c_across_setup()`

Fixes #6403

### DIFF
--- a/R/across.R
+++ b/R/across.R
@@ -137,11 +137,14 @@
 #' @export
 #' @seealso [c_across()] for a function that returns a vector
 across <- function(.cols = everything(), .fns = NULL, ..., .names = NULL) {
+  mask <- peek_mask()
+
   setup <- across_setup(
     {{ .cols }},
     fns = .fns,
     names = .names,
     .caller_env = caller_env(),
+    mask = mask,
     inline = FALSE
   )
 
@@ -168,8 +171,6 @@ across <- function(.cols = everything(), .fns = NULL, ..., .names = NULL) {
   }
   fns <- setup$fns
   names <- setup$names
-
-  mask <- peek_mask()
 
   if (is.null(fns)) {
     data <- mask$pick(vars)
@@ -278,9 +279,10 @@ if_across <- function(op, df) {
 #'  )
 c_across <- function(cols = everything()) {
   cols <- enquo(cols)
-  vars <- c_across_setup(!!cols)
 
-  mask <- peek_mask("c_across")
+  mask <- peek_mask()
+  vars <- c_across_setup(!!cols, mask = mask)
+
 
   cols <- mask$current_cols(vars)
   vec_c(!!!cols, .name_spec = zap())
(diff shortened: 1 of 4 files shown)


DavisVaughan (approved):
Looks good, maybe add a test for the original `select()` case? It looks nice now

``` r
library(dplyr)
# and link to (#6403)
df <- tibble(x = 1)
select(df, across(x))
#> Error in `select()`:
#> ! Problem while evaluating `across(x)`.
#> Caused by error in `across()`:
#> ! Must only be used inside data-masking verbs like `mutate()`,
#>   `filter()`, and `group_by()`.
```

---

I was a little worried that evaluating `caller_env()` on every group when you do something like `summarise(df, n = n())` was going to significantly slow the evaluation down when there are many groups. But I tried this example (1mil groups, each of size 2) before / after this PR and both times it ran in around 11secs (which isn't great anyways, but can't do better without a hybrid eval).

```r
library(dplyr)

df <- tibble(g = rep(seq_len(1e6), each = 2))
df <- group_by(df, g)

bench::mark(summarise(df, n = n()), iterations = 10)
```

I guess it might be passing `call` through as an unevaluated promise all the way through to `context_peek()`, and then `context_peek()` never uses it so it never gets evaluated. `caller_env()` is also relatively fast (1us) so it shouldn't hurt too much either way I guess.

DavisVaughan:
Yea the `call` is getting passed through unevaluated like this, so that's good

``` r
library(rlang)

top <- function(x, call = stop("oh no!")) {
  middle(x, call = call)
}
middle <- function(x, call = caller_env()) {
  bottom(x, call = call)
}
bottom <- function(x, call = caller_env()) {
  x %||% call
}

top(1) # `call` remains unevaluated all the way through
#> [1] 1

top(NULL) # `call` gets evaluated
#> Error in x %||% call: oh no!
```

<sup>Created on 2022-08-17 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

hadley:
I didn't add another test, just because it feels like that error message already shows up in so many of the other tests.

PR #6408, opened 2022-08-18
Use rlang compat files for friendly type and type checks

DavisVaughan:
This brings in two rlang compat files in an attempt to standardize some helpers we've been using.

- `compat-obj-type.R` has `obj_type_friendly()`, the up to date version of `friendly_type_of()`
- `compat-types-check.R` requires the above compat file, and also includes `check_bool()`, `check_string()`, and a number of other check helpers

I've:
- Switched to `obj_type_friendly()` and removed our old `friendly_type_of()`
- Used `check_bool()` and `check_string()` in a few places and removed our custom `check_bool()`

As we discover more uses for these I feel like we should push the changes upstream into rlang so we can end up with a fairly comprehensive and standardized compat file.

Lionel is hesitant to export these `check_*()` functions from rlang until the potentials for type annotations has been completely explored


hadley (approved):
Looks good. Generally nice improvement to the generated errors 😄

PR #6409, opened 2022-08-18
Clarify tests/docs for negative n/prop

hadley:
I think this does a better job of conveying how negative values work. The previous explanation made me think that `bind_rows(slice_head(df, n = 5), slice_head(df, n = -5))` would yield `df`.

### DIFF
--- a/R/slice.R
+++ b/R/slice.R
@@ -35,13 +35,14 @@
 #'   proportion of rows to select. If neither are supplied, `n = 1` will be
 #'   used.
 #'
-#'   If a negative value of `n` or `prop` is provided, the specified number or
-#'   proportion of rows will be removed.
+#'   A negative value of `n` or `prop` will count backwards from the group
+#'   size. For example, `n = -2` with a group of 5 rows will select 5 - 2 = 3
+#'   rows; `prop = -0.25` with 8 rows will select 8 * (1 - 0.25) = 6 rows.
 #'
 #'   If `n` is greater than the number of rows in the group (or `prop > 1`),
-#'   the result will be silently truncated to the group size. If the
-#'   `prop`ortion of a group size does not yield an integer number of rows, the
-#'   absolute value of `prop*nrow(.data)` is rounded down.
+#'   the result will be silently truncated to the group size; `n < 0` or
+#'   `prop < 0 ` then the result will contain zero rows. If `prop * group_size`
+#'   is not an integer, it is rounded towards zero.
 #' @return
 #' An object of the same type as `.data`. The output has the following
 #' properties:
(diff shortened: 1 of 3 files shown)


R/slice.R
```diff
@@ -35,13 +35,14 @@
...
+#'   A negative value of `n` or `prop` will count backwards from the group
+#'   size. For example, `n = -2` with a group of 5 rows will select 5 - 2 = 3
+#'   rows; `prop = -0.25` with 8 rows will select 8 * (1 - 0.25) = 6 rows.
 #'
 #'   If `n` is greater than the number of rows in the group (or `prop > 1`),
-#'   the result will be silently truncated to the group size. If the
-#'   `prop`ortion of a group size does not yield an integer number of rows, the
-#'   absolute value of `prop*nrow(.data)` is rounded down.
+#'   the result will be silently truncated to the group size; `n < 0` or
+#'   `prop < 0 ` then the result will contain zero rows. If `prop * group_size`
```
DavisVaughan, line 44: What is this bit about zero rows? `n < 0` is just the negative `n` case, so I'm not understanding something here

DavisVaughan: Oh do you mean when the value of `n` is negative AND `abs(n)` is greater than the number of rows in the group?

That seemed a little hard to get from this

tests/testthat/test-slice.r
```diff
@@ -255,6 +255,16 @@ test_that("min and max ignore NA's when requested (#4826)", {
   expect_equal(slice_max(df, tibble(a, b), n = 3, na_rm = TRUE)$id, 1)
 })
 
+test_that("slice_min/max() count from back with negative n/prop", {
+  df <- tibble(id = 1:4, x = c(2, 3, 1, 4))
+  expect_equal(slice_min(df, x, n = -1), slice_min(df, x, n = 3))
+  expect_equal(slice_max(df, x, n = -1), slice_max(df, x, n = 3))
+
+  # and can be larger that group size
```
DavisVaughan, line 263:
```suggestion
  # and can be larger than group size
```

tests/testthat/test-slice.r
```diff
@@ -336,6 +346,16 @@ test_that("slice_head/slice_tail keep positive values", {
   expect_equal(slice_tail(gf, n = 2)$id, c(1, 2, 3, 5, 6))
 })
 
+test_that("slice_head/tail() count from back with negative n/prop", {
+  df <- tibble(id = 1:4, x = c(2, 3, 1, 4))
+  expect_equal(slice_head(df, n = -1), slice_head(df, n = 3))
+  expect_equal(slice_tail(df, n = -1), slice_tail(df, n = 3))
+
+  # and can be larger that group size
```
DavisVaughan, line 354:
```suggestion
  # and can be larger than group size
```

DavisVaughan (approved):

R/slice.R
hadley: No, I just meant `==`, not `<` 😄

Issue #6388, opened 2022-08-04
File/test tidyup


hadley:
Need to do after PRs mostly closed

* Ensure all use `.R`
* Ensure `tests/` and `R/` align

PR #6412, opened 2022-08-18
Refactor window tests

hadley:
Part of #6388

The coverage is unchanged, and I tried to delete only test that were duplicated or for hybrid code.


R/lead-lag.R
```diff
@@ -107,8 +103,7 @@ shift <- function(x,
   }
 
   vec_assert(x, arg = "x", call = error_call)
-
-  n <- check_shift_n(n, error_call = error_call)
+  check_number(n)
```
DavisVaughan, line 106: Would you mind keeping the cast to integer in addition to `check_number()`? It makes me feel better

```r
n <- vec_cast(n, integer(), call = error_call)
```

I think just having this in the `shift()` helper is enough, even though you call `check_number()` elsewhere too (it all comes through here eventually)

DavisVaughan (approved):

PR #6414, opened 2022-08-18
Supersede transmute()

hadley:


R/transmute.R
```diff
@@ -0,0 +1,77 @@
...
+#' properties:
+#'
+#' * Columns created or modified through `...` will be returned in the order
+#'     specified by `...`.
+#' * Unmodified grouping columns will be placed at the front.
+#' * The number of rows is not affected.
+#' * Columns given the value `NULL` will be removed.
+#' * Groups will be recomputed if a grouping variable is mutated.
+#' * Data frame attributes are preserved.
+#' @keywords internal
```
hadley, line 29: This should have examples, but `mutate()` didn't have any and no one complained, which I think is a strong signal that few people were using it.

R/transmute.R
```diff
@@ -0,0 +1,77 @@
...
+#' implementations (methods) for other classes. See the documentation of
+#' individual methods for extra arguments and differences in behaviour.
+#'
+#' Methods available in currently loaded packages:
+#' \Sexpr[stage=render,results=rd]{dplyr:::methods_rd("transmute")}.
+#' @returns An object of the same type as `.data`. The output has the following
+#' properties:
+#'
+#' * Columns created or modified through `...` will be returned in the order
+#'     specified by `...`.
```
DavisVaughan, line 23:
```suggestion
#' * Columns created or modified through `...` will be returned in the order
#'   specified by `...`.
```

R/transmute.R
DavisVaughan: Works for me

R/transmute.R
```diff
@@ -0,0 +1,77 @@
...
+#' * Unmodified grouping columns will be placed at the front.
+#' * The number of rows is not affected.
+#' * Columns given the value `NULL` will be removed.
+#' * Groups will be recomputed if a grouping variable is mutated.
+#' * Data frame attributes are preserved.
+#' @keywords internal
+#' @export
+transmute <- function(.data, ...) {
+  lifecycle::signal_stage("superseded", "transmute()", I("mutate(.keep = 'none')"))
```
DavisVaughan, line 32: That `I()` is useful!

R/mutate.R
```diff
@@ -62,29 +62,22 @@
...
 #' individual methods for extra arguments and differences in behaviour.
 #'
 #' Methods available in currently loaded packages:
-#'
-#' * `mutate()`: \Sexpr[stage=render,results=rd]{dplyr:::methods_rd("mutate")}.
-#' * `transmute()`: \Sexpr[stage=render,results=rd]{dplyr:::methods_rd("transmute")}.
+#' \Sexpr[stage=render,results=rd]{dplyr:::methods_rd("mutate")}.
 #' @examples
 #' # Newly created variables are available immediately
 #' starwars %>%
```
DavisVaughan, line 83: One of the examples says `# same as transmute()` which you may want to remove

DavisVaughan (commented):
A few more places to remove it from:

- Switch to `mutate(.keep = "none")` in `arrange_rows()`, and also update the `` "Problem with the implicit `transmute()` step." `` error bullet

- Mentioned once in `?dplyr_extending` i.e. `` `transmute()` does the same then uses 1d `[` ``

- `dplyr.Rmd` i.e. `` If you only want to keep the new variables, use `transmute()` ``

- `grouping.Rmd` i.e. `` ### `mutate()` and `transmute()` ``. This doesn't even use transmute so you could just say `mutate()` 

I think it automatically got removed from pkgdown, so thats good

R/transmute.R
```diff
@@ -0,0 +1,77 @@
+#' Create, modify, and delete columns
+#'
+#' @description
+#' `r lifecycle::badge("superseded")`
+#'
+#' `transmute()` is like `mutate(.keep = "none")`; adds new variables and drops
+#' existing ones. New variables overwrite existing variables of the same name.
+#' Variables can be removed by setting their value to `NULL`. It's supersed
```
DavisVaughan, line 8:
```suggestion
#' Variables can be removed by setting their value to `NULL`. It's superseded
```

R/transmute.R
DavisVaughan, line 32: Maybe add a comment with the dplyr version?

hadley: TOUCHÉ

R/transmute.R
```diff
@@ -0,0 +1,77 @@
+#' Create, modify, and delete columns
+#'
+#' @description
+#' `r lifecycle::badge("superseded")`
+#'
+#' `transmute()` creates a new data frame containing only the specified
+#' computations. It's superseded because you can perform the same job
+#' with `mutate(.keep = "none")`; it
```
DavisVaughan, line 8: `; it` ?

hadley: 😬

Issue #4663, opened 2019-12-13
Move database generics to dbplyr


hadley:
Deprecate database generics, following basic strategy in https://gist.github.com/hadley/5aa40f955adace1ccaf45975050b47fa so that existing database backends don't break (but will still require an update).

hadley:
Also need to remove `sql()` and `ident()` (#4434), and `src_dbi()` and friends (#4431)

hadley:
For dplyr 1.0.0 I will:

* Deprecate `src_postgres()`, `src_mysql()`, and `src_sqlite()`. The database dependencies will need to stay in suggests until these functions can be made defunct, which is probably 2021 at the earliest.

* Remove the `setOldClass()` definitions

hadley:
Now have a new plan for this where the generics no longer need to move; but we will need to eventually deprecate and then remove the database generics here. See details in https://dbplyr.tidyverse.org/articles/backend-2.html

hadley:
Need to break this up into smaller issues.

hadley:
Also need to rename all db files consistently:

```R
# compute-collect.r -> db-compute-complete.R
# copy-to.r -> db-copy-to.R
# dbplyr.r -> db-generics.R
# deprec-dbi -> ?
# explain.R -> db-explain.R
# order-by -> db-order-by
# src_dbi.R -> db-src.R
# src.R -> db-src.R
```

And plan for shifting all dbplyr generics (e.g. `sql`, `ident`, `show_query`, `src`, `is.src`, `same_src`, `auto_copy`, `explain`) in dbplyr.

@krlmlr's process for pillar/tbl

* in downstream:
  * implement the generic, export it
  * in .onLoad() or equivalent, overwrite generic with upstream::generic if it still exists there
  * copy default methods from upstream without adding @export
  * register methods if they aren't registered yet, I used the naming convention ..._if_..._hasnt or ..._if_..._has, I'll need to dig up that code if needed
  * release
* in upstream
  * remove generic and default methods
  * revdepcheck, alert downstream maintainers
  * wait if needed
  * release
* in downstream
  * remove compatibility code
  * release

Will probably need to update advice in https://dbplyr.tidyverse.org/articles/backend-2.html.

hadley:
Need to start by re-doing #6367 when we have an appetite for more revdep breakages.

PR #6418, opened 2022-08-19
WIP: Add dots to many generics

krlmlr:
From painful experience with pillar, it seems advisable to *always* add dots to all generics, even if there is no use for them today. Once a generic is available and used by other packages, I have found it to be *impossible* to add dots to a generic without temporarily CRAN-breaking all downstream packages that implement methods for the generic.

Is there a downside to always adding dots? Will R7 make this easier?

Action advice:

- Add dots at least to all new methods
- Consider `tally()`
- Add dots to methods that aren't used by other CRAN packages yet

Omitting database generics due to #4663.

### DIFF
--- a/R/colwise.R
+++ b/R/colwise.R
@@ -249,12 +249,12 @@ tbl_if_syms <- function(.tbl, .p, .env, ..., .include_group_vars = FALSE, error_
 #'
 #' @export
 #' @keywords internal
-tbl_ptype <- function(.data) {
+tbl_ptype <- function(.data, ...) {
   UseMethod("tbl_ptype")
 }
 
 #' @export
-tbl_ptype.default <- function(.data) {
+tbl_ptype.default <- function(.data, ...) {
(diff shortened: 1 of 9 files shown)


DavisVaughan (changes requested):
I personally think that the _only_ generic we should consider adding `...` to here is `tally()`.

And even with that I think we'd have to add the dots to the end because I am confident that many people do `tally(df, col)` rather than `tally(df, wt = col)`

---

I don't think internal tools like `dplyr_reconstruct()` should get `...` because 99.9% of the time the only place this will ever be called from is dplyr, and we won't ever be calling it with extra arguments.

Even for things like `group_data()`, I don't think we want people adding arbitrary arguments here.

`as_join_by()` is an internal generic, so I don't think it needs them either

krlmlr:
Yeah, and adding dots to `tally()` is too late now because first we'd need to add dots to all methods in all packages.

This doesn't happen if all generics get dots out of general principle. We can do better next time.

PR #6420, opened 2022-08-19
WIP: New .multi argument to summarize()

### LINKED ISSUE #6382: summarize() with multi-row returns
As of dplyr 1.0.0, `summarize()` will create multiple rows per group, according to the length of the return value of the summary function. This new feature leads to unintended behavior if the vector return is accidental, and also can lead to data loss.

``` r
library(conflicted)
library(dplyr)

my_custom_summary_function <- function(n) {
  # Should return a scalar, but I accidentally return a vector
  rep(n, n)
}

tibble(n = 2:0) %>% 
  group_by(n) %>% 
  summarize(out = my_custom_summary_function(n), .groups = "drop") %>% 
  ungroup()
#> # A tibble: 3 × 2
#>       n   out
#>   <int> <int>
#> 1     1     1
#> 2     2     2
#> 3     2     2
```

Should we introduce a `.multi = c("allow", "require", "fail")` argument that supports the pre-1.0.0 strict mode of operation? Should `.multi = "fail"` even be the default?

``` r
library(conflicted)
library(dplyr)

my_custom_summary_function <- function(n) {
  # Should return a scalar, but I accidentally return a vector
  rep(n, n)
}

tibble(n = 2:0) %>% 
  group_by(n) %>% 
  summarize(out = my_custom_summary_function(n), .groups = "drop", .multi = "fail") %>% 
  ungroup()
## Error: `out` has length != 1 in groups 1, 3, use `.multi = "allow"` if this is intended
```

<sup>Imagined on 2022-08-01 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

krlmlr:
``` r
library(conflicted)
library(dplyr)

my_custom_summary_function <- function(n) {
  # Should return a scalar, but I accidentally return a vector
  rep(n, n)
}

tibble(n = 2:0) %>%
  group_by(n) %>%
  summarize(out = my_custom_summary_function(n), .groups = "drop", .multi = "fail")
#> Error in `summarize()`:
#> ! A summary function returned a vector of size != 1, use `.multi =
#>   "require"` or `.multi = "allow"`.

tibble(n = 2:0) %>%
  group_by(n) %>%
  summarize(out = my_custom_summary_function(n), .groups = "drop", .multi = "require")
#> # A tibble: 3 × 2
#>       n   out
#>   <int> <int>
#> 1     1     1
#> 2     2     2
#> 3     2     2

tibble(n = 2:0) %>%
  group_by(n) %>%
  summarize(out = n, .groups = "drop", .multi = "require")
#> Error in `summarize()`:
#> ! All summary functions returned a vector of size one, use `.multi =
#>   "fail"` or `.multi = "allow"`.
```

<sup>Created on 2022-08-19 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

Happy to discuss argument name and values. The error message can be made much prettier, I'll also add tests.

There seems to be no mention of multi-row `summarize()` in the vignettes.

Closes #6382.

### DIFF
--- a/R/summarise.R
+++ b/R/summarise.R
@@ -57,6 +57,17 @@
 #'   In addition, a message informs you of that choice, unless the result is ungrouped,
 #'   the option "dplyr.summarise.inform" is set to `FALSE`,
 #'   or when `summarise()` is called from a function in a package.
+#' @param .multi `r lifecycle::badge("experimental")` Support multiple resulting rows per group
+#'
+#'   As of dplyr 1.0.0, if one summary function returns a vector of size != 1,
+#'   multiple rows are created for each group.
+#'   This feature is currently supported only for data frames.
+#'   While useful occasionally, it may lead to accidental duplication or loss of output rows.
+#'   Specifying `.multi = "fail"` leads to an error if any summary function
+#'   has returned a vector of size != 1.
+#'   With `.multi = "require"`, an error is raised if the size of all summaries is 1.
+#'   The default, `.multi = "allow"`, means no checking and is kept for compatibility
+#'   with earlier versions of dplyr.
 #'
 #' @family single table verbs
 #' @return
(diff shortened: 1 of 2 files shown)


hadley:
We'll definitely fix this for 1.1, but we're planing to take a different approach — we'll add a new function so we can deprecate multi-row summaries in `summarise()`. Function name TBD.

PR #6426, opened 2022-08-21
Simplify implicit `transmute()` in `arrange()`

hadley:


lionel- (approved):
Nice!

romainfrancois (approved):
Thanks :-)

R/arrange.R
```diff
@@ -125,47 +125,11 @@ arrange_rows <- function(data,
...
-      bullets
-    )
-    abort(bullets, call = error_call, parent = parent)
-
-  })
-
-  directions <- directions[quo_names %in% names(data)]
+  if (length(quosures) > 0) {
+    names(quosures) <- paste0("..", seq_along(quosures))
+  }
```
DavisVaughan, line 130:
```suggestion
  names(quosures) <- vec_paste0("..", seq_along(quosures))
```

Could use this if you want to

R/arrange.R
```diff
@@ -125,47 +125,11 @@ arrange_rows <- function(data,
     quosure
   })
 
-  # give the quosures arbitrary names so that
-  # data has the right number of columns below after transmute()
-  quo_names <- vec_paste0("^^--arrange_quosure_", seq_along(quosures))
-  names(quosures) <- quo_names
-
-  # TODO: not quite that because when the quosure is some expression
-  #       it should be evaluated by groups.
```
DavisVaughan, line 134: I feel like it may be important to retain this comment about the quosure being evaluated by groups. Right now the `new_data_frame()` in the `transmute()` forces evaluation on the whole data frame

R/arrange.R
hadley: Oooh, I was looking for `str_` something and couldn't find it.

R/arrange.R
hadley: The docs say:

> Unlike other dplyr verbs, arrange() largely ignores grouping; you need to explicitly mention grouping variables (or use .by_group = TRUE) in order to group by them, and functions of variables are evaluated once per data frame, not once per group.

So it's not clear that the comment is correct.

PR #6427, opened 2022-08-22
Force dots before setting up error handlers

### LINKED ISSUE #6178: Inconsistent error when not finding bangbanged function
Using dev version

See below, the use case is I don't attach dplyr, use `dplyr::` notation and forget that I should namespace `sym` too, the error with `mutate()` is especially not helpful, but there is also an inconsistency between `transmute()` and `select()` 

``` r
dplyr::mutate(iris, !!sym("Sepal.Width"))
#> Error in local_error_context(dots = dots, .index = i, mask = mask): promise already under evaluation: recursive default argument reference or earlier problems?

dplyr::select(iris, !!sym("Sepal.Width"))
#> Error in `dplyr::select()`:
#> ! could not find function "sym"

dplyr::transmute(iris, !!sym("Sepal.Width"))
#> Error in sym("Sepal.Width"): could not find function "sym"
```

If we don't use `!!` errors are helpful and consistent.

``` r
dplyr::mutate(iris, foo("Sepal.Width"))
#> Error in `dplyr::mutate()`:
#> ! Problem while computing `..1 = foo("Sepal.Width")`.
#> Caused by error in `foo()`:
#> ! could not find function "foo"

dplyr::select(iris, foo("Sepal.Width"))
#> Error in `dplyr::select()`:
#> ! could not find function "foo"

dplyr::transmute(iris, foo("Sepal.Width"))
#> Error in `dplyr::transmute()`:
#> ! Problem while computing `..1 = foo("Sepal.Width")`.
#> Caused by error in `foo()`:
#> ! could not find function "foo"
```

<sup>Created on 2022-02-09 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

lionel-:
Fixes #6178

### DIFF
--- a/R/mutate.R
+++ b/R/mutate.R
@@ -273,6 +273,9 @@ check_transmute_args <- function(..., .keep, .before, .after, error_call = calle
 }
 
 mutate_cols <- function(.data, dots, caller_env, error_call = caller_env()) {
+  # Collect dots before setting up error handlers (#6178)
+  force(dots)
+
   error_call <- dplyr_error_call(error_call)
 
   mask <- DataMask$new(.data, caller_env, "mutate", error_call = error_call)
--- a/tests/testthat/_snaps/conditions.md
+++ b/tests/testthat/_snaps/conditions.md
@@ -198,3 +198,31 @@
     Output
       `c(1, 2, 3, 4, 5)` and 2 more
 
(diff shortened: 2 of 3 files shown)


hadley:
Looks like you need a little of lifecycle warning upkeep.

hadley (approved):

PR #6429, opened 2022-08-23
Implement `.unpack` argument for `across()`

### LINKED ISSUE #6360: Automatically unpack in `across()`
To support this sort of functionality:

```R

quantile_df <- function(x, probs = c(0.25, 0.5, 0.75)) {
  tibble(
    "{{x}}_val" := quantile(x, probs),
    "{{x}}_quant" := probs
  )
}

df <- tibble(
  grp = rep(1:3, each = 10),
  x = runif(30),
  y = rnorm(30)
)

df |> 
  group_by(grp) |> 
  summarise(across(x:y, quantile_df))
```

DavisVaughan:
Closes #6360 

Some expected usage:

``` r
library(dplyr)

quantile_df <- function(x, probs = c(0.25, 0.5, 0.75)) {
  tibble(
    val = quantile(x, probs),
    quant = probs
  )
}

multilag <- function(x, lags = 1:3) {
  names(lags) <- lags
  purrr::map_dfr(lags, lag, x = x)
} 

df <- tibble(
  grp = rep(1:3, each = 10),
  x = runif(30),
  y = rnorm(30)
)

df %>%
  group_by(grp) %>%
  summarise(across(x:y, quantile_df, .unpack = TRUE))
#> `summarise()` has grouped output by 'grp'. You can override using the `.groups`
#> argument.
#> # A tibble: 9 × 5
#> # Groups:   grp [3]
#>     grp x_val x_quant   y_val y_quant
#>   <int> <dbl>   <dbl>   <dbl>   <dbl>
#> 1     1 0.152    0.25 -0.749     0.25
#> 2     1 0.248    0.5   0.291     0.5 
#> 3     1 0.499    0.75  0.534     0.75
#> 4     2 0.123    0.25  0.0954    0.25
#> 5     2 0.204    0.5   0.536     0.5 
#> 6     2 0.527    0.75  0.944     0.75
#> 7     3 0.318    0.25 -0.439     0.25
#> 8     3 0.679    0.5  -0.100     0.5 
#> 9     3 0.794    0.75  0.250     0.75

df |> 
  group_by(grp) |> 
  mutate(across(x:y, multilag, .unpack = "{outer}.{inner}"))
#> # A tibble: 30 × 9
#> # Groups:   grp [3]
#>      grp      x      y     x.1     x.2     x.3    y.1    y.2    y.3
#>    <int>  <dbl>  <dbl>   <dbl>   <dbl>   <dbl>  <dbl>  <dbl>  <dbl>
#>  1     1 0.399   0.215 NA      NA      NA      NA     NA     NA    
#>  2     1 0.275  -1.16   0.399  NA      NA       0.215 NA     NA    
#>  3     1 0.221  -0.455  0.275   0.399  NA      -1.16   0.215 NA    
#>  4     1 0.142  -0.847  0.221   0.275   0.399  -0.455 -1.16   0.215
#>  5     1 0.0893  0.387  0.142   0.221   0.275  -0.847 -0.455 -1.16 
#>  6     1 0.766   0.783  0.0893  0.142   0.221   0.387 -0.847 -0.455
#>  7     1 0.183   0.367  0.766   0.0893  0.142   0.783  0.387 -0.847
#>  8     1 0.532   1.33   0.183   0.766   0.0893  0.367  0.783  0.387
#>  9     1 0.999  -1.19   0.532   0.183   0.766   1.33   0.367  0.783
#> 10     1 0.113   0.583  0.999   0.532   0.183  -1.19   1.33   0.367
#> # … with 20 more rows
```

Some notes:

- Not adding to `c_across()`, `if_any()` or `if_all()`. I don't feel like it would be useful there?
- Defaults to `FALSE` for no unpacking
- We have the limitation that `.unpack = TRUE / glue-spec` currently turns off any opportunity for "expansion" of `across()` calls, which likely slows it down.


R/across.R
```diff
@@ -662,3 +739,47 @@ is_inlinable_formula <- function(x, mask) {
     FALSE
   }
 }
+
+df_unpack <- function(x, spec, caller_env, error_call = caller_env()) {
```
DavisVaughan, line 743: Basically `tidyr::unpack()` with support for a glue spec rather than `names_sep`

tests/testthat/test-across.R
```diff
@@ -245,6 +339,26 @@ test_that("across(.names=) can use local variables in addition to {col} and {fn}
   expect_identical(res, data.frame(MEAN_x = 42))
 })
 
+test_that("across(.unpack=) can use local variables in addition to {outer} and {inner}", {
```
DavisVaughan, line 342: I don't love this behavior, but we do it in `.names` so I've carried it over here.

I'd personally limit interpolation to just `outer` and `inner` and otherwise evaluate in an empty environment

hadley (approved):

R/across.R
```diff
@@ -534,6 +599,18 @@ expand_across <- function(quo) {
   dplyr_mask <- peek_mask()
   mask <- dplyr_mask$get_rlang_mask()
 
+  if (".unpack" %in% names(expr)) {
+    unpack <- eval_tidy(expr$.unpack, mask, env = env)
```
lionel-, line 603:
```suggestion
    unpack <- eval_bare(expr$.unpack, env)
```

R/across.R
```diff
@@ -534,6 +599,18 @@ expand_across <- function(quo) {
...
+  } else {
+    unpack <- FALSE
+  }
+
+  # Abort expansion if unpacking as expansion makes named expressions and we
+  # need the expressions to remain unnamed
+  if (!is_false(unpack)) {
```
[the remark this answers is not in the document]
lionel-: We only know the outer names. We don't know what functions returns a df, or even the shape or inner names they would return, so it's not possible to generate individual expressions for the calling verb ahead of time.

lionel- (approved):

R/across.R
DavisVaughan: Do you think `.name` evaluation should also use `eval_bare()` then? I assume it is important for `.fns`?

https://github.com/tidyverse/dplyr/blob/73ffdbf29bdb5763b2788c7b41c03a4813c929a8/R/across.R#L548-L549

lionel-: oh hmm.... I guess this is emulating the fact that `across()` is run e.g. within `mutate()`, so evaluation is masked. You can scratch the commit suggestion above, your change was correct in the first place. Maybe add a comment about why we're evaluating in the mask?

R/across.R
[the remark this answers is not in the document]
DavisVaughan: To expand on what Lionel is saying (at least how I see it), in a case like this:

```r
mutate(across(x, list(lag = multilag, double = ~.x * 2, .unpack = "{inner}"))
```

we'd have to somehow know to expand to `mutate(multilag(x), x_double = x * 2)`, i.e. a mix of named and unnamed expansions. But that depends on the return value of the function (which one returns a df?), which you can't know ahead of time

---

Using `"{inner}"` also probably makes little sense when there is >1 column you are iterating across, because it likely results in duplicate names (which I currently have as an error). And with 1 column, you should just call it without `across()` and let normal unpacking kick in.

```r
library(dplyr)

multilag <- function(x, lags = 1:3) {
  names(lags) <- lags
  purrr::map_dfr(lags, lag, x = x)
} 

df <- tibble(
  x = runif(3),
  y = rnorm(3)
)

# With multiple columns it would likely give duplicate names and right now
# that is an error
df %>%
  mutate(across(x:y, multilag, .unpack = "{inner}"))
#> Error in `mutate()`:
#> ! Problem while computing `..1 = across(x:y, multilag, .unpack =
#>   "{inner}")`.
#> Caused by error in `across()`:
#> ! Names must be unique.
#> ✖ These names are duplicated:
#>   * "1" at locations 1 and 4.
#>   * "2" at locations 2 and 5.
#>   * "3" at locations 3 and 6.

# With one column it is the same as just calling `multilag()` without `across()`
df %>%
  mutate(across(x, multilag, .unpack = "{inner}"))
#> # A tibble: 3 × 5
#>        x      y    `1`    `2`   `3`
#>    <dbl>  <dbl>  <dbl>  <dbl> <dbl>
#> 1 0.277  -0.735 NA     NA        NA
#> 2 0.546  -0.106  0.277 NA        NA
#> 3 0.0435 -1.09   0.546  0.277    NA

df %>%
  mutate(multilag(x))
#> # A tibble: 3 × 5
#>        x      y    `1`    `2`   `3`
#>    <dbl>  <dbl>  <dbl>  <dbl> <dbl>
#> 1 0.277  -0.735 NA     NA        NA
#> 2 0.546  -0.106  0.277 NA        NA
#> 3 0.0435 -1.09   0.546  0.277    NA
```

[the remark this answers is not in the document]
hadley: Ok, I was just thinking of my original `quantile_df()` that produced unique var names; but we're assuming people won't need to do that with `unpack = TRUE`.

PR #6433, opened 2022-08-25
Supersede `recode()` and `recode_factor()` in favor of `case_match()`

DavisVaughan:
@hadley I didn't know if you wanted to soft-deprecate or just signal them as superseded, so I went with superseded like you did with `transmute()`.

I used the example section as a way to show how to transition to `case_match()` or `fct_recode()` and to point out the gap that the upcoming forcats function will fill


R/recode.R
```diff
@@ -54,66 +54,61 @@
...
 #' @export
 #' @examples
-#' # For character values, recode values with named arguments only. Unmatched
-#' # values are unchanged.
 #' char_vec <- sample(c("a", "b", "c"), 10, replace = TRUE)
-#' recode(char_vec, a = "Apple")
+#'
+#' # `recode()` is superseded by `case_match()`
 #' recode(char_vec, a = "Apple", b = "Banana")
+#' case_match(char_vec, "a" ~ "Apple", "b" ~ "Banana", .default = char_vec)
```
DavisVaughan, line 61: Now that I've written this, I do get the vibe that `replace_match()` and `replace_when()` might be useful as even more direct replacements for `recode()`, but I still can't decide if adding two more functions to dplyr is worth it or not

tests/testthat/test-recode.R
```diff
@@ -90,6 +90,9 @@ test_that(".dot argument works correctly (PR #2110)", {
   )
 })
 
+test_that("`recode()` signals that it is superseded", {
+  expect_snapshot(catch_cnd(recode("a", a = "A")))
```
DavisVaughan, line 94: I've never done this before, but it seemed like a useful way to test this?

NEWS.md
```diff
@@ -1,5 +1,10 @@
 # dplyr (development version)
 
+* `recode()` is superseded in favor of `case_match()`. `recode_factor()` is
```
hadley, line 3: I think we could deprecate since `recode()` is more obviously wrong, and we've marked as questioning for some time.

hadley: Oh I guess I've thought of it as questioning, but we never actually marked it as such 😬

R/recode.R
```diff
@@ -1,23 +1,23 @@
...
-#' numeric, character, and factors. For logical vectors, use [if_else()]. For
-#' more complicated criteria, use [case_when()].
+#' `r lifecycle::badge("superseded")`
 #'
-#' You can use `recode()` directly with factors; it will preserve the existing
-#' order of levels while changing the values. Alternatively, you can
-#' use `recode_factor()`, which will change the order of levels to match
-#' the order of replacements. See the [forcats](https://forcats.tidyverse.org/)
-#' package for more tools for working with factors and their levels.
+#' `recode()` is superseded in favor of [case_match()], which handles the most
```
hadley, line 6: I think you should also mention the major problem with `recode()` is that it has `old = new` which is the opposite pattern to `new = old` which we see almost everywhere else.

R/recode.R
```diff
@@ -54,66 +54,61 @@
...
+#'   num_vec,
+#'   1 ~ "z",
+#'   2 ~ "y",
+#'   3 ~ "x",
+#'   NA ~ "M",
+#'   .default = "D",
+#'   .ptype = factor(levels = c("z", "y", "x", "D", "M"))
+#' )
 recode <- function(.x, ..., .default = NULL, .missing = NULL) {
+  lifecycle::signal_stage("superseded", "recode()", "case_match()")
```
hadley, line 111: Maybe add a comment with the dplyr version?

tests/testthat/test-recode.R
hadley: Yeah, seems fine to me.

hadley (approved):

R/recode.R
DavisVaughan: But `case_when()` and `case_match()` are also `old ~ new` 😬. I think they work that way because that is how it works in SQL. It is also how `switch()` works

hadley: Yeah, but it also somehow feels different because of the matching context.

PR #6438, opened 2022-08-30
Extract `mutate_col()` from `mutate_cols()`

lionel-:
This PR extracts the body of the `mutate_cols()` loop into a separate function. This simplifies the `withCallingHandlers()` call and it is easier to see what's going on in the handled expression and in the condition handlers. Also reduces nesting/indentation inside the body, making it easier to read.


R/mutate.R
```diff
@@ -230,199 +230,204 @@ mutate_cols <- function(.data, dots, caller_env, error_call = caller_env()) {
...
 
-  rows <- mask$get_rows()
-
   new_columns <- set_names(list(), character())
 
-  withCallingHandlers({
+  withCallingHandlers(
     for (i in seq_along(dots)) {
       context_poke("column", old_current_column)
+      new_columns <- mutate_col(i, .data, dots, mask, new_columns)
```
DavisVaughan, line 238: I suppose you could do

```r
dot <- dots[[I]]
new_columns <- mutate_col(.data, dot, mask, new_columns)
```

That would really narrow the scope (i.e. `mutate_col()` doesn't need to know which iteration it is on, and it doesn't need all the `dots`, only the one it currently has to process)

It seems like you didn't change much code while extracting this out, which I appreciate because it made it easier to review 😄 but this might be worth it

DavisVaughan (approved):
Looks great, will probably make debugging in the IDE a little easier too

R/mutate.R
lionel-: oops good catch, somehow I thought the function depended on its iteration number.

PR #6443, opened 2022-09-01
Collect warnings and add `dplyr_last_warnings()`

### LINKED ISSUE #6005: Unenhanced warnings are emitted from `mutate()` when run in a `reprex()`
When run with `reprex::reprex()`:

``` r
library(dplyr)

foo <- function() {
  warning("foo")
  NA
}

df <- tibble(x = 1)

df <- mutate(df, x = foo())
#> Warning in foo(): foo
```

vs when run interactively:

```r
df <- mutate(df, x = foo())
#> Warning message:
#> Problem with `mutate()` column `x`.
#> ℹ `x = foo()`.
#> ℹ foo 
```

I think this is related to the changes from https://github.com/tidyverse/dplyr/pull/5686, in particular it is about the `maybe_restart()` bit here that occurs when the warning is muffled

https://github.com/tidyverse/dplyr/blob/87d1cecd7285f935ee3a202af0f74b9c1a640f7d/R/mutate.R#L427-L446

I guess in the stack of `reprex()`, someone is muffling the warnings as it captures them to be able to print them out later on?

Since it was slow to throw the enhanced warnings, maybe this is the best we can do, but it wasn't mentioned in #5686 as a potential issue so I figured I'd bring it up in case @lionel- has any thoughts here

### LINKED ISSUE #6236: `rowwise()` with a mutate() call with warnings is very slow
As identified by @debruine here https://github.com/tidyverse/dplyr/issues/6236#issuecomment-1103972470 this issue is actually caused when mutate() is used after rowwise() and generates warnings.

This code takes ~11 seconds to run in a clean R session

```r
library(dplyr)
library(tictoc)
tic()
foo <- tibble(
  id = 1:800
) %>% 
  rowwise() %>% 
  mutate(group_b_mean = mean(NULL))
toc()
# 11.584 sec elapsed
```

Please note that weirdly the same code takes <1second when run with the {reprex} addin.

Many thanks to debruine for figuring this out.

lionel-:
Branched from #6438.

Fixes #6005
Fixes #6236

* Warnings within mutate steps are now collected instead of being enriched right away.
* The new function `dplyr_last_warnings()` returns enriched warnings, the first 20 by default.
* Enriched warnings are now chained so as to keep a reference to the verb within which they were emitted.

```r
f <- function(x) {
  warning("foo")
  x
}

df <- tibble(id = 1:100)

df |>
  rowwise() |>
  mutate(x = f(1))
#> # A tibble: 100 × 2
#> # Rowwise:
#>      id     x
#>   <int> <dbl>
#> 1     1     1
#> 2     2     1
#> 3     3     1
#> 4     4     1
#> # … with 96 more rows
#> # ℹ Use `print(n = ...)` to see more rows
#> Warning message:
#> There were 100 warnings in `mutate()`.
#> ℹ Run `dplyr::dplyr_last_warnings()` to see them.

dplyr_last_warnings(3)
#> [[1]]
#> <warning/rlang_warning>
#> Warning in `mutate()`:
#> Problem in row 1 while computing `x = f(1)`.
#> Caused by warning in `f()`:
#> ! foo
#>
#> [[2]]
#> <warning/rlang_warning>
#> Warning in `mutate()`:
#> Problem in row 2 while computing `x = f(1)`.
#> Caused by warning in `f()`:
#> ! foo
#>
#> [[3]]
#> <warning/rlang_warning>
#> Warning in `mutate()`:
#> Problem in row 3 while computing `x = f(1)`.
#> Caused by warning in `f()`:
#> ! foo
#>
#> ... with 97 more warnings.
#> ℹ Use `dplyr_last_warnings(n = ...)` to show more.
```

The other way we could handle long lists of warnings is by detecting if there are more than 10 or 20. If that's the case, suggest to run `head(dplyr_last_warnings())`, otherwise just `dplyr_last_warnings()`. On hindsight I think that would make things simpler internally and for users.


DESCRIPTION
```diff
@@ -30,6 +30,7 @@ BugReports: https://github.com/tidyverse/dplyr/issues
 Depends: 
     R (>= 3.4.0)
 Imports: 
+    cli (>= 3.3.0),
```
lionel-, line 33: This PR uses cli right away to use features like pluralisation. Also I think it's better to write new code with cli right away since porting existing code is so tricky.

R/aaa.R
```diff
@@ -0,0 +1,4 @@
+on_load({
+  # Singletons
+  the <- new_environment()
```
lionel-, line 3: This introduces Jim's pattern for singletons in dplyr.

R/conditions.R
```diff
@@ -95,6 +90,26 @@ peek_error_context <- function() {
   context_peek("dplyr_error_context", "peek_error_context", "dplyr error handling")
 }
 
+dots_expr <- function(dots, i) {
```
lionel-, line 93: I didn't end up using this function elsewhere in the end, but it feels like it's better as its own separate function.

R/conditions.R
```diff
@@ -161,3 +176,99 @@ skip_internal_condition <- function(cnd) {
...
+  if (remaining) {
+    cli::cli_bullets(c(
+      "... with {remaining} more warning{?s}.",
+      "i" = "Use {.code dplyr_last_warnings(n = ...)} to show more."
+    ))
+  }
+}
+
+# rlang should export this routine
+error_call <- function(call) {
```
lionel-, line 269: This is a hack to get the result of `rlang:::error_call()`, which has special behaviour compared to `frame_call()`, for instance with S3 methods.

Tracked in https://github.com/r-lib/rlang/issues/1474.

R/aaa.R
hadley: I thought we just did this at build time? And in `dplyr.R`?

R/context.R
```diff
@@ -96,15 +96,52 @@ group_labels_details <- function(keys) {
...
+cur_group_label <- function(type = mask_type(),
+                            id = cur_group_id(),
+                            group = cur_group()) {
+  switch(
+    type,
+    ungrouped = "",
+    grouped = glue(
+      "group {id}: {label}",
+      label = group_labels_details(group)
+    ),
```
hadley, line 108:
```suggestion
    grouped = glue("group {id}: {label}", label = group_labels_details(group)),
```

R/context.R
```diff
@@ -96,15 +96,52 @@ group_labels_details <- function(keys) {
...
+  switch(
+    mask_type,
+    ungrouped = list(),
+    grouped = list(
+      id = cur_group_id(),
+      group = cur_group()
+    ),
+    rowwise = list(
+      id = cur_group_id()
+    ),
```
hadley, line 124:
```suggestion
    grouped = list(id = cur_group_id(), group = cur_group()),
    rowwise = list(id = cur_group_id()),
```

I feel like this makes it easier to skim the `switch()`

hadley (commented):
Also need to do the same thing with summarise? e.g.

```R

df <- tibble(id = 1:100, x = NA)
df |> 
  group_by(id) |> 
  summarise(y = min(x, na.rm = TRUE))
```

hadley:
Playing around with this a bit, I think it's probably worth showing maybe 1 warning by default? i.e. instead of:

```R
#> Warning message:
#> There were 2 warnings in `mutate()`.
#> ℹ Run `dplyr::dplyr_last_warnings()` to see them. 
```

Something like:

```R
#> Warning message:
#> There were warnings in `mutate()`.
#> The first warning was:
...
#> ℹ Run `dplyr::dplyr_last_warnings()` to see the remaining 1 warning. 
```

lionel-:
> Also need to do the same thing with summarise?

We don't currently enrich warnings in summarise, but it should be easy to add. I guess we should make sure all grouped evals collect warnings? This seems better scoped for a second PR though.

> Playing around with this a bit, I think it's probably worth showing maybe 1 warning by default

Good idea.

lionel-:
With the last commit:

- If there is only one warning, just emit it enriched.
- If there are multiple warnings, just show the first one unenriched (because we already mention the call in the outer warning).

```r
f <- function(x) {
  warning("foo")
  x
}
df <- tibble(id = 1:3)

df |>
   group_by(x = f(1):n()) |>
   mutate(x = f(1))
#> # A tibble: 3 × 2
#> # Groups:   x [1]
#>      id     x
#>   <int> <dbl>
#> 1     1     1
#> 2     2     1
#> 3     3     1
#> Warning messages:
#> 1: In group_by() : Problem while computing `x = f(1):n()`.
#> Caused by warning in `f()`:
#> ! foo
#> 2: There were 3 warnings in `mutate()`.
#> The first warning was:
#> ! foo
#> ℹ Run `dplyr::dplyr_last_warnings()` to see the 2 remaining warnings.
```

This is a bit messy because of the way our warning conditions interact with base display, but that's not too bad.

hadley:
The display of the one and two warning case feels rather different:

Screen Shot 2022-09-01 at 14 37 17

Can we make them more similar? Maybe the 1 warning case could just be:

```
Warning message:
There was 1 warning in `mutate()`.
! foo
```

i.e. drop the "the first warning" and "run ..." lines.

How hard would it be to make `dplyr_last_warnings()` show the warning backtraces?

lionel-:
@hadley Much better, thanks!

The problem with capturing backtraces is that it's very costly. We could do it for the first 5 warnings though. I've also changed:

```
in `group_by()`.
```

to

```
in a `group_by()` step.
```

Just as a reminder that there could be multiple group-by steps to look for. Is this helpful?

```r
df |>
  rowwise() |>
  mutate(x = f(1))
#> # A tibble: 100 × 2
#> # Rowwise:
#>      id     x
#>   <int> <dbl>
#> 1     1     1
#> 2     2     1
#> 3     3     1
#> 4     4     1
#> # … with 96 more rows
#> # ℹ Use `print(n = ...)` to see more rows
#> Warning message:
#> There were 100 warnings in a `mutate()` step.
#> The first warning was:
#> ! foo
#> ℹ Run `dplyr::dplyr_last_warnings()` to see the 99 remaining warnings.
```

```r
df |>
  group_by(x = f(1):n()) |>
  mutate(x = f(1))
#> # A tibble: 3 × 2
#> # Groups:   id [3]
#>      id     x
#>   <int> <dbl>
#> 1     1     1
#> 2     2     1
#> 3     3     1
#> Warning messages:
#> 1: There was 1 warning in a `group_by()` step.
#> ! foo
#> 2: There were 3 warnings in a `mutate()` step.
#> The first warning was:
#> ! foo
#> ℹ Run `dplyr::dplyr_last_warnings()` to see the 2 remaining warnings.
```

lionel-:
With the last commit, we now entrace the first 5 warnings. We also only display 5 warnings by default instead of 20, so that:

- It's less confusing why some warnings have backtraces and others don't.
- We show less output, since backtraces take up so much space.
- The interesting output (with backtraces) is at the end near the prompt.

The display opts into the new tree backtrace style with grayed out inner parts.

```r
df <- tibble(id = 1:2)

f <- function(...) g(...)
g <- function(...) h(...)
h <- function(x, base = TRUE) {
  if (base) {
    warning("foo")
  } else {
    warn("foo")
  }
  x
}

foo <- function() bar()
bar <- function() {
  df |>
    group_by(x = f(1):n()) |>
    mutate(x = f(1, base = FALSE))
}

foo()
#> # A tibble: 2 × 2
#> # Groups:   x [1]
#>      id     x
#>   <int> <dbl>
#> 1     1     1
#> 2     2     1
#> Warning messages:
#> 1: There was 1 warning in a `group_by()` step.
#> ! foo
#> 2: There were 2 warnings in a `mutate()` step.
#> The first warning was:
#> ! foo
#> ℹ Run `dplyr::dplyr_last_warnings()` to see the 1 remaining warning.

dplyr_last_warnings()
#> [[1]]
#> <warning/rlang_warning>
#> Warning in `group_by()`:
#> Problem while computing `x = f(1):n()`.
#> Caused by warning in `h()`:
#> ! foo
#> ---
#> Backtrace:
#>      ▆
#>   1. ├─utils (local) .ess.command(...)
#>   2. │ ├─base::withCallingHandlers(interrupt = restart, withVisible(expr))
#>   3. │ └─base::withVisible(expr)
#>   4. ├─base::local(...)
#>   5. │ └─base::eval.parent(substitute(eval(quote(expr), envir)))
#>   6. │   └─base::eval(expr, p)
#>   7. │     └─base::eval(expr, p)
#>   8. ├─base::eval(...)
#>   9. │ └─base::eval(...)
#>  10. │   └─global foo()
#>  11. │     └─global bar()
#>  12. │       ├─dplyr::mutate(group_by(df, x = f(1):n()), x = f(1, base = FALSE))
#>  13. │       ├─dplyr::group_by(df, x = f(1):n())
#>  14. │       └─dplyr:::group_by.data.frame(df, x = f(1):n()) at dplyr/R/group-by.r:135:2
#>  15. │         └─dplyr::group_by_prepare(...) at dplyr/R/group-by.r:140:2
#>  16. │           └─dplyr:::add_computed_columns(...) at dplyr/R/group-by.r:215:2
#>  17. │             └─dplyr:::mutate_cols(...) at dplyr/R/group-by.r:253:6
#>  18. │               ├─base::withCallingHandlers(...) at dplyr/R/mutate.R:239:2
#>  19. │               └─(function() {... at dplyr/R/mutate.R:239:2
#>  20. │                 └─dplyr:::mutate_col(i, .data, dots, mask, new_columns) at dplyr/R/mutate.R:244:6
#>  21. │                   └─mask$eval_all_mutate(quo) at dplyr/R/mutate.R:371:8
#>  22. │                     └─dplyr (local) eval() at dplyr/R/data-mask.R:69:6
#>  23. └─global f(1)
#>  24.   └─global g(...)
#>  25.     └─global h(...)
#>  26.       └─base::warning("foo")
#>
#> [[2]]
#> <warning/rlang_warning>
#> Warning in `mutate()`:
#> Problem in group 1: x = 1 while computing `x = f(1, base = FALSE)`.
#> Caused by warning:
#> ! foo
#> ---
#> Backtrace:
#>      ▆
#>   1. ├─utils (local) .ess.command(...)
#>   2. │ ├─base::withCallingHandlers(interrupt = restart, withVisible(expr))
#>   3. │ └─base::withVisible(expr)
#>   4. ├─base::local(...)
#>   5. │ └─base::eval.parent(substitute(eval(quote(expr), envir)))
#>   6. │   └─base::eval(expr, p)
#>   7. │     └─base::eval(expr, p)
#>   8. ├─base::eval(...)
#>   9. │ └─base::eval(...)
#>  10. │   └─global foo()
#>  11. │     └─global bar()
#>  12. │       ├─dplyr::mutate(group_by(df, x = f(1):n()), x = f(1, base = FALSE))
#>  13. │       └─dplyr:::mutate.data.frame(...) at dplyr/R/mutate.R:146:2
#>  14. │         └─dplyr:::mutate_cols(.data, dplyr_quosures(...), caller_env = caller_env()) at dplyr/R/mutate.R:175:2
#>  15. │           ├─base::withCallingHandlers(...) at dplyr/R/mutate.R:239:2
#>  16. │           └─(function() {... at dplyr/R/mutate.R:239:2
#>  17. │             └─dplyr:::mutate_col(i, .data, dots, mask, new_columns) at dplyr/R/mutate.R:244:6
#>  18. │               └─mask$eval_all_mutate(quo) at dplyr/R/mutate.R:371:8
#>  19. │                 └─dplyr (local) eval() at dplyr/R/data-mask.R:69:6
#>  20. └─global f(1, base = FALSE)
#>  21.   └─global g(...)
#>  22.     └─global h(...)
#>  23.       └─rlang::warn("foo")
#>
#> [[3]]
#> <warning/rlang_warning>
#> Warning in `mutate()`:
#> Problem in group 2: x = 2 while computing `x = f(1, base = FALSE)`.
#> Caused by warning:
#> ! foo
#> ---
#> Backtrace:
#>      ▆
#>   1. ├─utils (local) .ess.command(...)
#>   2. │ ├─base::withCallingHandlers(interrupt = restart, withVisible(expr))
#>   3. │ └─base::withVisible(expr)
#>   4. ├─base::local(...)
#>   5. │ └─base::eval.parent(substitute(eval(quote(expr), envir)))
#>   6. │   └─base::eval(expr, p)
#>   7. │     └─base::eval(expr, p)
#>   8. ├─base::eval(...)
#>   9. │ └─base::eval(...)
#>  10. │   └─global foo()
#>  11. │     └─global bar()
#>  12. │       ├─dplyr::mutate(group_by(df, x = f(1):n()), x = f(1, base = FALSE))
#>  13. │       └─dplyr:::mutate.data.frame(...) at dplyr/R/mutate.R:146:2
#>  14. │         └─dplyr:::mutate_cols(.data, dplyr_quosures(...), caller_env = caller_env()) at dplyr/R/mutate.R:175:2
#>  15. │           ├─base::withCallingHandlers(...) at dplyr/R/mutate.R:239:2
#>  16. │           └─(function() {... at dplyr/R/mutate.R:239:2
#>  17. │             └─dplyr:::mutate_col(i, .data, dots, mask, new_columns) at dplyr/R/mutate.R:244:6
#>  18. │               └─mask$eval_all_mutate(quo) at dplyr/R/mutate.R:371:8
#>  19. │                 └─dplyr (local) eval() at dplyr/R/data-mask.R:69:6
#>  20. └─global f(1, base = FALSE)
#>  21.   └─global g(...)
#>  22.     └─global h(...)
#>  23.       └─rlang::warn("foo")
```

lionel-:
I've also renamed `dplyr_last_warnings()` to `last_dplyr_warnings()`. This namespaces it inside `last_` for autocompletion.

R/conditions.R
```diff
@@ -161,3 +176,129 @@ skip_internal_condition <- function(cnd) {
...
+  warnings <- map(warnings, new_dplyr_warning)
+
+  structure(
+    warnings,
+    class = c("last_dplyr_warnings", "list"),
+    remaining = remaining
+  )
+}
+
+the$last_warnings <- list()
```
hadley, line 213: Oh, I would've put these in `dplyr.R` too — then you don't need `aaa.R` and all the elements of `the` are defined in one place. But I can also see the advantages of defined here. Although I can also imaging only setting these on load.

hadley (approved):

R/conditions.R
lionel-: There are advantages to defining everything in the same place, and to defining things in context. I tend to favour the latter though.

hadley: Then you could use `on_load()` to avoid the need for `aaa.R`?

lionel-: This doesn't work with `load_all()`.

Not sure if it would work with R CMD build. Even if it did, it would mean that you can't populate `the` on load in files that come prior to `dplyr.r` because of the order in which the onload hook is populated.

I thought that adding an `aaa.R` file is pretty standard. Does it pose a problem?

hadley: I think I didn't explain myself clearly. I meant put `the <- new.env()` in `dplyr.R` . Then in this file, write:

```R
on_load({
  the$last_warnings <- list()
  ...
})
```

That should work, right?

lionel-: oh gotcha, yes it would work. I'd be fine with this scheme too.

NEWS.md
```diff
@@ -1,5 +1,12 @@
...
+* Warnings emitted inside `mutate()` and variants are now collected and stashed
+  away. Run the new `last_dplyr_warnings()` function to see the warnings emitted
+  within dplyr verbs during the last top-level command.
+
+  This fixes performance issues when throusands of warnings are emitted with
+  rowwise and grouped data frames (#6236).
+
 * `transmute()` is superseded in favour of `mutate(keep = "none")`
 
 * `recode()` is superseded in favor of `case_match()`. `recode_factor()` is
```
DavisVaughan, line 12: I think it is a bit of a shame that the 1 warning case (and when showing the "first warning" when there are multiple) doesn't show the expression that the warning originated from. I think that is very useful information when there are multiple expressions in the same `mutate()` call and you don't know where the problem came from

```r
mtcars %>%
  mutate(mpg2 = sqrt(-mpg), disp2 = disp + 5)

# Main
Warning message:
Problem while computing `mpg2 = sqrt(-mpg)`.
ℹ NaNs produced 

# This PR
Warning message:
There was 1 warning in a `mutate()` step.
! NaNs produced 
```

R/conditions.R
DavisVaughan: Slightly simpler as `dot_expr()`?

```r
dot_expr <- function(expr) {
  if (quo_is_call(expr, "invisible")) {
    ""
  } else {
    quo_as_label(expr)
  }
}
```

NEWS.md
```diff
@@ -1,5 +1,12 @@
 # dplyr (development version)
 
+* Warnings emitted inside `mutate()` and variants are now collected and stashed
+  away. Run the new `last_dplyr_warnings()` function to see the warnings emitted
+  within dplyr verbs during the last top-level command.
+
+  This fixes performance issues when throusands of warnings are emitted with
```
DavisVaughan, line 7:
```suggestion
  This fixes performance issues when thousands of warnings are emitted with
```

NEWS.md
```diff
@@ -1,5 +1,12 @@
...
+  rowwise and grouped data frames (#6236).
```
DavisVaughan, line 8:
```suggestion
  rowwise and grouped data frames (#6005, #6236).
```

R/conditions.R
```diff
@@ -95,6 +90,26 @@ peek_error_context <- function() {
...
+ expr <- dots[[i]]
+  if (quo_is_call(expr, "invisible")) {
+    ""
+  } else {
+    quo_as_label(dots[[i]])
+  }
+}
+
+mask_type <- function(mask = peek_mask()) {
+  if (mask$get_size() > 0) {
```
DavisVaughan, line 103: What's wrong with zero row grouped / rowwise data frames?

R/conditions.R
```diff
@@ -161,3 +176,129 @@ skip_internal_condition <- function(cnd) {
     cnd
```
DavisVaughan, line 176: Neat that this stashes all 4 warnings, but it is a _little_ hard to know which sets of warnings came from which expressions in `fn()`. Not sure if there is anything to do, just wanted to mention it.

```r
library(dplyr)

fn <- function() {
  throws_warning <- mtcars %>%
    group_by(vs) %>%
    mutate(mpg = sqrt(-mpg))
  
  throws_warning2 <- mtcars %>%
    group_by(vs) %>%
    mutate(mpg = sqrt(-mpg))
}

fn()
```

R/conditions.R
```diff
@@ -161,3 +176,129 @@ skip_internal_condition <- function(cnd) {
...
+  push_dplyr_warnings(warnings)
+
+  first <- warnings[[1]][["cnd"]]
+  call <- format_error_call(error_call)
+
+  cli::cli_warn(c(
+    "There {cli::qty(n)} {?was/were} {n} warning{?s} in a {call} step.",
+    if (n > 1) "The first warning was:",
+    cnd_message_lines(first),
+    i = if (n > 1) "Run `dplyr::last_dplyr_warnings()` to see the {n - 1} remaining warning{?s}."
```
DavisVaughan, line 249:
```suggestion
    i = if (n > 1) "Run `dplyr::last_dplyr_warnings()` to see the {n - 1} remaining warning{?s}."
```

I WANT TO BE ABLE TO CLICK THIS `dplyr::last_dplyr_warnings()`

R/conditions.R
```diff
@@ -161,3 +176,129 @@ skip_internal_condition <- function(cnd) {
...
+#' with warnings. To see the original warnings, use
+#' `last_dplyr_warnings()`.
+#'
+#' @param n Passed to [head()] so that only the first `n` warnings are
+#'   displayed.
+#' @export
+last_dplyr_warnings <- function(n = 5) {
+  if (!identical(n, Inf)) {
+    check_number(n)
+    stopifnot(n >= 0)
```
DavisVaughan, line 197: I feel like `check_number()` should have a `positive` argument (maybe allowing for `0` optionally too) because we have used something like this a few times now. See also the `n` arg of `dplyr:::lead()`

R/conditions.R
```diff
@@ -161,3 +176,129 @@ skip_internal_condition <- function(cnd) {
...
+  local_options(
+    "rlang:::trace_display_tree" = TRUE
+  )
+  print(unstructure(x), ..., simplify = "none")
+
+  remaining <- attr(x, "remaining")
+  if (remaining) {
+    cli::cli_bullets(c(
+      "... with {remaining} more warning{?s}.",
+      "i" = "Use {.code last_dplyr_warnings(n = ...)} to show more."
```
DavisVaughan, line 285: You could use a clickable `last_dplyr_warnings(n = Inf)` and say

``Use `last_dplyr_warnings(n = Inf)` to show all warnings``

I think people will be able to intuit that they can put a different number there, but I like that this would be clickable code

R/conditions.R
```diff
@@ -161,3 +176,129 @@ skip_internal_condition <- function(cnd) {
...
+  )
+}
+
+cnd_message_lines <- function(cnd, ...) {
+  c(
+    "!" = cnd_header(cnd, ...),
+    cnd_body(cnd, ...),
+    cnd_footer(cnd, ...)
+  )
+}
```
DavisVaughan, line 304: Note:

I actually don't like the backtraces being emitted automatically when you call `dplyr_last_warnings()`, but I do like the idea of capturing a few to debug with.

I think my issue with them is that it makes scanning across multiple warnings difficult because the big backtraces get in the way.

What if you captured the first 5, but didn't show them unless the user did something like `dplyr_last_warnings(trace = TRUE)` (which you could also potentially provide advice for)?

DavisVaughan: Like, there is a lot of output here if I am just trying to quickly scan the 5 warning messages

``` r
library(dplyr)

fn <- function() {
  throws_warning <- mtcars %>%
    group_by(mpg) %>%
    mutate(mpg = sqrt(-mpg))
}

fn()
#> Warning: There were 25 warnings in a `mutate()` step.
#> The first warning was:
#> ! NaNs produced
#> ℹ Run `dplyr::last_dplyr_warnings()` to see the 24 remaining
#>   warnings.

last_dplyr_warnings()
#> [[1]]
#> <warning/rlang_warning>
#> Warning in `mutate()`:
#> Problem in group 1: mpg = 10.4 while computing `mpg =
#> sqrt(-mpg)`.
#> Caused by warning in `sqrt()`:
#> ! NaNs produced
#> ---
#> Backtrace:
#>      ▆
#>   1. ├─global fn()
#>   2. │ └─mtcars %>% group_by(mpg) %>% mutate(mpg = sqrt(-mpg))
#>   3. ├─dplyr::mutate(., mpg = sqrt(-mpg))
#>   4. └─dplyr:::mutate.data.frame(., mpg = sqrt(-mpg)) at dplyr/R/mutate.R:146:2
#>   5.   └─dplyr:::mutate_cols(.data, dplyr_quosures(...), caller_env = caller_env()) at dplyr/R/mutate.R:175:2
#>   6.     ├─base::withCallingHandlers(...) at dplyr/R/mutate.R:239:2
#>   7.     └─(function() {... at dplyr/R/mutate.R:239:2
#>   8.       └─dplyr:::mutate_col(i, .data, dots, mask, new_columns) at dplyr/R/mutate.R:244:6
#>   9.         └─mask$eval_all_mutate(quo) at dplyr/R/mutate.R:371:8
#>  10.           └─dplyr (local) eval() at dplyr/R/data-mask.R:69:6
#> 
#> [[2]]
#> <warning/rlang_warning>
#> Warning in `mutate()`:
#> Problem in group 2: mpg = 13.3 while computing `mpg =
#> sqrt(-mpg)`.
#> Caused by warning in `sqrt()`:
#> ! NaNs produced
#> ---
#> Backtrace:
#>      ▆
#>   1. ├─global fn()
#>   2. │ └─mtcars %>% group_by(mpg) %>% mutate(mpg = sqrt(-mpg))
#>   3. ├─dplyr::mutate(., mpg = sqrt(-mpg))
#>   4. └─dplyr:::mutate.data.frame(., mpg = sqrt(-mpg)) at dplyr/R/mutate.R:146:2
#>   5.   └─dplyr:::mutate_cols(.data, dplyr_quosures(...), caller_env = caller_env()) at dplyr/R/mutate.R:175:2
#>   6.     ├─base::withCallingHandlers(...) at dplyr/R/mutate.R:239:2
#>   7.     └─(function() {... at dplyr/R/mutate.R:239:2
#>   8.       └─dplyr:::mutate_col(i, .data, dots, mask, new_columns) at dplyr/R/mutate.R:244:6
#>   9.         └─mask$eval_all_mutate(quo) at dplyr/R/mutate.R:371:8
#>  10.           └─dplyr (local) eval() at dplyr/R/data-mask.R:69:6
#> 
#> [[3]]
#> <warning/rlang_warning>
#> Warning in `mutate()`:
#> Problem in group 3: mpg = 14.3 while computing `mpg =
#> sqrt(-mpg)`.
#> Caused by warning in `sqrt()`:
#> ! NaNs produced
#> ---
#> Backtrace:
#>      ▆
#>   1. ├─global fn()
#>   2. │ └─mtcars %>% group_by(mpg) %>% mutate(mpg = sqrt(-mpg))
#>   3. ├─dplyr::mutate(., mpg = sqrt(-mpg))
#>   4. └─dplyr:::mutate.data.frame(., mpg = sqrt(-mpg)) at dplyr/R/mutate.R:146:2
#>   5.   └─dplyr:::mutate_cols(.data, dplyr_quosures(...), caller_env = caller_env()) at dplyr/R/mutate.R:175:2
#>   6.     ├─base::withCallingHandlers(...) at dplyr/R/mutate.R:239:2
#>   7.     └─(function() {... at dplyr/R/mutate.R:239:2
#>   8.       └─dplyr:::mutate_col(i, .data, dots, mask, new_columns) at dplyr/R/mutate.R:244:6
#>   9.         └─mask$eval_all_mutate(quo) at dplyr/R/mutate.R:371:8
#>  10.           └─dplyr (local) eval() at dplyr/R/data-mask.R:69:6
#> 
#> [[4]]
#> <warning/rlang_warning>
#> Warning in `mutate()`:
#> Problem in group 4: mpg = 14.7 while computing `mpg =
#> sqrt(-mpg)`.
#> Caused by warning in `sqrt()`:
#> ! NaNs produced
#> ---
#> Backtrace:
#>      ▆
#>   1. ├─global fn()
#>   2. │ └─mtcars %>% group_by(mpg) %>% mutate(mpg = sqrt(-mpg))
#>   3. ├─dplyr::mutate(., mpg = sqrt(-mpg))
#>   4. └─dplyr:::mutate.data.frame(., mpg = sqrt(-mpg)) at dplyr/R/mutate.R:146:2
#>   5.   └─dplyr:::mutate_cols(.data, dplyr_quosures(...), caller_env = caller_env()) at dplyr/R/mutate.R:175:2
#>   6.     ├─base::withCallingHandlers(...) at dplyr/R/mutate.R:239:2
#>   7.     └─(function() {... at dplyr/R/mutate.R:239:2
#>   8.       └─dplyr:::mutate_col(i, .data, dots, mask, new_columns) at dplyr/R/mutate.R:244:6
#>   9.         └─mask$eval_all_mutate(quo) at dplyr/R/mutate.R:371:8
#>  10.           └─dplyr (local) eval() at dplyr/R/data-mask.R:69:6
#> 
#> [[5]]
#> <warning/rlang_warning>
#> Warning in `mutate()`:
#> Problem in group 5: mpg = 15 while computing `mpg = sqrt(-mpg)`.
#> Caused by warning in `sqrt()`:
#> ! NaNs produced
#> ---
#> Backtrace:
#>      ▆
#>   1. ├─global fn()
#>   2. │ └─mtcars %>% group_by(mpg) %>% mutate(mpg = sqrt(-mpg))
#>   3. ├─dplyr::mutate(., mpg = sqrt(-mpg))
#>   4. └─dplyr:::mutate.data.frame(., mpg = sqrt(-mpg)) at dplyr/R/mutate.R:146:2
#>   5.   └─dplyr:::mutate_cols(.data, dplyr_quosures(...), caller_env = caller_env()) at dplyr/R/mutate.R:175:2
#>   6.     ├─base::withCallingHandlers(...) at dplyr/R/mutate.R:239:2
#>   7.     └─(function() {... at dplyr/R/mutate.R:239:2
#>   8.       └─dplyr:::mutate_col(i, .data, dots, mask, new_columns) at dplyr/R/mutate.R:244:6
#>   9.         └─mask$eval_all_mutate(quo) at dplyr/R/mutate.R:371:8
#>  10.           └─dplyr (local) eval() at dplyr/R/data-mask.R:69:6
#> ... with 20 more warnings.
#> ℹ Use `last_dplyr_warnings(n = ...)` to show more.
```

R/conditions.R
DavisVaughan: Oh its mainly for `cur_group_label()` I see

R/mutate.R
```diff
@@ -223,20 +223,26 @@ mutate_cols <- function(.data, dots, caller_env, error_call = caller_env()) {
...
+    for (i in seq_along(dots)) local({
+      error_ctxt <<- local_error_context(dots = dots, .index = i, mask = mask)
       context_poke("column", old_current_column)
-      new_columns <- mutate_col(i, .data, dots, mask, new_columns)
-    },
+
+      new_columns <<- mutate_col(i, .data, dots, mask, new_columns)
+    }),
     error = function(e) {
       local_error_context(dots = dots, .index = i, mask = mask)
```
DavisVaughan, line 247: Do you need this anymore if you have `error_ctxt`?

R/mutate.R
```diff
@@ -254,26 +260,26 @@ mutate_cols <- function(.data, dots, caller_env, error_call = caller_env()) {
...
+      if (is_null(w$trace) && length(warnings) < 5) {
+        # The frame skipping assumes a regular `base::warning()` stack.
+        # Also works with warnings emitted from C with `Rf_warning()`.
+        w$trace <- trace_back(bottom = sys.frame(-5))
       }
+      new <- cnd_data(
+        cnd = w,
+        error_ctxt,
+        mask_type = mask_type,
+        call = error_call_forced
```
DavisVaughan, line 273: I guess you can't just call `error_call(error_call)` from here because the context would be wrong? That's why you delay-assign it in the `mutate_cols()` function environment? It just seems a little complicated, so I want to make sure I understand

tests/testthat/_snaps/conditions.md
```diff
@@ -226,3 +226,498 @@
...
+        1. +-dplyr::mutate(df, x = f())
+        2. +-dplyr:::mutate.data.frame(df, x = f())
+        3. | \-dplyr:::mutate_cols(.data, dplyr_quosures(...), caller_env = caller_env())
+        4. |   +-base::withCallingHandlers(...)
+        5. |   \-(function() {...
+        6. |     \-dplyr:::mutate_col(i, .data, dots, mask, new_columns)
+        7. |       \-mask$eval_all_mutate(quo)
+        8. |         \-dplyr (local) eval()
+        9. \-dplyr (local) f()
+       10.   \-base::warning("msg")
```
DavisVaughan, line 259: I guess this is likely to be a noisy snapshot test if we change the internals/signature of `mutate_cols()` (like if the path to get here changes then the snapshot will change)

DavisVaughan (approved):

R/conditions.R
lionel-: yeah I don't know how to do better here. If you source the file you'll get srcrefs buried in the backtraces of `dplyr_last_warnings()`.

Each `mutate()` sys-call should have its own srcref that we could show in the summary warnings, I'll look into making advantage of that. I'm not sure why this isn't the case already actually.

R/conditions.R
lionel-: Maybe we can wait for Gabor to finish implementing link syntax in the next cli version?

R/conditions.R
lionel-: I was thinking the same.

R/conditions.R
lionel-: I'm not sure we want to make it easy to print potentially 1000s of warnings.

R/conditions.R
lionel-: Good idea.

R/mutate.R
lionel-: I think so. I did not plan to refactor this mechanism entirely.

R/mutate.R
lionel-: `error_call()` does some environment climbing (e.g. to find generics) so I want to avoid doing this work n-group times.

tests/testthat/_snaps/conditions.md
lionel-: That's an annoyance with backtrace snapshots but I that find it's good to be mindful of what our stacks look like to end users.

NEWS.md
lionel-: This is a bit long but we could make it:

```
There was 1 warning in `mutate()` while computing `mpg2 = sqrt(-mpg)`.
```

Or use `cnd$call` and let R mention `mutate()`, but it looks a bit janky with a different quoting style and a space before the colon.

R/conditions.R
lionel-: Perhaps the following would work. We'd take `n = NA` by default. When `NA`, and interactive, and there is more than 20 warnings, we'd open a `utils::menu()` with these options:

```
There are 1000 warnings. How many do you want to show?

1. 20
2. All of them
3. None

Selection:
```

R/conditions.R
hadley: Yeah I think we could do it in one sweep.

R/conditions.R
hadley: Maybe we should brainstorm this bit in the dplyr meeting? And maybe pick some home for these check functions to live? Either rlang or cli or maybe some new package.

R/conditions.R
hadley: Alternatively, is there some easy way to collapse repeated warnings?

R/conditions.R
lionel-: They currently live in a compat file in rlang: https://github.com/r-lib/rlang/blob/main/R/compat-types-check.R

But this relies on `obj_type_friendly()` from https://github.com/r-lib/rlang/blob/main/R/compat-obj-type.R which is now also implemented in cli and has already diverged, so it'd be good to talk this through.

R/conditions.R
hadley: Or what if we made the top env of the trace back the mutate call?

lionel-: We should do that (not sure why it isn't the case already, I'll take a look) but this would not solve the problem in general because backtraces might potentially be verbose anyway.

hadley: Yeah, it's not perfect but it should help a bit. I'm not too worried about the lack of scannability; I think generally you'd start by resolving the first warning and then work your way down.

NEWS.md
DavisVaughan: I like what you suggested here

R/conditions.R
DavisVaughan: Or maybe just make the clickable link `last_dplyr_warnings(n = 20)`? People can intuit that they can change `n` if they need more than that

NEWS.md
lionel-: This is now:

```
Warning message:
There were 100 warnings in a `mutate()` step.
The first warning was:
! Problem in row 1 while computing `x = f()`.
Caused by warning in `f()`:
! foo
ℹ Run `dplyr::last_dplyr_warnings()` to see the 99 remaining warnings. 
```

DavisVaughan: I like that!

R/conditions.R
```diff
@@ -161,3 +175,147 @@ skip_internal_condition <- function(cnd) {
...
+  push_dplyr_warnings(warnings)
+
+  first <- new_dplyr_warning(warnings[[1]])
+  call <- format_error_call(error_call)
+
+  msg <- paste_line(
+    cli::format_warning(c(
+        "There {cli::qty(n)} {?was/were} {n} warning{?s} in a {call} step.",
+        if (n > 1) "The first warning was:"
+      )),
```
DavisVaughan, line 252: Slightly awkward tabbing here?

tests/testthat/_snaps/conditions.md
```diff
@@ -248,17 +248,9 @@
...
-        4. |   +-base::withCallingHandlers(...)
-        5. |   \-(function() {...
-        6. |     \-dplyr:::mutate_col(dots[[i]], .data, mask, new_columns)
-        7. |       \-mask$eval_all_mutate(quo)
-        8. |         \-dplyr (local) eval()
-        9. \-dplyr (local) f()
-       10.   \-base::warning("msg")
+          x
+       1. +-dplyr::mutate(df, x = f())
+       2. \-dplyr:::mutate.data.frame(df, x = f())
```
DavisVaughan: This looks a lot nicer!

R/context.R
```diff
@@ -96,15 +96,44 @@ group_labels_details <- function(keys) {
   }), ", ")
```
DavisVaughan, line 96: We should probably backtick the `{name} = {value}` result in `group_labels_details()` here so it shows up a little better in the error message. Fine if you don't want to do that here

DavisVaughan (approved):
Not sure if you are done, but I gave the changes a look and they look nice

I'm assuming you are planning on doing `summarise()` in another PR

PR #6450, opened 2022-09-03
Mention `join_by()` when `by` argument is omitted in joins

### LINKED ISSUE #6448: `by = NULL` should emit a message to use `join_by()` rather than a character vector
i.e. something like this, possibly with a clickable hyperlink that copies to clipboard if we can get that in the IDE

```r
library(dplyr)

df1 <- tibble(x = 1)
df2 <- tibble(x = c(1, 2))

inner_join(df1, df2)
# Joining using `by = join_by(x)`
#                ~~~~~~~~~~~~~~~ <- clickable hyperlink that copies to clipboard
```

I think this is going to be another place we need `dplyr:::tick_if_needed()`, like for `` join_by(`foo bar`) ``

krlmlr:
We already have `tick_if_needed()` .

Closes #6448.

### DIFF
--- a/R/join-by.R
+++ b/R/join-by.R
@@ -352,17 +352,13 @@ join_by_common <- function(x_names,
   if (length(by) == 0) {
     message <- c(
       "`by` must be supplied when `x` and `y` have no common variables.",
-      i = "Use `by = character()` to perform a cross-join."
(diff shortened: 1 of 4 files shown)


DavisVaughan:
Thanks!

PR #6451, opened 2022-09-07
Switch to radix ordering in more places

DavisVaughan:
- `nth(order_by =)`
- `with_order()`

This means that `vec_order_base()` is now only used in the `dplyr.legacy_locale` code path of `group_by()`. And I might be able to remove that in another PR by using `dplyr_order_legacy()`, which is used in the `arrange()` legacy path. I'll take a look.

I did not feel like this needs a news bullet, because it is highly unlikely to affect much code, and we don't provide a way to opt out of this.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,9 @@
 # dplyr (development version)
 
+* `nth()`, `first()`, `last()`, and `with_order()` now sort character `order_by`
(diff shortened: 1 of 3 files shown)


hadley (approved):
Probably worth a news bullet even if it's just to jog our memories.

PR #6455, opened 2022-09-13
Use `dplyr_order_legacy()` and remove `vec_order_base()`

DavisVaughan:
We already have `dplyr_order_legacy()` for `arrange()`, which is slightly different from `vec_order_base()` because it allows you to use a vectorized `direction` argument.

Since `arrange()` already needed a custom version of `vec_order_base()`, we might as well adapt it to be usable by the legacy path of `group_by()` as well.

That means:
- Both `arrange()` and `group_by()` use the exact same legacy ordering helper
- We don't need to import `vctrs:::vec_order_base()`, which simplifies things a bit and means we definitely don't need it in vctrs anymore

So now we just have to worry about managing `vec_order_radix()`

### DIFF
--- a/R/arrange.R
+++ b/R/arrange.R
@@ -187,7 +187,7 @@ sort_key_generator <- function(locale) {
 
 # ------------------------------------------------------------------------------
 
-dplyr_order_legacy <- function(data, direction) {
+dplyr_order_legacy <- function(data, direction = "asc") {
   if (ncol(data) == 0L) {
     # Work around `order(!!!list())` returning `NULL`
     return(seq_len(nrow(data)))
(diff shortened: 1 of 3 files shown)


R/arrange.R
```diff
@@ -227,6 +227,14 @@ dplyr_proxy_order_legacy <- function(x, direction) {
     return(out)
   }
 
+  if (!is_character(x) && !is_logical(x) && !is_integer(x) && !is_double(x) && !is_complex(x)) {
+    abort("Invalid type returned by `vec_proxy_order()`.", .internal = TRUE)
+  }
+
+  if (is.object(x)) {
+    x <- unstructure(x)
+  }
```
DavisVaughan, line 236: These were parts of `vec_order_base()` that weren't pulled over when we created the custom version of it for `arrange()`. I pulled them over for completeness, but I think they are mostly harmless

hadley (approved):

PR #6456, opened 2022-09-13
Tweak multi-match warning

### LINKED ISSUE #6445: Better warning for *-to-many join?
Currently:

```R
Warning message:
Each row in `x` should match at most 1 row in `y`.
ℹ Row 1 of `x` matches multiple rows.
ℹ If multiple matches are expected, specify `multiple = "all"` in the join call to silence this warning. 
```

Maybe it should be more like:

```R
Warning message:
Output is longer than input, because some rows in `x` match multiple rows in `y`.
ℹ Row 1 of `x` matches multiple rows.
ℹ If multiple matches are expected, specify `multiple = "all"` in the join call to silence this warning. 
```

hadley:
Fixes #6445

### DIFF
--- a/R/join-rows.R
+++ b/R/join-rows.R
@@ -150,7 +150,7 @@ rethrow_error_join_matches_multiple <- function(cnd, call) {
 
   stop_join(
     message = c(
-      glue("Each row in `x` can match at most 1 row in `y`."),
+      glue("Each row in `x` must match at most 1 row in `y`."),
       i = glue("Row {i} of `x` matches multiple rows.")
     ),
     class = "dplyr_error_join_matches_multiple",
--- a/R/join-rows.R
+++ b/R/join-rows.R
@@ -89,7 +89,7 @@ dplyr_locate_matches <- function(needles,
       rethrow_error_join_matches_multiple(cnd, error_call)
     },
     vctrs_warning_matches_multiple = function(cnd) {
-      rethrow_warning_join_matches_multiple(cnd)
+      rethrow_warning_join_matches_multiple(cnd, error_call)
     }
   )
 }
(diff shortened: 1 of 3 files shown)


R/join-rows.R
```diff
@@ -158,19 +158,20 @@ rethrow_error_join_matches_multiple <- function(cnd, call) {
...
 }
 
-rethrow_warning_join_matches_multiple <- function(cnd) {
+rethrow_warning_join_matches_multiple <- function(cnd, call) {
   i <- cnd$i
 
   warn_join(
     message = c(
-      glue("Each row in `x` should match at most 1 row in `y`."),
+      glue("Each row in `x` is expected to match at most 1 row in `y`."),
```
DavisVaughan, line 166: Would you please apply these changes to `rethrow_error_join_matches_multiple()` too?

R/join-rows.R
```diff
@@ -158,19 +158,20 @@ rethrow_error_join_matches_multiple <- function(cnd, call) {
...
       i = paste0(
-        "If multiple matches are expected, specify `multiple = \"all\"` in ",
-        "the join call to silence this warning."
+        "If multiple matches are expected, set `multiple = \"all\"` ",
+        "to silence this warning."
       )
     ),
-    class = "dplyr_warning_join_matches_multiple"
+    class = "dplyr_warning_join_matches_multiple",
+    call = call
```
DavisVaughan, line 174: `rlang::warn()` doesn't have a `call` argument, but it does look like this gets passed through to `rlang::warning_cnd()`, which does have a `call` argument.

@lionel- is this expected? Should `warn()` have a `call` arg?

DavisVaughan (approved):

R/join-rows.R
lionel-: I guess an explicit `call` arg wouldn't hurt. `abort()` has to have one because it does something with it.

PR #6458, opened 2022-09-13
Remove `preceding()` and `following()` in favor of `closest()`

### LINKED ISSUE #6439: Replace `preceding()` and `following()` with `closest()`
```r
preceding(x, y, inclusive = TRUE)  -> closest(x >= y)
preceding(x, y, inclusive = FALSE) -> closest(x > y)
following(x, y, inclusive = TRUE)  -> closest(x <= y)
following(x, y, inclusive = FALSE) -> closest(x < y)
```

This should be more intuitive for interactive exploration too, because you probably start out with `x >= y`, then realize you only care about the closest match, so you just wrap it in `closest()` rather than having to switch to `preceding()`

DavisVaughan:
Closes #6439 

I think this is much nicer. It was easier to talk about in the docs, and the progression from

```r
join_by(id, sale_date >= promo_date)
```

to

```r
join_by(id, closest(sale_date >= promo_date))
```

in the examples is now much easier to understand

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -301,10 +301,10 @@
       `join_by(sale_date >= commercial_date)` to find every commercial that
       aired before a particular sale.
       
-    * Rolling joins: For "rolling" the preceding value forward or the following
-      value backwards when there isn't an exact match, specified by using one of
-      the rolling helpers: `preceding()` or `following()`. For example,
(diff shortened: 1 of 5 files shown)


hadley (approved):
Looks great!

PR #6461, opened 2022-09-15
Update summarise.R

eogoodwin:
When the informing message first popped up (and then for the next what, year? until today anyway) I interpreted it to mean that summarise() had ignored the grouping set up by group_by(), and had instead chosen to work with the reported new grouping.  It was not clear (enough to me) that it had grouped /its/ output, ready for any following functions, by the reported grouping.  I'm proposing a slightly more verbose message, which might also be slightly more informative (although I concede that verbosity and informance are imperfectly correlated, and possibly even negatively correlated. Not "informance", informativity). Anyway, see what you think, run it past your test panels, workshop it, mull it over and make your call.


eogoodwin:
Thanks for getting back to me Hadley.  Yes, of course once you know how it works, it's "all good".  If you're happy, then I'm happy.

PR #6464, opened 2022-09-16
Improve sample_n() docs

### LINKED ISSUE #6462: Update `sample_n()` docs
To remove `by_cyl` and generally update examples
https://github.com/tidyverse/dplyr/blob/af88fa0f252842102aef1b69c7dea4bb7453dbb2/R/sample.R#L36

Probably need a better example dataset too:
- Should be a tibble / no row names
- With less columns

hadley:
Fixes #6462

### DIFF
--- a/R/sample.R
+++ b/R/sample.R
@@ -33,35 +33,33 @@
 #' @param .env DEPRECATED.
 #' @param ... ignored
 #' @examples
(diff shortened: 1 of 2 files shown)


DavisVaughan (approved):
I also just pushed snapshot changes to main which should fix the failing builds

Issue #6465, opened 2022-09-17
`left_join()` error message names wrong variable in `y`


cjyetman:
When joining two tables by differently named columns, if they are incompatible types the error message uses the name of the by column in table x as the name of the by column in table y, e.g. `y$var1` which does not exist in the example below.

``` r
library(dplyr)

tb1 <-
  tibble(
    var1 = 1,
    value = 2
  )

tb2 <-
  tibble(
    id1 = "1",
    value = 2
  )

left_join(tb1, tb2, by = c(var1 = "id1"))
#> Error in `left_join()`:
#> ! Can't join on `x$var1` x `y$var1` because of incompatible types.
#> ℹ `x$var1` is of type <double>>.
#> ℹ `y$var1` is of type <character>>.
```

using:
R version 4.2.1
dplyr     1.0.10
rlang     1.0.5
cli       3.4.0

hadley:
Still see the same failure with dev dplyr:

```R
library(dplyr, warn.conflicts = FALSE)

df1 <- tibble(var1 = 1, val = 2)
df2 <- tibble(var2 = "1", val = 2)

left_join(df1, df2, by = join_by(var1 == var2))
```

PR #6467, opened 2022-09-19
Use original `y` name in join common type error message

DavisVaughan:
Closes https://github.com/tidyverse/dplyr/issues/6465

This was slightly more work than expected. I ended up making the common cast operation more explicit, as currently it is implicit in the `vec_locate_matches()` call.

Pulling it out into `join_cast_common()` felt better than trying to pass `vars` through to `join_rows()`, which otherwise wouldn't need to know anything about `vars`. Casting early also allowed me to remove some code that was recomputing the common type again.

### DIFF
--- a/R/join-cols.R
+++ b/R/join-cols.R
@@ -16,7 +16,7 @@ join_cols <- function(x_names,
   suffix <- standardise_join_suffix(suffix, error_call = error_call)
 
   x_by <- set_names(match(by$x, x_names), by$x)
-  y_by <- set_names(match(by$y, y_names), by$x)
+  y_by <- set_names(match(by$y, y_names), by$y)
 
   x_loc <- seq_along(x_names)
   names(x_loc) <- x_names
--- a/R/join-cols.R
+++ b/R/join-cols.R
@@ -164,3 +164,42 @@ add_suffixes <- function(x, y, suffix) {
   }
   out
 }
+
+join_cast_common <- function(x, y, vars, error_call = caller_env()) {
+  # Explicit `x/y_arg = ""` to avoid auto naming in `cnd$x_arg`
+  ptype <- try_fetch(
+    vec_ptype2(x, y, x_arg = "", y_arg = "", call = error_call),
+    vctrs_error_incompatible_type = function(cnd) {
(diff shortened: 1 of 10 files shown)


R/join-cols.R
```diff
@@ -16,7 +16,7 @@ join_cols <- function(x_names,
   suffix <- standardise_join_suffix(suffix, error_call = error_call)
 
   x_by <- set_names(match(by$x, x_names), by$x)
-  y_by <- set_names(match(by$y, y_names), by$x)
+  y_by <- set_names(match(by$y, y_names), by$y)
```
DavisVaughan, line 19: This change in what is returned by `join_cols()` is actually the key here. It allows us to track the original column names from `y`.

R/join.R
```diff
@@ -506,7 +506,11 @@ nest_join.data.frame <- function(x,
   y_in <- as_tibble(y, .name_repair = "minimal")
 
   x_key <- set_names(x_in[vars$x$key], names(vars$x$key))
-  y_key <- set_names(y_in[vars$y$key], names(vars$y$key))
+  y_key <- set_names(y_in[vars$y$key], names(vars$x$key))
```
DavisVaughan, line 509: You'll see that any place that we originally used `names(vars$y$key)`, we now use `names(vars$x$key)`. I think this actually makes more sense, because the point is to give the columns from `x_in` and `y_in` a consistent set of column names, so it makes sense to always use `names(vars$x$key)`

R/join.R
```diff
@@ -622,26 +630,22 @@ join_mutate <- function(x,
...
+
       new_rows <- which(is.na(x_slicer))
       y_replacer <- y_slicer[new_rows]
 
-      out[new_rows, names(y_merge)] <- vec_cast(
-        x = vec_slice(y_merge, y_replacer),
-        to = key_type,
-        call = error_call
-      )
+      out[new_rows, merge] <- vec_slice(y_merge, y_replacer)
```
DavisVaughan, line 648: Nice simplification here from not needing to compute the common type again. The keys are now already the same type here.

hadley (approved):
Looks great! I tried to break with a few local examples and didn't see any problems.

PR #6468, opened 2022-09-19
Use lifecycle infrastructure for `filter()` deprecation warnings

DavisVaughan:
Set up for https://github.com/tidyverse/dplyr/issues/6091

### DIFF
--- a/R/filter.R
+++ b/R/filter.R
@@ -171,18 +171,25 @@ filter_expand <- function(dots, mask, error_call = caller_env()) {
 filter_eval <- function(dots, mask, error_call = caller_env()) {
   env_filter <- env()
 
-  withCallingHandlers({
-    mask$eval_all_filter(dots, env_filter)
-  }, error = function(e) {
-    local_error_context(dots = dots, .index = env_filter$current_expression, mask = mask)
+  withCallingHandlers(
+    expr = mask$eval_all_filter(dots, env_filter),
+    error = function(e) {
+      local_error_context(dots = dots, .index = env_filter$current_expression, mask = mask)
 
-    bullets <- c(
-      cnd_bullet_header("computing"),
-      filter_bullets(e)
-    )
-    abort(bullets, call = error_call, parent = skip_internal_condition(e))
+      bullets <- c(
+        cnd_bullet_header("computing"),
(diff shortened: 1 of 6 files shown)


R/filter.R
```diff
@@ -222,3 +229,22 @@ filter_bullets.default <- function(cnd, ...) {
   )
 }
 
+warn_filter_across <- function(call) {
+  lifecycle::deprecate_warn(
+    when = "1.0.8",
+    what = I("Using `across()` in `filter()`"),
+    with = I("`if_any()` or `if_all()`"),
+    always = TRUE,
```
DavisVaughan, line 237: I have decided to use `always = TRUE` here, because this aligns with the `Rf_warningcall()` we were using before

src/filter.cpp
```diff
@@ -21,6 +21,20 @@ void stop_filter_incompatible_type(R_xlen_t i, SEXP quos, SEXP column_name, SEXP
...
+  Rf_eval(call, dplyr::envs::ns_dplyr);
+  UNPROTECT(2);
+}
+
+void signal_filter_data_frame() {
+  SEXP cls = PROTECT(Rf_mkString("dplyr:::signal_filter_data_frame"));
+  SEXP call = PROTECT(Rf_lang2(dplyr::symbols::dplyr_internal_signal, cls));
+  Rf_eval(call, dplyr::envs::ns_dplyr);
+  UNPROTECT(2);
+}
```
DavisVaughan, line 36: I've decided to implement this through signals so I don't have to pass the `error_call` through C++ just to be able to throw a warning.

We already have `withCallingHandlers()` set up around the filter on the R side, so we just catch the signal and throw a warning with the `error_call` from there.

I think that works pretty well, but let me know if I'm missing something

lionel-: yeah perfect

lionel- (approved):

src/filter.cpp
lionel-: The other thing you could do is use the `caller_env()` implementation from https://github.com/tidyverse/purrr/commit/a86f4a57bc15dde7d7eb043652b5dcaa1309052a

PR #6473, opened 2022-09-21
Backtick error contexts in condition messages

lionel-:


R/context.R
```diff
@@ -91,9 +91,9 @@ cur_group_rows <- function() {
 }
 
 group_labels_details <- function(keys) {
-  glue_collapse(map2_chr(keys, names(keys), function(x, name) {
-    glue("{name} = {value}", value = pillar::format_glimpse(x))
-  }), ", ")
+  keys <- map_chr(keys, pillar::format_glimpse)
+  labels <- paste0(names(keys), " = ", keys)
+  cli::format_inline("{.code {labels}}")
```
hadley, line 96: Would be good to have one test where the input is grouped by >1 variable.

lionel-: We have one but it's incidental, I'll add a specific test.

DavisVaughan: I'm actually not sure we want to use `{.code {labels}}` here because I don't think we want its collapse behavior of separating the last two elements with `and`. I think we just want to collapse everything with `,`

tests/testthat/_snaps/summarise.md
```diff
@@ -54,7 +54,7 @@
       Error in `summarise()`:
       ! Problem while computing `a = rlang::env(a = 1)`.
       x `a` must be a vector, not an environment.
-      i The error occurred in group 1: x = 1, y = 1.
+      i The error occurred in group 1: `x = 1` and `y = 1`.
```
DavisVaughan, line 57: Like, I'd rather just see `` `x = 1`, `y = 1` `` here

R/context.R
lionel-: Why do we want that? I was thinking using "and" was fine here.

DavisVaughan: I don't think it matters too much, it just doesn't feel like you are trying to construct a complete sentence here, you are just listing the group keys, so the `and` didn't feel necessary. I don't care too much either way

lionel-: ok I've restored the original collapsing and added a specific test.

@gaborcsardi Looks like `cli_collapse()` might be useful after all.

R/utils.R
```diff
@@ -66,3 +66,7 @@ node_walk_replace <- function(node, old, new) {
     node <- node_cdr(node)
   }
 }
+
+cli_collapse <- function(x, last = " and ") {
```
DavisVaughan, line 70: I will fight you over the lack of oxford comma! 😆 

```r
> cli::format_inline('{cli_collapse(c("foo", "bar", "baz"))}')
[1] "foo, bar and baz"
```

lionel-: oh yeah sorry. But it's not really meant to be used with "and" and "or" as we can use cli collapsing easily for these.

PR #6475, opened 2022-09-21
Simplify contextual errors

lionel-:


tests/testthat/_snaps/conditions.md
```diff
@@ -81,7 +81,7 @@
       slice(mtcars, var = invisible(999 + ""))
     Condition
       Error in `slice()`:
-      ! Problem while evaluating `var`.
+      ! Can't evaluating `var`.
```
lionel-, line 84: oops

R/conditions.R
```diff
@@ -318,11 +318,11 @@ new_dplyr_warning <- function(data) {
     data$group_data$id,
     data$group_data$group
   )
-  if (nzchar(label)) {
-    label <- glue(" in {label}")
-  }
 
-  msg <- glue::glue("Problem{label} while computing `{data$name} = {data$expr}`.")
+  msg <- c(
+    glue::glue("Can't compute `{data$name} = {data$expr}`."),
```
DavisVaughan, line 323: Is it worth using `{.code }` from cli?

tests/testthat/_snaps/slice.md
```diff
@@ -24,7 +24,7 @@
       slice(tibble(), "a")
     Condition
       Error in `slice()`:
-      ! Problem while evaluating `..1 = "a"`.
+      ! Can't evaluate `..1 = "a"`.
       Caused by error:
       ! Must subset elements with a valid subscript vector.
       x Subscript `"a"` has the wrong type `character`.
```
DavisVaughan, line 30: Is `evaluate` the right word here? I mean, you can evaluate `"a"` but you can't subset with it

DavisVaughan (approved):
Looks very nice!

R/conditions.R
lionel-: yup.

But I was going to ask about this one. It's a warning, so "Can't compute" isn't quite right. Maybe I should keep the original phrasing? How could we do better?

tests/testthat/_snaps/slice.md
lionel-: I'll try to find something better.

Also the other instances of "evaluate" could be "compute" instead.

R/conditions.R
hadley: Maybe "Failed to compute"?

hadley: Oh ooops. Is there an example in the tests?

tests/testthat/_snaps/slice.md
lionel-: We now have "process" and "compute" depending on the error:

```r
# Note: I still plan to make this case less verbose
slice(tibble(), "a")
#> Error in `slice()`:
#> ! Can't process `..1 = "a"`.
#> Caused by error:
#> ! Must subset elements with a valid subscript vector.
#> ✖ Subscript `"a"` has the wrong type `character`.
#> ℹ It must be numeric.

slice(tibble(), 1 + "")
#> Error in `slice()`:
#> ! Can't compute `..1 = 1 + ""`.
#> Caused by error in `1 + ""`:
#> ! non-numeric argument to binary operator
```

R/conditions.R
lionel-: yup in `conditions.md`, e.g.:

```
There were 2 warnings in a `mutate()` step.
The first warning was:
! Can't compute `mpg = sqrt(-mpg)`.
ℹ In group 1: `vs = 0`.
Caused by warning in `sqrt()`:
! NaNs produced
ℹ Run `dplyr::last_dplyr_warnings()` to see the 1 remaining warning. 
```

tests/testthat/_snaps/mutate.md
```diff
@@ -66,7 +66,7 @@
     Condition
       Warning:
       There was 1 warning in a `mutate()` step.
-      ! Problem while computing `y = max(x)`.
+      ! Can't compute `y = max(x)`.
```
DavisVaughan, line 69:
```suggestion
      ! In `y = max(x)`.
```

What about just `In`?

Because it already tells you there was "1 warning", so you are just talking about the expression it came from?

DavisVaughan: `In` may also compose well with the groups like in the example you said above:

```r
There were 2 warnings in a `mutate()` step.
The first warning was:
! In `mpg = sqrt(-mpg)`.
ℹ In group 1: `vs = 0`.
Caused by warning in `sqrt()`:
! NaNs produced
ℹ Run `dplyr::last_dplyr_warnings()` to see the 1 remaining warning. 
```

tests/testthat/_snaps/slice.md
DavisVaughan: Seems reasonable!

tests/testthat/_snaps/mutate.md
lionel-: hmm... Then should it be a `i` bullet instead of `!`? Otherwise it doesn't feel consistent with the group bullet. We'd have:

```
There were 2 warnings in a `mutate()` step.
The first warning was:
ℹ In `mpg = sqrt(-mpg)`.
ℹ In group 1: `vs = 0`.
Caused by warning in `sqrt()`:
! NaNs produced
```

This is still a bit weird for some reason, like something is missing.

lionel-: But it's not too bad, I think we can try this.

lionel-:
Merging now so we can get a better feel, let me know if something feels off.

R/conditions.R
hadley: Hmmmm, could it just be:

```
The first warning was:
! In `mpg = sqrt(-mpg)`.
ℹ In group 1: `vs = 0`.
Caused by warning in `sqrt()`:
! NaNs produced
ℹ Run `dplyr::last_dplyr_warnings()` to see the 1 remaining warning. 
``

DavisVaughan: @hadley that is what we went with https://github.com/tidyverse/dplyr/pull/6475#discussion_r976715655

PR #6476, opened 2022-09-22
Enrich warnings in `filter()` and `summarise()`

lionel-:


R/conditions.R
```diff
@@ -86,18 +86,18 @@ quo_as_label <- function(quo)  {
   }
 }
 
-local_error_context <- function(dots, .index, mask, frame = caller_env()) {
-  expr <- dot_as_label(dots[[.index]])
+local_error_context <- function(dots, i, mask, frame = caller_env()) {
```
lionel-, line 89: I went ahead and changed this argument name as it was staring at me the whole time.

R/conditions.R
```diff
@@ -306,7 +331,7 @@ signal_warnings <- function(warnings, error_call) {
 
   msg <- paste_line(
     cli::format_warning(c(
-      "There {cli::qty(n)} {?was/were} {n} warning{?s} in a {call} step.",
+      "There {cli::qty(n)} {?was/were} {n} warning{?s} in {call}.",
```
lionel-, line 334: I've removed the need for an indefinite article so that the message works with `arrange()` (which is now tested).

tests/testthat/_snaps/conditions.md
```diff
@@ -262,7 +262,7 @@
       df %>% group_by(id) %>% mutate(x = f()) %>% invisible()
     Condition
       Warning:
-      There were 2 warnings in a `mutate()` step.
```
hadley, line 265: Random thought — should we put `mutate()` in the warning call?

```R
Warning in `mutate()`:
* There were 2 warnings across all groups
```

lionel-: We can but it'll look a bit different (no backticks, space before colon) because R itself prints the call:

```
cnd_signal(warning_cnd(message = "There were 2 warnings.", call = call("mutate")))
#> Warning message:
#> In mutate() : There were 2 warnings.
```

hadley: Oh yeah, no point then

R/mutate.R
```diff
@@ -224,55 +224,37 @@ mutate_cols <- function(.data, dots, error_call = caller_env()) {
...
   withCallingHandlers(
-    for (i in seq_along(dots)) local({
-      error_ctxt <<- local_error_context(dots = dots, .index = i, mask = mask)
+    for (i in seq_along(dots)) {
+      local_error_context(dots, i, mask = mask)
       context_poke("column", old_current_column)
 
-      new_columns <<- mutate_col(dots[[i]], .data, mask, new_columns)
-    }),
+      new_columns <- mutate_col(dots[[i]], .data, mask, new_columns)
```
DavisVaughan, line 241: This seems like a nice improvement. Feels cleaner

R/filter.R
```diff
@@ -170,21 +170,32 @@ filter_expand <- function(dots, mask, error_call = caller_env()) {
 
 filter_eval <- function(dots, mask, error_call = caller_env()) {
   env_filter <- env()
+  warnings_state <- env(warnings = list())
```
DavisVaughan, line 173: _Maybe_ worth wrapping into `new_warnings_state()`?

DavisVaughan (approved):

R/mutate.R
lionel-: I'm not sure it's cleaner. It feels better to read but it's not cleaner in semantics. I don't see an obvious failure mode though. Oh now I remember I also meant to check that the `local_` action does not pile up handlers on a stack in that loop, I'll take a look.

R/filter.R
lionel-: I thought about it and I think I prefer to keep this simple with bindings that are locally explicit.

Starting to wrap this leads us to an R6 object. I tried this and it made the whole thing more engineered than seemed warranted.

R/mutate.R
lionel-: It does pile up handlers. Also `context_local()` doesn't restore handlers in the right order if called multiple times.

In the last commit I added `poke_error_context()` for updating the context within loops. And added a comment regarding what's needed to fix the restoration order once we depend on R 3.5.

DavisVaughan: > It does pile up handlers

Oh good catch

PR #6479, opened 2022-09-26
Simplify contextual errors with "i" bullets

lionel-:
I realised that:

- The "compute" and "process" verbs don't bring any information.
- Having `!` bullets in contextual errors that only display contextual information without reformulating the causal error in a higher level way is distracting.

This PR simplifies contextual errors by using "i" bullets starting with "In argument", as in #6477.

I kept `!` bullets for recycling errors though because "Can't recycle" is informative.

### DIFF
--- a/tests/testthat/_snaps/filter.md
+++ b/tests/testthat/_snaps/filter.md
@@ -23,7 +23,7 @@
     Output
       <error/rlang_error>
       Error in `filter()`:
-      i In `..1 = matrix(TRUE, nrow = 3, ncol = 2)`.
+      i In argument `..1 = matrix(TRUE, nrow = 3, ncol = 2)`.
       Caused by error:
       ! `..1` must be a logical vector, not a logical matrix.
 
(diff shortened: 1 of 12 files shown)


hadley (approved):
This "in" formation seems super useful.

tests/testthat/_snaps/filter.md
```diff
@@ -137,7 +137,7 @@
     Output
       <error/rlang_error>
       Error in `filter()`:
-      ! Can't compute `..1 = data.frame(Sepal.Length > 3, 1:n())`.
+      i In argument `..1 = data.frame(Sepal.Length > 3, 1:n())`.
       i In group 1: `Species = setosa`.
```
DavisVaughan, line 141: I really like how natural it is to read `in filter...in argument...in group` to get finer and finer nesting

DavisVaughan (approved):
Looks great, I like that we are trying to "get out of the way" of the most useful information in the error a bit more

tests/testthat/_snaps/filter.md
lionel-: I agree, thanks for the suggestion!

PR #6484, opened 2022-09-27
Update to latest lifecycle

lionel-:
- Pass bullets instead of formatted strings. This is compatible with both CRAN and dev lifecycle.

- Pass `user_env` to `deprecate_warn()` to differentiate between direct and indirect calls. This requires dev lifecycle.

### DIFF
--- a/tests/testthat/test-colwise-mutate.R
+++ b/tests/testthat/test-colwise-mutate.R
@@ -291,7 +291,7 @@ test_that("colwise mutate handle named chr vectors", {
   expect_identical(res, tibble(x = 1:10, y = 5.5))
 })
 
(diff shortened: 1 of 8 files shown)


tests/testthat/test-colwise-mutate.R
```diff
@@ -292,11 +292,12 @@ test_that("colwise mutate handle named chr vectors", {
 })
 
 test_that("colwise verbs soft deprecate quosures (#4330)", {
```
hadley, line 294:
```suggestion
test_that("colwise verbs deprecate quosures (#4330)", {
```

hadley (approved):

PR #6492, opened 2022-10-05
Implement `pick()`

### LINKED ISSUE #6204: Dedicated function for selecting from current data [FR]
Would it be useful to have a dedicated function (say, `pick()`) to select columns from the current data? Currently, `across()` with only a `.cols` argument serves this role.

I would see a dedicated function having at least three advantages:

1.  Nicer syntax for union selections: `pick(1, last_col())` vs. `across(c(1, last_col()))`.
2.  Better semantics. `across()` makes sense when there’s functions to apply, but less so when it’s used just for selecting columns. `pick()` seems intuitive for only selecting columns.
3.  Reuse existing patterns: `across(c(1:2, 4), mean)` vs. `map_df(pick(1:2, 4), mean)`. The first requires you to know that `across()` can select columns *and* apply a function, latter can re-use existing function application methods.

The last point is particularly important if/when `...` is deprecated in `across()` (\#6073), as funtionality would not be identical anymore. For example:

``` r
# With no ..., need to use an anonymous function for na.rm
across(c(1, 3:4), ~ mean(., na.rm = TRUE))

# Could be avoided with `pick()`
map_df(pick(1, 3:4), mean, na.rm = TRUE)
```

I would see the primary uses for this as:

1.  Replace `across()` in e.g. `group_by()` selections `group_by(across(c(1, 3:5)))` vs. `group_by(pick(1, 3:5))`. Big semantic and syntactic win, IMO.
2.  Passing arguments to functions that take data frame or matrix arguments. For example common questions about taking means or sums over rows in data frames. In my experience people don’t think to `apply(across(1:5), 1, f)`, but `apply(pick(1:5), 1, f)` might be more intuitive.

I could think of two ways to implement this as a wrapper:

``` r
pick <- function(...) {
  across(.cols = c(...))
}
```

Or:

``` r
pick <- function(...) {
  select(cur_data(), ...)
}
```

Although, particularly with the `across()` route, it would seem nicer to reverse the dependency and extract the relevant parts from `across()` into`pick()` instead.

I appreciate your consideration for this feature request.

DavisVaughan:
Closes #6204

This is a full implementation of `pick()`, but it has a performance issue that we aren't sure how to resolve.

- Implemented as `pick(...)` as opposed to `pick(cols)`. This function feels like `select()`; it's main purpose is column selection so it should get the easiest interface to accommodate that - i.e. `pick(a, b)` vs `pick(c(a, b))`.

- Implemented _without_ support for selecting grouping columns. We are currently under the impression that this is never useful, because you'd just be selecting a constant valued column. This means `everything()` doesn't select grouping columns. This is the same behavior as `across()`.

---

`pick()` is intended to supersede `cur_data()`, `cur_data_all()`, and `across(.fns = NULL)`, but there are a few potential issues with that:

- Since `pick()` can't select group columns, it doesn't really replace `cur_data_all()`. But I'm convinced the initial request for `cur_data_all()` didn't really need it (https://github.com/tidyverse/dplyr/issues/5342). We just aren't convinced that selecting group columns would ever be useful.

- I think `cur_data()` is used somewhat frequently, and the equivalent `pick()` syntax is `pick(everything())`. Unfortunately, this is much slower than `cur_data()` with grouped data frames because the `everything()` tidyselection has to be evaluated _on every group_ (like `across()`) and that is slow. I can't figure out how to "fix" this, but ideally the tidyselection would be evaluated 1 time on the full data frame and then reused across every group (that seems semantically cleaner too, otherwise the tidyselection could potentially give different results per group).

```r
library(dplyr)
library(nycflights13)

df <- group_by(flights, year, month, day, origin)
n_groups(df)
#> [1] 1095

bench::mark(
  fast = mutate(df, x = cur_data()),
  slow = mutate(df, x = pick(everything()))
)
#> # A tibble: 2 × 6
#> 1 fast       161.36ms 169.84ms     5.90     68.1MB     7.86
#> 2 slow          1.72s    1.72s     0.582    78.8MB     6.98
```

---

Still to do:

- [ ] Fix `pick()` performance issues if possible
- [ ] Supersede (soft deprecate?) `cur_data()` and `cur_data_all()`
- [ ] Soft deprecate `across(.cols = everything(), .fns = NULL)` behavior. In particular, we want:
    - `.cols` to be a required argument
    - `.fns` to be a required argument
    - But we'd have some backup behavior that does `if (missing(cols)) cols <- quote(everything())` to maintain backwards compat support for now.
- [ ] Scour vignettes for places where `pick()` could be useful and update them


DavisVaughan:
@lionel- do you have any ideas about improving the performance here? I think we learned while working on `across()` that caching a result like this is quite hard, and I don't think there is an equivalent to `expand_across()` here because most of the time people will wrap `pick()` into something like `dense_rank(pick(...))`, so it won't be the top level call.

Maybe we can see if there are easy wins in optimizing tidyselect? I took a quick look and it looks like `withr::defer()` seems to be a decent chunk of the time (but it also seems spread pretty thinly across the whole tidyselect call so maybe it is hard to optimize without going to C)

R/pick.R
```diff
@@ -0,0 +1,78 @@
...
+
+  if (dots_n(...) == 0L) {
+    abort("`...` can't be empty.")
+  }
+
+  # `pick()` is evaluated in a data mask so we need to remove the
+  # mask layer from the quosure environments (same as `across()`) (#5460)
+  quos <- enquos(..., .named = NULL)
+  quos <- map(quos, quo_set_env_to_data_mask_top)
+  expr <- expr(c(!!!quos))
```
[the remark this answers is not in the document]
lionel-: I think there used to be some issues with nested data masks where a potential solution required setting a proper environment for `eval_tidy()`, but it's been years since that came up. I don't remember the details. It's probably fine to omit passing an env though.

DavisVaughan:
I have added (imperfect) caching to `pick()`. It generally works like this:

- We create a `key` by extracting out all of the expressions from the quosures captured in `...` and then hashing that list-of-expressions object with `hash()`

- The first time a `key` is seen, we perform the tidyselection required by that `pick()` call to select columns from the _original_ data frame, and then we `vec_chop()` them according to the group locations. This list-of-tibbles becomes out `value` in the cache.

- The next time that `key` is seen (whether in this `pick()` call or in another `pick()` call that uses the same expression), we get the `value` from the cache and return the element that corresponds to the current group.

---

A few notes:

- We greatly simplified `pick()` by making the assertion that it _only_ selects columns from the _original data frame_. Meaning that previous expressions in `...` that add or delete columns do not affect the columns that `pick()` chooses from. This allows us to preserve some important invariants too:
  - The columns selected by `pick()` are the same across all groups.
  - The number of rows returned by `pick()` is always equal to the number of rows in the current group.

- Because `pick()` uses the original data frame, it cannot use the environment which has the promises in it that point to the chopped versions of the "current" data. This means it has to call `vec_chop()` itself, which I don't actually think is a huge deal.

- It is _technically_ possible to set up a mask environment specifically for `pick()`, which would mean that `mutate(df, x = pick(a, b), y = pick(b, c))` could share the chopped `b` column. I tried this and it seemed like too much work for too little performance gain. I don't think there are going to be that many `pick()` calls in a single `mutate()` call anyways, and the chance that there is column overlap between multiple `pick()` calls is probably pretty low.

---

Now the important things holding this back. The caching system is not perfect because it relies purely on the expression passed to `pick()`. This means that simple (reasonable) `pick()` calls like this one can defeat it:

```r
df <- tibble(a = 1, b = 2)

my_pick1 <- function(x) {
  pick(all_of(x))
}

mutate(df, x = my_pick1("a"), y = my_pick1("b"))
#> # A tibble: 1 × 4
#>       a     b   x$a   y$a
#>   <dbl> <dbl> <dbl> <dbl>
#> 1     1     2     1     1
```

It seems like the main two culprits are `all_of()` and `any_of()`, because these allow evaluation using the surrounding environment, and that isn't captured by the cache key.

DavisVaughan:
Here is the original benchmark rerun with the caching:

``` r
library(dplyr)
library(nycflights13)

df <- group_by(flights, year, month, day, origin)
n_groups(df)
#> [1] 1095

bench::mark(
  cur_data = mutate(df, x = cur_data()),
  pick = mutate(df, x = pick(everything()))
)
#> # A tibble: 2 × 6
#> 1 cur_data      122ms    126ms      7.98    66.1MB     2.66
#> 2 pick          109ms    109ms      9.17    74.6MB    27.5
```

hadley (commented):
Some initial feedback, mostly on the docs.

R/pick.R
```diff
@@ -0,0 +1,130 @@
...
+#' For `pick()`, the tidyselection provided through `...` is only evaluated once
+#' on the _original_ data frame, and the selection is then reused for all
+#' groups. This allows `pick()` to enforce two important invariants:
+#'
+#' - The columns selected by `pick()` are the same across all groups.
+#'
+#' - The number of rows returned by `pick()` is always equal to the number of
+#'   rows in the current group.
+#'
+#' For `rowwise()` data frames, list-columns are returned as they appear in the
```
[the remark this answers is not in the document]
DavisVaughan: As discussed, we are going all in on the expansion approach, so this means `pick()`ing list-cols in a rowwise data frame typically errors because of the difference in size between normal columns and the list-col elements.

``` r
library(dplyr)

df <- tibble(x = 1:3, y = list(1:2, 1:3, 1:4))
df <- rowwise(df)

# `z = tibble(x = x)`
mutate(df, z = pick(x))
#> # A tibble: 3 × 3
#> # Rowwise: 
#>       x y           z$x
#>   <int> <list>    <int>
#> 1     1 <int [2]>     1
#> 2     2 <int [3]>     2
#> 3     3 <int [4]>     3

# `y` evaluates to list-col elements
# `z = tibble(x = x, y = y)`
mutate(df, z = pick(x, y))
#> Error in `mutate()`:
#> ℹ In argument: `z = pick(x, y)`.
#> ℹ In row 1.
#> Caused by error:
#> ! `z` must be size 1, not 2.
#> ℹ Did you mean: `z = list(pick(x, y))` ?
```

I think this is easy enough to explain by showing how it expands, so I didn't make this an error.

R/across.R
```diff
@@ -160,17 +160,6 @@
...
-#' # across() returns a data frame, which can be used as input of another function
-#' df <- data.frame(
-#'   x1  = c(1, 2, NA),
-#'   x2  = c(4, NA, 6),
-#'   y   = c("a", "b", "c")
-#' )
-#' df %>%
-#'   mutate(x_complete = complete.cases(across(starts_with("x"))))
-#' df %>%
-#'   filter(complete.cases(across(starts_with("x"))))
```
DavisVaughan, line 172: Doubly bad example because it uses `across()` in `filter()` and should be `pick()`. The `pick()` examples feel more useful so I removed this.

R/conditions.R
```diff
@@ -65,7 +65,8 @@ needs_group_context <- function(cnd) {
     "dplyr:::error_incompatible_combine",
     "dplyr:::mutate_mixed_null",
     "dplyr:::mutate_constant_recycle_error",
-    "dplyr:::summarise_mixed_null"
+    "dplyr:::summarise_mixed_null",
+    "dplyr:::error_pick_tidyselect"
```
DavisVaughan, line 69: Skipping group context when `eval_select()` errors are thrown from `pick()`'s expansion tooling, as the `pick()` call is evaluated outside the `...` and the group context won't be set up yet.

R/deprec-context.R
```diff
@@ -0,0 +1,33 @@
...
+#'
+#' @description
+#' `r lifecycle::badge("deprecated")`
+#'
+#' These functions were deprecated in dplyr 1.1.0 in favor of [pick()].
+#'
+#' * `cur_data()` gives the current data for the current group (excluding
+#'   grouping variables).
+#' * `cur_data_all()` gives the current data for the current group (including
+#'   grouping variables).
```
DavisVaughan, line 11: These are now soft-deprecated on their own help page in their own `deprec-context.R` file

R/deprec-context.R
```diff
@@ -0,0 +1,33 @@
...
+  lifecycle::deprecate_soft(when = "1.1.0", what = "cur_data()", with = "pick()")
+  mask <- peek_mask()
+  vars <- mask$current_non_group_vars()
+  mask$pick_current(vars)
+}
+
+#' @rdname deprec-context
+#' @export
+cur_data_all <- function() {
+  lifecycle::deprecate_soft(when = "1.1.0", what = "cur_data_all()", with = "pick()")
```
DavisVaughan, line 29: Calling `deprecate_soft()` does force a 2x slowdown in these functions because it is called at every group evaluation. I figure this is ok.

R/deprec-do.R
```diff
@@ -18,11 +18,11 @@
 #' @export
 #' @examples
 #' # do() with unnamed arguments becomes summarise()
-#' # . becomes across()
+#' # . becomes pick()
 #' by_cyl <- mtcars %>% group_by(cyl)
 #' by_cyl %>% do(head(., 2))
 #' # ->
-#' by_cyl %>% summarise(head(across(), 2))
+#' by_cyl %>% summarise(head(pick(everything()), 2))
```
DavisVaughan, line 25: I'm really appreciating the clarity of seeing an explicit `everything()` when that is really what you mean. Rather than just empty calls to `pick()` or `across()`

vignettes/grouping.Rmd
```diff
@@ -270,29 +270,3 @@ by_species %>%
   filter(!is.na(height)) %>% 
   slice_min(height, n = 2)
 ```
-
-## Computing on grouping information
```
DavisVaughan, line 274: This section felt a little out of place since the grouping vignette is otherwise so high level and targeted at beginners.

It especially doesn't make much sense now that `cur_data()` is soft-deprecated, so I've removed it altogether

DavisVaughan:
Ok, after much discussion `pick()` now uses an _expansion_ approach like `across()`.

The general idea is that:

```r
df %>% mutate(z = pick(x, y))
```

is expanded to:

```r
df %>% mutate(z = tibble(x = x, y = y))
```

before the actual evaluation of the `...` happens. In reality it gets expanded to something slightly different, because `tibble()` itself is too slow to be called at every group iteration.

When `pick()` is wrapped in a user defined function which is then called in a data-masking verb, it does _not_ expand (because it can't find the `pick()` call) and instead falls back to an evaluation based approach that is much slower. I don't expect this to be hit that often, because most people will wrap a full `mutate() + pick()` expression rather than just `pick()`, but it is still heavily tested to ensure it is consistent with the expansion approach.

This avoids most of the weirdness around caching and keys, which can never work correctly 100% of the time, and has some rather common failure cases.

We have gone back to `pick()` being evaluated on the _current_ view of the data. We have gone ALL IN on treating `pick()` like macro expansion from a theoretical perspective. Any weird edge case in `pick()` should be viewed as if the `pick()` call was directly replaced with an equivalent `tibble()` call. Most of the time, you can explain any weirdness by doing that.

---

A few other things I've done:

- Soft-deprecated `cur_data()` and `cur_data_all()`
- Replaced all usage of ^ with `pick()`
- Replaced all usage of `across(.fns = NULL)` with `pick()`. This occured quite a few times in the docs and vignettes, and the new usage of `pick()` feels much better.

I have _not_ change `across(.cols = everything(), .fns = NULL)` to `across(.cols, .fns)` yet. That will take a bit more work and this PR has already gotten quite large. So I'm going to wait and do that in another one.

R/filter.R
```diff
@@ -154,6 +154,7 @@ filter_expand <- function(dots, mask, error_call = caller_env()) {
   env_filter <-  env()
   filter_expand_one <- function(dot, index) {
     env_filter$current_expression <- index
+    dot <- expand_pick(dot, mask)
     expand_if_across(dot)
```
DavisVaughan, line 158: Unlike `across()`, it seems reasonable to use `pick()` directly in `filter()` for something like:

```r
filter(vec_detect_complete(pick(x, y)))
```

NEWS.md
```diff
@@ -1,5 +1,13 @@
 # dplyr (development version)
 
+* `cur_data()` and `cur_data_all()` are now soft-deprecated in favor of
+  `pick()` (#6204).
+
+* New `pick()` function which is intended as a replacement for
```
hadley, line 6:
```suggestion
* New `pick()` function to access full data frame of the current group, and is intended as a replacement for
```

This wording might not be quite right, but I think you should say what it does before saying what it replaces.

R/arrange.R
```diff
@@ -75,8 +75,8 @@
 #' }
 #' tidy_eval_arrange(mtcars, mpg)
 #'
-#' # use across() to access select()-style semantics
-#' iris %>% arrange(across(starts_with("Sepal")))
+#' # Use `across()` or `pick()` to access select()-style semantics
```
hadley, line 78:
```suggestion
#' # Use `across()` or `pick()` to select columns with tidy-select
```
?

R/colwise-distinct.R
```diff
@@ -27,15 +28,15 @@
 #'
 #' distinct_all(df)
 #' # ->
-#' distinct(df, across())
+#' distinct(df, pick(everything()))
 #'
 #' distinct_at(df, vars(x,y))
 #' # ->
-#' distinct(df, across(c(x, y)))
+#' distinct(df, pick(c(x, y)))
```
hadley, line 35:
```suggestion
#' distinct(df, pick(x, y))
```

R/colwise-group-by.R
```diff
@@ -26,17 +27,17 @@
...
 #' group_by_if(iris, is.factor)
 #' # ->
-#' iris %>% group_by(across(where(is.factor)))
+#' iris %>% group_by(pick(where(is.factor)))
 #'
 #' # Group by variables selected by name:
 #' group_by_at(mtcars, vars(vs, am))
 #' # ->
-#' mtcars %>% group_by(across(c(vs, am)))
+#' mtcars %>% group_by(pick(c(vs, am)))
```
hadley, line 40:
```suggestion
#' mtcars %>% group_by(pick(vs, am))
```

R/deprec-context.R
hadley: I think you should be clearer that there's no replacement for `cur_data_all()` because we don't believe it can work.

R/deprec-do.R
hadley: Oh interesting, I had assumed `pick()` would default to `everything()`

R/pick.R
```diff
@@ -0,0 +1,193 @@
+#' Select a subset of columns
+#'
+#' @description
+#' `pick()` provides a way to easily select a subset of columns from your data
```
hadley, line 4: Something about the current group?

R/pick.R
```diff
@@ -0,0 +1,193 @@
...
+#' - With `pick()`, you typically apply a function to the full data frame.
+#' - With `across()`, you typically apply a function to each column.
+#'
+#' @details
+#' Theoretically, `pick()` is intended to be replaceable with an equivalent call
+#' to `tibble()`. For example, `pick(a, c)` could be replaced with
+#' `tibble(a = a, c = c)`, and `pick(everything())` on a data frame with cols
+#' `a`, `b`, and `c` could be replaced with `tibble(a = a, b = b, c = c)`.
+#'
+#' @param ... <[`tidy-select`][dplyr_tidy_select]>
```
hadley, line 20: `@inheritParams across`?

R/pick.R
```diff
@@ -0,0 +1,193 @@
...
+#' df
+#'
+#' # `pick()` provides a way to select a subset of your columns using
+#' # tidyselect. It returns a data frame.
+#' df %>% mutate(cols = pick(x, y))
+#'
+#' # This is useful for functions that take data frames as inputs.
+#' # For example, you can compute a joint rank between `x` and `y`.
+#' df %>% mutate(rank = dense_rank(pick(x, y)))
+#'
```
hadley, line 48: Do we want an `lm()` example? (Or something else with an return value that needs to be wrapped in a list)

vignettes/programming.Rmd
```diff
@@ -409,7 +409,7 @@ You can use this same idea for multiple sets of input data-variables:
 ```{r}
 my_summarise <- function(data, group_var, summarise_var) {
```
hadley, line 410: Also need to update sentence on line 395 to mention `across()` and `pick()`

hadley (approved):
I focussed on the docs and interface, assuming that @lionel- will focus more on the implementation.

lionel- (approved):

R/pick.R
DavisVaughan: They are `...` in `pick()` but `.cols` in `across()`. Plus the wording is slightly different to emphasize that in `across()` you _transform_ the columns but in `pick()` you just select them

R/pick.R
DavisVaughan: I can't seem to come up with any good examples, any ideas besides `lm()`?

R/across.R
```diff
@@ -465,12 +454,17 @@ data_mask_top <- function(env, recursive = FALSE, inherit = FALSE) {
...
+  quo_set_env(quo, env)
+}
 
 c_across_setup <- function(cols, mask) {
   cols <- enquo(cols)
-  across_cols <- mask$across_cols()
+  data <- mask$get_current_data(groups = FALSE)
 
-  vars <- tidyselect::eval_select(expr(!!cols), across_cols)
+  vars <- tidyselect::eval_select(expr(!!cols), data)
```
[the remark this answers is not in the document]
DavisVaughan: This actually feels somewhat tricky and is specific to `c_across()` so I've opened https://github.com/tidyverse/dplyr/issues/6522 to look into it

R/pick.R
```diff
@@ -0,0 +1,193 @@
...
+  mask <- peek_mask()
+
+  # Evaluates `pick()` on current columns.
+  # Mimicking expansion as much as possible, which should match the idea of
+  # replacing the `pick()` call directly with `tibble()`, like:
+  # pick(a, b, starts_with("foo")) -> tibble(a = a, b = b, foo1 = foo1)
+  non_group_vars <- mask$current_non_group_vars()
+  data <- mask$current_cols(non_group_vars)
+
+  # `pick()` is evaluated in a data mask so we need to remove the
```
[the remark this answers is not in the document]
DavisVaughan: Yea like if `pick()` is just called at the top level it fails in `peek_mask()`

This is admittedly a weird part of `pick()` and `across()` that I don't fully understand

R/pick.R
```diff
@@ -0,0 +1,193 @@
...
+  if (is_call(expr, name = "pick", ns = c("", "dplyr"))) {
+    expr <- as_pick_selection(expr, error_call)
+    out <- eval_pick(expr, env, mask, error_call)
+    out <- as_pick_expansion(out)
+    return(out)
+  }
+
+  index <- seq2(2L, length(expr))
+
+  for (i in index) {
```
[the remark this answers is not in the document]
DavisVaughan: Good idea

R/deprec-do.R
DavisVaughan: I guess we think `pick(everything())` is something that people would do often?

DavisVaughan: We were planning on moving away from `across(.cols = everything()` in favor of `across(.cols`, if that makes any difference

R/pick.R
```diff
@@ -0,0 +1,200 @@
...
+    expr <- expand_pick_impl(expr, env, mask, error_call)
+  }
+
+  out <- new_quosure(expr, env = env)
+  out <- new_dplyr_quosure(out, !!!attr(quo, "dplyr:::data"))
+
+  out
+}
+
+expand_pick_impl <- function(expr, env, mask, error_call = caller_env()) {
```
lionel-, line 114: By the way, this should also traverse quosures.

R/deprec-do.R
DavisVaughan: Decided internally to currently make `pick()` with no arguments an error, and potentially relax that to mean "everything" in the future if heavily requested

PR #6496, opened 2022-10-10
Evaluate each expression in `arrange()` separately

### LINKED ISSUE #6495: error in arrange when using across(everything())
Using dplyr_1.0.99.9000:
``` r
    library(tidyverse)
    df <- tibble(
      letter1 = c("A", "C", "B"),
      letter2 = c("B", "C", "B"),
      number = c(3, 1, 2)
    )
    df %>%
      arrange(across(everything()))
#> # A tibble: 3 × 3
#>   letter1 letter2 number
#>   <chr>   <chr>    <dbl>
#> 1 A       B            3
#> 2 B       B            2
#> 3 C       C            1
    df %>%
      arrange(
        letter1 != letter2,
        across(everything())
        )
#> Error in `arrange()`:
#> ℹ In argument: `..2 = across(everything())`.
#> Caused by error:
#> ! ..1 used in an incorrect context, no ... to look in

#> Backtrace:
#>      ▆
#>   1. ├─df %>% arrange(letter1 != letter2, across(everything()))
#>   2. ├─dplyr::arrange(., letter1 != letter2, across(everything()))
#>   3. ├─dplyr:::arrange.data.frame(., letter1 != letter2, across(everything()))
#>   4. │ └─dplyr:::arrange_rows(.data, dots = dots, locale = .locale)
#>   5. │   ├─dplyr::mutate(new_data_frame(data), !!!quosures, .keep = "none")
#>   6. │   └─dplyr:::mutate.data.frame(...)
#>   7. │     └─dplyr:::mutate_cols(.data, dplyr_quosures(...))
#>   8. │       ├─base::withCallingHandlers(...)
#>   9. │       └─dplyr:::mutate_col(dots[[i]], .data, mask, new_columns)
#>  10. │         └─mask$eval_all_mutate(quo)
#>  11. │           └─dplyr (local) eval()
#>  12. ├─dplyr::across(everything())
#>  13. │ └─mask$pick(vars)
#>  14. │   └─self$current_cols(vars)
#>  15. │     └─rlang::env_get_list(parent.env(private$mask), vars)
#>  16. ├─`<fn>`()
#>  17. └─base::.handleSimpleError(...)
#>  18.   └─dplyr (local) h(simpleError(msg, call))
#>  19.     └─rlang::abort(message, class = error_class, parent = parent, call = error_call)
```

DavisVaughan:
Closes #6495 

I've ended up with an approach that just evaluates each expression provided to `arrange()` in its own `mutate()` call. I don't expect the performance here to matter because it is always on ungrouped data frames and the `mutate()` call itself has a fairly small amount of overhead.

This means that each `across()` call will be evaluated on the original data, preventing `everything()` from accessing dummy columns added after evaluating earlier expressions.

I'm actually still using `..1`, `..2`, etc as the column names here, because they need to be called _something_ and that something ends up as part of the error message, so this seemed like the best option.

### DIFF
--- a/R/arrange.R
+++ b/R/arrange.R
@@ -106,28 +106,64 @@ arrange_rows <- function(data,
                          error_call = caller_env()) {
   dplyr_local_error_call(error_call)
 
+  size <- nrow(data)
+
+  # `arrange()` implementation always uses bare data frames
+  data <- new_data_frame(data, n = size)
+
   # Strip out calls to desc() replacing with direction argument
   is_desc_call <- function(x) {
     quo_is_call(x, "desc", ns = c("", "dplyr"))
   }
-  directions <- map_chr(dots, function(quosure) {
-    if (is_desc_call(quosure)) "desc" else "asc"
+  directions <- map_chr(dots, function(dot) {
+    if (is_desc_call(dot)) "desc" else "asc"
   })
-  quosures <- map(dots, function(quosure) {
-    if (is_desc_call(quosure)) {
-      expr <- quo_get_expr(quosure)
+  dots <- map(dots, function(dot) {
+    if (is_desc_call(dot)) {
+      expr <- quo_get_expr(dot)
       if (!has_length(expr, 2L)) {
         abort("`desc()` must be called with exactly one argument.", call = error_call)
       }
 
-      quosure <- new_quosure(expr[[2]], quo_get_env(quosure))
+      dot <- new_quosure(expr[[2]], quo_get_env(dot))
     }
-    quosure
+    dot
   })
 
-  names(quosures) <- vec_paste0("..", seq_along(quosures))
-  data <- mutate(new_data_frame(data), !!!quosures, .keep = "none")
-  directions <- directions[names(quosures) %in% names(data)]
+  n_dots <- length(dots)
+  seq_dots <- seq_len(n_dots)
+
+  cols <- vector("list", length = n_dots)
+  names(cols) <- as.character(seq_dots)
+
+  for (i in seq_dots) {
+    name <- vec_paste0("..", i)
+    dot <- dots[[i]]
+
+    # Evaluate each `dot` on the original data separately, rather than
+    # evaluating all at once. We want to avoid the "sequential evaluation"
+    # feature of `mutate()` where the 2nd expression can access results of the
+    # 1st (#6495).
+    elt <- mutate(data, "{name}" := !!dot, .keep = "none")
+    elt <- elt[[name]]
+
+    if (is.null(elt)) {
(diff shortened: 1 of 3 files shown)


R/arrange.R
```diff
@@ -106,28 +106,64 @@ arrange_rows <- function(data,
...
+    # feature of `mutate()` where the 2nd expression can access results of the
+    # 1st (#6495).
+    elt <- mutate(data, "{name}" := !!dot, .keep = "none")
+    elt <- elt[[name]]
+
+    if (is.null(elt)) {
+      # In this case, `dot` evaluated to `NULL` for "column removal" so
+      # `elt[[name]]` won't exist, but we don't want to shorten `cols`.
+      next
+    }
```
DavisVaughan, line 154: Stupid tricky case

R/arrange.R
```diff
@@ -106,28 +106,64 @@ arrange_rows <- function(data,
...
+
+    cols[[i]] <- elt
+  }
+
+  if (vec_any_missing(cols)) {
+    # Drop `NULL`s that could result from `dot` evaluating to `NULL` above
+    not_missing <- !vec_detect_missing(cols)
+    cols <- vec_slice(cols, not_missing)
+    directions <- vec_slice(directions, not_missing)
+  }
```
DavisVaughan, line 164: i.e. `arrange(df, NULL, y, NULL)` where the mutate calls look like `mutate(df, ..1 = NULL)` meaning `elt[[..1]]` won't exist and the `cols` element will be left as `NULL`.

Careful to also slice `directions` here to keep them synced

R/arrange.R
```diff
@@ -106,28 +106,64 @@ arrange_rows <- function(data,
...
+
+  for (i in seq_dots) {
+    name <- vec_paste0("..", i)
+    dot <- dots[[i]]
+
+    # Evaluate each `dot` on the original data separately, rather than
+    # evaluating all at once. We want to avoid the "sequential evaluation"
+    # feature of `mutate()` where the 2nd expression can access results of the
+    # 1st (#6495).
+    elt <- mutate(data, "{name}" := !!dot, .keep = "none")
```
hadley, line 147: Another place where we need a function that uses mutate to compute a single column.

hadley: Do we need a test of the error message when `!!dot` errors? Or does that already exist?

hadley (approved):

R/arrange.R
DavisVaughan: The snapshot errors here display the `{name}`
https://github.com/tidyverse/dplyr/blob/98f754ca20e501282912ded7f154de5b0e1e9ac8/tests/testthat/test-arrange.R#L27-L43

This test is actually how I remembered that we need to handle the name specially.

PR #6500, opened 2022-10-12
Use `enquos()` + `quo_squash()` in `join_by()`

### LINKED ISSUE #6469: Consider how folks will want to program with `join_by()`
i.e. it's neither tidy-select not data-masking

DavisVaughan:
Closes #6469 

To allow `{{ }}`, since `enexprs()` doesn't allow that.

I spent quite a bit of time yesterday re-convincing myself that this is the best way to handle this, and I think it is.

### DIFF
--- a/R/join-by.R
+++ b/R/join-by.R
@@ -354,18 +354,18 @@ join_by_common <- function(x_names,
 # these functions don't have an `error_call` argument.
 
 parse_join_by_expr <- function(expr, i, error_call) {
-  if (length(expr) == 0L) {
-    message <- c(
-      "Join by expressions can't be empty.",
-      x = glue("Expression {i} is empty.")
-    )
-    abort(message, call = error_call)
-  }
-
   if (is_missing(expr)) {
     message <- c(
       "Join by expressions can't be missing.",
       x = glue("Expression {i} is missing.")
+    )
+    abort(message, call = error_call)
+  }
+
+  if (length(expr) == 0L) {
+    message <- c(
+      "Join by expressions can't be empty.",
+      x = glue("Expression {i} is empty.")
     )
     abort(message, call = error_call)
   }
--- a/R/join-by.R
+++ b/R/join-by.R
@@ -189,10 +189,10 @@
 #' by <- join_by(chromosome, overlaps(x$start, x$end, y$start, y$end))
 #' full_join(segments, reference, by)
 join_by <- function(...) {
-  # `join_by()` captures pure expressions, but we want to allow `{{ }}`. Since
-  # embracing doesn't work with `enexprs()`, the second best option is to use
-  # `enquos()` and immediately squash all quosures to recursively extract the
-  # expressions.
+  # `join_by()` works off pure expressions with no evaluation in the user's
+  # environment, but we want to allow `{{ }}` to make it easier to program with.
+  # The best way to do this is to capture quosures with `enquos()`, and then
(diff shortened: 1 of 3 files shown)


R/join-by.R
```diff
@@ -357,6 +362,14 @@ parse_join_by_expr <- function(expr, i, error_call) {
     abort(message, call = error_call)
   }
 
+  if (is_missing(expr)) {
+    message <- c(
+      "Join by expressions can't be missing.",
+      x = glue("Expression {i} is missing.")
+    )
```
lionel-, line 369: "Join by" is redundant with the error call IIRC?

Maybe follow the new pattern used in `check_` functions for absent arguments?

```
Error in `join_by()`:
! `..2` must be a join specification, it can't be absent.
```

lionel-: By the way, if `expr` is the missing arg, the `length(expr)` call above will fail.

tests/testthat/test-join-by.R
```diff
@@ -152,13 +152,63 @@ test_that("between / overlaps / within / closest can use named arguments", {
...
+  fn <- function(x, y) {
+    join_by({{x}}, {{y}})
+  }
+  expect_snapshot(error = TRUE, fn(a))
+
+  fn <- function(x, y) {
+    join_by({{x}} == {{y}})
+  }
+  expect_snapshot(error = TRUE, fn(a))
+})
```
lionel-, line 210: It's nice to include the definitions in the snapshot so it can be understood in isolation.

lionel- (approved):

R/join-by.R
```diff
@@ -189,7 +189,12 @@
 #' by <- join_by(chromosome, overlaps(x$start, x$end, y$start, y$end))
 #' full_join(segments, reference, by)
 join_by <- function(...) {
-  exprs <- enexprs(..., .named = NULL)
+  # `join_by()` captures pure expressions, but we want to allow `{{ }}`. Since
+  # embracing doesn't work with `enexprs()`, the second best option is to use
```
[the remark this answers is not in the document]
DavisVaughan: Oh duh of course it supports it

``` r
fn <- function(...) {
  rlang::enexprs(...)
}
x <- 1
fn({{x}} + 2)
#> [[1]]
#> (~1) + 2
```

<sup>Created on 2022-10-13 with [reprex v2.0.2.9000](https://reprex.tidyverse.org)</sup>

R/join-by.R
DavisVaughan: Will it?

```r
> length(rlang::missing_arg())
[1] 1
```

DavisVaughan: I'll do a round of error updates in another PR, I think quite a few could be updated

tests/testthat/test-join-by.R
DavisVaughan: Unfortunately `{{ }}` gets parsed like this though, making it essentially useless

Screen Shot 2022-10-13 at 2 52 02 PM

R/join-by.R
lionel-: yup because you're evaluating a SYMSXP not a LANGSXP:

```r
expr <- missing_arg()
length(expr)
#> Error:
#> ! argument "expr" is missing, with no default
```

tests/testthat/test-join-by.R
lionel-: ooops, reported at https://github.com/r-lib/rlang/issues/1511

PR #6503, opened 2022-10-12
Disallow `keep = FALSE` with non-equi conditions

### LINKED ISSUE #6499: Issue with non equi joins, repeated column names, and `keep = FALSE`
With non equi joins we allow the same column to be used in multiple expressions, because you need to do this to specify bounds correctly. Seems that doesn't play well with `keep = FALSE`.

This, of course, also affects `between()` because it generates expressions exactly like these

``` r
library(dplyr)

df1 <- tibble(x = 1)
df2 <- tibble(lower = 0, upper = 2)

# Fine (default)
left_join(df1, df2, join_by(x >= lower, x <= upper), keep = TRUE)
#> # A tibble: 1 × 3
#>       x lower upper
#>   <dbl> <dbl> <dbl>
#> 1     1     0     2

# Hmm
left_join(df1, df2, join_by(x >= lower, x <= upper), keep = FALSE)
#> Error in `left_join()`:
#> ! Join columns must be unique.
#> ✖ Problem with `x`.

#> Backtrace:
#>     ▆
#>  1. ├─dplyr::left_join(df1, df2, join_by(x >= lower, x <= upper), keep = FALSE)
#>  2. └─dplyr:::left_join.data.frame(df1, df2, join_by(x >= lower, x <= upper), keep = FALSE)
#>  3.   └─dplyr:::join_mutate(...)
#>  4.     └─dplyr:::join_cols(...)
#>  5.       └─dplyr:::check_join_vars(by$x, x_names, by$condition, keep, error_call = error_call)
#>  6.         └─rlang::abort(bullets, call = error_call)
```

DavisVaughan:
Closes #6499 

See https://github.com/tidyverse/dplyr/issues/6499#issuecomment-1276346314 for rationale and good examples of why we shouldn't allow `keep = FALSE` with non-equi joins. Those examples have been turned into snapshot tests that now error informatively.

### DIFF
--- a/R/join-cols.R
+++ b/R/join-cols.R
@@ -9,7 +9,7 @@ join_cols <- function(x_names,
 
   if (is_false(keep) && any(by$condition != "==")) {
     abort(
-      "Can't set `keep = FALSE` when using a non-equi, rolling, or overlap join.",
+      "Can't set `keep = FALSE` when using an inequality, rolling, or overlap join.",
       call = error_call
     )
   }
--- a/R/join-cols.R
+++ b/R/join-cols.R
@@ -7,11 +7,18 @@ join_cols <- function(x_names,
                       error_call = caller_env()) {
   check_dots_empty0(...)
 
+  if (is_false(keep) && any(by$condition != "==")) {
+    abort(
+      "Can't set `keep = FALSE` when using a non-equi, rolling, or overlap join.",
+      call = error_call
+    )
+  }
+
   check_duplicate_vars(x_names, "x", error_call = error_call)
   check_duplicate_vars(y_names, "y", error_call = error_call)
 
-  check_join_vars(by$x, x_names, by$condition, keep, "x", error_call = error_call)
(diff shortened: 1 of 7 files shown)


R/join-cols.R
```diff
@@ -7,11 +7,18 @@ join_cols <- function(x_names,
                       error_call = caller_env()) {
   check_dots_empty0(...)
 
+  if (is_false(keep) && any(by$condition != "==")) {
+    abort(
+      "Can't set `keep = FALSE` when using a non-equi, rolling, or overlap join.",
```
hadley, line 12: Maybe use inequality instead of non-equi since rolling and overlap are types of non-equi join?

hadley (approved):

R/join-cols.R
DavisVaughan: I should probably update the `join_by()` docs to reflect that too. I think `Equality, Inequality, Rolling, and Overlap` sound nice as the 4 main join types

PR #6504, opened 2022-10-13
Add `between(bounds =)` and `overlaps(bounds =)`

### LINKED ISSUE #6488: Should `between()`, `overlaps()`, and `within()` use `[ )` bounds?
```r
> parties %>%
+   mutate(
+     start = replace(party, 1, as.Date("2022-01-01")),
+     end = lead(start, default = as.Date("2022-12-31"))
+   )
# A tibble: 4 × 4
      q party      start      end       
  <int> <date>     <date>     <date>    
1     1 2022-01-10 2022-01-01 2022-04-04
2     2 2022-04-04 2022-04-04 2022-07-11
3     3 2022-07-11 2022-07-11 2022-10-03
4     4 2022-10-03 2022-10-03 2022-12-31
```

These bounds are easy to make with lag/lead but aren't quite right for `overlaps()` because they will produce duplicate matches on the end->start endpoints because it uses `[,]`

DavisVaughan:
Closes #6488 

Highlights of my analysis in #6488:

- Kept `[]` default in `between()` and `overlaps()` for compatibility with `dplyr::between()` and how I think these are most likely to be used

- `between()` gains `bounds = "[]"` because `[]`, `[)`, and `()` all have their uses (and `(]` is then for completeness)

- `overlaps()` gains `closed = TRUE` because `xl <= yu, xu >= yl` covers `[]` and `xl < yu, xu > yl` covers the other 3 partially/fully open range cases.

- The binary conditions for `within()` are the same regardless of the inclusiveness of the range boundaries, so it doesn't gain any new arguments.

This would allow me to talk about `between(bounds = "[)")` and `overlaps(closed = FALSE)` in ivs, which would align with `iv_locate_between()` and `iv_locate_overlaps()`.

And R4DS could use `between(bounds = "[)")` which makes generation of `start` and `end` simpler and less error prone. Plus I find it a little easier to read row-by-row (i.e. `end` of row `i` is the same as the `start` of row `i+1`)

``` r
employees <- tibble(
  name = wakefield::name(100),
  birthday = lubridate::ymd("2022-01-01") + (sample(365, 100, replace = TRUE) - 1)
)

parties <- tibble(
  q = 1:4,
  party = lubridate::ymd(c("2022-01-10", "2022-04-04", "2022-07-11", "2022-10-03"))
)

# Use lag/lead/diff techniques to generate `start` and `end`
# (this one is a little complicated from the specifics of the R4DS example)
parties <- parties %>%
  mutate(
    start = replace(party, 1L, as.Date("2022-01-01")),
    end = lead(start, default = as.Date("2022-12-31"))
  )
parties
#> # A tibble: 4 × 4
#>       q party      start      end       
#>   <int> <date>     <date>     <date>    
#> 1     1 2022-01-10 2022-01-01 2022-04-04
#> 2     2 2022-04-04 2022-04-04 2022-07-11
#> 3     3 2022-07-11 2022-07-11 2022-10-03
#> 4     4 2022-10-03 2022-10-03 2022-12-31

employees |> 
  inner_join(
    parties, 
    join_by(between(birthday, start, end, bounds = "[)")), 
    unmatched = "error"
  )
#> # A tibble: 100 × 6
#>    name       birthday       q party      start      end       
#>    <variable> <date>     <int> <date>     <date>     <date>    
#>  1 Optimus    2022-09-09     3 2022-07-11 2022-07-11 2022-10-03
#>  2 Brij       2022-12-13     4 2022-10-03 2022-10-03 2022-12-31
#>  3 Kaytin     2022-03-11     1 2022-01-10 2022-01-01 2022-04-04
#>  4 Turki      2022-03-27     1 2022-01-10 2022-01-01 2022-04-04
#>  5 Panfilo    2022-11-19     4 2022-10-03 2022-10-03 2022-12-31
#>  6 Catori     2022-02-22     1 2022-01-10 2022-01-01 2022-04-04
#>  7 Jeselle    2022-01-22     1 2022-01-10 2022-01-01 2022-04-04
#>  8 Donte      2022-09-13     3 2022-07-11 2022-07-11 2022-10-03
#>  9 Tishonda   2022-07-07     2 2022-04-04 2022-04-04 2022-07-11
#> 10 Darlien    2022-05-07     2 2022-04-04 2022-04-04 2022-07-11
#> # … with 90 more rows
```

### DIFF
--- a/R/join-by.R
+++ b/R/join-by.R
@@ -80,16 +80,16 @@
 #'   The binary conditions used to build `within()` are the same regardless of
 #'   the inclusiveness of the supplied ranges.
 #'
-#' - `overlaps(x_lower, x_upper, y_lower, y_upper, ..., closed = TRUE)`
+#' - `overlaps(x_lower, x_upper, y_lower, y_upper, ..., bounds = "[]")`
 #'
 #'   For each range in `[x_lower, x_upper]`, this finds everywhere that range
 #'   overlaps `[y_lower, y_upper]` in any capacity. Equivalent to `x_lower <=
 #'   y_upper, x_upper >= y_lower` by default.
 #'
-#'   `closed` controls whether or not the ranges are treated as fully closed
-#'   intervals, like \code{"[]"}. If `FALSE`, the binary conditions `x_lower <
-#'   y_upper, x_upper > y_lower` are used, which work for \code{"[)"},
-#'   \code{"(]"}, and \code{"()"}.
+#'   `bounds` can be one of \code{"[]"}, \code{"[)"}, \code{"(]"}, or
+#'   \code{"()"} to alter the inclusiveness of the lower and upper bounds.
+#'   \code{"[]"} uses `<=` and `>=`, but the 3 other options use `<` and `>`
+#'   and generate the exact same binary conditions.
 #'
 #'   Dots are for future extensions and must be empty.
 #'
@@ -210,9 +210,8 @@
 #'
 #' # It is common to have right-open ranges with bounds like `[)`, which would
 #' # mean an end value of `415` would no longer overlap a start value of `415`.
-#' # Setting `closed = FALSE` allows you to compute overlaps with those kinds
-#' # of ranges.
-#' by <- join_by(chromosome, overlaps(x$start, x$end, y$start, y$end, closed = FALSE))
+#' # Setting `bounds` allows you to compute overlaps with those kinds of ranges.
+#' by <- join_by(chromosome, overlaps(x$start, x$end, y$start, y$end, bounds = "[)"))
 #' full_join(segments, reference, by)
 join_by <- function(...) {
   # `join_by()` works off pure expressions with no evaluation in the user's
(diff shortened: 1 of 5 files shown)


R/join-by.R
```diff
@@ -58,25 +58,44 @@
...
+#'
+#' - `overlaps(x_lower, x_upper, y_lower, y_upper, ..., closed = TRUE)`
 #'
 #'   For each range in `[x_lower, x_upper]`, this finds everywhere that range
 #'   overlaps `[y_lower, y_upper]` in any capacity. Equivalent to `x_lower <=
-#'   y_upper, x_upper >= y_lower`.
+#'   y_upper, x_upper >= y_lower` by default.
+#'
+#'   `closed` controls whether or not the ranges are treated as fully closed
+#'   intervals, like \code{"[]"}. If `FALSE`, the binary conditions `x_lower <
```
hadley, line 90: Are you sure it’s not worth using `bounds` here too, just in the interests of reducing the total number of parameters?

hadley (approved):

R/join-by.R
DavisVaughan: Hmm, in which case `bounds = "[]"` would do `closed = TRUE`, and `bounds = "[)" / "(]" / "()"` would do `closed = FALSE`.

So 3 of the possible values would map to the same behavior but it might be easier to understand in the long term when comparing with `between()` and when writing it as an end user.

hadley: Yeah, that's what I was thinking.

hadley (approved):

PR #6513, opened 2022-10-20
Improve vars() documentation

hadley:


msberends:
Wow, your dedication is awesome!!

I think it’s a great improvement. What do you think of: “`vars()` is superseded for its need in the scoped verbs…”, instead of: “`vars()` is superseded because it was needed for the scoped verbs…”?

I’m still searching for a way to lose the “was”, since it will stay around for probably a long time.

hadley:
Oh I think I can just make that "is"

msberends:
👍
Good to go!

PR #6523, opened 2022-11-02
Quietly deprecate optional `.cols` and `.fns` cases

DavisVaughan:
Follow up to #6492 

`.cols` and `.fns` are now required for `across()`, `c_across()`, `if_all()`, and `if_any()`.

We recommend in the NEWS that you use `pick()` instead of `across(.fns = NULL)`, which was the most common case of how people relied on these defaults

~This deprecation is currently hidden, all calls to these functions that rely on the previous defaults still work with no messages or warnings. I think that is definitely the right thing to do for `.fns = NULL`, because people would often do `across(c(x, y))` to mimic what `pick()` now does, and previously there was no other option. But I wonder if we could go ahead and soft deprecate the `.cols` case?~ The deprecation is hidden for `.fns = NULL`, but we now use `deprecate_warn()` for missing `.cols`

### DIFF
--- a/R/across.R
+++ b/R/across.R
@@ -12,6 +12,9 @@
 #' the predicate is `TRUE` for *any* of the selected columns, `if_all()`
 #' is `TRUE` when the predicate is `TRUE` for *all* selected columns.
 #'
+#' If you just need to select columns without applying a transformation to each
+#' of them, then you probably want to use [pick()] instead.
+#'
 #' `across()` supersedes the family of "scoped variants" like
 #' `summarise_at()`, `summarise_if()`, and `summarise_all()`.
 #'
@@ -27,9 +30,6 @@
 #'     `list(mean = mean, n_miss = ~ sum(is.na(.x))`. Each function is applied
 #'     to each column, and the output is named by combining the function name
 #'     and the column name using the glue specification in `.names`.
-#'   - `NULL`: the default, returns the selected columns in a data frame
-#'     without applying a transformation. This is useful for when you want to
-#'     use a function that takes a data frame.
 #'
 #'   Within these functions you can use [cur_column()] and [cur_group()]
 #'   to access the current column and grouping keys respectively.
@@ -168,13 +168,26 @@
 #'
 #' @export
 #' @seealso [c_across()] for a function that returns a vector
-across <- function(.cols = everything(),
-                   .fns = NULL,
+across <- function(.cols,
+                   .fns,
                    ...,
                    .names = NULL,
                    .unpack = FALSE) {
   mask <- peek_mask()
   caller_env <- caller_env()
+
+  .cols <- enquo(.cols)
+
+  if (quo_is_missing(.cols)) {
+    # Silent restoration to old defaults of `.cols` for now.
+    # TODO: Escalate this to formal deprecation.
+    .cols <- quo_set_expr(.cols, expr(everything()))
+  }
+  if (is_missing(.fns)) {
+    # Silent restoration to old defaults of `.fns` for now.
+    # TODO: Escalate this to formal deprecation.
+    .fns <- NULL
+  }
 
   if (!is_bool(.unpack) && !is_string(.unpack)) {
     stop_input_type(.unpack, "`TRUE`, `FALSE`, or a single string")
@@ -188,7 +201,7 @@ across <- function(.cols = everything(),
   }
 
   setup <- across_setup(
-    {{ .cols }},
+    cols = !!.cols,
     fns = .fns,
     names = .names,
     .caller_env = caller_env,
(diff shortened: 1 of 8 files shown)


R/across.R
```diff
@@ -168,14 +168,27 @@
...
                    ...,
                    .names = NULL,
                    .unpack = FALSE) {
   mask <- peek_mask()
   caller_env <- caller_env()
 
+  .cols <- enquo(.cols)
+
+  if (quo_is_missing(.cols)) {
+    # Silent restoration to old defaults of `.cols` for now.
```
hadley, line 182: I think you should do a soft deprecation here. I don't think it will be too painful for interactive use because you'll only get the warnings every 8 hours.

DavisVaughan: Yea I think you're right

R/across.R
```diff
@@ -738,6 +766,45 @@ is_inlinable_formula <- function(x, mask) {
...
+  })
+
+  if (is_null(cnd)) {
+    # Condition wasn't signaled
+    return(NULL)
+  }
+
+  # Subclassed so we can skip computing group context info when the warning
+  # is thrown from `expand_across()` outside of any group
+  class(cnd) <- c("dplyr:::warning_across_missing_cols_deprecated", class(cnd))
```
DavisVaughan, line 794: This is a bit more complicated than I'd like, but I think we have to do it.

Without this, dplyr will try and add group context to the warning, but this warning is thrown from `expand_across()` which is outside of any group context, and an error is thrown.

We have a system in place to handle a few other cases like this, which involves checking if the condition is one of a set of known errors/warnings to "skip" computing group context for, but that requires adding a custom class to check against.

tests/testthat/_snaps/across.md
```diff
@@ -230,6 +230,161 @@
...
+      ! Using `across()` without supplying `.cols` was deprecated in dplyr 1.1.0.
+      i Please supply `.cols` instead.
+
+---
+
+    Code
+      out <- mutate(gdf, z = across())
+    Condition
+      Warning:
+      There were 2 warnings in `mutate()`.
```
DavisVaughan, line 251: Only 1 warning is actually thrown when called interactively. The verbosity option is set to always warn, so we get 2 here when there are 2 groups

DavisVaughan: I originally didn't have tests for the grouped case, but since I had to make a few internal changes to actually support throwing the warning in the grouped case, it felt important to add some tests cases too.

R/across.R
lionel-: I guess we should add a `class` argument to `deprecate_warn()` to make this sort of things easier? Probably too late to add `...` for attributes, but we could add `fields = list()` instead.

R/across.R
```diff
@@ -738,6 +766,45 @@ is_inlinable_formula <- function(x, mask) {
...
 }
 
+across_missing_cols_deprecate_warn <- function() {
+  across_if_fn <- context_peek_bare("across_if_fn") %||% "across"
+
+  # Passing the correct `user_env` through `expand_across()` to here is
+  # complicated, so instead we force the global environment. This means users
+  # won't ever see the "deprecated feature was likely used in the {pkg}"
+  # message, but the warning will still fire and that is more important.
+  user_env <- global_env()
```
lionel-, line 776: Nice pattern, this should be documented in lifecycle: https://github.com/r-lib/lifecycle/issues/146

It'd be nice to restore the infrastructure to obtain the user env though.

lionel- (approved):

R/across.R
```diff
@@ -168,14 +168,26 @@
 #'
 #' @export
 #' @seealso [c_across()] for a function that returns a vector
-across <- function(.cols = everything(),
-                   .fns = NULL,
+across <- function(.cols,
```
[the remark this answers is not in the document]
DavisVaughan: > we made across() without args work in dplyr 1.0 

That was the version `across()` was introduced in though, I don't think we decided that it should be `everything()` after hearing from users that they wanted it to be that way. It was just this commit https://github.com/tidyverse/dplyr/commit/57f2c27095e2b93ecff175acb952b354fc9faf12

We want to discourage people from doing `across(, .fns = mean)` to apply mean to every column in an ugly way, so I do think this is the right change to make jointly with the `.fns` change.

---

But I guess you are talking about the fact that transitioning from `across()` could look like this for anyone who previously used it to select everything.

```r
across() -> pick(everything())
```

I still don't mind this too much, I like that it is explicit, but if you feel strongly about it I would be willing to make `pick()` with no arguments select everything.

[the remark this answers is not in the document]
lionel-: I think materialising the current data seems like such a fundamental operation (this does not mean common) that it's strange that it requires another argument that is so much longer. I don't feel very strongly about it though.

hadley (approved):

R/across.R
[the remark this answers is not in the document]
hadley: I think we should keep it strict for now, and then if we hear a lot of complaints we can relax it so `pick()` is equivalent to `pick(everything())`.

PR #6524, opened 2022-11-02
Only throw the `multiple` warning on "direct" calls to join functions

DavisVaughan:
This uses the same code as `lifecycle::deprecate_soft()`, which means that the following two cases _do_ still get the warning:
- dplyr users that directly call the join functions will see the warning
- Package authors that directly call join functions in their package code will see the warning in their tests

But users or package authors who _indirectly_ call a join function through some other package will not see the join warning, because they likely won't have access to `multiple` to silence it (there are probably some exceptions to this, like packages that write S3 methods for the join functions and would expose `multiple` in their method, but I'm willing to live with accidental silence there).

For example, pretty much any package author or user who called `tidyr::complete()` was probably getting a multiple match warning because it called `full_join()` and often produced multiple matches on purpose. It will no longer produce warnings in those cases, but when we run the tidyr tests ourselves we should still see the warnings.

### DIFF
--- a/R/join-rows.R
+++ b/R/join-rows.R
@@ -8,7 +8,8 @@ join_rows <- function(x_key,
                       cross = FALSE,
                       multiple = NULL,
                       unmatched = "drop",
-                      error_call = caller_env()) {
+                      error_call = caller_env(),
+                      user_env = caller_env()) {
   check_dots_empty0(...)
 
   type <- arg_match(type, error_call = error_call)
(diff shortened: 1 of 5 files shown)


R/join-rows.R
```diff
@@ -8,7 +8,8 @@ join_rows <- function(x_key,
                       cross = FALSE,
                       multiple = NULL,
                       unmatched = "drop",
-                      error_call = caller_env()) {
+                      error_call = caller_env(),
+                      user_env = caller_env()) {
```
DavisVaughan, line 12: I purposefully use `user_env = caller_env()`, not `caller_env(2)`, here. I think that makes sense for `deprecate_soft()` because that is an exported function and lines up with where it is going to be called from 99% of the time, but here I have _a lot_ of tests that call `join_rows()` directly, so the `user_env` needs to be meaningful when `join_rows()` is called on its own.

DavisVaughan: > join_rows() is not exported and is never called directly

The problem is that we have a _ton_ of tests that do call `join_rows()` directly. 
https://github.com/tidyverse/dplyr/blob/main/tests/testthat/test-join-rows.R

When called directly, the `user_env` of `join_rows()` is equivalent to the `caller_env()`, so that needs to be the default unless we want to update most/all of the tests with an explicit `user_env = current_env()` argument

The trade off is that we then have to specify `user_env = caller_env()` in the few places that we call `join_rows()` in the join functions themselves, but this felt acceptable to me

Issue #6509, opened 2022-10-18
`summarise()` edge case recycling bug


DavisVaughan:
``` r
library(dplyr)
#> 
#> Attaching package: 'dplyr'
#> The following objects are masked from 'package:stats':
#> 
#>     filter, lag
#> The following objects are masked from 'package:base':
#> 
#>     intersect, setdiff, setequal, union

df <- tibble(x = 1:3)

# Why doesn't this recycle to size 0?
summarise(df, sum = sum(x), empty = tibble())
#> Error in `dplyr_col_modify()`:
#> ! Can't recycle `empty` (size 0) to size 1.

#> Backtrace:
#>     ▆
#>  1. ├─dplyr::summarise(df, sum = sum(x), empty = tibble())
#>  2. ├─dplyr:::summarise.data.frame(df, sum = sum(x), empty = tibble())
#>  3. │ └─dplyr:::summarise_build(.data, cols)
#>  4. │   ├─dplyr::dplyr_col_modify(out, cols$new)
#>  5. │   └─dplyr:::dplyr_col_modify.data.frame(out, cols$new)
#>  6. │     └─vctrs::vec_recycle_common(!!!cols, .size = nrow(data))
#>  7. └─vctrs:::stop_recycle_incompatible_size(...)
#>  8.   └─vctrs:::stop_vctrs(...)
#>  9.     └─rlang::abort(message, class = c(class, "vctrs_error"), ..., call = vctrs_error_call(call))

# This does
summarise(df, sum = sum(x), empty = tibble(a = integer()))
#> # A tibble: 0 × 2
#> # … with 2 variables: sum <int>, empty <tibble[,1]>

# And these work
summarise(df, sum = sum(x), tibble())
#> # A tibble: 1 × 1
#>     sum
#>   <int>
#> 1     6

summarise(df, sum = sum(x), tibble(a = integer()))
#> # A tibble: 0 × 2
#> # … with 2 variables: sum <int>, a <int>
```

<sup>Created on 2022-10-18 with [reprex v2.0.2.9000](https://reprex.tidyverse.org)</sup>

PR #6527, opened 2022-11-04
Recycle 0-column data frames in `summarise()` when they are named

DavisVaughan:
Closes https://github.com/tidyverse/dplyr/issues/6509

``` r
df <- tibble(x = 1:3)

# Now correctly recycles rather than erroring
summarise(df, sum = sum(x), empty = tibble())
#> # A tibble: 0 × 2
#> # … with 2 variables: sum <int>, empty <tibble[,0]>

summarise(df, sum = sum(x), empty = new_tibble(list(), nrow = 3))
#> # A tibble: 3 × 2
#>     sum empty       
#>   <int> <tibble[,0]>
#> 1     6             
#> 2     6             
#> 3     6
```

This reverts most of https://github.com/tidyverse/dplyr/pull/5088, but the tests that were added there still pass. This is because _unnamed_ data frames are actually already handled on the R side these days. They get auto-spliced ahead of time into their individual columns, which effectively makes 0-column data frames disappear.

https://github.com/tidyverse/dplyr/blob/02ef1d7c2de2bd70252a124004b2fe7c1ee9f5d8/R/summarise.R#L256-L270

So we don't need the complicated C code that was trying to exclude "useless" data frames that have 0 columns from the size computation. Those data frames actually do matter when they are named, because they will result in a df-col in the result and need to be considered during recycling.

I had a tough time trying to work through the C code for the recycling helper, so I've rewritten it. I think it is much easier to read now, and I think it should be faster because it uses more linear access of the lists.

The C function used to return `1L` if all of the chunks were size 1, but I've tried to make it more type stable by always returning the `sizes` vector. We just do `all(sizes == 1L)` on the R side now, which I'm not expecting to have any performance impact.

I've also tried to make it clearer that the function modifies its input by reference, which took me awhile to figure out. I do think it makes sense to do that though, for performance.

### DIFF
--- a/R/summarise.R
+++ b/R/summarise.R
@@ -216,9 +216,6 @@ summarise_cols <- function(.data, dots, error_call = caller_env()) {
   cols <- list()
 
   sizes <- 1L
-  chunks <- vector("list", length(dots))
-  types <- vector("list", length(dots))
-
   chunks <- list()
   results <- list()
   types <- list()
@@ -276,21 +273,19 @@ summarise_cols <- function(.data, dots, error_call = caller_env()) {
           results <- append(results, list(results_k))
           out_names <- c(out_names, name)
         }
-
       }
     }
 
-    recycle_info <- .Call(`dplyr_summarise_recycle_chunks`, chunks, mask$get_rows(), types, results)
-    chunks <- recycle_info$chunks
-    sizes <- recycle_info$sizes
-    results <- recycle_info$results
+    # Recycle horizontally across sets of chunks.
+    # Modifies `chunks` and `results` in place for efficiency!
+    sizes <- .Call(`dplyr_summarise_recycle_chunks_in_place`, chunks, results)
 
-    # materialize columns
+    # Materialize columns, regenerate any `results` that were `NULL`ed
+    # during the recycling process.
     for (i in seq_along(chunks)) {
       result <- results[[i]] %||% vec_c(!!!chunks[[i]], .ptype = types[[i]])
       cols[[ out_names[i] ]] <- result
     }
-
   },
   error = function(cnd) {
     if (inherits(cnd, "dplyr:::summarise_incompatible_size")) {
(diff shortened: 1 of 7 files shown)


R/summarise.R
```diff
@@ -216,9 +216,6 @@ summarise_cols <- function(.data, dots, error_call = caller_env()) {
   cols <- list()
 
   sizes <- 1L
-  chunks <- vector("list", length(dots))
-  types <- vector("list", length(dots))
-
   chunks <- list()
   results <- list()
   types <- list()
```
DavisVaughan, line 221: Not strictly part of this PR, but these were immediately being overwritten, so I removed them to clean things up

R/summarise.R
```diff
@@ -317,7 +312,7 @@ summarise_cols <- function(.data, dots, error_call = caller_env()) {
 
   signal_warnings(warnings_state, error_call)
 
-  list(new = cols, size = sizes, all_one = identical(sizes, 1L))
+  list(new = cols, sizes = sizes, all_one = all(sizes == 1L))
```
DavisVaughan, line 315: No longer have the special case where `sizes` could be `1L` to specially signal that all chunks were size 1. I don't think it makes much of a difference.

src/summarise.cpp
```diff
@@ -56,101 +56,104 @@ SEXP dplyr_mask_eval_all_summarise(SEXP quo, SEXP env_private) {
...
-    UNPROTECT(2);
-    return res;
+  SEXP first_chunks = v_list_of_chunks[0];
+  const SEXP* v_first_chunks = VECTOR_PTR_RO(first_chunks);
+  const R_xlen_t n_chunks = Rf_xlength(first_chunks);
+
+  SEXP sizes = PROTECT(Rf_allocVector(INTSXP, n_chunks));
+  int* v_sizes = INTEGER(sizes);
+
+  // Initialize `sizes` with first set of chunks
```
lionel-, line 89: Could this be computed on the R side with `list_sizes()`?

Though that wouldn't make a bit difference.

lionel- (approved):
I'm not sure what "chunks" are. If I'm not mistaken: We have a rectangular list of lists of vectors and we're recycling the vectors rowwise? If so, should we extract that to vctrs, e.g. as something like `rows_common_size()`?

DavisVaughan:
> If so, should we extract that to vctrs, e.g. as something like rows_common_size()?

We do also do something similar in `df_unchop()` (i.e. if you unchop 2 list-cols at the same time you have to rowwise recycle them). But I'm looking at `df_unchop()` and we have so much custom code there that it would probably be hard to incorporate something like `rows_common_size()`. We should probably consider it if it comes up a 3rd distinct time though

PR #6528, opened 2022-11-04
Implement `.by`

DavisVaughan:
Closes https://github.com/tidyverse/dplyr/issues/6214

This feels...pretty good? It definitely needs more testing, but I think the implementation is definitely going in the right direction. I think I've found the right abstraction with `compute_by()` which takes the `by` tidy-selection and the `data` and returns a list of:
- The grouping type (ungrouped, rowwise, grouped)
- The group column names
- The group data, i.e. what typically comes from `group_data()`

Notes:
- Only implemented for `summarise()` and `mutate()` right now.
- Can't be used with existing rowwise or grouped data frames
- Can't specify "by row"
- `summarise()` prevents you from combining `.by` and `.groups`, which shouldn't ever make sense.
- Not exposing this for `transmute()` because it is superseded.

It is implemented in a way that all tests still pass, even though there is only support for it in `mutate()` and `summarise()`.

I think it feels so much nicer than our existing syntax! Especially in the case where `summarise()` returns >1 row per group and you are forced to specify `.groups = "drop"` or call `ungroup()`, like with this common ivs example:

``` r
library(dplyr)
library(ivs)

users <- tribble(
  ~user, ~from, ~to,
  1L, "2019-01-01", "2019-01-05", 
  1L, "2019-01-12", "2019-01-13", 
  1L, "2019-01-03", "2019-01-10", 
  2L, "2019-01-02", "2019-01-03", 
  2L, "2019-01-03", "2019-01-04", 
  2L, "2019-01-05", "2019-01-07"
)
users <- users %>%
  mutate(from = as.Date(from), to = as.Date(to)) %>%
  mutate(range = iv(from, to), .keep = "unused")

users
#> # A tibble: 6 × 2
#>    user                    range
#>   <int>               <iv<date>>
#> 1     1 [2019-01-01, 2019-01-05)
#> 2     1 [2019-01-12, 2019-01-13)
#> 3     1 [2019-01-03, 2019-01-10)
#> 4     2 [2019-01-02, 2019-01-03)
#> 5     2 [2019-01-03, 2019-01-04)
#> 6     2 [2019-01-05, 2019-01-07)

# Long, don't like that I have to use `.groups = "drop"` because I rarely want that
users %>%
  group_by(user) %>%
  summarise(range = iv_groups(range), .groups = "drop")
#> # A tibble: 4 × 2
#>    user                    range
#>   <int>               <iv<date>>
#> 1     1 [2019-01-01, 2019-01-10)
#> 2     1 [2019-01-12, 2019-01-13)
#> 3     2 [2019-01-02, 2019-01-04)
#> 4     2 [2019-01-05, 2019-01-07)

# SO MUCH NICER
users %>%
  summarise(range = iv_groups(range), .by = user)
#> # A tibble: 4 × 2
#>    user                    range
#>   <int>               <iv<date>>
#> 1     1 [2019-01-01, 2019-01-10)
#> 2     1 [2019-01-12, 2019-01-13)
#> 3     2 [2019-01-02, 2019-01-04)
#> 4     2 [2019-01-05, 2019-01-07)
```


mine-cetinkaya-rundel:
Some reactions/questions:

- It says "I rarely want `.groups = "drop"` but this implementation always does `.groups = "drop"`, no? So, that doesn't seem like the right justification. Unless it's about that syntax `.groups = "drop"`, which I'm also not the biggest fan of.

- Is the idea that if this existed before, we'd never have `group_by()`? I've always thought group_by |> summarize is so powerful, that this feels like a step back for the simplest scenario (single grouping variable, calculating a summary statistic once). But I see the upsides for longer pipelines. I imagine y'all have done the "cost-benefit analysis" of that?

- `.by` sounds like the right argument name but what about the `by` in `*_join()` functions?

- What will this look like for multiple grouping variables? `.by = c(var1, var2)`? Is the result ungrouped regardless of the number of grouping variables?

DavisVaughan:
oh regarding this comment of mine:

```r
don't like that I have to use `.groups = "drop"` because I rarely want that
```

I think I just made some kind of typo. I actually meant that I almost always want `.groups = "drop"`, but I hate that I have to type it out all the time to get it. This does `.groups = "drop"` unconditionally.

DavisVaughan:
> Is the idea that if this existed before, we'd never have group_by()?

That is kind of how I feel about it, yea. More powerfully is that we probably never would have needed the `grouped_df` subclass, we could have always used bare data frames or bare tibbles. Which would make extending dplyr a lot easier for people like the tsibble authors.

---

> I've always thought group_by |> summarize is so powerful, that this feels like a step back for the simplest scenario (single grouping variable, calculating a summary statistic once)

I actually felt like it was a step forward, especially for teaching. You could introduce `summarise()` with:

```r
mtcars %>% 
  summarise(mean = mean(mpg))
```

And then grouped summaries use the same syntax, no new functions, just a new argument, _and you never have to worry about ungrouping afterwards_

```r
mtcars %>% 
  summarise(mean = mean(mpg), .by = vs)
```

---

> what about the by in *_join() functions?

Reasonable question, we'd have to see how confusing that is, I guess

---

With multiple variables, yes `.by = c(var1, var2)` would be valid syntax. It accepts anything supported by tidyselect, so you can use `.by = all_of(character_vector)` or `.by = starts_with("grp_")` too.

> Is the result ungrouped regardless of the number of grouping variables?

Yes and I feel fairly strongly that this is the right design decision, because otherwise we end up back with the `.groups = ` argument of `summarise()` along with having to emit messages to users about what is happening.

hadley:
@mine-cetinkaya-rundel another advantage of this approach is that you never have to explain what a `grouped_df` is and how it affects the results of all subsequent operations.

R/summarise.R
```diff
@@ -116,27 +117,47 @@
...
 }
 
 #' @export
-summarise.grouped_df <- function(.data, ..., .groups = NULL) {
-  cols <- summarise_cols(.data, dplyr_quosures(...))
-  out <- summarise_build(.data, cols)
+summarise.grouped_df <- function(.data, ..., .by = NULL, .groups = NULL) {
+  # TODO: Is it right to even expose `.by` here? It lets us catch errors.
+  # Will always error if `.by != NULL` b/c you can't use it with grouped/rowwise dfs.
+  by <- compute_by({{ .by }}, .data, by_arg = ".by", data_arg = ".data")
```
DavisVaughan, line 157: But this is kind of weird. `.by` is here in the grouped-df method because it is in the generic (although I dont think it _has_ to be here) but it will always error informatively if non-null.

I can't decide if removing it from here would be better or not. If we removed it then `.by` would slip by on grouped-dfs and potentially would become a column of the result through being captured by `...`

R/summarise.R
```diff
@@ -401,9 +426,19 @@ summarise_bullets <- function(cnd, ...) {
...
+
+  inform <- getOption("dplyr.summarise.inform")
+
+  if (is_true(inform) || is_false(inform)) {
+    # User supplied global option
+    return(inform)
+  }
+
+  is_reference(topenv(.env), global_env())
 }
```
DavisVaughan, line 442: I tweaked this a little so I could force verbosity in the `by.Rmd` doc

tests/testthat/test-mutate.R
```diff
@@ -75,6 +81,7 @@ test_that("assignments don't overwrite variables (#315)", {
 })
 
 test_that("can mutate a data frame with zero columns and `NULL` column names", {
+  skip("Do we really care about this gremlin case? This data frame is unusable.")
   df <- new_data_frame(n = 2L)
   colnames(df) <- NULL
   expect_equal(mutate(df, x = 1), data.frame(x = c(1, 1)))
```
DavisVaughan, line 87: Very weird edge case test that I don't think we care about, but I wanted to check. It breaks with some of the `.by` changes

man/rmd/by.Rmd
```diff
@@ -0,0 +1,167 @@
...
+```
+
+As of dplyr 1.1.0, an additional option is available that lets you specify the grouping *inline* within the verb:
+
+```{r}
+expenses %>%
+  summarise(cost = mean(cost), .by = region)
+```
+
+This great idea comes from [data.table](https://CRAN.R-project.org/package=data.table), where the equivalent syntax looks something like `expenses[, .(cost = mean(cost)), by = region]`.
```
hadley, line 58: Move this up to intro.

man/rmd/by.Rmd
```diff
@@ -0,0 +1,167 @@
...
+Imagine that you wanted to compute the average cost per region.
+You'd probably write something like this:
+
+```{r}
+expenses %>%
+  group_by(region) %>%
+  summarise(cost = mean(cost))
+```
+
+As of dplyr 1.1.0, an additional option is available that lets you specify the grouping *inline* within the verb:
```
hadley, line 51:
```suggestion
Instead, you can now specify the grouping *inline* within the verb:
```

man/rmd/by.Rmd
```diff
@@ -0,0 +1,167 @@
...
+  group_by(id, region) %>%
+  summarise(cost = mean(cost), .groups = "drop")
+```
+
+Because `.by` grouping is temporary, you don't need to worry about ungrouping and it never needs to emit a message to remind you what is happening:
+
+```{r}
+expenses %>%
+  summarise(cost = mean(cost), .by = c(id, region))
+```
```
hadley, line 83: I think you could reduce these two paragraphs + code chunks this to a single sentence.

man/rmd/by.Rmd
```diff
@@ -0,0 +1,167 @@
...
+There are two ways to group in dplyr:
+
+-   Persistent grouping with [group_by()]
+
+-   Temporary grouping with `.by`
+
+This help page is dedicated to explaining where and why you might want to use the latter.
+The most important reason to use `.by` is that you *never* have to remember to use [ungroup()] after it.
+The grouping is always temporary, meaning that if an ungrouped data frame goes in, then an ungrouped data frame comes out, regardless of the number of grouping columns that are specified.
+
```
hadley, line 20: Maybe here would be a good place to summarise the differences between `group_by` and `.by`?

* Grouping only affects a single verb
* Select variables with tidyselect
* Summaries are ordered by appearance

Or maybe a table with one column for `group_by` and one column for `.by`?

man/rmd/by.Rmd
```diff
@@ -0,0 +1,167 @@
...
+```
+
+Or you could slice out the maximum cost per combination of id and region:
+
+```{r}
+expenses %>%
+  slice_max(cost, n = 1, by = c(id, region))
+```
+
+Again, note that the result of this `slice_max()` is an ungrouped data frame, which is probably what you'd expect here.
```
hadley, line 111: Again, feels a bit too duplicative

man/rmd/by.Rmd
```diff
@@ -0,0 +1,167 @@
...
+
+```{r}
+expenses %>%
+  group_by(id, region) %>%
+  slice_max(cost, n = 1)
+```
+
+### Result ordering
+
+When used with `.by`, `summarise()` and `slice()` both return results where the group keys are ordered by their existing ordering in the original data rather than by their values.
```
hadley, line 122:
```suggestion
When used with `.by`, `summarise()` and `slice()` both maintain ordering of the existing results.
```

man/rmd/by.Rmd
```diff
@@ -0,0 +1,167 @@
...
+# Uses ordering by "first appearance" in the original data
+df %>%
+  summarise(average_temp = mean(temp), .by = month)
+
+# Sorts in ascending order
+df %>%
+  group_by(month) %>%
+  summarise(average_temp = mean(temp))
+
+# Same with `slice()`
```
hadley, line 140: I think you can skip the slice example

R/by.R
```diff
@@ -0,0 +1,126 @@
...
+                     error_call = caller_env()) {
+  check_dots_empty0(...)
+
+  if (quo_is_null(by)) {
+    return(invisible(NULL))
+  }
+
+  if (is_grouped_df(data)) {
+    message <- paste0(
+      "{.arg {by_arg}} can't be supplied when ",
```
hadley, line 95:
```suggestion
      "Can't supply {.arg {by_arg}} when ",
```

R/by.R
```diff
@@ -0,0 +1,126 @@
...
+      "{.arg {data_arg}} is a grouped data frame."
+    )
+    cli::cli_abort(message, call = error_call)
+  }
+
+  if (is_rowwise_df(data)) {
+    message <- paste0(
+      "{.arg {by_arg}} can't be supplied when ",
```
hadley, line 103:
```suggestion
      "Can't supply {.arg {by_arg}} when ",
```

tests/testthat/test-mutate.R
hadley: I'd suggest just changing this to a test of `df <- new_data_frame(n = 2L)`; i.e. no columns but normal names.

man/rmd/by.Rmd
hadley: It's relatively unimportant but it would be nice to mention somewhere that `df |> summarise(.by)` preserves the class of `df`, and `df |> mutate(.by)` preserves the attributes.

R/summarise.R
DavisVaughan: Yea, that's fair

It is easier for `mutate()` because there is only `mutate.data.frame()`, there is not `mutate.grouped_df()`, so we don't have to make the same kind of call there

NEWS.md
```diff
@@ -1,5 +1,50 @@
 # dplyr (development version)
 
+* `.by` is a new experimental inline alternative to `group_by()` that supports
+  _temporary_ grouping in the following key dplyr verbs: `mutate()`,
```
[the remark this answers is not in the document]
DavisVaughan: I think I went with temporary in documentation because it seemed like an easier verb for users to understand, but I am not tied to it if you feel like transient is clearer

R/slice.R
```diff
@@ -117,52 +118,62 @@
 #' filter(mtcars, row_number() == 1L)
 #' filter(mtcars, row_number() == n())
 #' filter(mtcars, between(row_number(), 5, n()))
-slice <- function(.data, ..., .preserve = FALSE) {
+slice <- function(.data, ..., .by = NULL, .preserve = FALSE) {
```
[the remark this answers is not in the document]
DavisVaughan: Yea, `.preserve` only makes sense when you have an external grouping structure

``` r
df <- tibble(g = c(1, 1, 2, 2), x = c(1, 2, 3, 4))

slice(df, which(x > 2), .by = g)
#> # A tibble: 2 × 2
#>       g     x
#>   <dbl> <dbl>
#> 1     2     3
#> 2     2     4

# drops all g==1, but kept it in grouping structure
df %>%
  group_by(g) %>%
  slice(which(x > 2), .preserve = TRUE)
#> # A tibble: 2 × 2
#> # Groups:   g [2]
#>       g     x
#>   <dbl> <dbl>
#> 1     2     3
#> 2     2     4

# shows up in summarise() downstream
df %>%
  group_by(g) %>%
  slice(which(x > 2), .preserve = TRUE) %>%
  summarise(x = mean(x))
#> # A tibble: 2 × 2
#>       g     x
#>   <dbl> <dbl>
#> 1     1 NaN  
#> 2     2   3.5
```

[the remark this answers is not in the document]
DavisVaughan: I think you could still get around this by making `g` a factor so `slice()` would still keep the level around, then when you pipe to `summarise()` you'd use `.modifier = no_drop()` or whatever the `.drop = FALSE` equivalent turns out to be

man/rmd/by.Rmd
DavisVaughan: > df |> summarise(.by) preserves the class of df

I think this is only true for classes that `summarise()` knows about. i.e. `summarise()` a data.table and you get a data.frame back

I think I'd prefer to leave these details out for the moment if that is okay with you

R/by.R
```diff
@@ -0,0 +1,126 @@
...
+#' ```{r, echo = FALSE, results = "asis"}
+#' result <- rlang::with_options(
+#'   knitr::knit_child("man/rmd/by.Rmd"),
+#'   dplyr.summarise.inform = TRUE
+#' )
+#' cat(result, sep = "\n")
+#' ```
+#'
+#' @name dplyr_by
+#' @keywords internal
```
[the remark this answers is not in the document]
DavisVaughan: I put it under `group_by()` in the index

man/rmd/by.Rmd
hadley: Yeah, sure.

DavisVaughan:
@hadley thanks for the in depth `by.Rmd` review! I think it is way more concise now

PR #6531, opened 2022-11-08
Docs: Add note about `head()`

eitsupi:
Since there is a note about `tail()`, I think it would be better if about `head()` was also written.


hadley:
Thanks!

PR #6543, opened 2022-11-17
Error on argument-less `n_distinct()`

### LINKED ISSUE #6535: `n_distinct()` with no arguments should error
``` r
dplyr::n_distinct()
#> [1] 0
```

<sup>Created on 2022-11-11 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

hadley:
Fixes #6535


R/n-distinct.R
```diff
@@ -30,6 +28,9 @@
 #' # Also works with data frames
 #' n_distinct(data.frame(x, y))
 n_distinct <- function(..., na.rm = FALSE) {
+  if (missing(...)) {
+    abort("... is absent, but must be supplied.")
```
DavisVaughan, line 32:
```suggestion
    abort("`...` is absent, but must be supplied.")
```

Minor. Or maybe consider using `cli_abort()` and `{.code ...}`

DavisVaughan (approved):

R/n-distinct.R
DavisVaughan: I think we started relying on cli when Lionel was working on the group-context errors, so now I've just been using `cli_abort()` in new error messages. We'd eventually have to spend a significant amount of time updating to cli everywhere.

https://github.com/search?q=repo%3Atidyverse%2Fdplyr%20cli_abort&type=code

DavisVaughan: We can also roll that back if we don't want to start using cli anywhere yet

Issue #6534, opened 2022-11-11
Grouped `across()` doesn't give good error with bad input


hadley:
``` r
library(dplyr, warn.conflicts = FALSE)

df <- tibble(g = 1, x = 1, y = 2, z = 3)

df |> summarise(across(everything(), median()))
#> Error in `summarise()`:
#> ℹ In argument: `..1 = across(everything(), median())`.
#> Caused by error in `is.factor()`:
#> ! argument "x" is missing, with no default

df |> group_by(g) |> summarise(across(everything(), median()))
#> Error in vapply(.x, .f, .mold, ..., USE.NAMES = FALSE): values must be length 1,
#>  but FUN(X[[1]]) result is length 0
```

<sup>Created on 2022-11-11 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

hadley:
The problem seems to occur because it occurs during the across splatting, i.e. in group "0".

DavisVaughan:
Yea the problem is likely that we are trying to add group context to the error message, but there isn't any to add because we haven't set up the group information yet.

Determining whether or not to attempt to add group context to the error message has been a little painful so far, so I feel like we should rethink how it works a little bit. If we are "before" the groups have been set up, we have to specially register the error class as one that doesn't need group context, which seems hard to unconditionally get right (as seen in your example)
https://github.com/tidyverse/dplyr/blob/50c58dd9980b9216c60a1de4da521ee5aa16e2de/R/conditions.R#L63-L72

Maybe we can explicitly set something like `the$needs_group_context` to `TRUE` as soon as we set up the groups, but otherwise default it to `FALSE`, which would greatly simplify some things.

For example, this is the reason `pick()` has to subclass tidyselect errors
https://github.com/tidyverse/dplyr/blob/50c58dd9980b9216c60a1de4da521ee5aa16e2de/R/pick.R#L167-L182

And it is why we have to do this crazy catch-subclass-rethrow pattern with deprecation conditions
https://github.com/tidyverse/dplyr/blob/50c58dd9980b9216c60a1de4da521ee5aa16e2de/R/across.R#L778-L794

DavisVaughan:
@lionel- what do you think about that?

lionel-:
Using a boolean flag in the condition with `%||% FALSE` default instead subclassing seems like a good idea.

hadley:
Are you sure we should just have someway to let the mask know that we've started/ended grouping? Then we could have:

[...]

PR #6550, opened 2022-11-18
Look up symbols as functions in `across()`

### LINKED ISSUE #6545: `across()` confuses function name with column name
Apparently, any function whose name appears as a column name of a tibble cannot be used within `across()` even when it is specified as the `.fns` argument: 

``` r
library(dplyr)

# column called "mean"
tibble(mean = 1:5) %>% 
  summarize(across(.fns = mean))
#> Error in `summarize()`:
#> ! Problem while computing `..1 = across(.fns = mean)`.
#> Caused by error:
#> ! attempt to select less than one element in integerOneIndex

#> Backtrace:
#>      ▆
#>   1. ├─tibble(mean = 1:5) %>% summarize(across(.fns = mean))
#>   2. ├─dplyr::summarize(., across(.fns = mean))
#>   3. ├─dplyr:::summarise.data.frame(., across(.fns = mean))
#>   4. │ └─dplyr:::summarise_cols(.data, dplyr_quosures(...), caller_env = caller_env())
#>   5. │   ├─base::withCallingHandlers(...)
#>   6. │   └─dplyr:::expand_across(dots[[i]])
#>   7. │     ├─dplyr:::across_setup(...)
#>   8. │     └─rlang::eval_tidy(expr$.fns, mask, env = env)
#>   9. ├─`<fn>`()
#>  10. └─base::.handleSimpleError(...)
#>  11.   └─dplyr (local) h(simpleError(msg, call))
#>  12.     └─rlang::abort(bullets, call = error_call, parent = skip_internal_condition(e))

# column called "average"
tibble(average = 1:5) %>% 
  summarize(across(.fns = mean))
#> # A tibble: 1 × 1
#>   average
#>     <dbl>
#> 1       3
```

<sup>Created on 2022-11-17 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

<details style="margin-bottom:10px;">
<summary>
Session info
</summary>

``` r
sessioninfo::session_info()
#> ─ Session info ───────────────────────────────────────────────────────────────
#>  setting  value
#>  version  R version 4.2.0 (2022-04-22)
#>  os       macOS 13.0.1
#>  system   aarch64, darwin20
#>  ui       X11
#>  language (EN)
#>  collate  en_US.UTF-8
#>  ctype    en_US.UTF-8
#>  tz       Europe/Berlin
#>  date     2022-11-17
#>  pandoc   2.19.2 @ /Applications/RStudio.app/Contents/MacOS/quarto/bin/tools/ (via rmarkdown)
#> 
#> ─ Packages ───────────────────────────────────────────────────────────────────
#>  package     * version date (UTC) lib source
#>  assertthat    0.2.1   2019-03-21 [1] CRAN (R 4.2.0)
#>  cli           3.4.1   2022-09-23 [1] CRAN (R 4.2.0)
#>  DBI           1.1.3   2022-06-18 [1] CRAN (R 4.2.0)
#>  digest        0.6.30  2022-10-18 [1] CRAN (R 4.2.0)
#>  dplyr       * 1.0.10  2022-09-01 [1] CRAN (R 4.2.0)
#>  evaluate      0.18    2022-11-07 [1] CRAN (R 4.2.0)
#>  fansi         1.0.3   2022-03-24 [1] CRAN (R 4.2.0)
#>  fastmap       1.1.0   2021-01-25 [1] CRAN (R 4.2.0)
#>  fs            1.5.2   2021-12-08 [1] CRAN (R 4.2.0)
#>  generics      0.1.3   2022-07-05 [1] CRAN (R 4.2.0)
#>  glue          1.6.2   2022-02-24 [1] CRAN (R 4.2.0)
#>  highr         0.9     2021-04-16 [1] CRAN (R 4.2.0)
#>  htmltools     0.5.3   2022-07-18 [1] CRAN (R 4.2.0)
#>  knitr         1.40    2022-08-24 [1] CRAN (R 4.2.0)
#>  lifecycle     1.0.3   2022-10-07 [1] CRAN (R 4.2.0)
#>  magrittr      2.0.3   2022-03-30 [1] CRAN (R 4.2.0)
#>  pillar        1.8.1   2022-08-19 [1] CRAN (R 4.2.0)
#>  pkgconfig     2.0.3   2019-09-22 [1] CRAN (R 4.2.0)
#>  purrr         0.3.5   2022-10-06 [1] CRAN (R 4.2.0)
#>  R.cache       0.16.0  2022-07-21 [1] CRAN (R 4.2.0)
#>  R.methodsS3   1.8.2   2022-06-13 [1] CRAN (R 4.2.0)
#>  R.oo          1.25.0  2022-06-12 [1] CRAN (R 4.2.0)
#>  R.utils       2.12.2  2022-11-11 [1] CRAN (R 4.2.0)
#>  R6            2.5.1   2021-08-19 [1] CRAN (R 4.2.0)
#>  reprex        2.0.2   2022-08-17 [1] CRAN (R 4.2.0)
#>  rlang         1.0.6   2022-09-24 [1] CRAN (R 4.2.0)
#>  rmarkdown     2.18    2022-11-09 [1] CRAN (R 4.2.0)
#>  rstudioapi    0.14    2022-08-22 [1] CRAN (R 4.2.0)
#>  sessioninfo   1.2.2   2021-12-06 [1] CRAN (R 4.2.0)
#>  stringi       1.7.8   2022-07-11 [1] CRAN (R 4.2.0)
#>  stringr       1.4.1   2022-08-20 [1] CRAN (R 4.2.0)
#>  styler        1.8.1   2022-11-07 [1] CRAN (R 4.2.0)
#>  tibble        3.1.8   2022-07-22 [1] CRAN (R 4.2.0)
#>  tidyselect    1.2.0   2022-10-10 [1] CRAN (R 4.2.0)
#>  utf8          1.2.2   2021-07-24 [1] CRAN (R 4.2.0)
#>  vctrs         0.5.0   2022-10-22 [1] CRAN (R 4.2.0)
#>  withr         2.5.0   2022-03-03 [1] CRAN (R 4.2.0)
#>  xfun          0.34    2022-10-18 [1] CRAN (R 4.2.0)
#>  yaml          2.3.6   2022-10-18 [1] CRAN (R 4.2.0)
#> 
#>  [1] /Library/Frameworks/R.framework/Versions/4.2-arm64/Resources/library
#> 
#> ──────────────────────────────────────────────────────────────────────────────
```

</details>

lionel-:
Closes #6545

Symbols supplied to `.fns` are now looked up step by step in the chain of environments, discarding any value that's not a function, formula, or list. This way, we prevent data mask columns from masking a function.

This is only the case for bare symbols like `mean`, not for complex calls like `list(mean)` which I think we shouldn't special-case as that would create surprising semantics. The error message that is produced in this case is suboptimal:

```r
df <- tibble(mean = 1:5)
summarize(df, across(.fns = list(mean)))
#> Error in `summarize()`:
#> ℹ In argument: `..1 = across(.fns = list(mean))`.
#> Caused by error:
#> ! attempt to select less than one element in integerOneIndex
```

I think that is because the group chunks are not set in the data mask at expansion time. @DavisVaughan do you think that could be worked around easily?

An alternative solution would have been to evaluate `.fns` outside of the mask. This would have fixed the `list(mean)` issues but would have broken lambdas that refer to mask columns:

```r
mtcars |> mutate(across(1:2, \(x) x + mean(cyl)))
```


DavisVaughan:
> I think that is because the group chunks are not set in the data mask at expansion time

Yea, group context isn't set up yet at expansion time because we haven't started working on any groups yet.

I'm fairly certain that is going to be resolved with https://github.com/tidyverse/dplyr/issues/6534#issuecomment-1312215910

It ends up failing in `cnd_bullet_cur_group_label()` because `needs_group_context()` returned `TRUE` by mistake, but we plan to fix that

lionel-:
hmm it's the active binding for `mean` that fails, so I'm not sure the group context info for errors will help here.

DavisVaughan:
Oh I see what you mean. Right, so the active binding for `mean` the column is trying to be triggered, which ends up basically trying to do `.subset2(mean, .current_group)`. The chops and active bindings are all set up at this point, and `.current_group` is technically there, it is just set to `0L` which gives us this error.

I don't think we could actually set it to anything else. Like setting it to `1L` or something would probably give an even more obscure error down the line

If `list(mean)` is going to have to evaluate in the mask, then we probably do want this to error in some way right? I'm not sure we can improve on the error if that is the case

tests/testthat/_snaps/across.md
```diff
@@ -408,3 +408,22 @@
...
+      Caused by error:
+      ! attempt to select less than one element in integerOneIndex
+    Code
+      (expect_error(summarize(df, (across(everything(), list(mean))))))
+    Output
+      <error/rlang_error>
+      Error in `summarize()`:
+      i In argument: `..1 = (across(everything(), list(mean)))`.
+      Caused by error in `across()`:
+      ! Can't convert `X[[i]]`, an integer vector, to a function.
```
DavisVaughan, line 428: Yea I wouldn't say this error is actually "better" than the other one

If theoretically `across()` is a pure expansion function, the fact that an integer vector is even returned by `mean` in `list(mean)` is kind of buggy

R/across.R
```diff
@@ -455,7 +459,7 @@ across_setup <- function(cols,
   )
 
   if (!inline) {
-    fns <- map(fns, as_function)
+    fns <- map(fns, as_function, call = quote(across()))
```
DavisVaughan, line 462:
```suggestion
    fns <- map(fns, as_function, call = call(across_if_fn))
```

I think you want this? i.e. if `if_any/all()` go through here

DavisVaughan (approved):
Looks reasonable

tests/testthat/_snaps/across.md
lionel-: Agreed, I'll take another look at the issue and see if we can do better.

lionel-:
I've reviewed the approach to evaluate the `.fns` argument outside the mask. Non-inlinable lambdas are post-processed after evaluation so they are still maskable when run.

I've also refactored this part of the code to make things more consistent and simpler, and improved the internal documentation. In particular:

- `quo_eval_fns()` now returns either a function or a list of functions, so that we no longer have to worry about formulas later on. Inlinable lambdas are set to the empty environment so they can be recognised and inlined.

- Calls to `\()` and `function()` can now be inlined as well. Previously only `~` calls were inlined.

The updated revdeps show that 3 packages now fail because of the `function` inlining because they call `return()` in their lambdas. The last commit checks for `return()` usage and disable inlining in that case. I'm rerunning the failures.

R/across.R
```diff
@@ -373,15 +384,15 @@ across_setup <- function(cols,
                          names,
                          .caller_env,
                          mask,
-                         inline = FALSE) {
+                         error_call = caller_env()) {
   cols <- enquo(cols)
+  quo_env <- quo_get_env(cols)
+  across_if_fn <- context_peek_bare("across_if_fn") %||% "across"
```
DavisVaughan, line 390: This seems to be used in two places:

1. The `abort()` if branch right below this
2. The return value

I think this could actually be removed from the return value. Only `across()` uses it and that now also has a local copy of `across_if_fn` so really I don't think we need to pass it around anymore.

And then I think you could move this into the `if` branch itself right before you construct the `bullets`, since that is the only time it will be used

DavisVaughan: Oh, you could actually make this an argument to `across_setup()`. The two places where `across_setup()` is called from (`across()` and `expand_across()`) already have local copies of `across_if_fn` so they could just pass them through to here (in `expand_across()` it is part of creation of the `error_call`)

R/across.R
```diff
@@ -845,3 +855,132 @@ apply_unpack_spec <- function(col, outer, spec, caller_env) {
...
+      # that was directly supplied and evaluated here. We set its
+      # environment to the data mask so it can refer to columns.
+      if (identical(get_env(x), sentinel_env)) {
+        out <- set_env(out, mask)
+      }
+
+      out
+    } else {
+      abort(
+        "`.fns` must be a function, a formula, or a list of functions/formulas.",,
```
DavisVaughan, line 932:
```suggestion
        "`.fns` must be a function, a formula, or a list of functions/formulas.",
```

R/across.R
```diff
@@ -451,14 +461,29 @@ across_setup <- function(cols,
...
+    fn <- as_function(fn, call = error_call)
+  }
 
-  if (!inline) {
-    fns <- map(fns, as_function)
+  # Reset environment of inlinable lambdas which are set to the empty
+  # env sentinel
+  if (is_function(fn) && identical(get_env(fn), empty_env())) {
+    fn <- set_env(fn, env)
   }
```
DavisVaughan, line 484: Is the `is_formula(fn)` branch ever hit? I thought `quo_eval_fns()` should return a function or list of functions? So once it got here this should always be a function?

Which brings me to my next point - is the `is_function(fn)` condition ever `FALSE`? It seems like `fn` would always be a function.

So then this function would just be checking the empty-env sentinel, which makes this much easier to understand.

R/across.R
```diff
@@ -232,7 +240,10 @@ across <- function(.cols,
   fns <- setup$fns
   names <- setup$names
 
-  if (is.null(fns)) {
+  fns_quo_env <- quo_get_env(fns_quo)
+  fns <- map(fns, function(fn) uninline(fn, fns_quo_env, error_call = current_env()))
```
DavisVaughan, line 244: In the (deprecated) `.fns = NULL` case, is this doing `map(NULL, )`? I know that works but I feel a little uncomfortable with it.

`across_setup()` has a branch that specifically early returns on `.fns = NULL`, I wonder if that could return `fns = list()` in the `setup` list that it returns to fix this.

It probably requires tweaking `expand_across()` too, because that has an `if(is.null(fns))` branch too.

Worth looking into but might be too much work

R/across.R
```diff
@@ -845,3 +855,132 @@ apply_unpack_spec <- function(col, outer, spec, caller_env) {
...
+# To distinguish inlinable lambdas from non-inlinable ones, we set
+# their environments to the empty env.
+#
+# None of the lambdas passed through `list()` are inlinable and there
+# are some other cases where we can't inline (e.g. lambdas that are
+# passed additional arguments through `...`). We still want
+# non-inlinable lambdas to be maskable so that they can refer to
+# data-mask columns. So we set them (a) in the evaluation case, to
+# their original quosure environment which is the data mask, or (b) in
+# the expansion case, to the unitialised data mask.
```
DavisVaughan, line 876:
```suggestion
# the expansion case, to the uninitialised data mask.
```

R/across.R
```diff
@@ -845,3 +855,132 @@ apply_unpack_spec <- function(col, outer, spec, caller_env) {
...
+    } else {
+      abort(
+        "`.fns` must be a function, a formula, or a list of functions/formulas.",,
+        call = error_call
+      )
+    }
+  }
+
+  if (vec_is_list(out)) {
+    map(out, function(elt) validate(elt))
```
DavisVaughan, line 939:
```suggestion
    map(out, validate)
```

Does the anonymous function matter here?

R/across.R
```diff
@@ -845,3 +855,132 @@ apply_unpack_spec <- function(col, outer, spec, caller_env) {
...
+  fmls <- formals(fn)
+
+  # Don't inline if there are additional arguments even if they have
+  # defaults or are passed through `...`
+  if (length(fmls) != 1) {
+    return(FALSE)
+  }
+
+  # Don't inline lambdas that call `return()` at the moment a few
+  # packages do things like `across(1, function(x) return(x))`
```
DavisVaughan, line 962: Can you please add a short note about why we can't do this? I assume it has something to do with `return()` needing to know where it can be "returned from" and if you inline the function that gets screwed up

R/across.R
```diff
@@ -845,3 +855,132 @@ apply_unpack_spec <- function(col, outer, spec, caller_env) {
...
+    return(FALSE)
+  }
+
+  TRUE
+}
+
+quo_is_inlinable_formula <- function(quo) {
+  expr <- quo_get_expr(quo)
+
+  if (!is_formula(expr, scoped = FALSE)) {
```
DavisVaughan, line 973: Should you use the `lhs` argument here?

tests/testthat/_snaps/across.md
```diff
@@ -260,6 +260,29 @@
...
+      ! the ... list contains fewer than 2 elements
+    Code
+      (expect_error(df %>% mutate((across(1:2, ~ .y + mean(bar))))))
+    Output
+      <error/dplyr:::mutate_error>
+      Error in `mutate()`:
+      i In argument: `(across(1:2, ~.y + mean(bar)))`.
+      Caused by error in `across()`:
+      ! Can't compute column `foo`.
+      Caused by error in `fn()`:
```
DavisVaughan, line 283: Interesting that this says `in fn()`

tests/testthat/test-across.R
```diff
@@ -1221,3 +1288,47 @@ test_that("group variables are in scope when passed in dots (#5832)", {
...
+  out <- mutate(df, across(1:2, function(foo) return(foo)))
+  expect_equal(out, df)
+})
+
+test_that("non-inlinable but maskable lambdas can refer to their lexical environment", {
+  foo <- "OK"
+  df <- tibble(bar = "a")
+
+  expect_equal(
+    mutate(df, across(1, function(x) return(paste(x, foo)))),
```
DavisVaughan, line 1331: Should you also test that the inlinable variant can refer to the lexical env too? i.e. this version?

```
    mutate(df, across(1, function(x) paste(x, foo)))
```

DavisVaughan (approved):
I'm fairly certain I understand what is happening here. It does seem easier to follow now, but it is obviously by no means simple. I like the empty-env trick

R/across.R
```diff
@@ -232,7 +240,10 @@ across <- function(.cols,
   fns <- setup$fns
   names <- setup$names
 
-  if (is.null(fns)) {
+  fns_quo_env <- quo_get_env(fns_quo)
+  fns <- map(fns, function(fn) uninline(fn, fns_quo_env, error_call = current_env()))
+
+  if (!length(fns)) {
```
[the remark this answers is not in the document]
lionel-: I think we discussed it but didn't reach a conclusion as there were different opinions on the matter.

R/across.R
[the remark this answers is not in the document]
lionel-: hmm I think it's a mistake after refactoring to lambda form.

But also there's no longer formulas passing through that code path, so we don't need to call `as_function()` and we don't need to pass `error_call`.

R/across.R
lionel-: Also we should pass down an environment from the across-if-fn context, so that backtraces can be appropriately truncated.

R/across.R
lionel-: yup, already fixed

R/across.R
lionel-: I think mapping over `NULL` simplifies things here.

R/across.R
lionel-: I'm trying out using lambdas as a rule.

R/across.R
lionel-: I've added a note that it's because even though `eval()` sets a return point, `eval_tidy()` doesn't.

tests/testthat/_snaps/across.md
lionel-: I was wondering about that too, not sure how to do better.

tests/testthat/test-across.R
lionel-: I thought this was tested elsewhere but I can't find it. I also added tests for formulas which revealed an issue when the formula has a `return()` call. We no longer inline those.

R/across.R
lionel-: We now set the correct error frame in all cases.

For the `expand_` functions, we set the error call to their frames too, but override the call with `local_error_call()` so that the name of the _expanded_ function appears in errors, instead of the name of the _expanding_ function. It's neat that we can leverage this existing functionality in rlang to do that.

PR #6552, opened 2022-11-18
Simplify group context detection

### LINKED ISSUE #6534: Grouped `across()` doesn't give good error with bad input

DavisVaughan:
Closes #6534

We now detect whether or not we need to add group context by checking if the `mask$get_current_group()` value is `0` or not. If it is `0`, we are either before or after the actual computation of the group results. Otherwise we must be inside some kind of group computation, or we have hardcoded the group using `mask$set_current_group()`, which we do in a few places.

The most important change was probably resetting the group index to `0` in `DPLYR_MASK_FINALISE()` so that it is `0` _after_ the group computations too.

This allowed me to remove the special error wrapping in `pick()` and in the `across()` deprecation.


src/summarise.cpp
```diff
@@ -50,7 +50,16 @@ SEXP dplyr_mask_eval_all_summarise(SEXP quo, SEXP env_private) {
...
+    const SEXP* p_chunks = VECTOR_PTR_RO(chunks);
+
+    for (R_xlen_t i = 0; i < ngroups; i++) {
+      if (p_chunks[i] == R_NilValue) {
+        // Find out the first time the group was `NULL`
+        // so that the error will be associated with this group
+        DPLYR_MASK_SET_GROUP(i);
+        dplyr::stop_summarise_mixed_null();
+      }
+    }
```
DavisVaughan, line 62: This is the error path that occurs when you happen to return a `NULL` from one group and a non-`NULL` value in another group

In `mutate()` we already do something similar (report the group with the first problematic `NULL`), so this is just us being consistent with that now

tests/testthat/_snaps/across.md
```diff
@@ -339,6 +368,7 @@
       There were 2 warnings in `filter()`.
       The first warning was:
       i In argument `..1 = (if_any())`.
+      i In group 1: `g = 1`.
```
DavisVaughan, line 371: Minor improvement

We used to always skip adding the group context if we saw a `warning_across_missing_cols_deprecated` warning class. But during the evaluation path (i.e. not expansion) we actually do have the group context, so we can report it

tests/testthat/_snaps/mutate.md
```diff
@@ -164,6 +164,18 @@
       <error/dplyr:::mutate_error>
       Error in `mutate()`:
       i In argument: `..1 = if (a == 1) NULL else "foo"`.
+      i In group 1: `a = 1`.
```
DavisVaughan, line 167: Another small improvement

Again, we used to always skip adding the group context if `mutate_mixed_null` was the error class, but we actually do set the group value explicitly in the C++ code for this case, and we can now let that bubble up

tests/testthat/test-across.R
```diff
@@ -978,17 +995,27 @@ test_that("across() applies old `.cols = everything()` default with a warning",
...
   gdf <- group_by(df, g)
 
+  times_two <- function(x) x * 2
+
   # Expansion path
-  expect_snapshot(out <- mutate(df, z = across()))
-  expect_identical(out$z, df)
-  expect_snapshot(out <- mutate(gdf, z = across()))
-  expect_identical(out$z, df[c("x", "y")])
+  expect_snapshot(out <- mutate(df, across(.fns = times_two)))
```
DavisVaughan, line 1001: I realized this wasn't actually doing the "expansion path" of `across()` because it was a named call to `across()`, i.e. the `z = `, so I fixed that here. It ends up tweaking the warning in the snapshot test a little

R/across.R
```diff
@@ -775,25 +775,12 @@ across_missing_cols_deprecate_warn <- function() {
...
-  # is thrown from `expand_across()` outside of any group
-  class(cnd) <- c("dplyr:::warning_across_missing_cols_deprecated", class(cnd))
-
-  cnd_signal(cnd)
+  lifecycle::deprecate_warn(
+    when = "1.1.0",
+    what = I(glue("Using `{across_if_fn}()` without supplying `.cols`")),
+    details = "Please supply `.cols` instead.",
+    user_env = user_env
+  )
```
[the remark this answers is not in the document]
lionel-: Nice!

R/conditions.R
```diff
@@ -60,18 +60,10 @@ or_1 <- function(x) {
   }
 }
 
-needs_group_context <- function(cnd) {
-  !inherits_any(cnd, c(
-    "dplyr:::error_incompatible_combine",
-    "dplyr:::mutate_mixed_null",
-    "dplyr:::mutate_constant_recycle_error",
-    "dplyr:::summarise_mixed_null",
```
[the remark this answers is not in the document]
lionel-: Cool

src/summarise.cpp
```diff
@@ -50,7 +50,16 @@ SEXP dplyr_mask_eval_all_summarise(SEXP quo, SEXP env_private) {
   if (n_null == ngroups) {
     return R_NilValue;
   } else if (n_null != 0) {
-    dplyr::stop_summarise_mixed_null();
+    const SEXP* p_chunks = VECTOR_PTR_RO(chunks);
```
lionel-, line 53: `v_chunks`?

tests/testthat/_snaps/summarise.md
```diff
@@ -149,6 +149,18 @@
       <error/rlang_error>
       Error in `summarise()`:
       i In argument: `x = if (g == 1) 42`.
+      i In group 2: `g = 2`.
```
lionel-, line 152: It's useful to now have this information.

lionel- (approved):
Nice improvement!

PR #6122, opened 2021-12-09
Allow hiding variable names and expressions in error messages

lionel-:
Useful to hide internal implementations in verb wrappers.

```r
mutate(mtcars, invisible(1 + ""))
#> Error in `mutate()`:
#>   Problem while computing `..1`.
#> Caused by error in `+`:
#>   non-numeric argument to binary operator

mutate(mtcars, var = invisible(1 + ""))
#> Error in `mutate()`:
#>   Problem while computing `var`.
#> Caused by error in `+`:
#>   non-numeric argument to binary operator

mutate(mtcars, .__var__. = 1 + "")
#> Error in `mutate()`:
#>   Problem while computing `1 + ""`.
#> Caused by error in `+`:
#>   non-numeric argument to binary operator

mutate(mtcars, .__var__. = invisible(1 + ""))
#> Error in `mutate()`:
#>   Problem while computing result.
#> Caused by error in `+`:
#>   non-numeric argument to binary operator
```

PR #6556, opened 2022-11-21
Don't show argument number

lionel-:
* Don't show `..i` in context info bullets for unnamed expressions.
* Consistently refer to name or expression in error messages.
* Remove support for `invisible()` syntax introduced in #6122. I don't think we ended up using it because we used the error borrowing approach instead, and it made things more complex.


R/conditions.R
```diff
@@ -370,18 +369,20 @@ signal_warnings <- function(state, error_call) {
...
+    group_label <- ""
   }
 
+  label <- error_label_named(data$name, data$expr)
+
   msg <- c(
-    "i" = glue::glue("In argument `{data$name} = {data$expr}`."),
-    "i" = if (nzchar(label)) glue("In {label}.")
+    "i" = glue::glue("In argument `{label}`."),
+    "i" = if (nzchar(group_label)) glue("In {group_label}.")
```
DavisVaughan, line 385: Are we not doing the `:` here? i.e. `In argument:`

R/mutate.R
```diff
@@ -429,7 +429,7 @@ mutate_bullets <- function(cnd, ...) {
 #' @export
 `mutate_bullets.dplyr:::mutate_incompatible_size` <- function(cnd, ...) {
   error_context <- peek_error_context()
-  error_name <- error_context$error_name
+  error_name <- ctxt_error_label(error_context)
```
DavisVaughan, line 432:
```suggestion
  label <- ctxt_error_label(error_context)
```

Or `error_label`?

This would just be for consistency with what the function is named. There are a few other places in this file you'd also have to do this renaming if you did this

DavisVaughan (approved):
I think it looks nice because it improves the 99% case.

The 1% case where it can be more confusing is when the error-label gets truncated _and_ you didn't provide a name, like:


```r
library(dplyr)

df <- tibble(x = 1:5)

mutate(df, {
  pick(everything()) 
  stop("oh no")
})
#> Error in `mutate()`:
#> ℹ In argument: `{ ... }`.
#> Caused by error:
#> ! oh no
```

But I think we are ok with that?

hadley:
@lionel- yup

PR #6557, opened 2022-11-21
Implement `reframe()`

### LINKED ISSUE #6382: summarize() with multi-row returns

DavisVaughan:
Closes #6382 

Name is still pending, see https://github.com/tidyverse/dplyr/issues/6382#issuecomment-1320992575

- New `morph()` that allows an arbitrary number of rows to be returned per group computation
  - Always returns an ungrouped tibble/data.frame
  - Never messages, so no need for `.groups`
  - Does have support for `.by`
- Soft-deprecated `summarise()` with 0 or >1 row results in favor of `morph()`
- Updated all documentation and tests to reflect this, including vignettes

`morph()` is similar enough to `summarise()` that they can just share the underlying code. `summarise()` needs to continue to support the multi-row behavior for now anyways, as we slowly deprecate it. Eventually we can transition to having `summarise()` error immediately if it sees a `size != 1` result, but that is a long way off.

``` r

table <- c("a", "b", "d", "f")

df <- tibble(
  g = c(1, 1, 1, 2, 2, 2, 2),
  x = c("e", "a", "b", "c", "f", "d", "a")
)

df %>%
  morph(x = intersect(x, table))
#> # A tibble: 4 × 1
#>   x    
#>   <chr>
#> 1 a    
#> 2 b    
#> 3 f    
#> 4 d

df %>%
  morph(x = intersect(x, table), .by = g)
#> # A tibble: 5 × 2
#>       g x    
#>   <dbl> <chr>
#> 1     1 a    
#> 2     1 b    
#> 3     2 f    
#> 4     2 d    
#> 5     2 a

df %>%
  summarise(x = intersect(x, table), .by = g)
#> Warning: Returning more (or less) than 1 row per `summarise()` group was deprecated in
#> dplyr 1.1.0.
#> ℹ Please use `morph()` instead.
#> ℹ When switching from `summarise()` to `morph()`, remember that `morph()`
#>   always returns an ungrouped data frame and adjust accordingly.
#> # A tibble: 5 × 2
#>       g x    
#>   <dbl> <chr>
#> 1     1 a    
#> 2     1 b    
#> 3     2 f    
#> 4     2 d    
#> 5     2 a
```

<sup>Created on 2022-11-21 with [reprex v2.0.2.9000](https://reprex.tidyverse.org)</sup>


NEWS.md
```diff
@@ -1,5 +1,25 @@
 # dplyr (development version)
 
+* `morph()` is a new experimental verb that creates a new data frame by applying
+  functions to columns of an existing data frame. It is very similar to
+  `summarise()`, with two big differences:
+
+  * `morph()` can return an arbitrary number of rows per group, while
+    `summarise()` reduces each group down to a single row.
```
hadley, line 8:
```suggestion
    `summarise()` now reduces each group down to a single row.
```

NEWS.md
```diff
@@ -1,5 +1,25 @@
...
+
+  * `morph()` always returns an ungrouped data frame, while `summarise()` might
+    return a grouped or rowwise data frame, depending on the scenario.
+    
+  `morph()` has been added in response to valid concern from the community that
+  allowing `summarise()` to return 0 or >1 rows per group increases the chance
```
hadley, line 14:
```suggestion
  allowing `summarise()` to any number of rows per group increases the chance
```

R/morph.R
```diff
@@ -0,0 +1,114 @@
+#' Create a new data frame
```
hadley, line 1: Transform each group with an arbitrary number of rows.

R/morph.R
```diff
@@ -0,0 +1,114 @@
...
+#'
+#' @description
+#' `r lifecycle::badge("experimental")`
+#'
+#' `morph()` creates a new data frame by applying functions to columns of an
```
hadley, line 6: I think this could use a lead in like:

`summarise()` requires that each argument return a single value, and `mutate()` requires that each argument produces the same number of rows as the input. `morph()` is a more general workhorse with no requirements on the number of rows (except that they be consistent across arguments).

hadley (approved):
Only thing is to confirm the name; I'll put some more thoughts in the issue.

NEWS.md
DavisVaughan: Yea, I was trying to talk about `summarise()` here in the way that it should _ideally_ work, and I was relying on the other news bullet to mention the multirow deprecation

PR #6526, opened 2022-11-03
WIP: demo for relational API

krlmlr:
Needs https://github.com/duckdb/duckdb and https://github.com/duckdblabs/relational.

CC @hannes.

``` r
library(tidyverse)

options(duckdb.materialize_message = TRUE)

df <- palmerpenguins::penguins

# No materialization should happen here!
out <-
  df %>%
  slice_head(n = 10)

# Materializing should happen only here!
out
#> [1] TRUE
#> materializing:
#> ---------------------
#> --- Relation Tree ---
#> ---------------------
#> Limit 10
#>   r_dataframe_scan(0x108df73c8)
#> 
#> ---------------------
#> -- Result Columns  --
#> ---------------------
#> - species (species)
#> - island (island)
#> - bill_length_mm (DOUBLE)
#> - bill_depth_mm (DOUBLE)
#> - flipper_length_mm (INTEGER)
#> - body_mass_g (INTEGER)
#> - sex (sex)
#> - year (INTEGER)
#> 
#> # A tibble: 10 × 8
#>    species island    bill_length_mm bill_depth_mm flipper_…¹ body_…² sex    year
#>    <fct>   <fct>              <dbl>         <dbl>      <int>   <int> <fct> <int>
#>  1 Adelie  Torgersen           39.1          18.7        181    3750 male   2007
#>  2 Adelie  Torgersen           39.5          17.4        186    3800 fema…  2007
#>  3 Adelie  Torgersen           40.3          18          195    3250 fema…  2007
#>  4 Adelie  Torgersen           NA            NA           NA      NA <NA>   2007
#>  5 Adelie  Torgersen           36.7          19.3        193    3450 fema…  2007
#>  6 Adelie  Torgersen           39.3          20.6        190    3650 male   2007
#>  7 Adelie  Torgersen           38.9          17.8        181    3625 fema…  2007
#>  8 Adelie  Torgersen           39.2          19.6        195    4675 male   2007
#>  9 Adelie  Torgersen           34.1          18.1        193    3475 <NA>   2007
#> 10 Adelie  Torgersen           42            20.2        190    4250 <NA>   2007
#> # … with abbreviated variable names ¹​flipper_length_mm, ²​body_mass_g
```

<sup>Created on 2022-11-12 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

PR #6559, opened 2022-11-23
Move responsibility to compute dots and by to method

krlmlr:
Required for #6526.

Also fixes default argument of internal helper.


DavisVaughan:
@krlmlr do you still need this one? Right now `filter_rows()` and `slice_rows()` are rather symmetrical, which I like, so I'd rather only do this if we have to. 

We can definitely fix the bad default arg either way, of course

krlmlr:
Good catch, `slice_rows()` has the same problem, fixed that 🤪 

I've added comments to the code to explain the reasoning for the separation.

DavisVaughan:
@krlmlr would you please take a look at the CI?

DavisVaughan:
Actually I see the problem, I'll take over and manage merging, thanks!

krlmlr:
Thanks for taking over, I pushed too hastily.

PR #6563, opened 2022-11-24
Refactor mutate.data.frame()

krlmlr:
This method does four things:

1. Argument conversion
2. Execution
3. Getting input for postprocessing from execution results
4. Data conversion for postprocessing
5. Postprocessing

With #6526 and this PR, steps 1, 4 and 5 would remain unchanged, we'd substitute only step 2 and 3. Merging this early would support this effort.


krlmlr:
Thanks, done.

DavisVaughan:
Thanks @krlmlr!

PR #6569, opened 2022-11-29
Modified across dots deprecation warning

### LINKED ISSUE #6567: across `...` deprecation warning
Hi dplyr developers! I am so thrilled to see this release - the additional functionality is incredibly useful.

When I was working with the dev version today I saw the very helpful messaging in [across](https://github.com/tidyverse/dplyr/blob/f5ed6618c0d97ebf8910c4c6a39bdf9cff282e94/R/across.R#L213):

```
Warning message:
There was 1 warning in `mutate()`.
Caused by warning:
! The `...` argument of `across()` is deprecated as of dplyr 1.1.0.
Supply arguments directly to `.fns` through a lambda instead.

  # Previously
  across(a:b, mean, na.rm = TRUE)

  # Now
  across(a:b, ~mean(.x, na.rm = TRUE))
```

I was wondering:
1. If you would want to suggest  `across(a:b, \(x) mean(x, na.rm = TRUE))` to be consistent with [purrr examples](https://github.com/tidyverse/purrr/issues/936), and,
2. If the phrasing "anonymous function" might be more readily understood to users than "lambda" (though if they are using this syntax, they would likely be familiar).

shannonpileggi:
Closes #6567 

 - Modified dots deprecation warning in `across()`
 - Updated snapshot test accordingly

### DIFF
--- a/R/across.R
+++ b/R/across.R
@@ -210,13 +210,13 @@ across <- function(.cols,
 
   if (!missing(...)) {
     details <- c(
-      "Supply arguments directly to `.fns` through a lambda instead.",
+      "Supply arguments directly to `.fns` through an anonymous function instead.",
       "",
       " " = "# Previously",
       " " = "across(a:b, mean, na.rm = TRUE)",
       "",
       " " = "# Now",
-      " " = "across(a:b, ~mean(.x, na.rm = TRUE))"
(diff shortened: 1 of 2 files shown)


DavisVaughan:
Perfect, thanks so much!

Issue #2252, opened 2016-11-12
Revisit definition of slice(df)


krlmlr:
Currently returns all rows, but it's not tested anywhere. I think it should return an empty data frame instead, and slice(df, i1, i2) should be the same as slice(df, c(i1, i2)).

@hadley: Thoughts?

hadley:
I think the motivation was for `filter(x)` and `slice(x)` to give identical results, but maybe that's wrong because an empty logical should be treated as selecting all rows, whereas an empty integer should be treated as selecting no row.

Equally, `filter()` ANDs together the inputs, and here you're proposing (effectively) that slice OR's together the inputs.

krlmlr:
Yes, that's the way I'd see it -- perhaps not even OR, but UNION ALL. FWIW, slice(df, i1, i2) currently throws an error.

hadley:
Can you tackle this one too please? I think it should be a one-line change in `slice_impl()`.

krlmlr:
I have now found the [test](https://github.com/hadley/dplyr/blob/f39db50921110c3d23612cc81a7b3e027c0b3d1c/tests/testthat/test-slice.r#L23-L25). It seems that `slice()` intentionally accepts either 0 or one args, negative values are treated as exclusions. The current behavior actually makes some sense, I see no reason to change it.

We could still consider supporting multiple arguments to slice(), but only after #2311.

PR #6573, opened 2022-11-29
Drop all rows with empty `slice()` calls

DavisVaughan:
This revisits a point brought up in https://github.com/tidyverse/dplyr/issues/2252

I went to remove the early exit in `slice_rows()` and discovered it was actually being used to ensure that `slice()` with no `...` returns all rows. I think this is theoretically inconsistent.

I think that if we have:

```r
df <- data.frame(x = 1:2)
```

then:

```r
filter(df)

# Starts with this slice index
c(TRUE, TRUE)

# Uses `&` to combine it with the locations specified by `...`

# So it returns all rows
```

and:

```r
slice(df)

# Starts with this slice index
integer()

# Uses `c()` to combine it with the locations specified by `...`

# So it returns no rows
```

This is inline with the logic mentioned here:
https://github.com/tidyverse/dplyr/issues/2252#issuecomment-260128765

And with these two examples:

``` r
df <- data.frame(x = 1:2)

slice(df, c())
#> [1] x

slice(df, 1, 1)
#>   x
#> 1 1
#> 2 1
```

And it allows us to remove the early exit, which to me has always been an indication that the theoretical logic is sound

`"e53f5c67-a3fa-4b29-93b0-4df2710f40fa"`

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,8 @@
 # dplyr (development version)
 
+* `slice()` with no inputs now returns 0 rows. This is mostly for theoretical
+  consistency (#6573).
+
 * `rename_with()` now disallows renaming in the `.cols` tidy-selection (#6561).
 
 * `rename_with()` now checks that the result of `.fn` is the right type and size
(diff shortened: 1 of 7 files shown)


DavisVaughan:
I took a look through the 137 broken revdeps, and the only one that looks related to this is rsample, where I actually added some overly strict tests. I will go fix that, otherwise I think this is fine to merge

hadley (approved):

PR #6574, opened 2022-11-29
Add more error checking to `rename_with()`

### LINKED ISSUE #6561: `rename_with()` doesn't handle factors well
`rename_with.data.frame()` has _unexpected_ results when `.fn` returns a `factor` -- or at least unexpected for me.  Documentation does _suggest_ that `.fn` should return a `character` vector, but I wouldn't expect factors to be coerced to integers first.

The coercion occurs within `[<-`, which could be replaced with `vec_slice()`?

https://github.com/tidyverse/dplyr/blob/434caa2db5c1f48b9f0becc9295e547c1072260b/R/rename.R#L69

``` r
df <- data.frame(
  a = 1,
  b = 2,
  c = 3
)

foo <- function(x) {
  factor(toupper(x))
}

dplyr::rename_with(df, foo)
#>   1 2 3
#> 1 1 2 3


# name reassignment within rename_with.data.frame()
x <- c("a", "b", "c")
x[1:3] <- foo(x)
x
#> [1] "1" "2" "3"


# vctrs seems fine
x <- c("a", "b", "c")
vctrs::vec_slice(x, 2:3) <- foo(x[2:3])
x
#> [1] "a" "B" "C"
```

<sup>Created on 2022-11-23 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

DavisVaughan:
Closes #6561

### DIFF
--- a/R/rename.R
+++ b/R/rename.R
@@ -75,9 +75,9 @@ rename_with.data.frame <- function(.data, .fn, .cols = everything(), ...) {
       "{.arg .fn} must return a character vector, not {.obj_type_friendly {new}}."
     )
   }
-  if (vec_size(new) != vec_size(sel)) {
+  if (length(new) != length(sel)) {
     cli::cli_abort(
-      "{.arg .fn} must return a vector of size {vec_size(sel)}, not {vec_size(new)}."
+      "{.arg .fn} must return a vector of length {length(sel)}, not {length(new)}."
     )
   }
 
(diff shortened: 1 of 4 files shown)


R/rename.R
```diff
@@ -63,10 +63,25 @@ rename_with <- function(.data, .fn, .cols = everything(), ...) {
...
+
+  sel <- vec_slice(names, cols)
+  new <- .fn(sel, ...)
+
+  if (!is_character(new)) {
+    cli::cli_abort(
+      "{.arg .fn} must return a character vector, not {.obj_type_friendly {new}}."
+    )
+  }
+  if (vec_size(new) != vec_size(sel)) {
```
hadley, line 78: I can't tell if it would be clearer to use `length()` here, since we know that new and old are both character vectors.

hadley (approved):

PR #6576, opened 2022-11-29
Simplify `mutate()` even further

DavisVaughan:
@krlmlr I think I've managed to simplify the `mutate()` refactor even further.

This merges directly into your branch, would you mind taking a look?

### DIFF
--- a/R/mutate.R
+++ b/R/mutate.R
@@ -179,6 +179,7 @@ mutate.data.frame <- function(.data,
   by <- compute_by({{ .by }}, .data, by_arg = ".by", data_arg = ".data")
 
   cols <- mutate_cols(.data, dplyr_quosures(...), by)
+  used <- attr(cols, "used")
 
   out <- dplyr_col_modify(.data, cols)
 
@@ -191,7 +192,6 @@ mutate.data.frame <- function(.data,
     names_original = names_original
   )
 
-  used <- attr(cols, "used")
   names_exprs <- names(cols)
   names_groups <- by$names
 
--- a/R/mutate.R
+++ b/R/mutate.R
@@ -192,14 +192,14 @@ mutate.data.frame <- function(.data,
     names_original = names_original
   )
 
-  names_exprs <- names(cols)
+  names_new <- names(cols)
   names_groups <- by$names
 
   out <- mutate_keep(
     out = out,
     keep = keep,
     used = used,
-    names_exprs = names_exprs,
+    names_new = names_new,
     names_groups = names_groups
   )
 
@@ -229,7 +229,7 @@ mutate_relocate <- function(out, before, after, names_original) {
   )
 }
 
-mutate_keep <- function(out, keep, used, names_exprs, names_groups) {
+mutate_keep <- function(out, keep, used, names_new, names_groups) {
   names <- names(out)
 
   if (keep == "all") {
@@ -242,7 +242,7 @@ mutate_keep <- function(out, keep, used, names_exprs, names_groups) {
       none = character(),
       abort("Unknown `keep`.", .internal = TRUE)
     )
-    names_out <- intersect(names, c(names_exprs, names_groups, names_keep))
+    names_out <- intersect(names, c(names_new, names_groups, names_keep))
   }
 
   dplyr_col_select(out, names_out)
--- a/R/mutate.R
+++ b/R/mutate.R
@@ -222,7 +222,7 @@ mutate_relocate <- function(out, before, after, names_original) {
   names <- setdiff(names, names_original)
 
   relocate(
-    .data = out,
+    out,
     all_of(names),
     .before = !!before,
     .after = !!after
--- a/R/mutate.R
+++ b/R/mutate.R
@@ -174,70 +174,76 @@ mutate.data.frame <- function(.data,
                               .keep = c("all", "used", "unused", "none"),
                               .before = NULL,
                               .after = NULL) {
-  # dplyr front-end
   keep <- arg_match(.keep)
 
   by <- compute_by({{ .by }}, .data, by_arg = ".by", data_arg = ".data")
 
-  before <- enquo(.before)
-  after <- enquo(.after)
-
-  # Execution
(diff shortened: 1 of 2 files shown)


R/mutate.R
```diff
@@ -174,71 +174,77 @@ mutate.data.frame <- function(.data,
...
     return(out)
   }
 
-  # Only change the order of new columns
-  cols_expr_new <- setdiff(cols_expr, cols_expr_modified)
-  relocate(out, all_of(cols_expr_new), .before = !!before, .after = !!after)
+  # Only change the order of completely new columns that
+  # didn't exist in the original data
+  names <- names(out)
+  names <- setdiff(names, names_original)
```
DavisVaughan, line 222: I think this is now much easier to understand. The only columns that we should ever relocate are ones that weren't in the original data

R/mutate.R
```diff
@@ -174,71 +174,77 @@ mutate.data.frame <- function(.data,
...
-    cols_retain <- setdiff(cols_out, cols_unused)
+    names_out <- intersect(names, c(names_exprs, names_groups, names(used)[used]))
   } else if (keep == "unused") {
-    cols_used <- setdiff(cols_data, c(cols_group, cols_expr_modified, names(used)[!used]))
-    cols_retain <- setdiff(cols_out, cols_used)
+    names_out <- intersect(names, c(names_exprs, names_groups, names(used)[!used]))
   } else if (keep == "none") {
-    cols_any <- setdiff(cols_data, c(cols_group, cols_expr_modified))
-    cols_retain <- setdiff(cols_out, cols_any)
+    names_out <- intersect(names, c(names_exprs, names_groups))
```
DavisVaughan, line 242: In retrospect it seems somewhat obvious to me that this double `setdiff()` is really one `intersect()` call

It seems like much more straightforward code to me now

R/mutate.R
```diff
@@ -174,71 +174,77 @@ mutate.data.frame <- function(.data,
...
-  cols_expr_new <- setdiff(cols_expr, cols_expr_modified)
-  relocate(out, all_of(cols_expr_new), .before = !!before, .after = !!after)
+  # Only change the order of completely new columns that
+  # didn't exist in the original data
+  names <- names(out)
+  names <- setdiff(names, names_original)
+
+  relocate(
+    .data = out,
+    all_of(names),
```
hadley, line 226: Feels weird to name the first but not the second argument. (I understand why you did it, it just looks weird)

R/mutate.R
hadley: I think you could simplify this even further with something like:

```R
if (keep == "all") {
  names_keep <- names
} else if (keep == "used") {
  names_keep <- names(used)[used]
} ...

names_out <- intersect(names, c(names_groups, names_exprs, names_keep))
```

But I might've missed something in the logic.

DavisVaughan: I was trying to avoid the `intersect()` call in the 99% `"all"` case, just for a minimal performance boost

R/mutate.R
DavisVaughan: Oh this is a case where I'd feel comfortable not naming the first arg, I agree

R/mutate.R
hadley: Then maybe this?

```R
if (keep == "all") {
  names_out <- names
} else {
  names_keep <- switch(keep
    used = names(used)[used],
    unused =  names(used)[!used],
    none = character()
  )
  names_out <- intersect(names, c(names_groups, names_exprs, names_keep))
}
```

DavisVaughan: Oh yea I think that would be a little nicer

R/mutate.R
```diff
@@ -174,71 +174,80 @@ mutate.data.frame <- function(.data,
...
+    out = out,
+    before = {{ .before }},
+    after = {{ .after }},
+    names_original = names_original
+  )
 
-  # Convert input for postprocessing
-  cols_group <- by$names
-  cols_expr <- names(cols)
+  used <- attr(cols, "used")
```
krlmlr, line 194: I moved `used <-` up, because it's going to be provided by the implementation and not always by some obscure attribute in `cols` .

R/mutate.R
```diff
@@ -174,71 +174,80 @@ mutate.data.frame <- function(.data,
...
-  # Compact out `NULL` columns that got removed.
-  # These won't exist in `out`, but we don't want them to look "new".
-  # Note that `dplyr_col_modify()` makes it impossible to `NULL` a group column,
-  # which we rely on below.
-  cols <- compact_null(cols)
+  out <- mutate_relocate(
+    out = out,
+    before = {{ .before }},
+    after = {{ .after }},
+    names_original = names_original
```
krlmlr, line 191: Should this perhaps be

```suggestion
    names_original = names(.data)
```

with removing the extracted variable above?

krlmlr (approved):
I love that `compact_null(cols)` seems gone now without replacement.

R/mutate.R
```diff
@@ -174,71 +174,80 @@ mutate.data.frame <- function(.data,
...
+    out,
+    all_of(names),
+    .before = !!before,
+    .after = !!after
+  )
 }
 
-mutate_keep <- function(out, keep, used, cols_data, cols_group, cols_expr_modified) {
-  cols_out <- names(out)
+mutate_keep <- function(out, keep, used, names_exprs, names_groups) {
```
hadley, line 232: Instead of `names_exprs` , maybe `names_new`?

DavisVaughan: I had that at one point but since you often use `mutate()` to overwrite existing variables it ended up being a little confusing to me, since those aren't technically new

hadley: I agree that it's not ideal, but it still seems mildly less confusing than `names_exprs` to me.

R/mutate.R
DavisVaughan: I think I prefer having the local variable there because I think it is a little easier to read. In general I like collecting the variables up front right before calling the function

PR #6584, opened 2022-12-02
Changed function argument alias

llayman12:
Changed the function argument `.fn` to `.fns` to align with across.R


hadley:
`across()` can accept multiple functions, hence the plural.

PR #6590, opened 2022-12-07
Add more basic explanation to case-when documentation and indicate order of evaluation

jpeacock29:


R/case-when.R
```diff
@@ -146,7 +147,7 @@
 #'   mutate(type = case_character_type(height, mass, species, robots = FALSE)) %>%
 #'   pull(type)
 case_when <- function(...,
-                      .default = NULL,
+                      .default = NA,
```
hadley, line 150: I changed the default to `NA` here — this seems intuitively correct to me, doesn't break any unit tests, and IMO removes one layer of indirection.

DavisVaughan: I think we should consider this in a broader issue, as it also affects `case_match()`, `vec_case_when()`, and `vec_case_match()`. I'm also not sure I agree.

I've reverted for now just so we can more easily merge this, but I'll open an issue

DavisVaughan:
Thanks @jpeacock29!

jpeacock29:
Cheers, thank you as well!

PR #6594, opened 2022-12-10
Add comments to separate common from backend-specific code

krlmlr:
For duckplyr I'm implementing machinery that extracts code from dplyr. This will allow keeping duckplyr development in sync with dplyr. We're basically feature-complete from day one. We now "only" need to provide duckdb implementations verb by verb.

A lot of argument checking and transformation code will be reused too. I wonder if we could keep anchor comments of the form provided in this PR in the dplyr package. My machinery would pick up those comments, so that synchronization can remain idempotent and pick up future changes to dplyr code.

If we can agree on those comments, I would supply further PRs as duckplyr development progresses. Occasionally, refactorings in the spirit of https://github.com/tidyverse/dplyr/pull/6563 and https://github.com/tidyverse/dplyr/pull/6559 may be helpful too.

CC @hannes.


krlmlr:
Not necessary.

PR #6608, opened 2022-12-15
Doc tweaks for `.by`

### LINKED ISSUE #6582: Should `with_groups()` be deprecated in favor of `.by` argument?
In my understanding, most of the usages of `with_groups()` will be superseded by `.by`. Shouldn't it be marked as deprecated or superseded?

### LINKED ISSUE #6585: Suggested wording for `.by` documentation: "per-operation" instead of "temporary"
Currently, `.by` documentation at [R/by.R](https://github.com/tidyverse/dplyr/blob/HEAD/R/by.R) calls `.by` temporary and `group_by()`. 

I think calling `.by` "temporary" only makes sense in the presence of, or as a newly developed alternative to `group_by()`. However, if you're learning dplyr today for the first time (or after this release is out) and you learn `.by` first instead of `group_by()`, and have never heard of `group_by()` or don't know how it works, the word "temporary" doesn't make as much sense, I think.

The [blog post](https://www.tidyverse.org/blog/2022/11/dplyr-1-1-0-is-coming-soon/#temporary-grouping-with-by) uses the following wording for data.table:

> This feature was inspired by [data.table](https://cran.r-project.org/package=data.table), which has always used per-operation grouping.

I like the term "per-operation" grouping better than temporary and I think it's easier to understand even if you don't know about `group_by()` behaviour.

I would suggest something like the following:

> There are two ways to group in dplyr:
>
> - Persistent grouping with `group_by()`
> - Per operation grouping with `.by`
>
> This help page is dedicated to explaining where and why you might want to use the latter. Grouping radically affects the computation of the dplyr verb you use it with, and one of the goals of `.by` is to allow you to place that grouping specification within the function that actually uses it. As an added benefit, with `.by` don't need to remember to `ungroup()` after `summarise()`, and `summarise()` won't message you about how it's handling the groups!

hadley:
Fixes #6582. Fixes #6585.


NEWS.md
```diff
@@ -1,5 +1,7 @@
 # dplyr (development version)
 
+* `with_groups()` is superseded in favour of `.by` (#6582)
```
```suggestion
* `with_groups()` is superseded in favour of `.by` (#6582).
```

R/by.R
```diff
@@ -5,14 +5,14 @@
 #' @param .by `r lifecycle::badge("experimental")`
 #'
 #'   <[`tidy-select`][dplyr_tidy_select]> Optionally, a selection of columns to
-#'   temporarily group by using an inline alternative to [group_by()]. For
+#'   group by for just this operation, as an alternative to [group_by()]. For
```
DavisVaughan, line 8:
```suggestion
#'   group by for just this operation, functioning as an alternative to [group_by()]. For
```

For some reason it sounds kind of strange without this. Might need to reflow to line wrap it

man/rmd/by.Rmd
```diff
@@ -12,7 +12,7 @@ There are two ways to group in dplyr:
 
 -   Persistent grouping with [group_by()]
 
--   Temporary grouping with `.by`
+-   Per-operation, temporary, grouping with `.by`
```
DavisVaughan, line 15:
```suggestion
-   Per-operation grouping with `.by`
```

I think it is fine without this

DavisVaughan (approved):

PR #6612, opened 2022-12-16
Implement `cross_join()`

### LINKED ISSUE #6604: `cross_join()` in place of `by = character()`
Rationale for adding an explicit `cross_join()` function in favor of the existing `by = character()` method:

* I think `by = character()` is fairly unintuitive, and I’ve seen 2 cases recently where people didn’t even know it existed

* I think `by = character()` should really be an error. I think its current behavior of doing a cross join is actually a little scary. Like, supplying `c("x", "y")` matches on two _columns_, supplying `"x"` matches on one _column_, but `character()` all of the sudden matches all _rows_? Seems very odd to me.

* Importantly, currently only `inner_join(by = character())` is a _true_ cross join. SQL cross joins must result in `nrow(x) * nrow(y)` number of rows, and only the inner cross join does this in the edge case where one of the inputs has 0 rows (i.e. the left join would retain all x rows). I know this is edge case behavior, but the weirdness here feels worth resolving by just adding `cross_join()`

* I think it would be more straightforward for dbplyr and friends to convert `cross_join()` to SQL. Converting `left_join(by = character())` to SQL is probably quite odd.

* It would be nice that `cross_join()` wouldn’t have a `by` argument at all, making it very clear it doesn’t have anything to do with columns

* We could document `cross_join()` on its own help page, raising awareness of it

* `join_by()` wouldn’t have to support cross joins, which it currently does by allowing empty `join_by()` calls to be treated like cross joins. I could turn that into an error for 1.1.0 if we also added `cross_join()` at the same time, which would be much nicer than having to deprecate it slowly in the future.

DavisVaughan:
Closes #6604 

And soft deprecate `by = character()`


R/join-by.R
```diff
@@ -243,40 +251,35 @@ join_by <- function(...) {
...
   condition <- flat_map_chr(bys, function(by) by$condition)
 
-  # Cross join for empty `join_by()` calls
-  cross <- n == 0L
-
   new_join_by(
     exprs = exprs,
     condition = condition,
     filter = filter,
-    cross = cross,
```
DavisVaughan, line 253: Felt "right" to be able to remove cross join handling from `join_by()`

R/join-by.R
```diff
@@ -243,40 +251,35 @@ join_by <- function(...) {
...
 
   invisible(x)
 }
 
-new_join_by <- function(exprs, condition, filter, cross, x, y) {
+new_join_by <- function(exprs = list(),
+                        condition = character(),
+                        filter = character(),
+                        x = character(),
+                        y = character()) {
```
DavisVaughan, line 278: Users can never create empty `join_by()` calls, but we can internally make an empty `new_join_by()` object, which is useful for `cross_join()` and our backwards compatible support for `by = character()`

R/join-by.R
```diff
@@ -324,22 +327,35 @@ as_join_by.list <- function(x, error_call = caller_env()) {
   # TODO: check lengths
   x_names <- x[["x"]]
   y_names <- x[["y"]]
+
+  if (!is_character(x_names)) {
+    abort("`by$x` must evaluate to a character vector.")
+  }
+  if (!is_character(y_names)) {
+    abort("`by$y` must evaluate to a character vector.")
+  }
```
DavisVaughan, line 336: Added some more error handling for some obscure support for `by = list(x = , y = )`, which I don't think many people use but it was causing a weird error without this

R/join-cross.R
```diff
@@ -0,0 +1,93 @@
...
+  y_in <- as_tibble(y, .name_repair = "minimal")
+
+  x_size <- vec_size(x_in)
+  y_size <- vec_size(y_in)
+
+  x_out <- set_names(x_in, names(vars$x$out))
+  y_out <- set_names(y_in, names(vars$y$out))
+
+  x_out <- vec_rep_each(x_out, times = y_size)
+  y_out <- vec_rep(y_out, times = x_size)
```
DavisVaughan, line 88: This should be faster than the previous approach too. No need to actually "locate" the matches using `vec_locate_matches()`.

R/join-rows.R
```diff
@@ -19,19 +19,27 @@ join_rows <- function(x_key,
   y_unmatched <- unmatched$y
 
   if (cross) {
+    # TODO: Remove this section when `by = character()` is defunct
```
DavisVaughan, line 22: Unfortunately we do still need this section in `join_rows()` for backwards compatible support, we can't just call out to `cross_join()` or something.

_Technically_ the `unmatched = "error"` and `multiple = "error"` args are supposed to do something even when you set `by = character()`. These aren't meaningful things to do IMO, which is why `cross_join()` doesn't have these args, but the simplest way to keep backwards compatible support here seemed to be to just continue to let it all run through `vec_locate_matches()`. I've added tests of the `unmatched` and `multiple` interaction with `by = character()`.

I did go ahead and aggregate any usage of `cross` below into a single `if (cross)` branch that will be easy to remove in the future.

R/join.R
```diff
@@ -759,3 +780,18 @@ check_keep <- function(keep, error_call = caller_env()) {
       call = error_call)
   }
 }
+
+is_cross_by <- function(x) {
+  identical(x, character()) || identical(x, list(x = character(), y = character()))
+}
```
DavisVaughan, line 786: The 2nd option here is a technically valid way to do a cross join in CRAN dplyr

NEWS.md
```diff
@@ -1,5 +1,9 @@
 # dplyr (development version)
 
+* New `cross_join()` that is a more explicit and slightly more correct
```
hadley, line 3: Can you break this into two bullets, since with the news re-org, one will go in "new features" and one will go in "newly deprecated"

hadley (approved):

PR #6642, opened 2023-01-17
Vectorize `add_suffixes()`

DavisVaughan:
Closes https://github.com/tidyverse/dplyr/pull/6636

Alternate approach that truly vectorizes the calculation. I'm about 95% sure this is correct. It passes all tests and we do have a decent number of tests that check corner cases of `suffix` so I'm fairly confident about it.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -339,6 +339,9 @@ package, bringing greater consistency and improved performance.
 * Joins now reference the correct column in `y` when a type error is thrown
   while joining on two columns with different names (#6465).
 
+* Joins on very wide tables are no longer limited by the application of `suffix`
+  (#6642).
+
 * `*_join()` now error if you supply them with additional arguments that
   aren't used (#6228).
 
--- a/NEWS.md
+++ b/NEWS.md
@@ -339,8 +339,8 @@ package, bringing greater consistency and improved performance.
 * Joins now reference the correct column in `y` when a type error is thrown
   while joining on two columns with different names (#6465).
 
-* Joins on very wide tables are no longer limited by the application of `suffix`
-  (#6642).
+* Joins on very wide tables are no longer bottlenecked by the application of
+  `suffix` (#6642).
 
 * `*_join()` now error if you supply them with additional arguments that
   aren't used (#6228).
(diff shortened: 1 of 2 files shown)


R/join-cols.R
```diff
@@ -157,21 +157,27 @@ standardise_join_suffix <- function(x,
...
-  out <- rep_along(x, na_chr)
-  for (i in seq_along(x)) {
-    nm <- x[[i]]
-    while (nm %in% y || nm %in% out[seq_len(i - 1)]) {
-      nm <- paste0(nm, suffix)
-    }
+  x <- c(y, x)
+
+  # Never marks the "first" duplicate (i.e. never anything in `y`)
+  dup <- duplicated(x)
```
hadley, line 170: Since we're doing performance optimisations, I think this might be a little faster: 

```R
dup <- anyDuplicated(x)
while (length(dup) > 0) { 
  x[dup] <- paste0(x[dup], suffix)
  dup <- anyDuplicated(x)
}
```

NEWS.md
```diff
@@ -339,6 +339,9 @@ package, bringing greater consistency and improved performance.
 * Joins now reference the correct column in `y` when a type error is thrown
   while joining on two columns with different names (#6465).
 
+* Joins on very wide tables are no longer limited by the application of `suffix`
```
hadley, line 342:
```suggestion
* Joins on very wide tables are no longer bottlenecked by the application of `suffix`
```

hadley (approved):

R/join-cols.R
DavisVaughan: Would it? If there are multiple columns that are duplicated between `x` and `y` then my approach tackles them all in one pass. We only ever do a 2nd pass at all in the super rare case where adding a suffix created another duplication.

`anyDuplicated()` returns exactly 1 value, the location of the first duplicated value or `0` if there are no dups. So we'd have to do (at least) as many passes as there are duplicates in `x` and `y`

DavisVaughan: Related to https://github.com/r-lib/vctrs/issues/1239 since it would be nice to be able to use `vec_detect_duplicate(x, ignore = "first")` here

hadley: Oh yeah, good point.

PR #6646, opened 2023-01-19
error in `across()` when supplied `...` without a `.fns` argument

### LINKED ISSUE #6638: across() fails silently when using .funs (used in scoped mutate) instead of .fns
[across](https://dplyr.tidyverse.org/reference/across.html#arguments)() uses `.fns =`  as the argument to supply an anonymous function. 
This is different than the [scoped mutate functions](https://dplyr.tidyverse.org/reference/mutate_all.html?q=mutate_at#arguments), which used `.funs =`, with an extra "u".

This causes some problems when rewriting code from scoped syntax `mutate_at(.funs = ~)` to the new `mutate(across(.fns = ~))`, because supplying the wrong argument in `across()` (namely, `.funs` instead of `.fns`) fails silently, as shown below.

``` r
library(tidyverse)

mtcars <- as_tibble(mtcars) %>% select(mpg)

# Using the named argument ".fns" applies the function, as expected
mtcars %>% mutate(across(mpg, .fns = ~ . * 1000))
#> # A tibble: 32 × 1
#>      mpg
#>    <dbl>
#>  1 21000
#>  2 21000
#>  3 22800
#>  4 21400
#>  5 18700
#>  6 18100
#>  7 14300
#>  8 24400
#>  9 22800
#> 10 19200
#> # … with 22 more rows

# Using the named argument ".funs" fails, silently
mtcars %>% mutate(across(mpg, .funs = ~ . * 1000))
#> # A tibble: 32 × 1
#>      mpg
#>    <dbl>
#>  1  21  
#>  2  21  
#>  3  22.8
#>  4  21.4
#>  5  18.7
#>  6  18.1
#>  7  14.3
#>  8  24.4
#>  9  22.8
#> 10  19.2
#> # … with 22 more rows
```

<sup>Created on 2023-01-12 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

As the argument is ignored, the result is no different than failing to supply any function at all.
(Supplying the function without naming the argument works as expected.)

Perhaps `across` should throw a warning when it does not detect an anonymous function being supplied?

simonpcouch:
Closes #6638. cc @DavisVaughan

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -355,6 +355,9 @@ package, bringing greater consistency and improved performance.
 * Functions supplied to `across()` are no longer masked by columns (#6545). For
   instance, `across(1:2, mean)` will now work as expected even if there is a
   column called `mean`.
+  
+* `across()` will now error when supplied `...` without a `.fns` argument 
+  (#6638).
 
 * `arrange()` now correctly ignores `NULL` inputs (#6193).
 
--- a/R/across.R
+++ b/R/across.R
@@ -192,6 +192,10 @@ across <- function(.cols,
     # Silent restoration to old defaults of `.fns` for now.
     # TODO: Escalate this to formal deprecation.
     .fns <- NULL
+
+    # Catch if dots are non-empty with no `.fns` supplied.
+    # Mainly catches typos, e.g. `.funs` (#6638).
+    check_dots_empty0(...)
   } else {
     .fns <- quo_eval_fns(fns_quo, mask = fns_quo_env, error_call = error_call)
   }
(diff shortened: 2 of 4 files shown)


DavisVaughan:
@hadley this is the outcome of today's pairing session. Seem reasonable?

In a perfect world we'd mention "`...` must be empty _when `.fns` isn't supplied_" but I still feel like the error is an improvement on what we had before (which was the deprecation warning about supplying `...` in `across()`). The nice part about this error is that `check_dots_empty0()` will show you which arg caused the problem, and that should be enough for you to figure out you had a typo.

I'm not hugely worried about throwing a great error here anyways. When we escalate the deprecation so that not supplying `.fns` is an error, we can remove all of this anyways.

hadley (approved):

Issue #6639, opened 2023-01-12
Mark `nest_by()` and `group_*()` as questioning rather than experimental


DavisVaughan:
- Since we will probably have `nest(.by =)` in tidyr
- And the `group_*()` family doesn't fit super well in conjunction with the new dplyr `.by` ideas

manhnguyen48:
Hi, `dplyr::nest_by` returns a rowwise data frame while `tidyr::nest(.by=)` doesn't. Those are useful to organise tibbles within tibble and not having to use `purrr::map` on every `mutate `calls. If `nest_by` is being phased out perhaps `tidyr::nest()` could gain a `.rowwise=TRUE` argument for backward compatibility?

hadley:
@manhnguyen48 we are considering a different way to achieve a `rowwise()` since `mutate()` is really the only function that `rowwise()` affects, suggesting that creating a custom class isn't the right approach here.

PR #6657, opened 2023-01-24
Note instability of `nest_by()` and `group_nest/split()`

DavisVaughan:
Closes https://github.com/tidyverse/dplyr/issues/6639

- No NEWS bullet (I'm not really sure what it would say)
- All still experimental
- The `Life cycle` section is something rlang does. i.e. `?rlang::int`
- Marked `group_split()` as internal and removed pkgdown reference index

### DIFF
--- a/R/group-nest.R
+++ b/R/group-nest.R
@@ -10,7 +10,7 @@ group_nest_impl <- function(.tbl, .key, keep = FALSE){
 #'
 #' Nest a tibble using a grouping specification
 #'
-#' @section Life cycle:
+#' @section Lifecycle:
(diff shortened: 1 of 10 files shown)


R/group-nest.R
```diff
@@ -10,6 +10,10 @@ group_nest_impl <- function(.tbl, .key, keep = FALSE){
 #'
 #' Nest a tibble using a grouping specification
 #'
+#' @section Life cycle:
```
hadley, line 13:
```suggestion
#' @section Lifecycle:
```

?

hadley (approved):

R/group-nest.R
DavisVaughan: Oh i copied rlang but yea that's better

PR #6658, opened 2023-01-25
Explicitly test reusing column with `rename()`

krlmlr:
This only adds a test for behavior that is tested implicitly in the tests for `relocate()` .


DavisVaughan:
Thanks!

PR #6667, opened 2023-01-28
Add braces to methods

krlmlr:
This slightly simplifies extraction from dplyr's implementation needed elsewhere.


DavisVaughan:
Thanks!

PR #6690, opened 2023-02-06
Add `rename_with()` example of using `paste0(recycle0 = TRUE)`

### LINKED ISSUE #6688: `rename_with` with empty column selection fails w/ dplyr 1.1.0.
Hi,

due to the size check in the new dplyr version, dplyr 1.1.0, renames with empty column selections don't work anymore (since it expects a vector of functions of size 0).
This is of course not a big problem when used interactively, since one can simply omit the whole statement.
But when used programmatically in a data pipeline it forces the programmer to always treat this case separately (if for instance the .cols argument is defined by external dynamical parameters).

I would expect this function to simply do nothing if no columns are selected, and not throw an error.

``` r

tibble(col = c(1,3)) %>%
  rename_with(
    .cols = contains("test"),
    .fn = ~ paste0("blah.", .)
  )
#> Error in `rename_with()`:
#> ! `.fn` must return a vector of length 0, not 1.

#> Backtrace:
#>     ▆
#>  1. ├─tibble(col = c(1, 3)) %>% ...
#>  2. ├─dplyr::rename_with(...)
#>  3. └─dplyr:::rename_with.data.frame(., .cols = contains("test"), .fn = ~paste0("blah.", .))
#>  4.   └─cli::cli_abort("{.arg .fn} must return a vector of length {length(sel)}, not {length(new)}.")
#>  5.     └─rlang::abort(...)
```

<sup>Created on 2023-02-06 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

DavisVaughan:
Closes #6688

I think that documentation is really the best we can do here? It seems worth calling this specific case out, at least.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,8 @@
 # dplyr (development version)
 
+* `rename_with()` now includes an example of using `paste0(recycle0 = TRUE)` to
+  correctly handle empty selections (#6688).
+
 # dplyr 1.1.0
 
 ## New features
--- a/R/rename.R
+++ b/R/rename.R
@@ -44,6 +44,21 @@
 #' rename_with(iris, toupper)
 #' rename_with(iris, toupper, starts_with("Petal"))
 #' rename_with(iris, ~ tolower(gsub(".", "_", .x, fixed = TRUE)))
+#'
+#' @examplesIf getRversion() > "4.0.1"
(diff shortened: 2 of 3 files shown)


hadley (commented):
Yeah, if we do anything, this is the best we can do. And I think it's worth doing _something_.

Issue #6696, opened 2023-02-08
Can't install: chop.cpp:70:36: error: ‘LOGICAL_ELT’ was not declared in this scope


pur80a:
Please briefly describe your problem and what output you expect. If you have a question, please don't use this form. Instead, ask on <https://stackoverflow.com/> or <https://community.rstudio.com/>.
```
$ uname -a
Linux login02.lisc.univie.ac.at 4.18.0-425.10.1.el8_7.x86_64 #1 SMP Thu Jan 12 11:31:50 PST 2023 x86_64 x86_64 x86_64 GNU/Linux

$ gcc --version | head -1
gcc (GCC) 8.5.0 20210514 (Red Hat 8.5.0-16.0.2)

$ R --version | head -1
R version 3.4.4 (2018-03-15) -- "Someone to Lean On"

$ echo 'install.packages("dplyr")' | R --no-save >> /dev/null
trying URL 'https://cloud.r-project.org/src/contrib/dplyr_1.1.0.tar.gz'
Content type 'application/x-gzip' length 1068347 bytes (1.0 MB)
==================================================
downloaded 1.0 MB

* installing *source* package ‘dplyr’ ...
** package ‘dplyr’ successfully unpacked and MD5 sums checked
** libs
chop.cpp: In function ‘SEXPREC* dplyr_lazy_vec_chop(SEXP, SEXP, SEXP, SEXP)’:
chop.cpp:70:36: error: ‘LOGICAL_ELT’ was not declared in this scope
   bool grouped = static_cast<bool>(LOGICAL_ELT(ffi_grouped, 0));
                                    ^~~~~~~~~~~
chop.cpp:70:36: note: suggested alternative: ‘LOGICAL_RO’
   bool grouped = static_cast<bool>(LOGICAL_ELT(ffi_grouped, 0));
                                    ^~~~~~~~~~~
                                    LOGICAL_RO
make: *** [/home/apps/R/3.4.4/lib64/R/etc/Makeconf:168: chop.o] Error 1
ERROR: compilation failed for package ‘dplyr’
* removing ‘/home/apps/R/3.4.4/lib64/R/library/dplyr’
* restoring previous ‘/home/apps/R/3.4.4/lib64/R/library/dplyr’

The downloaded source packages are in
        ‘/tmp/RtmpAoq2Ps/downloaded_packages’
Updating HTML index of packages in '.Library'
Making 'packages.html' ... done
Warning message:
In install.packages("dplyr") :
[...]

PR #6700, opened 2023-02-08
Require R >=3.5.0

DavisVaughan:
Closes https://github.com/tidyverse/dplyr/issues/6696

@hadley do you agree with this?


hadley (approved):
Yup

Issue #6695, opened 2023-02-08
Two updates to `?dplyr_extending`


DavisVaughan:
https://dplyr.tidyverse.org/reference/dplyr_extending.html

---

> If you have data frame attributes that don't depend on the rows or columns (and should unconditionally be preserved), you don't need to do anything.

If you extend a `"data.frame"` rather than a tibble, then you need to provide a `[.class` method to make it preserve attributes. See #6689 and https://github.com/tidyverse/dplyr/issues/6691

---

Current usage section for `mutate()` should be updated to also include that it uses 1-D `[` to implement `.keep`, and that is unconditionally called even if keeping everything.

---

Maybe do a pass through everything in Current usage to see if we are up to date

PR #6701, opened 2023-02-08
Refresh `?dplyr_extending`

DavisVaughan:
Closes https://github.com/tidyverse/dplyr/issues/6695


hadley (commented):
Just a note that if https://dplyr.tidyverse.org/articles/compatibility.html becomes `vignette("dplyr-in-packages")` then we should make sure to mention this doc there too.

Issue #6666, opened 2023-01-26
Expressions using `<-` can overwrite active bindings in the data mask


DavisVaughan:
Seen in metacore

``` r
library(dplyr, warn.conflicts = FALSE)

foo <- tibble(
   x = c("CHG", "CHG", "CHG", "ABLFL", "ABLFL"),
   y = c("float", "integer", "float", "text", "text")
)
foo <- group_by(foo, x)

mutate(foo, y = {
   attr(y, "label") <- NULL
   y
})
#> Error in `mutate()`:
#> ℹ In argument: `y = { ... }`.
#> ℹ In group 2: `x = "CHG"`.
#> Caused by error:
#> ! `y` must be size 3 or 1, not 2.

#> Backtrace:
#>      ▆
#>   1. ├─dplyr::mutate(...)
#>   2. ├─dplyr:::mutate.data.frame(...) at dplyr/R/mutate.R:146:2
#>   3. │ └─dplyr:::mutate_cols(.data, dplyr_quosures(...), by) at dplyr/R/mutate.R:181:2
#>   4. │   ├─base::withCallingHandlers(...) at dplyr/R/mutate.R:268:2
#>   5. │   └─dplyr:::mutate_col(dots[[i]], data, mask, new_columns) at dplyr/R/mutate.R:273:6
#>   6. │     └─mask$eval_all_mutate(quo) at dplyr/R/mutate.R:380:8
#>   7. │       └─dplyr (local) eval() at dplyr/R/data-mask.R:69:6
#>   8. ├─dplyr:::dplyr_internal_error(...)
#>   9. │ └─rlang::abort(class = c(class, "dplyr:::internal_error"), dplyr_error_data = data) at dplyr/R/conditions.R:194:2
#>  10. │   └─rlang:::signal_abort(cnd, .file)
#>  11. │     └─base::signalCondition(cnd)
#>  12. └─dplyr (local) `<fn>`(`<dpl:::__>`)
#>  13.   └─rlang::abort(message, class = error_class, parent = parent, call = error_call) at dplyr/R/conditions.R:235:4
```

<sup>Created on 2023-01-26 with [reprex v2.0.2.9000](https://reprex.tidyverse.org)</sup>

DavisVaughan:
Simpler example, still weird. Seems like `<-` is messing up the active binding or something

``` r
library(dplyr, warn.conflicts = FALSE)

foo <- tibble(
  x = c("a", "b"),
  y = c("c", "d")
)
[...]

PR #6704, opened 2023-02-09
Evaluate each group expression in a fresh data mask

DavisVaughan:
Closes https://github.com/tidyverse/dplyr/issues/6666

This is a bit of a tricky one.

The issues in #6666 are caused by the fact that every evaluation within a single `mutate()` call (both across multiple expressions and across groups within the same expression) share the same data mask.

This means that assignments with `<-` in one expression can affect downstream expressions or group evaluations. This is most problematic when the assignment is like `x <- ` where `x` is a column in the original data frame. That assignment will _mask_ the active binding we set up for `x`, so it will no longer evaluate to the correct expression in the following expressions or groups.

My idea here is to create a _fresh_ data mask for each individual evaluation (again, both across expressions and across groups). This ensures that each evaluation is done in a clean environment that shouldn't be polluted by previous evaluations.

This works, but the main downside is that it was slower due to the repeated data mask creation. With https://github.com/r-lib/rlang/pull/1553 and https://github.com/r-lib/rlang/pull/1555, I'm hopeful that I've fixed most of the performance issues as long as the user has R >=4.1.0 (which includes the C function `R_NewEnv()`, otherwise mask creation involves an R callback to `new.env()`). In that case, the overhead is barely noticeable.

Below is a set of the absolute worst possible cases, which involves an expression that essentially does nothing, like `identity(x)`, so we basically just measure the overhead of `mutate()`. I've tested over various group sizes. As soon as you add in any expressions that do any "work," the difference in performance isn't noticeable because the expression itself dominates the timings (I show one of these as the last example).

As long as the user has R >=4.1.0 (released May 2021), then they probably won't notice any changes at all. 

I've included timings with CRAN rlang to show that those 2 rlang PRs really do make a difference, and that we should wait for dev rlang to hit CRAN before sending dplyr in, even though _technically_ we don't have to.

One way to interpret these benchmarks is that with 100k groups, this "fresh" data mask idea adds ~50ms of overhead, which seems reasonable for the safety it gets us.

``` r
library(dplyr)

# 1,000 groups
df <- tibble(g = seq_len(1000), x = seq_along(g))
bench::mark(mutate(df, y = identity(x), .by = g), iterations = 1000)

# CRAN dplyr, CRAN rlang
#> # A tibble: 1 × 6
#>   expression                                min   median itr/s…¹ mem_a…² gc/se…³
#>   <bch:expr>                           <bch:tm> <bch:tm>   <dbl> <bch:b>   <dbl>
#> 1 mutate(df, y = identity(x), .by = g)   4.12ms   5.02ms    194.  2.36MB    11.3
#> # … with abbreviated variable names ¹​`itr/sec`, ²​mem_alloc, ³​`gc/sec`

# This PR, dev rlang, R >=4.1.0
#> # A tibble: 1 × 6
#>   expression                                min   median itr/s…¹ mem_a…² gc/se…³
#>   <bch:expr>                           <bch:tm> <bch:tm>   <dbl> <bch:b>   <dbl>
#> 1 mutate(df, y = identity(x), .by = g)   4.33ms   5.27ms    188.  3.52MB    14.8
#> # … with abbreviated variable names ¹​`itr/sec`, ²​mem_alloc, ³​`gc/sec`

# This PR, dev rlang, R <4.1.0
#> # A tibble: 1 × 6
#>   expression                                min   median itr/s…¹ mem_a…² gc/se…³
#>   <bch:expr>                           <bch:tm> <bch:tm>   <dbl> <bch:b>   <dbl>
#> 1 mutate(df, y = identity(x), .by = g)    3.8ms   5.75ms    173.  2.45MB    20.8
#> # … with abbreviated variable names ¹​`itr/sec`, ²​mem_alloc, ³​`gc/sec`

# This PR, CRAN rlang
#> # A tibble: 1 × 6
#>   expression                                min   median itr/s…¹ mem_a…² gc/se…³
#>   <bch:expr>                           <bch:tm> <bch:tm>   <dbl> <bch:b>   <dbl>
#> 1 mutate(df, y = identity(x), .by = g)   5.96ms   7.42ms    134.  4.06MB    19.4
#> # … with abbreviated variable names ¹​`itr/sec`, ²​mem_alloc, ³​`gc/sec`
```

```r
# 10,000 groups
df <- tibble(g = seq_len(10000), x = seq_along(g))
bench::mark(mutate(df, y = identity(x), .by = g), iterations = 100)

# CRAN dplyr, CRAN rlang
#> # A tibble: 1 × 6
#>   expression                                min   median itr/s…¹ mem_a…² gc/se…³
#>   <bch:expr>                           <bch:tm> <bch:tm>   <dbl> <bch:b>   <dbl>
#> 1 mutate(df, y = identity(x), .by = g)   22.8ms   24.8ms    40.3  1.56MB    30.4
#> # … with abbreviated variable names ¹​`itr/sec`, ²​mem_alloc, ³​`gc/sec`

# This PR, dev rlang, R >=4.1.0
#> # A tibble: 1 × 6
#>   expression                                min   median itr/s…¹ mem_a…² gc/se…³
#>   <bch:expr>                           <bch:tm> <bch:tm>   <dbl> <bch:b>   <dbl>
#> 1 mutate(df, y = identity(x), .by = g)   23.8ms   26.1ms    38.0  1.56MB    80.8
#> # … with abbreviated variable names ¹​`itr/sec`, ²​mem_alloc, ³​`gc/sec`

# This PR, dev rlang, R <4.1.0
#> # A tibble: 1 × 6
#>   expression                                min   median itr/s…¹ mem_a…² gc/se…³
#>   <bch:expr>                           <bch:tm> <bch:tm>   <dbl> <bch:b>   <dbl>
#> 1 mutate(df, y = identity(x), .by = g)   25.2ms   28.9ms    35.5  1.56MB    976.
#> # … with abbreviated variable names ¹​`itr/sec`, ²​mem_alloc, ³​`gc/sec`

# This PR, CRAN rlang
#> disabled.
#> # A tibble: 1 × 6
#>   expression                                min   median itr/s…¹ mem_a…² gc/se…³
#>   <bch:expr>                           <bch:tm> <bch:tm>   <dbl> <bch:b>   <dbl>
#> 1 mutate(df, y = identity(x), .by = g)   47.9ms   56.1ms    17.5  12.3MB    26.7
#> # … with abbreviated variable names ¹​`itr/sec`, ²​mem_alloc, ³​`gc/sec`
```

```r
# 100,000 groups
df <- tibble(g = seq_len(100000), x = seq_along(g))
bench::mark(mutate(df, y = identity(x), .by = g), iterations = 10)

# CRAN dplyr, CRAN rlang
#> disabled.
#> # A tibble: 1 × 6
#>   expression                                min   median itr/s…¹ mem_a…² gc/se…³
#>   <bch:expr>                           <bch:tm> <bch:tm>   <dbl> <bch:b>   <dbl>
#> 1 mutate(df, y = identity(x), .by = g)    204ms    217ms    4.26  15.1MB    11.9
#> # … with abbreviated variable names ¹​`itr/sec`, ²​mem_alloc, ³​`gc/sec`

# This PR, dev rlang, R >=4.1.0
#> disabled.
#> # A tibble: 1 × 6
#>   expression                                min   median itr/s…¹ mem_a…² gc/se…³
#>   <bch:expr>                           <bch:tm> <bch:tm>   <dbl> <bch:b>   <dbl>
#> 1 mutate(df, y = identity(x), .by = g)    234ms    263ms    3.67  15.1MB    17.1
#> # … with abbreviated variable names ¹​`itr/sec`, ²​mem_alloc, ³​`gc/sec`

# This PR, dev rlang, R <4.1.0
#> disabled.
#> # A tibble: 1 × 6
#>   expression                                min   median itr/s…¹ mem_a…² gc/se…³
#>   <bch:expr>                           <bch:tm> <bch:tm>   <dbl> <bch:b>   <dbl>
#> 1 mutate(df, y = identity(x), .by = g)    256ms    311ms    3.11  15.1MB    19.7
#> # … with abbreviated variable names ¹​`itr/sec`, ²​mem_alloc, ³​`gc/sec`

# This PR, CRAN rlang
#> disabled.
#> # A tibble: 1 × 6
#>   expression                                min   median itr/s…¹ mem_a…² gc/se…³
#>   <bch:expr>                           <bch:tm> <bch:tm>   <dbl> <bch:b>   <dbl>
#> 1 mutate(df, y = identity(x), .by = g)    476ms    540ms    1.78   123MB    10.7
#> # … with abbreviated variable names ¹​`itr/sec`, ²​mem_alloc, ³​`gc/sec`
```

Here is an example of "work" being done by the `summarise()` call by computing a `mean()`. Can't really notice the data mask creation overhead here, even on <4.1.0.

```r
library(dplyr)

set.seed(123)

# 100k possible groups
dictionary <- seq_len(100000)
g <- sample(dictionary, 10000000, replace = TRUE)
x <- sample(10, length(g), replace = TRUE)

# 10,000,000 rows
df <- tibble(g = g, x = x)

# 100k groups
vctrs::vec_unique_count(df$g)
#> [1] 100000

bench::mark(summarise(df, x = mean(x), .by = g), iterations = 50)

# CRAN dplyr, CRAN rlang
#> disabled.
#> # A tibble: 1 × 6
#>   expression                               min   median itr/se…¹ mem_a…² gc/se…³
#>   <bch:expr>                          <bch:tm> <bch:tm>    <dbl> <bch:b>   <dbl>
#> 1 summarise(df, x = mean(x), .by = g)    1.26s    1.62s    0.610   306MB    2.89
#> # … with abbreviated variable names ¹​`itr/sec`, ²​mem_alloc, ³​`gc/sec`

# This PR, dev rlang, R >=4.1.0
#> disabled.
#> # A tibble: 1 × 6
#>   expression                               min   median itr/se…¹ mem_a…² gc/se…³
#>   <bch:expr>                          <bch:tm> <bch:tm>    <dbl> <bch:b>   <dbl>
#> 1 summarise(df, x = mean(x), .by = g)    1.32s    1.48s    0.642   306MB    3.76
#> # … with abbreviated variable names ¹​`itr/sec`, ²​mem_alloc, ³​`gc/sec`

# This PR, dev rlang, R <4.1.0
#> disabled.
#> # A tibble: 1 × 6
#>   expression                               min   median itr/se…¹ mem_a…² gc/se…³
#>   <bch:expr>                          <bch:tm> <bch:tm>    <dbl> <bch:b>   <dbl>
#> 1 summarise(df, x = mean(x), .by = g)    1.22s    1.42s    0.665   306MB    4.21
#> # … with abbreviated variable names ¹​`itr/sec`, ²​mem_alloc, ³​`gc/sec`
```

### DIFF
--- a/R/data-mask.R
+++ b/R/data-mask.R
@@ -24,7 +24,7 @@ DataMask <- R6Class("DataMask",
       private$rowwise <- by$type == "rowwise"
 
       private$chops <- .Call(dplyr_lazy_vec_chop_impl, data, rows, private$grouped, private$rowwise)
-      private$mask <- .Call(dplyr_data_masks_setup, private$chops, data, rows)
+      private$env_bindings <- .Call(dplyr_make_column_bindings, private$chops, data)
 
       private$keys <- group_keys0(by$data)
       private$by_names <- by$names
@@ -41,11 +41,11 @@ DataMask <- R6Class("DataMask",
         }
       }
 
-      .Call(`dplyr_mask_add`, private, name, result, chunks)
+      .Call(`dplyr_binding_add`, private, name, result, chunks)
     },
 
     remove = function(name) {
-      .Call(`dplyr_mask_remove`, private, name)
+      .Call(`dplyr_binding_remove`, private, name)
     },
 
     resolve = function(name) {
@@ -93,7 +93,7 @@ DataMask <- R6Class("DataMask",
     },
 
     current_cols = function(vars) {
-      env_get_list(parent.env(private$mask), vars)
+      env_get_list(private$env_bindings, vars)
     },
 
     current_rows = function() {
@@ -190,11 +190,16 @@ DataMask <- R6Class("DataMask",
     },
 
     get_env_bindings = function() {
-      parent.env(private$mask)
+      private$env_bindings
     },
 
     get_rlang_mask = function() {
-      private$mask
+      # Mimicking the data mask that is created during typical
+      # expression evaluations, like in `DataMask$eval_all_mutate()`.
+      # Important to insert a `.data` pronoun!
+      mask <- new_data_mask(private$env_bindings)
+      mask[[".data"]] <- as_data_pronoun(private$env_bindings)
+      mask
     }
 
   ),
@@ -208,10 +213,12 @@ DataMask <- R6Class("DataMask",
     # - .current_group: scalar integer that identifies the current group
     chops = NULL,
 
-    # dynamic data mask, with active bindings for each column
-    # this is an rlang data mask, as such the bindings are actually
-    # in the parent environment of `mask`
-    mask = NULL,
+    # Environment with active bindings for each column.
+    # Expressions are evaluated in a fresh data mask created from this
+    # environment. Each group gets its own newly created data mask to avoid
+    # cross group contamination of the data mask by lexical side effects, like
+    # usage of `<-` (#6666).
+    env_bindings = NULL,
 
     # ptypes of all the variables
     current_data = list(),
--- a/R/data-mask.R
+++ b/R/data-mask.R
@@ -24,7 +24,7 @@ DataMask <- R6Class("DataMask",
       private$rowwise <- by$type == "rowwise"
 
       private$chops <- .Call(dplyr_lazy_vec_chop_impl, data, rows, private$grouped, private$rowwise)
-      private$env_bindings <- .Call(dplyr_make_column_bindings, private$chops, data)
+      private$env_mask_bindings <- .Call(dplyr_make_mask_bindings, private$chops, data)
 
       private$keys <- group_keys0(by$data)
       private$by_names <- by$names
@@ -41,11 +41,11 @@ DataMask <- R6Class("DataMask",
         }
       }
 
-      .Call(`dplyr_binding_add`, private, name, result, chunks)
+      .Call(`dplyr_mask_binding_add`, private, name, result, chunks)
     },
 
     remove = function(name) {
-      .Call(`dplyr_binding_remove`, private, name)
+      .Call(`dplyr_mask_binding_remove`, private, name)
     },
 
     resolve = function(name) {
@@ -93,7 +93,7 @@ DataMask <- R6Class("DataMask",
     },
 
     current_cols = function(vars) {
-      env_get_list(private$env_bindings, vars)
+      env_get_list(private$env_mask_bindings, vars)
     },
 
     current_rows = function() {
@@ -166,10 +166,10 @@ DataMask <- R6Class("DataMask",
       }
 
       promises <- map(names_bindings, function(.x) expr(osbolete_promise_fn(!!.x)))
-      bindings <- self$get_env_bindings()
+      env_mask_bindings <- private$env_mask_bindings
       suppressWarnings({
-        rm(list = names_bindings, envir = bindings)
-        env_bind_lazy(bindings, !!!set_names(promises, names_bindings))
+        rm(list = names_bindings, envir = env_mask_bindings)
+        env_bind_lazy(env_mask_bindings, !!!set_names(promises, names_bindings))
       })
     },
 
@@ -189,16 +189,12 @@ DataMask <- R6Class("DataMask",
       private$size
     },
 
-    get_env_bindings = function() {
-      private$env_bindings
-    },
-
     get_rlang_mask = function() {
       # Mimicking the data mask that is created during typical
       # expression evaluations, like in `DataMask$eval_all_mutate()`.
       # Important to insert a `.data` pronoun!
-      mask <- new_data_mask(private$env_bindings)
-      mask[[".data"]] <- as_data_pronoun(private$env_bindings)
+      mask <- new_data_mask(private$env_mask_bindings)
+      mask[[".data"]] <- as_data_pronoun(private$env_mask_bindings)
       mask
     }
 
@@ -218,7 +214,7 @@ DataMask <- R6Class("DataMask",
     # environment. Each group gets its own newly created data mask to avoid
     # cross group contamination of the data mask by lexical side effects, like
     # usage of `<-` (#6666).
-    env_bindings = NULL,
+    env_mask_bindings = NULL,
 
     # ptypes of all the variables
     current_data = list(),
--- a/src/chop.cpp
+++ b/src/chop.cpp
@@ -89,7 +89,7 @@ SEXP dplyr_lazy_vec_chop(SEXP data, SEXP rows, SEXP ffi_grouped, SEXP ffi_rowwis
   return chops_env;
 }
 
-void add_mask_binding(SEXP name, SEXP env_bindings, SEXP env_chops) {
+void add_column_binding(SEXP name, SEXP env_bindings, SEXP env_chops) {
   SEXP body = PROTECT(Rf_lang3(dplyr::functions::dot_subset2, name, dplyr::symbols::dot_current_group));
   SEXP fun  = PROTECT(Rf_lang3(dplyr::functions::function, R_NilValue, body));
   SEXP binding = PROTECT(Rf_eval(fun, env_chops));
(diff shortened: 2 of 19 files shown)


R/data-mask.R
```diff
@@ -41,11 +41,11 @@ DataMask <- R6Class("DataMask",
         }
       }
 
-      .Call(`dplyr_mask_add`, private, name, result, chunks)
+      .Call(`dplyr_binding_add`, private, name, result, chunks)
     },
 
     remove = function(name) {
-      .Call(`dplyr_mask_remove`, private, name)
+      .Call(`dplyr_binding_remove`, private, name)
```
DavisVaughan, line 48: These functions are specific to `env_bindings`, so I've tweaked their names

R/data-mask.R
```diff
@@ -93,7 +93,7 @@ DataMask <- R6Class("DataMask",
     },
 
     current_cols = function(vars) {
-      env_get_list(parent.env(private$mask), vars)
+      env_get_list(private$env_bindings, vars)
```
DavisVaughan, line 96: A number of places were simplified by the removal of `parent.env()` or `ENCLOS()` now that we create the data mask on the fly. This "feels" better to me.

R/data-mask.R
```diff
@@ -190,11 +190,16 @@ DataMask <- R6Class("DataMask",
...
 
     get_rlang_mask = function() {
-      private$mask
+      # Mimicking the data mask that is created during typical
+      # expression evaluations, like in `DataMask$eval_all_mutate()`.
+      # Important to insert a `.data` pronoun!
+      mask <- new_data_mask(private$env_bindings)
+      mask[[".data"]] <- as_data_pronoun(private$env_bindings)
+      mask
     }
```
DavisVaughan, line 203: This is only needed for `across()`, I've added an extra `across()` test to ensure that this is set up correctly (including a test for the `.data` pronoun being there)

src/chop.cpp
```diff
@@ -98,25 +98,25 @@ void add_mask_binding(SEXP name, SEXP env_bindings, SEXP env_chops) {
...
+
   for (R_xlen_t i = 0; i < n_columns; i++) {
     SEXP name = PROTECT(rlang::str_as_symbol(p_names[i]));
-    add_mask_binding(name, env_bindings, env_chops);
+    add_column_binding(name, env_bindings, env_chops);
     UNPROTECT(1);
   }
-  SEXP mask = PROTECT(rlang::new_data_mask(env_bindings, R_NilValue));
-  SEXP pronoun = PROTECT(rlang::as_data_pronoun(env_bindings));
-  Rf_defineVar(dplyr::symbols::dot_data, pronoun, mask);
```
DavisVaughan, line 116: No longer returning a data mask. We now return the environment containing the active bindings. This is what serves as `bottom` in the data masks we create at each evaluation

src/dplyr.h
```diff
@@ -109,33 +109,45 @@ SEXP dplyr_summarise_recycle_chunks_in_place(SEXP list_of_chunks, SEXP list_of_r
...
-*p_current_group = 0
+  UNPROTECT(6);                                                \
+  *p_current_group = 0
 
-#define DPLYR_MASK_SET_GROUP(INDEX) *p_current_group = INDEX + 1
+// At each iteration, we create a fresh data mask so that lexical side effects,
+// such as using `<-` in a `mutate()`, don't persist between groups
+#define DPLYR_MASK_ITERATION_INIT()                                    \
+  SEXP mask = PROTECT(rlang::new_data_mask(env_bindings, R_NilValue)); \
+  Rf_defineVar(dplyr::symbols::dot_data, pronoun, mask)
```
DavisVaughan, line 141: This is the meat of the change. `DPLYR_MASK_INIT()` no longer gets the `mask`. Instead, it is created fresh on each iteration by this macro

tests/testthat/_snaps/join-rows.md
```diff
@@ -252,7 +252,7 @@
       join_rows(df, df, unmatched = 1)
     Condition
       Error:
-      ! `unmatched` must be a character vector, not a number.
+      ! `unmatched` must be a character vector, not the number 1.
```
DavisVaughan, line 255: Unrelated change due to changes in dev rlang with `arg_match()`

tests/testthat/test-across.R
```diff
@@ -748,6 +748,30 @@ test_that("list of lambdas work", {
   expect_equal(df %>% mutate((across(1:2, list(~ .x + mean(bar))))), exp)
 })
 
+test_that("anonymous function `.fns` can access the `.data` pronoun even when not inlined", {
```
DavisVaughan, line 751: This is the extra test I added to ensure `get_rlang_mask()` is still working right for where it is used by `across()` to evaluate `.fns`

tests/testthat/test-mutate.R
```diff
@@ -120,6 +120,85 @@ test_that("mutate() supports constants (#6056, #6305)", {
...
 })
 
+test_that("can't overwrite column active bindings (#6666)", {
+  skip_if(getRversion() < "3.6.3", message = "Active binding error changed")
+
+  df <- tibble(g = 1:2, x = 3:4)
+  gdf <- group_by(df, g)
+
+  # The error seen here comes from trying to `<-` to an active binding when
+  # the active binding function has 0 arguments.
```
DavisVaughan, line 130: We could change `add_column_binding()` so that the active binding function it creates takes 1 argument, which we'd then check with something like:

```
if (!missing(x)) {
  abort("Can't assign to dplyr active binding")
}
```

But it is so hard for a user to actually do this, I don't think it is worth the performance penalty (however minor). I think the important thing is that it errors.

src/chop.cpp
lionel-: Note: I think I'd have kept "mask" in the name, e.g. `mask_bindings`. A data mask comprises multiple envs, so this environment is still part of the mask.

tests/testthat/_snaps/across.md
```diff
@@ -283,6 +283,18 @@
...
+
+    Code
+      mutate(df, across(y, fn))
+    Condition
+      Error in `mutate()`:
+      i In argument: `across(y, fn)`.
+      Caused by error in `across()`:
+      ! Can't compute column `y`.
+      Caused by error:
+      ! Can't subset `.data` outside of a data mask context.
```
lionel-, line 296: Unrelated but should we try to simplify the error like this?

```
      Error in `mutate()`:
      i In argument: `across(y, fn)`.
      i In column: `y`.
      Caused by error:
      ! Can't subset `.data` outside of a data mask context.
```

lionel- (approved):
Looks like a good trade off! Happy to get faster environment creation in rlang.

tests/testthat/_snaps/across.md
DavisVaughan: https://github.com/tidyverse/dplyr/issues/6712

Issue #6674, opened 2023-02-01
v1.1.0 runtime for case_when with grouping variable is slow


fawda123:
Using `case_when` in a `mutate` call with a grouping variable is much, much slower in v1.1.0 compared to v1.0.10.  The code works but it's causing a tremendous slowdown in many of the packages I maintain (see [here](https://www.r-project.org/nosvn/R.check/r-devel-linux-x86_64-debian-clang/MassWateR-00check.html), many examples have elapsed time >5s).  

Here's a reprex for v1.1.0.
``` r
library(dplyr, warn.conflicts = F)
library(microbenchmark)

n <- 1000
dat <- data.frame(
    x = seq(1:n), 
    y = rnorm(n)
)

microbenchmark(
    dat %>% 
        group_by(x) %>% 
        mutate(
                 z = case_when(
                    y < 0 ~ '-',
                    T ~ '+', 
                 )
        ), 
    times = 100
)
#> Unit: seconds
#>                                                                        expr
#>  dat %>% group_by(x) %>% mutate(z = case_when(y < 0 ~ "-", T ~      "+", ))
#>       min       lq     mean   median       uq      max neval
#>  2.376748 2.537896 2.650869 2.625663 2.723655 3.170204   100
```

<sup>Created on 2023-02-01 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

<details style="margin-bottom:10px;">
<summary>
Session info
</summary>

``` r
sessioninfo::session_info()
#> - Session info  --------------------------------------------------------------
#>  hash: person in steamy room: medium-dark skin tone, goat, black small square
#> 
#>  setting  value
#>  version  R version 4.1.3 (2022-03-10)
#>  os       Windows 10 x64 (build 22000)
#>  system   x86_64, mingw32
#>  ui       RTerm
#>  language (EN)
#>  collate  English_United States.1252
#>  ctype    English_United States.1252
#>  tz       America/New_York
#>  date     2023-02-01
#>  pandoc   2.19.2 @ C:/Program Files/RStudio/resources/app/bin/quarto/bin/tools/ (via rmarkdown)
#> 
#> - Packages -------------------------------------------------------------------
#>  package        * version date (UTC) lib source
#>  cli              3.6.0   2023-01-09 [1] CRAN (R 4.1.3)
#>  digest           0.6.31  2022-12-11 [1] CRAN (R 4.1.3)
[...]

Issue #6678, opened 2023-02-02
`dplyr::if_else()` fails when `condition` is a labelled logical vector


salim-b:
Starting with dplyr 1.1.0, [`dplyr::if_else()`](https://dplyr.tidyverse.org/reference/if_else.html) fails with a meaningless error message when `condition` is a **labelled** logical vector:

``` r
d <-
    mtcars |>
    dplyr::mutate(mileage = mpg > 20) |>
    labelled::set_variable_labels(mileage = "mileage dummy")

dplyr::if_else(d$mileage,
               "guzzler",
               "moderate")
#> Error in `dplyr::if_else()`:
#> ! `condition` must be a vector with type <logical>.
#> Instead, it has type <logical>.

#> Backtrace:
#>     ▆
#>  1. └─dplyr::if_else(d$mileage, "guzzler", "moderate")
#>  2.   └─vctrs::vec_assert(x = condition, ptype = logical(), arg = "condition")
#>  3.     └─rlang::abort(...)
```

<sup>Created on 2023-02-02 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

Explicitly providing the correct prototype gives the same result:

``` r
dplyr::if_else(d$mileage,
               "guzzler",
               "moderate",
               ptype = vctrs::vec_ptype(d$mileage))
#> Error in `dplyr::if_else()`:
#> ! `condition` must be a vector with type <logical>.
#> Instead, it has type <logical>.

#> Backtrace:
#>     ▆
#>  1. └─dplyr::if_else(d$mileage, "guzzler", "moderate", ptype = vctrs::vec_ptype(d$mileage))
#>  2.   └─vctrs::vec_assert(x = condition, ptype = logical(), arg = "condition")
#>  3.     └─rlang::abort(...)
```

PR #6711, opened 2023-02-10
Improve performance of `case_when()` and `case_match()`

DavisVaughan:
Closes https://github.com/tidyverse/dplyr/issues/6674

Three improvements to make these decently fast again:

- Minor: Calls to `current_env()` and `caller_env()` are now lazy (only evaluated when needed) (needs https://github.com/r-lib/rlang/issues/1558)

- Major: We no longer call `as_label()` to turn the expressions into labels which were being used as informative names that showed up in error messages. This stinks, but `as_label()` is just too slow to be used eagerly like this. We can revisit this again in the future if we add a lazy ALTREP character vector which could materialize the `as_label()` calls on demand (i.e. at error time). In the meantime, we now use positional names that also include LHS/RHS information, like `..2 (RHS)`, which ends up being decent.

- Major: I was previously wrapping the evaluation of the lhs/rhs of the formulas in `with_case_errors()` to give an informative indexed error about which formula failed to evaluate. This got slow as you increased the number of expressions in `case_when()` (2 * N expressions = number of `with_case_errors()` calls) because it uses `try_fetch()`, which is built on `tryCatch()`, which is in R not C (probably a good Luke project!). I've changed it up so 1 call to `with_case_errors()` now wraps the entire evaluation loop, and we track some index information to use in the error if one occurs.

Sadly the private option added in rlang (https://github.com/r-lib/rlang/commit/33db700d556b0b85a1fe78e14a53f95ac9248004) wasn't enough to make `as_label()` fast enough to be used eagerly like this. But the `infix_overflows()` call there is a big part of the performance degradation we see in my benchmarks below.

When I work on https://github.com/tidyverse/dplyr/issues/6678 that will end up also knocking off another 50us or so from switching away from `vec_assert()` in `vec_case_when()` and `if_else()`.

After that, any other improvements will come from moving `vec_case_when()` to C in vctrs.

Right now we are in the 2-4x slower range. After https://github.com/tidyverse/dplyr/issues/6678 I expect to be 2x slower most of the time. I think we can do a release with that. Then the C implementation in vctrs should fix the rest.

```r
library(dplyr)

# Size 1
x <- 1L

bench::mark(
  one = case_when(x == 1L ~ TRUE), 
  two = case_when(x == 1L ~ TRUE, x == 2L ~ NA),
  three = case_when(x == 1L ~ TRUE, x == 2L ~ NA, TRUE ~ FALSE),
  iterations = 50000,
  check = FALSE
)

# dplyr 1.0.10
#> # A tibble: 3 × 6
#> 1 one          56.8µs   74.7µs    12802.     168KB     18.2
#> 2 two          75.5µs   95.4µs    10514.        0B     22.5
#> 3 three        97.4µs  120.2µs     8390.        0B     23.9

# This PR
#> # A tibble: 3 × 6
#> 1 one           209µs    278µs     3388.   503.8KB     11.3
#> 2 two           242µs    326µs     2879.     1.3KB     12.2
#> 3 three         272µs    350µs     2787.     1.3KB     14.3

# dplyr 1.1.0 (smacked by the infix-overflow `as_label()` performance issue)
# (note that i had to go down to only 5,000 iterations here)
#> # A tibble: 3 × 6
#> 1 one          1.67ms   2.12ms      477.    1.01MB     16.6
#> 2 two          3.14ms   3.86ms      261.    3.36KB     18.2
#> 3 three        3.27ms   4.32ms      228.    3.36KB     16.5
```

```r
library(dplyr)

# Size 1000
x <- 1:1000

bench::mark(
  one = case_when(x < 200 ~ TRUE), 
  two = case_when(x < 200 ~ TRUE, x < 500 ~ NA),
  three = case_when(x < 200 ~ TRUE, x < 500 ~ NA, TRUE ~ FALSE),
  iterations = 50000,
  check = FALSE
)

# dplyr 1.0.10
#> # A tibble: 3 × 6
#> 1 one          73.2µs   92.2µs    10873.    60.1KB     17.0
#> 2 two         108.6µs  136.9µs     7383.   100.9KB     17.3
#> 3 three       136.1µs  171.2µs     5914.   130.6KB     18.3

# This PR
#> # A tibble: 3 × 6
#> 1 one           222µs    271µs     3720.    44.1KB     14.4
#> 2 two           279µs    334µs     3011.    60.1KB     15.4
#> 3 three         318µs    382µs     2649.      70KB     16.5

# dplyr 1.1.0
#> # A tibble: 3 × 6
#> 1 one          2.03ms   2.39ms      400.      45KB     15.0
#> 2 two          3.98ms   4.45ms      222.    62.2KB     16.6
#> 3 three         4.1ms   4.64ms      211.      72KB     16.6
```

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,7 @@
 # dplyr (development version)
 
+* A major performance regression in `case_when()` has been fixed (#6674).
+
 * Fixed an issue where expressions involving infix operators had an abnormally
   large amount of overhead (#6681).
 
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,6 +1,8 @@
 # dplyr (development version)
 
-* A major performance regression in `case_when()` has been fixed (#6674).
+* A major performance regression in `case_when()` has been fixed. It is still a
+  little slower than in dplyr 1.0.10, but we plan to improve this further in the
+  future (#6674).
 
 * Fixed an issue where expressions involving infix operators had an abnormally
   large amount of overhead (#6681).
--- a/R/case-match.R
+++ b/R/case-match.R
@@ -128,15 +128,11 @@ case_match <- function(.x,
                        .ptype = NULL) {
   args <- list2(...)
 
-  default_env <- caller_env()
-  dots_env <- current_env()
-  error_call <- current_env()
-
   args <- case_formula_evaluate(
     args = args,
-    default_env = default_env,
-    dots_env = dots_env,
-    error_call = error_call
+    default_env = caller_env(),
+    dots_env = current_env(),
+    error_call = current_env()
   )
 
   haystacks <- args$lhs
@@ -152,6 +148,6 @@ case_match <- function(.x,
     default = .default,
     default_arg = ".default",
     ptype = .ptype,
-    call = error_call
+    call = current_env()
   )
 }
--- a/R/case-when.R
+++ b/R/case-when.R
@@ -213,9 +213,9 @@ case_formula_evaluate <- function(args,
 
   env_error_info <- new_environment()
 
-  # Using 1 call to `with_case_errors()` that wraps all `eval_tidy()`
-  # evaluations to avoid repeated calls to `try_fetch()` (#6674)
-  with_case_errors(error_call = error_call, env_error_info = env_error_info, {
+  # Using 1 call to `withCallingHandlers()` that wraps all `eval_tidy()`
+  # evaluations to avoid repeated handler setup (#6674)
+  withCallingHandlers(
     for (i in seq_args) {
       env_error_info[["i"]] <- i
       pair <- pairs[[i]]
@@ -232,8 +232,15 @@ case_formula_evaluate <- function(args,
       if (!is.null(elt_rhs)) {
         rhs[[i]] <- elt_rhs
       }
+    },
+    error = function(cnd) {
+      message <- glue::glue_data(
+        env_error_info,
+        "Failed to evaluate the {side}-hand side of formula {i}."
+      )
+      abort(message, parent = cnd, call = error_call)
     }
-  })
+  )
 
   # TODO: Ideally we'd name the lhs/rhs values with their `as_label()`-ed
   # expressions. But `as_label()` is much too slow for that to be useful in
@@ -292,17 +299,5 @@ validate_and_split_formula <- function(x,
   list(
     lhs = new_quosure(f_lhs(x), env),
     rhs = new_quosure(f_rhs(x), env)
-  )
-}
-
-with_case_errors <- function(expr, error_call, env_error_info) {
-  try_fetch(
-    expr,
-    error = function(cnd) {
-      i <- env_error_info$i
-      side <- env_error_info$side
-      message <- glue("Failed to evaluate the {side}-hand side of formula {i}.")
-      abort(message, parent = cnd, call = error_call)
-    }
   )
 }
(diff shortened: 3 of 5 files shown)


tests/testthat/_snaps/case-when.md
```diff
@@ -73,20 +73,20 @@
...
-      ! Can't recycle `c(TRUE, FALSE)` (size 2) to match `1:3` (size 3).
+      ! Can't recycle `..1 (LHS)` (size 2) to match `..1 (RHS)` (size 3).
     Code
       (expect_error(case_when(c(TRUE, FALSE) ~ 1, c(FALSE, TRUE, FALSE) ~ 2, c(FALSE,
         TRUE, FALSE, NA) ~ 3)))
     Output
       <error/vctrs_error_incompatible_size>
       Error in `case_when()`:
-      ! Can't recycle `c(TRUE, FALSE)` (size 2) to match `c(FALSE, TRUE, FALSE)` (size 3).
+      ! Can't recycle `..1 (LHS)` (size 2) to match `..2 (LHS)` (size 3).
```
DavisVaughan, line 83: Particularly important to show LHS vs RHS in these errors, otherwise it can be very confusing

NEWS.md
```diff
@@ -1,5 +1,7 @@
 # dplyr (development version)
 
+* A major performance regression in `case_when()` has been fixed (#6674).
```
hadley, line 3: I'd add: "It's still a little slower than in dplyr 1.0.10, but we plan to resolve this further in the future"

R/case-match.R
```diff
@@ -128,15 +128,11 @@ case_match <- function(.x,
...
-  default_env <- caller_env()
-  dots_env <- current_env()
-  error_call <- current_env()
-
   args <- case_formula_evaluate(
     args = args,
-    default_env = default_env,
-    dots_env = dots_env,
-    error_call = error_call
+    default_env = caller_env(),
```
hadley, line 133: It's interesting that this is expensive enough to care about.

R/case-when.R
```diff
@@ -283,10 +295,12 @@ validate_and_split_formula <- function(x,
   )
 }
 
-with_case_errors <- function(expr, side, i, error_call) {
+with_case_errors <- function(expr, error_call, env_error_info) {
```
hadley, line 298: Are you taking advantage of `try_fetch()` features here? Is it worth changing to a bare `withCallingHandlers()`?

R/case-when.R
```diff
@@ -196,52 +192,68 @@ case_formula_evaluate <- function(args,
...
+  # lhs_names <- map(quos_pairs, function(pair) pair$lhs)
+  # lhs_names <- map_chr(lhs_names, as_label)
+  # names(lhs) <- lhs_names
+  #
+  # rhs_names <- map(quos_pairs, function(pair) pair$rhs)
+  # rhs_names <- map_chr(rhs_names, as_label)
+  # names(rhs) <- rhs_names
+  if (n_args > 0L) {
+    names(lhs) <- paste0("..", seq_args, " (LHS)")
+    names(rhs) <- paste0("..", seq_args, " (RHS)")
```
hadley, line 255: Given that people don't seem to general understand LHS/RHS might be better to just use left/right.

hadley (approved):

lionel- (approved):

R/case-match.R
DavisVaughan: It is super minor honestly, but we've worked hard in vctrs to make env evaluation lazy and it seems a waste to throw that away

Issue #6685, opened 2023-02-06
Let `pick()` support `NULL`


psychelzh:
In the documentation:
> `pick()` is useful as a bridge between data-masking functions (like `mutate()` or `group_by()`) and functions with tidy-select behavior (like `select()`).

Therefore, it will be as expected if they both support `NULL` to fall to no grouping.

``` r
library(dplyr)
#> 
#> Attaching package: 'dplyr'
#> The following objects are masked from 'package:stats':
#> 
#>     filter, lag
#> The following objects are masked from 'package:base':
#> 
#>     intersect, setdiff, setequal, union
df <- data.frame(x = 1)
df |> 
  group_by(across(NULL))
#> # A tibble: 1 × 1
#>       x
#>   <dbl>
#> 1     1
df |> 
  group_by(pick(NULL))
#> Error in `group_by()`:
#> ℹ In argument: `pick(NULL)`.
#> Caused by error:
#> ! `pick(NULL)` must be size 1, not 0.

#> Backtrace:
#>      ▆
#>   1. ├─dplyr::group_by(df, pick(NULL))
#>   2. ├─dplyr:::group_by.data.frame(df, pick(NULL))
#>   3. │ └─dplyr::group_by_prepare(.data, ..., .add = .add, error_call = current_env())
#>   4. │   └─dplyr:::add_computed_columns(.data, new_groups, error_call = error_call)
#>   5. │     └─dplyr:::mutate_cols(...)
#>   6. │       ├─base::withCallingHandlers(...)
#>   7. │       └─dplyr:::mutate_col(dots[[i]], data, mask, new_columns)
#>   8. │         └─mask$eval_all_mutate(quo)
#>   9. │           └─dplyr (local) eval()
#>  10. ├─dplyr:::dplyr_internal_error(...)
#>  11. │ └─rlang::abort(class = c(class, "dplyr:::internal_error"), dplyr_error_data = data)
#>  12. │   └─rlang:::signal_abort(cnd, .file)
#>  13. │     └─base::signalCondition(cnd)
#>  14. └─dplyr (local) `<fn>`(`<dpl:::__>`)
#>  15.   └─rlang::abort(message, class = error_class, parent = parent, call = error_call)
```

<sup>Created on 2023-02-06 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

<details style="margin-bottom:10px;">
<summary>
Session info
</summary>

``` r
sessioninfo::session_info()
[...]

PR #6727, opened 2023-02-14
Make the current group size a variable updated by reference

DavisVaughan:
This PR:

- Provides access to the current group size through `$get_current_group_size()`
- Uses this helper in `n()`, which makes it a little faster

This also gives us an easy private way to access the current group size for https://github.com/tidyverse/dplyr/issues/6685#issuecomment-1430392850

It also gets rid of the need to do `parent.env(chops)` internally, which always felt a little clunky to me.

``` r
library(dplyr)

set.seed(123)

# ~100k groups, so that many evaluations of `n()`
x <- sample(1e5, 1e6, replace = TRUE)
df <- tibble(x = x)
df <- group_by(df, x)
df
#> # A tibble: 1,000,000 × 1
#> # Groups:   x [99,995]
#>        x
#>    <int>
#>  1 51663
#>  2 57870
#>  3  2986
#>  4 29925
#>  5 95246
#>  6 68293
#>  7 62555
#>  8 45404
#>  9 65161
#> 10 46435
#> # … with 999,990 more rows

bench::mark(summarise(df, n = n()), iterations = 100)

# Main
#> disabled.
#> # A tibble: 1 × 6
#>   expression                  min   median `itr/sec` mem_alloc `gc/sec`
#>   <bch:expr>             <bch:tm> <bch:tm>     <dbl> <bch:byt>    <dbl>
#> 1 summarise(df, n = n())    919ms    1.14s     0.873    7.15MB     13.8

# This PR
#> disabled.
#> # A tibble: 1 × 6
#>   expression                  min   median `itr/sec` mem_alloc `gc/sec`
#>   <bch:expr>             <bch:tm> <bch:tm>     <dbl> <bch:byt>    <dbl>
#> 1 summarise(df, n = n())    596ms    742ms      1.35    7.16MB     15.1
```

<sup>Created on 2023-02-15 with [reprex v2.0.2.9000](https://reprex.tidyverse.org)</sup>

### DIFF
--- a/R/context.R
+++ b/R/context.R
@@ -48,7 +48,7 @@ NULL
 #' @rdname context
 #' @export
 n <- function() {
-  duplicate(peek_mask()$get_current_group_size())
+  peek_mask()$get_current_group_size()
 }
 
 #' @rdname context
@@ -60,11 +60,7 @@ cur_group <- function() {
 #' @rdname context
 #' @export
 cur_group_id <- function() {
-  # `duplicate()` to copy the private `dplyr:::current_group_id` variable we
-  # modify by reference. If we don't get a copy, code like this won't give
-  # correct results:
-  # summarise(id = cur_group_id())
-  duplicate(peek_mask()$get_current_group_id())
+  peek_mask()$get_current_group_id()
 }
 
 #' @rdname context
--- a/R/context.R
+++ b/R/context.R
@@ -48,7 +48,7 @@ NULL
 #' @rdname context
 #' @export
 n <- function() {
-  length(peek_mask()$current_rows())
+  duplicate(peek_mask()$get_current_group_size())
 }
 
 #' @rdname context
@@ -60,10 +60,11 @@ cur_group <- function() {
 #' @rdname context
 #' @export
 cur_group_id <- function() {
-  # [] to get a copy because the current group is dealt with internally
-  # if we don't get a copy, code like this won't give correct result:
+  # `duplicate()` to copy the private `dplyr:::current_group_id` variable we
+  # modify by reference. If we don't get a copy, code like this won't give
+  # correct results:
   # summarise(id = cur_group_id())
-  peek_mask()$get_current_group()[]
+  duplicate(peek_mask()$get_current_group_id())
 }
 
 #' @rdname context
--- a/R/data-mask.R
+++ b/R/data-mask.R
@@ -23,7 +23,12 @@ DataMask <- R6Class("DataMask",
       private$grouped <- by$type == "grouped"
       private$rowwise <- by$type == "rowwise"
 
-      private$chops <- .Call(dplyr_lazy_vec_chop_impl, data, rows, private$grouped, private$rowwise)
+      private$env_current <- new_environment(data = list(
+        `dplyr:::current_group_id` = 0L,
+        `dplyr:::current_group_size` = 0L
+      ))
+
+      private$chops <- .Call(dplyr_lazy_vec_chop_impl, data, rows, private$env_current, private$grouped, private$rowwise)
       private$env_mask_bindings <- .Call(dplyr_make_mask_bindings, private$chops, data)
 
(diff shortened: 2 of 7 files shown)


R/data-mask.R
```diff
@@ -23,7 +23,12 @@ DataMask <- R6Class("DataMask",
       private$grouped <- by$type == "grouped"
       private$rowwise <- by$type == "rowwise"
 
-      private$chops <- .Call(dplyr_lazy_vec_chop_impl, data, rows, private$grouped, private$rowwise)
+      private$env_current <- new_environment(data = list(
+        `dplyr:::current_group_id` = 0L,
+        `dplyr:::current_group_size` = 0L
+      ))
+
+      private$chops <- .Call(dplyr_lazy_vec_chop_impl, data, rows, private$env_current, private$grouped, private$rowwise)
```
DavisVaughan, line 31: We previously created an environment sort of like `env_current` from inside `dplyr_lazy_vec_chop_impl` and set it as the parent env of `chops`, but that meant we would always access that end on the R side using `parent.env(private$chops)` and that was a little clunky.

Creating the env on the R side and passing it in seems a little cleaner, and means we have easy access to `env_current` and its "current" group information variables for use in `n()` and `cur_group_id()`.

R/data-mask.R
```diff
@@ -121,12 +126,29 @@ DataMask <- R6Class("DataMask",
...
+      # (like in `cur_group_id()`), then it must be duplicated on the way out.
+      private[["env_current"]][["dplyr:::current_group_id"]]
+    },
+
+    get_current_group_size = function() {
+      # `dplyr:::current_group_size` is modified by reference at the C level.
+      # If the result of `get_current_group_size()` is used in a persistent way
+      # (like in `n()`), then it must be duplicated on the way out.
+      private[["env_current"]][["dplyr:::current_group_size"]]
     },
```
DavisVaughan, line 141: Important to note that these return the _exact_ reference to the object we modify by reference at the C level.

This is why we `duplicate()` the result before returning from `n()` and `cur_group_id()`. If we just returned the variable without duplicating then it could get modified by reference accidentally

R/context.R
```diff
@@ -60,10 +60,11 @@ cur_group <- function() {
...
 #' @export
 cur_group_id <- function() {
-  # [] to get a copy because the current group is dealt with internally
-  # if we don't get a copy, code like this won't give correct result:
+  # `duplicate()` to copy the private `dplyr:::current_group_id` variable we
+  # modify by reference. If we don't get a copy, code like this won't give
+  # correct results:
   # summarise(id = cur_group_id())
-  peek_mask()$get_current_group()[]
+  duplicate(peek_mask()$get_current_group_id())
```
lionel-, line 67: I think it'd be a bit safer to have two variants for group-id and group-size:

- `get_current_group_id()` that returns a duplicated value
- `get_current_group_id_unsafe()` that returns a shared value

R/data-mask.R
lionel-: Nice improvement. I agree this makes the code clearer.

R/data-mask.R
```diff
@@ -121,12 +126,29 @@ DataMask <- R6Class("DataMask",
...
 
     set_current_group = function(group) {
-      parent.env(private$chops)$.current_group[] <- group
+      # Only to be used right before throwing an error.
+      # We `duplicate()` group to be extremely conservative, because there is an
+      # extremely small chance we could modify this by reference and cause
+      # issues with the `group` variable in the caller, but this has never been
+      # seen. `length()` always returns a fresh variable so we don't duplicate
```
lionel-, line 148: Ideally we'd check for references in the return value.

R/data-mask.R
lionel-: Can we find a more informative name? `current_group_info`?

lionel- (approved):

PR #6743, opened 2023-02-19
Generate better code in get_slice_size()

krlmlr:
Only refer to our `clamp()` and base functions `floor()`, `ceiling()` and arithmetics.


DavisVaughan (commented):
Can you help me understand exactly what is necessary here?

I can get behind branching on `allow_outsize` early and inlining `slice_input$n` related computations when we can from a performance perspective

But is the `clamp_env` bit strictly necessary for you?

Issue #6731, opened 2023-02-16
On `multiple` matches, and how we can improve


DavisVaughan:
After reading https://github.com/tidyverse/dplyr/issues/6717, I've gone back to the drawing board to see if we are missing something with how `multiple` works.

I am now convinced that we have gotten `multiple` slightly wrong, making it more painful than it should be. I think the intentions are right, the execution of what it warns on is just slightly off.

There are 4 types of joins worth considering in this discussion:
- `1-to-1`
- `many-to-1`
- `1-to-many`
- `many-to-many`

With equi-joins, `multiple` will warn on `1-to-many` and `many-to-many`, i.e. anything with `*-to-many` where something in `x` matches multiple rows in `y`.

My big proposal here is that:
- Warning on `many-to-many` is correct.
- Warning on `1-to-many` is _wrong_.

---

I think that `many-to-1` and `1-to-many` are both very useful and safe joins. But more importantly, these are essentially _symmetrical_ styles of joins, so it doesn't make much sense to warn on one of these and not the other. Most of the time, you actually see these listed online as _just_ `1-to-many`, with the opposite being inferred, because they are so similar. It's just a matter of how you frame your question.

Our original thought process was:

> It is confusing for users when a join results in more rows than there are in `x`.

This is what can happen in `1-to-many` and `many-to-many`, which explains our current behavior, but I think a more appropriate assertion is:

> It is confusing for users when a join can result in a (possibly cartesian) explosion of `nrow(x) * nrow(y)` rows.

This can only happen with `many-to-many`, and is worth guarding against because it typically means there is a misspecified join. This assertion is supported by three trustworthy references online:

- [IBM's database relationship guide](https://www.ibm.com/docs/en/cdfsp/7.6.0?topic=structure-database-relationships) which says this about `many-to-many`, suggesting they aren't a good idea:
  > Each record in both tables can relate to none or any number of records in the other table. These relationships require a third table, called an associate or linking table, because relational systems cannot directly accommodate the relationship.

- [Stata's merge guide](https://www.stata.com/manuals13/dmerge.pdf) which does `m:m` merges slightly differently than we do, but the principle still stands, I think:
  > Because m:m merges are such a bad idea, we are not going to show you an example. If you think
that you need an m:m merge, then you probably need to work with your data so that you can use a
1:m or m:1 merge.

[...]

Issue #6717, opened 2023-02-10
Is it possible to set an option to globally turn off the multiple matches warning on joins?


eipi10:
`dplyr` 1.1 has a new argument `multiple` in the join functions that, by default, throws a warning if a row in `x` matches more than one row in `y`. You have to explicitly set `multiple="all"` in every join in order to silence the warning. According to the help, this default was chosen because multiple matches are "usually surprising." In the data I work with (education) multiple matches are the norm (students attend in multiple terms and take multiple courses in each term; instructors have multiple students, teach more than one course and perhaps multiple sections of a course, etc.). The constant warnings are really annoying, but it would likewise be annoying to have to explicitly set the `multiple` argument every time I do a join (including many, many joins across existing functions in our internal packages that would need to be updated). 

Ideally, the default would be not to warn. If you're not open to that change, is there any way to set an option at the top of a script that disables warnings related to multiple matches, and, if not, would you be willing to add such an option?

DavisVaughan:
I see that you gave some text based description of your use cases, but could you please provide a simple reprex that shows a few of your use cases?

eipi10:
Below is an example of a typical use case: I have data frames with (1) student data (one row per student), (2) term data (one row per student per term), and (3) course enrollment data (one row per student per term per course enrollment), and I want join the student data to the term data and then join the student+term data to the course enrollment data. 

I can silence the warning if I put the course enrollment data  in the `x` position. But why add that extra cognitive load when I already have a workflow that works the way I want it to? I also have built-in checks to warn me if something weird (like a student ID appearing twice in the same term in the term data, or a student having two enrollments in the same course in the same term) happens. 

The example below is just one among many use cases where I expect multiple matches. Over the years I've worked with data in several fields and I'm struggling to find an instance where I *didn't* typically expect multiple matches.

``` r
library(tidyverse)

student = structure(list(
  student.id = structure(c(3L, 1L, 4L, 2L), levels = c("S1", "S2", "S3", "S4"), class = "factor"),
  cohort = c("First-time", "Transfer", "First-time", "Transfer"),
  entering.term = c("Fall 2019", "Fall 2020", "Fall 2020", "Spring 2021")),
  row.names = c(NA, -4L), class = c("tbl_df", "tbl", "data.frame"))

term = structure(list(
  student.id = structure(c(2L, 2L, 2L, 2L, 3L, 3L, 3L, 3L, 3L, 1L, 1L, 1L,
[...]

PR #6753, opened 2023-02-22
Implement `relationship` and refine join match warning

DavisVaughan:
Closes https://github.com/tidyverse/dplyr/issues/6731
Closes https://github.com/tidyverse/dplyr/issues/6717
Joint PR on the vctrs side https://github.com/r-lib/vctrs/pull/1791, which contains even more detail

This PR hopefully fixes most of the complaints we've seen about the multiple match warning. At a high level:
- We previously warned on one-to-many and many-to-many relationships
- We now only warn on many-to-many relationships

As outlined in https://github.com/tidyverse/dplyr/issues/6731, only warning on many-to-many makes much more sense, as one-to-many is generally pretty useful, and is symmetric with many-to-one which we don't warn on, and that didn't make much sense. As further proof that we _should_ warn on many-to-many, most SQL dialects won't even let you create a many-to-many "relationship" between two tables; instead you have to create a 3rd junction table to break it into two one-to-many relationships.

This PR accomplishes this in two steps by:

- Changing the `multiple` default to `"all"` (same as SQL), and deprecating the `NULL`, `"error"`, and `"warning"` options
- Introducing a new `relationship` argument which replaces the `multiple = "error"` case and holistically handles multiple matches between `x` and `y`

The `relationship` argument takes on various options:

- `NULL` (default, chooses `"none"` or `"warn_many_to_many"` automatically)
- `"none"` (default for inequality, rolling, overlap joins)
- `"one_to_one"`
- `"one_to_many"`
- `"many_to_one"`
- `"many_to_many"`
- `"warn_many_to_many"` (default for equality joins)

These are inspired by [database table relationships](https://www.metabase.com/learn/databases/table-relationships), which end up being orthogonal to the idea of the "kind" of join you are doing (i.e. left/right/full/inner), so we can add an argument for this without any ambiguity or conflicts.

The choice of `relationship` activates a constraint on `x` and `y`, for example, `"many_to_one"` says that:
- Each row of `x` can match _at most 1_ row in `y`
- Each row of `y` can match _any number_ of rows in `x`

Which makes `relationship = "many_to_one"` a nice replacement for `multiple = "error"`. But we also now have `"one_to_one"` and `"one_to_many"` too!

The `"warn_many_to_many"` doesn't assume any expected relationship, but instead will check to see if a many-to-many relationship is present and will warn if it is, encouraging you to either look at your data or explicitly set `"many_to_many"` if that is expected.

Note that _at most 1_ means this doesn't handle the 0 match case, which is instead handled jointly by `unmatched` and your choice of join. Using _at most 1_ aligns exactly with the typical SQL definition for these options.

### DIFF
--- a/R/join.R
+++ b/R/join.R
@@ -41,7 +41,7 @@
 #' rows returned from the join.
 #'
 #' If a many-to-many relationship is expected, silence this warning by
-#' explicitly setting `relationship = "many_to_many"`.
+#' explicitly setting `relationship = "many-to-many"`.
 #'
 #' In production code, it is best to preemptively set `relationship` to whatever
 #' relationship you expect to exist between the keys of `x` and `y`, as this
@@ -55,7 +55,7 @@
 #'
 #' Rolling joins don't warn on many-to-many relationships either, but many
 #' rolling joins follow a many-to-one relationship, so it is often useful to
-#' set `relationship = "many_to_one"` to enforce this.
+#' set `relationship = "many-to-one"` to enforce this.
 #'
 #' Note that in SQL, most database providers won't let you specify a
 #' many-to-many relationship between two tables, instead requiring that you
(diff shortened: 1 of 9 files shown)


R/join.R
```diff
@@ -147,6 +166,40 @@
...
+#'
+#'   - `NULL`, the default, chooses:
+#'     - `"warn_many_to_many"` for equality joins
+#'     - `"none"` for inequality joins, rolling joins, and overlap joins
+#'
+#'     See the _Many-to-many relationships_ section for more details.
+#'
+#'   - `"none"` doesn't perform any relationship checks.
+#'
+#'   - `"one_to_one"` expects:
```
hadley, line 181: Should these be `one-to-one`, `many-to-many` etc? That's how we refer to them in text, and they're a little easier to type since you don't need to put your finger on the shift key for a single press.

hadley (approved):
Thanks for the detailed analysis on this!

DavisVaughan:
After further offline discussion, we've removed `"none"` and `"warn-many-to-many"` from the user facing API, since they are both handled internally by `NULL`, so now we have:

- `NULL` (default, chooses `"none"` or `"warn_many_to_many"` automatically)
- `"one_to_one"`
- `"one_to_many"`
- `"many_to_one"`
- `"many_to_many"`

Which is much cleaner

Issue #6762, opened 2023-02-24
Something wrong with nested `mutate()` calls


DavisVaughan:
``` r
library(dplyr)

df <- tibble(x = 1:2)

fn <- function() {
  df <- tibble(x = 1)
  mutate(df, y = x + 1)
}

mutate(df, z = {fn(); x})
#> Error in `mutate()`:
#> ℹ In argument: `z = { ... }`.
#> Caused by error:
#> ! attempt to select less than one element in integerOneIndex

#> Backtrace:
#>      ▆
#>   1. ├─dplyr::mutate(...)
#>   2. ├─dplyr:::mutate.data.frame(...) at dplyr/R/mutate.R:146:2
#>   3. │ └─dplyr:::mutate_cols(.data, dplyr_quosures(...), by) at dplyr/R/mutate.R:181:2
#>   4. │   ├─base::withCallingHandlers(...) at dplyr/R/mutate.R:268:2
#>   5. │   └─dplyr:::mutate_col(dots[[i]], data, mask, new_columns) at dplyr/R/mutate.R:273:6
#>   6. │     └─mask$eval_all_mutate(quo) at dplyr/R/mutate.R:380:8
#>   7. │       └─dplyr (local) eval() at dplyr/R/data-mask.R:90:6
#>   8. ├─`<fn>`()
#>   9. └─base::.handleSimpleError(...)
#>  10.   └─dplyr (local) h(simpleError(msg, call))
#>  11.     └─rlang::abort(message, class = error_class, parent = parent, call = error_call) at dplyr/R/conditions.R:235:4

fn <- function() {
  df <- tibble(x = 1)
  # doesn't cause the error? something to do with evaluation path
  mutate(df, y = x)
}

mutate(df, z = {fn(); x})
#> # A tibble: 2 × 2
#>       x     z
#>   <int> <int>
#> 1     1     1
#> 2     2     2
```

<sup>Created on 2023-02-24 with [reprex v2.0.2.9000](https://reprex.tidyverse.org)</sup>

PR #6763, opened 2023-02-25
Always duplicate variables we plan to modify by reference

DavisVaughan:
Closes https://github.com/tidyverse/dplyr/issues/6762

@lionel- this was a new one for me. The issue in #6762 was caused by a nested `mutate()` getting a reference to the exact same `dplyr:::current_group_id` as the outer `mutate()`. It seems that the `0L` value we initialize the id and size with isn't created "fresh" in each DataMask, so that is how they can get shared. Super tricky, took a long time to debug 😅 

Does this make sense to you? (No rush since you are OOO, just letting you know really)

### DIFF
--- a/R/data-mask.R
+++ b/R/data-mask.R
@@ -27,9 +27,13 @@ DataMask <- R6Class("DataMask",
       private$grouped <- by$type == "grouped"
       private$rowwise <- by$type == "rowwise"
 
+      # `duplicate(0L)` is necessary to ensure that the value we modify by
+      # reference is "fresh" and completely owned by this instance of the
+      # `DataMask`. Otherwise nested `mutate()` calls can end up modifying
+      # the same value (#6762).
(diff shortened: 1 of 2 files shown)


DavisVaughan:
Oh yea this is so strange. It looks like the inner `mutate()` was creating a new list SEXP for `group_info` (it has a different object address), but the `dplyr:::current_group_id` integer SEXP is a reference to the exact same one that was used in the outer SEXP.

So when you run line 31 once it gets to the inner `mutate()` and then print out `group_info`, it gives you the super strange result of `id = 1, size = 2` even though it looks like you just initialized those to `0`.


https://user-images.githubusercontent.com/19150088/221578529-846f620f-54bf-46a7-bcb8-932d94b60278.mov

lionel-:
> Does this make sense to you?

yup it looks like we were modifying the object created by the parser/deserialiser which end up being shared among closure instances.

PR #6771, opened 2023-03-02
Update variable name in name injection example

Bisaloo:
I think that the current example doesn't help understand fully where the injection happens by using the same variable `x` as the function argument symbol and as the value passed as this argument.

This edit hopefully makes it clearer. I tried to choose a neutral variable name but I'm happy to edit for any other name.


DavisVaughan:
Thanks!

PR #6782, opened 2023-03-08
Fix typo

fkohrt:
Not sure whether the other reference was meant to be R4DS, but it would make sense. Also there's no `vignette("one-table")` (well, it's this `vignette("base")`), so removing that.

Maybe @sastoudt can comment?


sastoudt:
Good catch! I think I was originally referring to [this](https://dplyr.tidyverse.org/articles/dplyr.html), but the reference to R4DS works too.

PR #6786, opened 2023-03-12
Require data frame compatibility for `setequal()`

### LINKED ISSUE #6785: Consider making `setequal()` error if data frames aren't compatible
It currently returns `FALSE`, but all of the other setops error on incompatible data frames

Sounds like it would help duckdb out if we could do this, but will likely cause a breakage or two so we should do it in a minor release. CC @krlmlr

krlmlr:
Running revdepchecks now.

Closes #6785.

### DIFF
--- a/tests/testthat/_snaps/sets.md
+++ b/tests/testthat/_snaps/sets.md
@@ -96,6 +96,25 @@
     Output
       Incompatible types for column `x`: double vs character.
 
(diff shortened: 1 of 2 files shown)


krlmlr:
I see only two failures, exuber might be a known (?) failure with dplyr 1.1.1, and modelplotr is a false positive. Would you consider including it in the 1.1.1 release?

PR #6797, opened 2023-03-20
Delay `dplyr_quosures()` auto naming until needed

DavisVaughan:
This mostly affects `filter()`, because expressions are never named there but we have been auto naming them anyways.

```r
library(dplyr)

df <- tibble(x = 1, y = 2:3)

bench::mark(filter(df, x == 1, y == 2), iterations = 10000)

# Before
#> # A tibble: 1 × 6
#>   expression                      min   median `itr/sec` mem_alloc `gc/sec`
#>   <bch:expr>                 <bch:tm> <bch:tm>     <dbl> <bch:byt>    <dbl>
#> 1 filter(df, x == 1, y == 2)    987µs   1.14ms      847.    1.35MB     13.7

# After
#> # A tibble: 1 × 6
#>   expression                      min   median `itr/sec` mem_alloc `gc/sec`
#>   <bch:expr>                 <bch:tm> <bch:tm>     <dbl> <bch:byt>    <dbl>
#> 1 filter(df, x == 1, y == 2)    653µs    835µs     1218.    1.75MB     12.9
```

There are some benefits for `mutate()` and `summarise()`. If an `across()` expansion occurs, then we previously had auto-named the expansion for nothing. This is less important though.

### DIFF
--- a/R/across.R
+++ b/R/across.R
@@ -545,25 +545,35 @@ new_dplyr_quosure <- function(quo, ...) {
   quo
 }
 
+dplyr_quosure_name <- function(quo_data) {
+  if (quo_data$is_named) {
+    # `name` is a user-supplied or known character string
+    quo_data$name
+  } else {
+    # `name` is a quosure that must be auto-named
+    with_no_rlang_infix_labeling(as_label(quo_data$name))
+  }
+}
+
 dplyr_quosures <- function(...) {
   # We're using quos() instead of enquos() here for speed, because we're not defusing named arguments --
   # only the ellipsis is converted to quosures, there are no further arguments.
   quosures <- quos(..., .ignore_empty = "all")
-  names_given <- names2(quosures)
+  names <- names2(quosures)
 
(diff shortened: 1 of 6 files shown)


R/across.R
```diff
@@ -545,25 +545,35 @@ new_dplyr_quosure <- function(quo, ...) {
...
-    if (is_named) {
-      name_auto <- name_given
-    } else {
-      name_auto <- with_no_rlang_infix_labeling(as_label(quosure))
+    name <- names[[i]]
+    is_named <- (name != "")
+
+    if (!is_named) {
+      # Will be auto-named by `dplyr_quosure_name()` only as needed
+      name <- quosure
```
DavisVaughan, line 571: The fact that we capture and store the _original_ quosure is somewhat important here, as there are cases with `pick()` where we expand the quosure into something else, so we need a reference to the original one to be able to auto-name it correctly. I've added a test for this.

lionel- (approved):

Issue #6774, opened 2023-03-04
dplyr 1.1.0 ungroup() drops data frame attributes


mkoohafkan:
This is referring to **dataframe/tibble** attributes, not tibble *column* attributes. The CRAN install of dplyr 1.1.0 sometimes drops attributes:

```r
d = mtcars
attr(d, "foo") = "bar"
attr(d, "foo")
#> [1] "bar"

## preserved
d |> select(cyl) |> attr("foo")
#> [1] "bar"
d |> rename(cyl2 = cyl) |> attr("foo")
#> [1] "bar"
d |> arrange(cyl) |> attr("foo")
#> [1] "bar"
d |> group_by(cyl) |> attr("foo")
#> [1] "bar"
d |> mutate(mpg2 = mpg) |> attr("foo")
#> [1] "bar"
d |> transmute(mpg2 = mpg) |> attr("foo")
#> [1] "bar"
d |> filter(cyl == 1L) |> attr("foo")
#> [1] "bar"
d |> count(cyl) |> attr("foo") 
#> [1] "bar"
d |> bind_rows(d) |> attr("foo")
#> [1] "bar"
d |> bind_cols(select(d, cyl2 = cyl)) |> attr("foo")
#> [1] "bar"


## dropped
d |> group_by(cyl) |> ungroup() |> attr("foo")
#> NULL
d |> summarize(mpg = max(mpg)) |> attr("foo")
#> NULL
```

I can understand why it might make sense for `summarize()` to drop attributes (but then why does `count()` preserve them?), but I don't see why `ungroup(group_by(...))` should drop attributes unilaterally. Given that almost all `dplyr` verbs preserve the attribute, I'm wondering if this is a bug/regression.

There are a variety of issues regarding attributes but they are all closed.

- https://github.com/tidyverse/dplyr/issues/2432
- https://github.com/tidyverse/dplyr/issues/5132
- https://github.com/tidyverse/dplyr/issues/3429

DavisVaughan:
FWIW this happens in dplyr 1.0.10 and looks to have happened since at least 2020, so I don't consider this an immediate regression.

``` r
# pak::pak("tidyverse/dplyr@v1.0.10")
library(dplyr, warn.conflicts = FALSE)

d = mtcars
attr(d, "foo") = "bar"
attr(d, "foo")
#> [1] "bar"

# kept after group-by
d |> group_by(cyl) |> attr("foo")
#> [1] "bar"

# dropped here
[...]

PR #6817, opened 2023-04-07
replace as_tibble with new_tibble

mkoohafkan:
Make `ungroup()` handle attributes in a way that is consistent with other `dplyr` functions by using `new_tibble()` instead of `as_tibble()`.

This is also future-proof if it is later decided that `new_tibble()` should drop attributes.

Fixes https://github.com/tidyverse/dplyr/issues/6774 without affecting discussion in https://github.com/tidyverse/tibble/pull/769

### DIFF
--- a/R/group-by.R
+++ b/R/group-by.R
@@ -156,7 +156,7 @@ ungroup <- function(x, ...) {
 #' @export
 ungroup.grouped_df <- function(x, ...) {
   if (missing(...)) {
-    as_tibble(x)
+    new_tibble(x, groups = NULL)
   } else {
     old_groups <- group_vars(x)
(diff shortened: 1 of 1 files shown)


krlmlr (approved):
Thanks, this looks good to me. It seems that other attributes are preserved here:

``` r
tibble::new_tibble(
  structure(list(a = 1), foo = "bar", let_it = "go"),
  let_it = NULL
) |>
  attributes()
#> $class
#> 
#> $row.names
#> [1] 1
#> 
#> $names
#> [1] "a"
#> 
#> $foo
#> [1] "bar"
```

<sup>Created on 2023-08-23 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

PR #6832, opened 2023-04-24
Update case-when.R

arielhasidim:
Since it stated:
> `case_when()` is an R equivalent of the SQL "searched" `CASE WHEN` statement.
It makes sense to complete the analogy with `.default` as an else `statment`.


DavisVaughan:
Thanks!

PR #6840, opened 2023-05-04
Update starwars.csv

robcreel:
Correct the capitalization of Beru's last name from "lars" to "Lars".


DavisVaughan:
If we are going to fix this, it needs to happen in `starwars.R` that generates this CSV, not in the CSV itself.
https://github.com/tidyverse/dplyr/blob/main/data-raw/starwars.R

PR #6875, opened 2023-06-29
Fix typos plus minor tweaks

salim-b:


krlmlr (commented):
Thanks. The "be able to" might seem redundant, and we need to update the `.Rd` files?

salim-b:
> The "be able to" might seem redundant

I added this because it's optional that the user supplies a tidyselect specification when wrapping the function argument in double braces (`{{ arg }}`); `arg` could still be a "regular" symbol or vector, right?

> and we need to update the .Rd files

You want me to add this to this PR? Or do you use some GH action for this anyways?

krlmlr:
I'm not sure, I'm not a native speaker.

There are other PRs without `.Rd` updates, it's fine.

DavisVaughan:
Thanks!

PR #6880, opened 2023-07-09
Add closing backtick in message

krlmlr:
Should this be `deprecate_soft()` ?


DavisVaughan:
Probably not worth the mental effort to switch to `deprecate_soft()` since `sample_n()` itself is superseded

PR #6882, opened 2023-07-16
Add tests for closest()

krlmlr:
I haven't found any behavior tests in the test suite.


DavisVaughan:
I think this was because I have a massive amount of behavior testing in vctrs for `vec_locate_matches()`, and ended up mostly testing interface stuff here, but some basic tests seem fine / useful since you use them in duckplyr

Issue #6869, opened 2023-06-21
Deprecated feature `summarize_each_()`


jpmonteagudo28:
---
The `summarize_each_( )` feature is still being used in dplyr even though it was deprecated in a previous version. 

```r
humanAR_df$activities <- factor(humanAR_df$activities, levels = activities[,1], labels = activities[,2])
humanAR_df <- humanAR_df %>% group_by(subject,activities) %>% summarize_each(mean)

## Warning message:
## `summarise_each_()` was deprecated in dplyr 0.7.0.
## ℹ Please use `across()` instead.
## ℹ The deprecated feature was likely used in the dplyr package.
## Please report the issue at <https://github.com/tidyverse/dplyr/issues>.
## This warning is displayed once every 8 hours.
## Call `lifecycle::last_lifecycle_warnings()` to see where this warning was
## generated. 
```
A warning message basically shows that `across ( )` should be used instead of `summarize_each_( )`

DavisVaughan:
Really its `summarise_each()` that is deprecated https://github.com/tidyverse/dplyr/blob/16b472fb2afc50a87502c2b4ed803e2f5f82b9d6/R/deprec-lazyeval.R#L356-L362

PR #6884, opened 2023-07-17
Fix `mutate_each()` and `summarise_each()` deprecation warnings

DavisVaughan:
Closes https://github.com/tidyverse/dplyr/issues/6869


moodymudskipper:
For static analysis purpose, and general exploration by printing functions, shouldn't the deprecation warnings be triggered directly by the exported functions ?

PR #6901, opened 2023-08-04
Fix duplicate test names

krlmlr:
Cases like this can cause mild confusion. Should testthat warn here?


DavisVaughan:
It can also sometimes mess up snapshot tests, so that's not a bad idea. Maybe we'd make the warning optional though?

krlmlr:
Thanks, filed issue.

PR #6904, opened 2023-08-04
Add tests for behavior of count()

krlmlr:
None of the existing tests was triggered when the sort order wasn't established.

I see that `summarize(.by = ...)` has a different default (of returning in the order of appearance), I wonder what the long-term intention for `count()` is.

### DIFF
--- a/tests/testthat/test-count-tally.R
+++ b/tests/testthat/test-count-tally.R
@@ -12,6 +12,12 @@ test_that("count can sort output", {
   expect_equal(out, tibble(x = c(2, 1), n = c(3, 2)))
 })
 
+test_that("count can rename grouping columns", {
+  df <- tibble(x = c(2, 1, 1, 2, 1))
+  out <- count(df, y = x)
+  expect_equal(out, tibble(y = c(1, 2), n = c(3, 2)))
+})
+
 test_that("informs if n column already present, unless overridden", {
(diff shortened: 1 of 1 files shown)


tests/testthat/test-count-tally.R
```diff
@@ -1,5 +1,23 @@
...
+test_that("count can sort output", {
+  df <- tibble(x = c(1, 1, 2, 2, 2))
+  out <- count(df, x, sort = TRUE)
+  expect_equal(out, tibble(x = c(2, 1), n = c(3, 2)))
+})
+
+test_that("count can rename grouping columns", {
+  df <- tibble(x = c(2, 1, 1, 2, 1))
+  out <- count(df, y = x)
+  expect_equal(out, tibble(y = c(1, 2), n = c(3, 2)))
```
DavisVaughan, line 18: 😬 I didn't even know this worked and I'm not sure I like it

krlmlr: A CRAN package might use that feature.

krlmlr:
A random sort order means that `count()` behaves somewhat differently from `summarize()`, though. I don't have a strong opinion, we can leave it open for now.

DavisVaughan:
I'm happy to merge this one as is, I do agree we should at least test it so if we ever change things in the future it forces us to acknowledge that we are breaking something

PR #6909, opened 2023-08-12
Materialize results in collect()

krlmlr:
for duckplyr.


DavisVaughan:
Do you think we can delay this one for now and keep this behavior in the `collect.duckplyr_df()` method that you have here?
https://github.com/duckdblabs/duckplyr/blob/main/R/collect.R

With the goal of eventually moving it to dplyr in the future if we can't come up with any better solution

krlmlr:
Sure. This is just what I think this will need to look like long-term, we don't need to act now.

PR #6938, opened 2023-10-08
`starwars` data - Replace defunct API

steveharoz:
1. The API used to generate the `starwars` data is no longer operational. I found 2 possible replacements (https://swapi.dev/about and https://swapi.py4e.com/about). The py4e site was more complete, so I used that one.
2. A typo was pointed out in #6840. The issue was fixed in swapi.dev but not in swapi.py4e.com, so I just fixed it myself in the code.
3. One character's details (Jek Tono Porkins) were missing from the new API, so I added them in code using the old data.
4. The order of the films was arbitrary in the old API. The new API sorts it by release date. (This is the bulk of the diff)
5. There are only very minor changes to the other data coming from the new API. So there shouldn't be much impact.


data-raw/starwars.csv
```diff
@@ -30,33 +30,35 @@ Arvel Crynyd,NA,NA,brown,fair,brown,NA,male,masculine,NA,Human,Return of the Jed
...
 Saesee Tiin,188,NA,none,pale,orange,NA,male,masculine,Iktotch,Iktotchi,"The Phantom Menace, Revenge of the Sith",,
 Yarael Poof,264,NA,none,white,yellow,NA,male,masculine,Quermia,Quermian,The Phantom Menace,,
-Plo Koon,188,80,none,orange,black,22,male,masculine,Dorin,Kel Dor,"Attack of the Clones, The Phantom Menace, Revenge of the Sith",,Jedi starfighter
-Mas Amedda,196,NA,none,blue,blue,NA,male,masculine,Champala,Chagrian,"Attack of the Clones, The Phantom Menace",,
-Gregar Typho,185,85,black,dark,brown,NA,male,masculine,Naboo,Human,Attack of the Clones,,Naboo fighter
-Cordé,157,NA,brown,light,brown,NA,female,feminine,Naboo,Human,Attack of the Clones,,
+Plo Koon,188,80,none,orange,black,22,male,masculine,Dorin,Kel Dor,"The Phantom Menace, Attack of the Clones, Revenge of the Sith",,Jedi starfighter
+Mas Amedda,196,NA,none,blue,blue,NA,male,masculine,Champala,Chagrian,"The Phantom Menace, Attack of the Clones",,
+Gregar Typho,185,85,black,dark,brown,NA,NA,NA,Naboo,NA,Attack of the Clones,,Naboo fighter
+Cordé,157,NA,brown,light,brown,NA,NA,NA,Naboo,NA,Attack of the Clones,,
```
DavisVaughan, line 61: Lost data here too

DavisVaughan (commented):
There are a few other places where we lost data in the switch to the new API, but we also gained some too. I don't _think_ we want to do too much manual tweaking, so I've removed your Jek adjustment for now (I think we'd either try to do all or nothing, and since Gregar and Corde also lost info, I think it is better to try to consistently do nothing right now)

DavisVaughan:
Thanks!

Issue #6838, opened 2023-05-02
Unhelpful error message when using `dplyr::between()` in `join_by()`


mgirlich:
I accidentally used `dplyr::between()` instead of `between()` in `join_by()`. Not sure whether you want to accept it with a namespace or not but the error message isn't very helpful.

``` r
dplyr::join_by(dplyr::between(y, a, b))
#> Error in `as_string()`:
#> ! Can't convert a call to a string.
#> Backtrace:
#>     ▆
#>  1. └─dplyr::join_by(dplyr::between(y, a, b))
#>  2.   └─dplyr:::parse_join_by_expr(exprs[[i]], i, error_call = error_call)
#>  3.     └─rlang::as_string(expr[[1]])
#>  4.       └─rlang:::abort_coercion(x, "a string")
#>  5.         └─rlang::abort(msg, call = call)
```

<sup>Created on 2023-05-02 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

hadley:
Yeah, probably.

balthasars:
> I accidentally used `dplyr::between()` instead of `between()` in `join_by()`. Not sure whether you want to accept it with a namespace or not but the error message isn't very helpful.
> 
> ```r
> dplyr::join_by(dplyr::between(y, a, b))
> #> Error in `as_string()`:
> #> ! Can't convert a call to a string.
> #> Backtrace:
> #>     ▆
> #>  1. └─dplyr::join_by(dplyr::between(y, a, b))
> #>  2.   └─dplyr:::parse_join_by_expr(exprs[[i]], i, error_call = error_call)
> #>  3.     └─rlang::as_string(expr[[1]])
> #>  4.       └─rlang:::abort_coercion(x, "a string")
> #>  5.         └─rlang::abort(msg, call = call)
> ```
> 
> Created on 2023-05-02 with [reprex v2.0.2](https://reprex.tidyverse.org)

Would be super neat if dplyr would support that so the new `join_by()` feature can also be used in situations where calling `dplyr::between()` is the only safe option. Thanks for the great work on dplyr!

DavisVaughan:
What do you mean by safe?

balthasars:
> What do you mean by safe?

Safe as in not loading the entire dplyr namespace using `library()` in legacy code projects that have the potential of namespace conflicts.

I suppose a potential workaround would be to use `library(dplyr, include.only = c("between"))`, which sort of invalidates my argument.

[...]

PR #6951, opened 2023-11-03
Allow for namespaced `join_by()` helpers

DavisVaughan:
Closes https://github.com/tidyverse/dplyr/issues/6838

@lionel- what do you think? This seemed pretty nice to me

### DIFF
--- a/R/join-by.R
+++ b/R/join-by.R
@@ -436,7 +436,17 @@ parse_join_by_expr <- function(expr, i, error_call) {
     expr[[1]] <- sym(call_name(expr))
   }
 
-  op <- as_string(expr[[1]])
+  op <- expr[[1]]
+
+  if (!is_symbol(op)) {
+    if (is_call(op, name = "::")) {
+      stop_invalid_namespaced_expression(expr, i, error_call)
+    } else {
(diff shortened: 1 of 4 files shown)


R/join-by.R
```diff
@@ -431,6 +431,13 @@ parse_join_by_expr <- function(expr, i, error_call) {
     abort(message, call = error_call)
   }
 
+  if (is_string(call_ns(expr), "dplyr")) {
```
lionel-, line 434: This could be: `is_call(expr, ns = "dplyr")`.

R/join-by.R
```diff
@@ -431,6 +431,13 @@ parse_join_by_expr <- function(expr, i, error_call) {
...
+    # Normalize by removing the `dplyr::`
```
lionel-, line 435: I think I'd just do:

```r
expr[[1]] <- sym(call_name(expr))
```

lionel- (approved):

lionel-:
Perhaps worth checking that `expr[[1]]` is a symbol and throw a user-facing error message if not the case. Currently the `as_string()` error checking is rather internal-facing.

Issue #6895, opened 2023-07-25
 Confirm if `dplyr::across` can pass same length of vector to  `.names` 


ynsec37:
Dear developer,

I found the below example work well but no related description in current [official document](https://dplyr.tidyverse.org/reference/across.html).

> .names	 A glue specification that describes how to name the output columns. This can use {.col} to stand for the selected column name, and {.fn} to stand for the name of the function being applied. The default (NULL) is equivalent to "{.col}" for the single function case and "{.col}_{.fn}" for the case where a list is used for .fns.

I just want to confirm whether this way is right or accepted, especilly when the new variable names can not simply be computed by `{.col}` or `{.fn}`

```r
library(dplyr, warn.conflicts = FALSE)

df <- tibble::tibble(a = 1, b = 2, c = 3)

cols <- c("a", "b", "c")
new_cols <- c("x", "y", "z")

df %>%
  mutate(across(
    .cols = all_of(cols),
    .fns = ~ .x + 10,
    .names = "{new_cols}"
  ))
#> # A tibble: 1 × 6
#>       a     b     c     x     y     z
#>   <dbl> <dbl> <dbl> <dbl> <dbl> <dbl>
#> 1     1     2     3    11    12    13
```
Thank very much for your reply.

moodymudskipper:
Since the doc says it's a glue specification I believe the behavior is accepted (though happy to see it confirmed), but maybe the intent might be made clearer by either of those 2, though these might also slightly bite into the undocumented territory. 

```
cols <- c("a", "b", "c")
new_cols <- c("x", "y", "z")

df %>%
  mutate(
    across(
    .cols = all_of(cols),
    .fns = ~ .x + 10
  ) %>% set_names(new_cols)
)

df %>%
  mutate(
    across(
      .cols = all_of(set_names(cols, new_cols)),
      .fns = ~ .x + 10
    )
  )
```

The former uses the fact that across returns a data frame, so should be stable, unless `across()` is changed to only be recognised at the top level in dplyr verbs (it's not very clear from the doc if only top level is expected).

The latter is not 100% documented either but pretty close through #6655

PR #6953, opened 2023-11-03
Add tests to ensure `across()` allows tidyselect renaming

DavisVaughan:
Closes https://github.com/tidyverse/dplyr/issues/6895

@hadley I think this is a valid usage of `across()` + renaming-in-tidyselect, do you agree?

``` r
library(rlang)

df <- tibble::tibble(a = 1, b = 2, c = 3)

cols <- set_names(c("a", "b"), c("x", "y"))

df %>%
  mutate(
    across(
      .cols = all_of(cols),
      .fns = ~ .x + 10
    )
  )
#> # A tibble: 1 × 5
#>       a     b     c     x     y
#>   <dbl> <dbl> <dbl> <dbl> <dbl>
#> 1     1     2     3    11    12
```

Extracted from https://github.com/tidyverse/dplyr/issues/6895#issuecomment-1660770953, adding a test because I didn't see any and want to ensure this continues to work.

### DIFF
--- a/tests/testthat/test-across.R
+++ b/tests/testthat/test-across.R
@@ -841,6 +841,29 @@ test_that("across() can access lexical scope (#5862)", {
   )
 })
 
+test_that("across() allows renaming in `.cols` (#6895)", {
+  df <- tibble(x = 1, y = 2, z = 3)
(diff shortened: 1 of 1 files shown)


hadley:
Yeah, seems fine to me.

Issue #6912, opened 2023-08-14
Catch and rethrow the "overflow" error from vctrs during joins


DavisVaughan:
See https://github.com/r-lib/vctrs/pull/1869, requires dev vctrs right now.

We could throw a much more informative error. Probably advising the user to check their join keys, because a misspecified join is typically the source of this overflow (as seen by many people opening vctrs issues about this)

PR #6956, opened 2023-11-06
Rethrow "overflow" join error with a better message

DavisVaughan:
Closes https://github.com/tidyverse/dplyr/issues/6912


R/join-rows.R
```diff
@@ -137,6 +140,28 @@ dplyr_locate_matches <- function(needles,
   )
 }
 
+rethrow_error_join_matches_overflow <- function(cnd, call) {
+  size <- cnd$size
+
+  stop_join(
+    message = c(
+      "This join would result in more rows than R can handle.",
```
DavisVaughan, line 148: Technically R can handle more with long vectors, but I didn't want to get into all that. Alternatively I can say `more rows than dplyr can handle` if we think that is better

lionel-: I like "dplyr" instead of "R"

R/join-rows.R
```diff
@@ -137,6 +140,28 @@ dplyr_locate_matches <- function(needles,
...
+      i = glue("{size} rows would be returned."),
+      i = "2147483647 is the maximum number of rows allowed.",
```
lionel-, line 150: There's a lot of bullets in the message. Maybe merge these?

R/join-rows.R
```diff
@@ -137,6 +140,28 @@ dplyr_locate_matches <- function(needles,
...
+      i = paste0(
+        "Double check your join keys. This error commonly occurs due to a ",
+        "missing join key, or an improperly specified join condition."
+      ),
+      i = cli::format_inline(paste0(
```
lionel-, line 155: Do we need this bullet? It feels like everytime this was reported in vctrs it was a user issue.

lionel- (approved):

PR #6987, opened 2024-02-04
Align snapshot tests

krlmlr:
Changes in cli, perhaps?


DavisVaughan:
cli I think, in this case, thanks!

PR #6988, opened 2024-02-04
Test that `summarize()` strips off the subclass

krlmlr:
I saw these tests in at least two tidyverse/tidymodels packages, but no equivalent test in dplyr. If this is intended, should we test and perhaps also document here?


krlmlr:
Test failures are fixed by #6987.

PR #7029, opened 2024-05-20
Extract `join_ptype_common()`

krlmlr:


DavisVaughan:
@krlmlr I moved the `vec_ptype_finalize()` call into `join_ptype_common()`, as I think it makes more sense there. Does that still work for you?

krlmlr:
Thanks, Davis. I think it should make no difference because `vec_ptype_finalise()` doesn't touch `x` or `y` .

Issue #6861, opened 2023-05-30
`mutate` superseding `transmute` should allow ordering columns


epruesse:
I recently noticed that `transmute` has been marked as superseded by `mutate(.keep="none")`. However, it turned out that `mutate` doesn't replicate column ordering behavior of `transmute`, but does something odd:

```r
> data.frame(a=1, b=2) %>% transmute(a, x=b*2, b)
  a x b
1 1 4 2
> data.frame(a=1, b=2) %>% mutate(a, x=b*2, b, .keep="none")
  a b x
1 1 2 4
```
With more complex examples, the ordering becomes pretty confusing and difficult to explain. I'm guessing this may have to do with the `.keep = "used"` use case resorting things. For `.keep = "none"`, explicit column ordering as given, replicating or approximating `transmute` behavior,  would be much more useful (e.g. order of first LHS mention or last LHS mention).

mgirlich:
This is one of the reasons why I still use `transmute()` over `mutate(.keep = "none")`.

DavisVaughan:
Two very related PRs:
- Unifying the implementation of `.keep` for `mutate()` https://github.com/tidyverse/dplyr/pull/6035
- Which then affected `transmute()`, so we gave it its own separate implementation here https://github.com/tidyverse/dplyr/issues/6086

Extremely important paragraph:

> The dev behavior of .keep = "none" is overall more consistent with the rest of the mutate() options, makes it easier to predict the output when combined with .before and .after, and simplifies the implementation because it means that .keep never affects the column ordering, it is mainly about which columns get dropped (https://github.com/tidyverse/dplyr/pull/6035 goes into this in great detail).

So we should be extremely careful when considering if we want to make any changes here. #6035's big insight is that `.keep` should not affect the column ordering _at all_, and I don't think we should go back to that.

> An important invariant that falls out here is that .keep plays no role in the column ordering, and I think that is valuable. I think giving keep = "none" special behavior in a few places that changed column order is what made this so hard to get correct before.

DavisVaughan:
I've refreshed myself on the logic in https://github.com/tidyverse/dplyr/pull/6035#issue-1012288146, and I am confident that the current implementation of `mutate(.keep = )` with its current 4 variants is correct.

A key principle is that only 1 argument should be able to affect the output ordering. As of right now, `.keep` does not affect the output ordering in any way. Only `.before` and `.after` affect the output ordering, and even they are _mutually exclusive_ because anything else is ambiguous.

[...]

PR #7038, opened 2024-06-02
.keep = "transmute" is implemented in mutate.

apalacio9502:
Hi @hadley and @DavisVaughan,

This merge request aims to align the mutate functionality with that of transmute. Presently, the documentation states that the equivalent of transmute is mutate(.keep="none"). However, it's important to note that this function retains the column order, unlike transmute.

After reviewing the bug reported here: https://github.com/tidyverse/dplyr/issues/6861, I concur that there should be a fifth variant of keep, named "transmute."

This proposal should also be implemented in the following tidyverse libraries:
- [ ] [dtplyr](https://github.com/tidyverse/dtplyr)  
- [ ] [dbplyr](https://github.com/tidyverse/dbplyr)  

And in the following third party libraries:
- [ ] [duckplyr](https://github.com/duckdblabs/duckplyr)
- [ ] [arrow](https://github.com/apache/arrow/r)
- [ ] [tidypolars](https://github.com/etiennebacher/tidypolars)

This change opens the possibility of un-superseded transmute. For example, you could implement that when a user calls transmute, it internally executes mutate(.keep="transmute"). This would greatly benefit veteran users who have extensively used transmute in their projects. Additionally, it improves code readability, as a simple transmute call is clearer than mutate(.keep="transmute").

Below I leave an example comparing transmute with mutate(.keep = "transmute")

```
example <- tibble(
  group = c("a", "a", "b", "b"),
  old = c(1, 2, 3, 4)
)

# transmute --------------------------------------------------------------------

example %>% transmute(
  group,
  new = old * 2,
  old
)

# A tibble: 4 × 3
  group   new   old
  <chr> <dbl> <dbl>
1 a         2     1
2 a         4     2
3 b         6     3
4 b         8     4

# mutate(.keep = "transmute") --------------------------------------------------

example %>% mutate(
  .keep = "transmute",
  group,
  new = old * 2,
  old
)

# A tibble: 4 × 3
  group   new   old
  <chr> <dbl> <dbl>
1 a         2     1
2 a         4     2
3 b         6     3
4 b         8     4

# group_by + transmute ---------------------------------------------------------

example %>% 
  group_by(group) %>% 
  transmute(
    new = sum(old),
    old
) %>% 
 ungroup()

# A tibble: 4 × 3
  group   new   old
  <chr> <dbl> <dbl>
1 a         3     1
2 a         3     2
3 b         7     3
4 b         7     4

# group_by + mutate(.keep = "transmute") ---------------------------------------

example %>% 
  group_by(group) %>% 
  mutate(
    .keep = "transmute",
    new = sum(old),
    old
) %>% 
 ungroup()

# A tibble: 4 × 3
  group   new   old
  <chr> <dbl> <dbl>
1 a         3     1
2 a         3     2
3 b         7     3
4 b         7     4

# mutate(.keep = "transmute", by = ...) ----------------------------------------

example %>%  
  mutate(
    .keep = "transmute",
    .by = "group",
    new = sum(old),
    old
  )

# A tibble: 4 × 3
  group   new   old
  <chr> <dbl> <dbl>
1 a         3     1
2 a         3     2
3 b         7     3
4 b         7     4

```

Regards,

### DIFF
--- a/R/mutate.R
+++ b/R/mutate.R
@@ -163,6 +163,8 @@ mutate <- function(.data, ...) {
 #'     the columns used to generate them.
 #'   * `"none"` doesn't retain any extra columns from `.data`. Only the grouping
 #'     variables and columns created by `...` are kept.
+#'   * `"transmute"` equivalent to "none," but preserves the column order in the
+#'     mutate.
 #' @param .before,.after
 #'   <[`tidy-select`][dplyr_tidy_select]> Optionally, control where new columns
 #'   should appear (the default is to add to the right hand side). See
@@ -171,10 +173,10 @@ mutate <- function(.data, ...) {
 mutate.data.frame <- function(.data,
                               ...,
                               .by = NULL,
-                              .keep = c("all", "used", "unused", "none"),
+                              .keep = c("all", "used", "unused", "none", "transmute"),
                               .before = NULL,
                               .after = NULL) {
-  keep <- arg_match0(.keep, values = c("all", "used", "unused", "none"))
+  keep <- arg_match0(.keep, values = c("all", "used", "unused", "none", "transmute"))
 
   by <- compute_by({{ .by }}, .data, by_arg = ".by", data_arg = ".data")
 
@@ -187,6 +189,7 @@ mutate.data.frame <- function(.data,
 
   out <- mutate_relocate(
     out = out,
+    keep = keep,
     before = {{ .before }},
     after = {{ .after }},
     names_original = names_original
@@ -208,11 +211,11 @@ mutate.data.frame <- function(.data,
 
 # Helpers -----------------------------------------------------------------
 
-mutate_relocate <- function(out, before, after, names_original) {
+mutate_relocate <- function(out, keep, before, after, names_original) {
   before <- enquo(before)
   after <- enquo(after)
 
-  if (quo_is_null(before) && quo_is_null(after)) {
+  if (keep == "transmute" || (quo_is_null(before) && quo_is_null(after))) {
     return(out)
   }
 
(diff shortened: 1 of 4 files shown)


DavisVaughan:
We may simply un-supercede `transmute()` since many people seem to like it and have pushed back on that, in which case we probably would not add this because it does have very different properties than the theoretically consistent set of existing `.keep` options. I am not sure yet though, we will need to have an internal discussion about it and we aren't currently doing a large amount of dplyr dev at the moment.

apalacio9502:
Hi @DavisVaughan,

Thank you for your response and I appreciate your perspective.

From my standpoint as a user, I think it would be beneficial to consider the following alternative.

- Incorporating the parameter mutate(.keep = "transmute"): I am convinced that adding this parameter would enrich the process, providing greater flexibility. This would allow for quick adaptation of the outcome through the option .keep = c("all", "used", "unused", "none", "transmute"). If we only opt for "un-superseded transmute" and a user wishes to transform a "mutate" into a "transmute", it would entail changing the function name and removing parameters that are not compatible with "transmute".

- Un-superseded transmute: This would be valuable for those users who prefer using transmute due to its clarity in the code, or for those who encounter difficulties when migrating to mutate(.keep = "transmute"), as it involves more than just a direct replacement (for example, if one wishes to add .keep at the end of the code or adopt a particular style). Additionally, I believe it would be interesting to include the parameter .by in transmute, which would add value for users.

Regards,

DavisVaughan:
> Additionally, I believe it would be interesting to include the parameter .by in transmute, which would add value for users.

If we unsupercede, definitely!

PR #7046, opened 2024-06-30
Create quosures in `tally_n()`

krlmlr:


R/count-tally.R
```diff
@@ -213,9 +213,9 @@ tally_n <- function(x, wt) {
   }
 
   if (quo_is_null(wt)) {
-    expr(n())
+    new_quosure(expr(n()), asNamespace("dplyr"))
   } else {
-    expr(sum(!!wt, na.rm = TRUE))
+    new_quosure(expr(sum(!!wt, na.rm = TRUE)), quo_get_env(wt))
```
krlmlr, line 218: This probably should be:

```suggestion
    new_quosure(expr(sum(!!wt, na.rm = TRUE)), baseenv())
```

DavisVaughan:
@krlmlr do you still use this feature in duckplyr? Would switching to `dplyr::n()` and `base::sum()` work instead?

krlmlr:
`dplyr::n()` and `base::sum()` would be fine too, yes.

DavisVaughan:
So to be super clear, instead of `expr(n())` we could put `expr(dplyr::n())`, and instead of `expr(sum(!!wt, na.rm = TRUE))` we could put `expr(base::sum(!!wt, na.rm = TRUE))`, and that'd please duckplyr? That sounds much preferable to me!

PR #7049, opened 2024-07-08
Implement own `df_n_col()` to avoid calling `ncol()` (and `dim()`)

krlmlr:
This leads to premature materialization in `union_all()` and others with duckplyr 0.4.0 . I'll have to work around by vendoring everything that uses `ncol()` directly or indirectly into duckdb.

### DIFF
--- a/R/data-storms.R
+++ b/R/data-storms.R
@@ -10,7 +10,7 @@
 #'   <https://github.com/tidyverse/dplyr/blob/main/data-raw/storms.R>
 #'
 #' @format A tibble with `r format(nrow(storms), big.mark = ",")` observations
-#'   and `r ncol(storms)` variables:
+#'   and `r df_n_col(storms)` variables:
 #' \describe{
 #' \item{name}{Storm Name}
 #' \item{year,month,day}{Date of report}
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,8 @@
 # dplyr (development version)
 
+* Fixed an issue where duckplyr's ALTREP data frames were being materialized
+  early due to internal usage of `ncol()` (#7049).
+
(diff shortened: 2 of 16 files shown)


DavisVaughan:
@krlmlr I've introduced our own version of `df_n_col()` and _blocked_ usage of base `ncol()` to try and prevent this from happening in the future.

We do still need to call `base::ncol()` on a matrix in a single case (that's why your commit had test failures), so I also added in `mat_n_col()` as well to allow us to use that on matrices as required.

krlmlr:
Thanks for taking this on, Davis. I love the "for future us" bit!

R/data-storms.R
```diff
@@ -10,7 +10,7 @@
 #'   <https://github.com/tidyverse/dplyr/blob/main/data-raw/storms.R>
 #'
 #' @format A tibble with `r format(nrow(storms), big.mark = ",")` observations
-#'   and `r ncol(storms)` variables:
+#'   and `r df_n_col(storms)` variables:
```
krlmlr, line 13:
```suggestion
#'   and `r ncol(storms)` variables:
```

krlmlr (commented):
Minor omission.

lionel- (approved):
LG!

R/data-storms.R
DavisVaughan: It's actually correct as is. This is inline markdown so it gets evaluated and you get this otherwise (as expected)

```r
Quitting from lines 1-1
✖ [data-storms.R:12]: @format failed to evaluate inline markdown code.
Caused by error in `map_chr()`:
ℹ In index: 2.
Caused by error in `ncol()`:
! Use `df_n_col()` or `mat_n_col()` instead.
```

Issue #6848, opened 2023-05-12
slice_sample() uses statistically incorrect sampling algorithms


skolenik:
`slice_sample()` relies on `base::sample.int()` which in turn relies on incorrect sampling algorithms. The difficulty of obtaining appropriate unequal probability samples is obscure and is not known outside of survey statistics. I don't want to go into base R code although they have to fix that, too (somewhat more surprising given that Thomas Lumley, one of the core R people, is a survey statistician, at least part time). The appropriate algorithms are implemented in `sampling::UPmaxentropy()`.

```r
# example from Tille (2006)
set.seed(25)
uneq_p = c(0.07,0.17,0.41,0.61,0.83,0.91)
sim_uneq_p <- as.data.frame(matrix(rep(0,6*20000),ncol=6))
for (k in 1:nrow(sim_uneq_p)) {
  this <- sample.int(n=length(uneq_p),size=sum(uneq_p),prob=uneq_p)
  sim_uneq_p[k,this] <- 1
}
colSums(sim_uneq_p)/nrow(sim_uneq_p)
uneq_p
# done right
sim_uneq_p_done_right <- as.data.frame(matrix(rep(0,6*20000),ncol=6))
for (k in 1:nrow(sim_uneq_p_done_right)) {
   sim_uneq_p_done_right[k,] <- sampling::UPmaxentropy(uneq_p)
}
colSums(sim_uneq_p_done_right)/nrow(sim_uneq_p_done_right)
uneq_p
```

@krlmlr @DavisVaughan 

Code in `sampling` is (kinda) ugly, and the development most likely does not satisfy the `tidyverse` standards. I don't know if you want to rely on this as a dependency. Relevant parts may need to be taken over and internalized. (I have written the unequal probability sampling code from scratch in Stata, so I am closely familiar with the methodology and what it takes to code it.)

skolenik:
Code in `sampling` is ugly, and the development most likely does not satisfy the `tidyverse` standards. I don't know if you want to rely on this as a dependency. Relevant parts may need to be taken over and internalized. (I have written the unequal probability sampling code from scratch in Stata, so I am closely familiar with the methodology and what it takes to code it.)

DavisVaughan:
Thanks for the report! We don't currently have any plans to switch away from the base R algorithms here. I'd encourage you to advocate for this change in base R itself, and see if they will be open to changing it there.

krlmlr:
[...]

PR #7052, opened 2024-07-14
Add documentation clarifying appropriate use of weights in `slice_sample()`

apeterson91:
Users may be led astray by the current language describing weighted sampling via `slice_sample()`. As discussed in https://stats.stackexchange.com/q/639211/ and elaborated on in issue #6848, there's more nuance to how weights might be used for sampling. The current base R algorithm which `slice_sample()` relies on does not use the sampling weights in a way that conform to how a user might suspect the weights ought to work, as in Horvitz Thompson estimation for example. 

I've added appropriate details here in the documentation including the above link for further details.

### DIFF
--- a/R/slice.R
+++ b/R/slice.R
@@ -19,6 +19,12 @@
 #' intrinsic notion of row order. If you want to perform the equivalent
 #' operation, use [filter()] and [row_number()].
 #'
+#' For `slice_sample()`, note that the weights provided in `weight_by` are
+#' passed through to the `prob` argument of [base::sample.int()]. This means
+#' they cannot be used to reconstruct summary statistics from the underlying
+#' population. See [this discussion](https://stats.stackexchange.com/q/639211/)
+#' for more details.
+#'
 #' @family single table verbs
 #' @inheritParams args_by
 #' @inheritParams arrange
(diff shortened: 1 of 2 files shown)


etiennebacher:
Hi, I'm not a member of the `tidyverse` team, but I just want to point out that you need to redocument the package to include this change in the `.Rd` file. You can use `devtools::document()`, or comment "/document" (without the quotes) in this PR to trigger a github action workflow (I'm not sure the latter will work).

apeterson91:
Thanks @etiennebacher! Been a bit since I've built an R package locally appreciate the reminder.

skolenik (changes requested):
The description of the problem is insufficient. I can poke at a more accurate answer with better references to both methodology and the existing R libraries.

(I would also point out that it is a somewhat better practice to name your branches informatively rather than `main` -- I would have called this `slice_sample_7052` to indicate the function affected and the issue number.)

apeterson91:
I don't completely follow the first part of your comment, but your edits LGTM. 

> I would also point out that it is a somewhat better practice to name your branches informatively

Acknowledged. I'll follow this convention in the future.

DavisVaughan:
Thanks @apeterson91!

PR #7069, opened 2024-08-15
add error message when passing a matrix

### LINKED ISSUE #6862: `case_when()` needs to throw a better error when conditions are matrices
Please briefly describe your problem and what output you expect. If you have a question, please don't use this form. Instead, ask on <https://stackoverflow.com/> or <https://community.rstudio.com/>.

Please include a minimal reproducible example (AKA a reprex). If you've never heard of a [reprex](http://reprex.tidyverse.org/) before, start by reading <https://www.tidyverse.org/help/#reprex>.

---

I had some code that was passing matrices as the conditions to a `case_when` statement. Previously, `case_when` would implicitly convert these matrices to vectors, but at some point it stopped doing that. However, I don't think this was an intentional behavior change, because depending on how exactly one calls `case_when` on matrix conditions, it can produce one of several different error messages or return a value of the wrong length (see reprex below demonstrating 3 different behaviors). None of the error messages hint at the underlying problem, which made debugging this issue after a package update very difficult.

If passing matrices as the conditions to `case_when` is not an intended usage, I would expect it to produce an interpretable error message to that effect.

``` r
library(dplyr)
#> 
#> Attaching package: 'dplyr'
#> 
#>     filter, lag
#> 
library(assertthat)

x <- matrix(rnorm(36), ncol = 6)

## Each of the following generates a different error
{
  y1 <- case_when(
    !is.finite(x) ~ "Invalid",
    .default = "Default"
  )
  assert_that(length(y1) == length(x))
}
#> Error: length(y1) not equal to length(x)

{
  y2 <- case_when(
    !is.finite(x) ~ "Invalid",
    x < 0 ~ "Negative",
    .default = "Default"
  )
  assert_that(length(y2) == length(x))
}
#> Error in `vec_slice()`:
#> ! Can't subset elements past the end.
#> ℹ Locations 8, 10, 13, …, 35, and 36 don't exist.
#> ℹ There are only 6 elements.
#> Backtrace:
#>     ▆
#>  1. ├─dplyr::case_when(...)
#>  2. │ └─dplyr:::vec_case_when(...)
#>  3. │   └─vctrs::vec_slice(value, loc)
#>  4. └─vctrs (local) `<fn>`()
#>  5.   └─vctrs:::stop_subscript_oob(...)
#>  6.     └─vctrs:::stop_subscript(...)
#>  7.       └─rlang::abort(...)

{
  y3 <- case_when(
    !is.finite(x) ~ "Invalid",
    x < 0 ~ "Negative",
    x > 0 ~ "Positive",
    x == 0 ~ "Zero"
  )
  assert_that(length(y3) == length(x))
}
#> Error in if (!any(are_unused)) {: missing value where TRUE/FALSE needed

## Everything works fine if you convert x to a vector
{
  xvec <- as.vector(x)
  yvec <- case_when(
    !is.finite(xvec) ~ "Invalid",
    xvec < 0 ~ "Negative",
    xvec > 0 ~ "Positive",
    xvec == 0 ~ "Zero"
  )
  assert_that(length(yvec) == length(xvec))
}
#> [1] TRUE
```

<sup>Created on 2023-05-30 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

ilovemane:
Closes #6862

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,8 @@
 # dplyr (development version)
 
+* `case_when()` now throws a better error if one of the conditions is an array
+  (#6862, @ilovemane).
+
 * `between()` gains a new `ptype` argument, allowing users to specify the
   desired output type. This is particularly useful for ordered factors and other
   complex types where the default common type behavior might not be ideal
--- a/R/vec-case-when.R
+++ b/R/vec-case-when.R
@@ -52,6 +52,9 @@ vec_case_when <- function(conditions,
     condition <- conditions[[i]]
     condition_arg <- condition_args[[i]]
     check_logical(condition, arg = condition_arg, call = call)
+    check_matrix(condition, call = call)
+
+
   }
 
   size <- vec_size_common(
@@ -206,4 +209,12 @@ names_as_error_names <- function(names, arg = "") {
 vec_paste0 <- function (...) {
   args <- vec_recycle_common(...)
   exec(paste0, !!!args)
+}
+
+check_matrix <- function(condition, call) {
+  if (inherits(condition, "matrix")) {
+    abort("`conditions` cannot be a matrix.", call = call)
+
+  }
+
 }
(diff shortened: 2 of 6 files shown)


DavisVaughan:
Thanks @ilovemane!

Issue #7030, opened 2024-05-21
group_by() documentation (v1.1.4) doesn't indicate whether ... is <tidy-select> or <data-masking> (it's <data-masking>)


DanielBraddock:
I'm looking at dplyr documentation for version 1.1.4.

The vignette https://dplyr.tidyverse.org/articles/programming.html states that "To determine whether a function argument uses data masking or tidy selection, look at the documentation: in the arguments list, you’ll see \<data-masking\> or \<tidy-select\>." (4th paragraph)

The vignette does tell you which verb is which including that group_by() is \<data-masking\>.

However, group_by()'s documentation doesn't tell you which `...` is.

PR #7071, opened 2024-08-15
[document] group_by works with data-masking, issue #7030

nmercadeb:
Closes https://github.com/tidyverse/dplyr/issues/7030


DavisVaughan:
Thanks @nmercadeb!

PR #7072, opened 2024-08-15
Fix `if_any()` and `if_all()` behavior with zero-column selections

### LINKED ISSUE #7059: `if_any` on zero-column selection yields `TRUE`
Please briefly describe your problem and what output you expect. If you have a question, please don't use this form. Instead, ask on <https://stackoverflow.com/> or <https://community.rstudio.com/>.

`if_any(<zero-column selection>, .fns)` will produce `TRUE`.  This contradicts expectations from, e..g, `any()`.

Please include a minimal reproducible example (AKA a reprex). If you've never heard of a [reprex](http://reprex.tidyverse.org/) before, start by reading <https://www.tidyverse.org/help/#reprex>.

---

Brief description of the problem
``` r
suppressPackageStartupMessages({
  library(dplyr)
})

tbl <- tibble(
  x1 = 1:5,
  x2 = c(-1, 4, 5, 4, 1),
  y = c(1, 4, 2, 4, 9),
)

# Basic example: this seems surprising:
tbl %>% filter(if_any(c(), ~ is.na(.x)))
#> # A tibble: 5 × 3
#>      x1    x2     y
#>   <int> <dbl> <dbl>
#> 1     1    -1     1
#> 2     2     4     4
#> 3     3     5     2
#> 4     4     4     4
#> 5     5     1     9

# It seems more natural to return zero rows, like this workaround:
tbl %>% filter(!if_all(c(), ~ !is.na(.x)))
#> # A tibble: 0 × 3
#> # ℹ 3 variables: x1 <int>, x2 <dbl>, y <dbl>

# Debugging: checking what is directly output from `if_any`:
tbl %>% filter({tmp <- if_any(c(), ~ is.na(.x)); print(tmp); tmp}) %>% invisible()
#> [1] TRUE

# Less minimal, motivating example: I don't recall the original context where I
# originally encountered this. But we might imagine models where the user can
# manually specify certain variables for special treatment, or when variables
# are selected for special treatment automatically.

predict1 <- function(tbl, regular_predictors, rate_predictors, outcomes) {
  problematic_rows <- tbl %>%
    filter(if_any({{rate_predictors}}, ~ .x < 0))
  if (nrow(problematic_rows) != 0L) {
    stop(paste0(
      "Rate predictors should always be < 0; some problematic rows shown below:",
      paste(collapse = "\n", paste0("  ", capture.output(print(problematic_rows))))
    ))
  }
  return (42L)
}

# Output here is good:
predict1(tbl, x1, x2, y)
#> Error in predict1(tbl, x1, x2, y): Rate predictors should always be < 0; some problematic rows shown below:  # A tibble: 1 × 3
#>        x1    x2     y
#>     <int> <dbl> <dbl>
#>   1     1    -1     1

# Output here is bad; there shouldn't be a problem flagged:
predict1(tbl, c(x1, x2), c(), y)
#> Error in predict1(tbl, c(x1, x2), c(), y): Rate predictors should always be < 0; some problematic rows shown below:  # A tibble: 5 × 3
#>        x1    x2     y
#>     <int> <dbl> <dbl>
#>   1     1    -1     1
#>   2     2     4     4
#>   3     3     5     2
#>   4     4     4     4
#>   5     5     1     9
```

<sup>Created on 2024-07-26 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

jrwinget:
Fixes #7059

This commit addresses the unexpected behavior of `if_any()` and `if_all()` when no columns are selected. Specifically, `if_any()` now returns `FALSE`, and `if_all()` returns `TRUE` in cases where no columns match the selection criteria.

Additionally, tests have been added to ensure the correct behavior of these functions in both `filter()` and `mutate()` contexts, including the scenarios highlighted by the `reprex` example. This ensures that empty selections behave consistently with the logical expectations.

### DIFF
--- a/R/across.R
+++ b/R/across.R
@@ -658,7 +658,7 @@ expand_if_across <- function(quo) {
   # Select all rows if there are no inputs for if_all(),
   # but select no rows if there are no inputs for if_any().
   if (!length(quos)) {
-    return(list(quo(empty)))
+    return(list(quo(!!empty)))
   }
 
   combine <- function(x, y) {
--- a/R/across.R
+++ b/R/across.R
@@ -12,18 +12,19 @@
 #' the predicate is `TRUE` for *any* of the selected columns, `if_all()`
 #' is `TRUE` when the predicate is `TRUE` for *all* selected columns.
 #'
-#' **Important Behavior:** When there are no selected columns:
-#'
-#' - `if_any()` will return `FALSE`, consistent with the behavior of
-#'   `any()` when called without inputs.
-#' - `if_all()` will return `TRUE`, consistent with the behavior of
-#'   `all()` when called without inputs.
-#'
 #' If you just need to select columns without applying a transformation to each
 #' of them, then you probably want to use [pick()] instead.
 #'
 #' `across()` supersedes the family of "scoped variants" like
 #' `summarise_at()`, `summarise_if()`, and `summarise_all()`.
+#'
+#' @details
+#' When there are no selected columns:
+#'
+#' - `if_any()` will return `FALSE`, consistent with the behavior of
+#'   `any()` when called without inputs.
+#' - `if_all()` will return `TRUE`, consistent with the behavior of
+#'   `all()` when called without inputs.
 #'
 #' @param .cols <[`tidy-select`][dplyr_tidy_select]> Columns to transform.
 #'   You can't select grouping columns because they are already automatically
--- a/R/across.R
+++ b/R/across.R
@@ -4,13 +4,20 @@
 #' `across()` makes it easy to apply the same transformation to multiple
 #' columns, allowing you to use [select()] semantics inside in "data-masking"
 #' functions like [summarise()] and [mutate()]. See `vignette("colwise")` for
-#'  more details.
+#' more details.
 #'
 #' `if_any()` and `if_all()` apply the same
 #' predicate function to a selection of columns and combine the
 #' results into a single logical vector: `if_any()` is `TRUE` when
 #' the predicate is `TRUE` for *any* of the selected columns, `if_all()`
 #' is `TRUE` when the predicate is `TRUE` for *all* selected columns.
+#'
+#' **Important Behavior:** When there are no selected columns:
+#'
+#' - `if_any()` will return `FALSE`, consistent with the behavior of
+#'   `any()` when called without inputs.
+#' - `if_all()` will return `TRUE`, consistent with the behavior of
+#'   `all()` when called without inputs.
 #'
 #' If you just need to select columns without applying a transformation to each
 #' of them, then you probably want to use [pick()] instead.
@@ -133,9 +140,16 @@
 #' iris %>%
 #'   group_by(Species) %>%
 #'   summarise(across(starts_with("Sepal"), mean, .names = "mean_{.col}"))
+#'
 #' iris %>%
 #'   group_by(Species) %>%
-#'   summarise(across(starts_with("Sepal"), list(mean = mean, sd = sd), .names = "{.col}.{.fn}"))
+#'   summarise(
+#'     across(
+#'       starts_with("Sepal"),
+#'       list(mean = mean, sd = sd),
+#'       .names = "{.col}.{.fn}"
+#'     )
+#'   )
 #'
 #' # If a named external vector is used for column selection, .names will use
 #' # those names when constructing the output names
@@ -146,7 +160,9 @@
 #' # When the list is not named, .fn is replaced by the function's position
 #' iris %>%
 #'   group_by(Species) %>%
-#'   summarise(across(starts_with("Sepal"), list(mean, sd), .names = "{.col}.fn{.fn}"))
+#'   summarise(
+#'     across(starts_with("Sepal"), list(mean, sd), .names = "{.col}.fn{.fn}")
+#'   )
 #'
 #' # When the functions in .fns return a data frame, you typically get a
 #' # "packed" data frame back
@@ -164,7 +180,9 @@
 #'
 #' # .unpack can utilize a glue specification if you don't like the defaults
 #' iris %>%
-#'   reframe(across(starts_with("Sepal"), quantile_df, .unpack = "{outer}.{inner}"))
+#'   reframe(
+#'     across(starts_with("Sepal"), quantile_df, .unpack = "{outer}.{inner}")
+#'   )
 #'
 #' # This is also useful inside mutate(), for example, with a multi-lag helper
 #' multilag <- function(x, lags = 1:3) {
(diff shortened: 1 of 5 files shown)


tests/testthat/test-across.R
```diff
@@ -888,6 +888,32 @@ test_that("if_any() and if_all() expansions deal with no inputs or single inputs
...
+test_that("if_any() on zero-column selection behaves as expected in filter", {
+  tbl <- tibble(
+    x1 = 1:5,
+    x2 = c(-1, 4, 5, 4, 1),
+    y = c(1, 4, 2, 4, 9),
+  )
+
+  expect_equal(
+    filter(tbl, if_any(c(), ~ is.na(.x))),
+    filter(tbl, FALSE)
```
krlmlr, line 900: Expected values seem slightly better without too much logic, not sure if it's easier to read though:

```suggestion
    tbl[0, ]
```

tests/testthat/test-across.R
```diff
@@ -888,6 +888,32 @@ test_that("if_any() and if_all() expansions deal with no inputs or single inputs
...
+test_that("if_all() on zero-column selection behaves as expected in filter", {
+  tbl <- tibble(
+    x1 = 1:5,
+    x2 = c(-1, 4, 5, 4, 1),
+    y = c(1, 4, 2, 4, 9),
+  )
+
+  expect_equal(
+    filter(tbl, if_all(c(), ~ is.na(.x))),
+    filter(tbl, TRUE)
```
krlmlr, line 913:
```suggestion
    tbl
```

R/across.R
```diff
@@ -634,9 +654,10 @@ expand_if_across <- function(quo) {
   call[[1]] <- quote(across)
   quos <- expand_across(quo_set_expr(quo, call))
 
-  # Select all rows if there are no inputs
+  # Select all rows if there are no inputs for if_all(),
+  # but select no rows if there are no inputs for if_any().
   if (!length(quos)) {
-    return(list(quo(TRUE)))
+    return(list(quo(empty)))
```
krlmlr, line 660: We need `!!` to plug in the actual value. Do we need the `empty` variable at all?

```suggestion
    return(list(quo(!!(if_fn == "if_all"))))
```

krlmlr (commented):
Thanks, looks great! My proposals are non-authoritative.

R/across.R
```diff
@@ -133,9 +140,16 @@
...
 #' iris %>%
 #'   group_by(Species) %>%
-#'   summarise(across(starts_with("Sepal"), list(mean = mean, sd = sd), .names = "{.col}.{.fn}"))
+#'   summarise(
+#'     across(
+#'       starts_with("Sepal"),
+#'       list(mean = mean, sd = sd),
+#'       .names = "{.col}.{.fn}"
+#'     )
+#'   )
```
DavisVaughan, line 152: Totally fine for today, but in general when we write PRs we try to keep them focused only on the bug at hand (see also https://code-review.tidyverse.org/author/focused.html). So, in this case, you would have done a specific PR just for documentation improvements like this. But all good!

R/across.R
DavisVaughan, line 660:
```suggestion
    return(list(quo(!!empty)))
```

@krlmlr is right but i still think that having the `empty` variable is a little clearer, so here is the change you can make

tests/testthat/test-across.R
```diff
@@ -888,6 +888,32 @@ test_that("if_any() and if_all() expansions deal with no inputs or single inputs
   )
 })
 
+test_that("if_any() on zero-column selection behaves as expected in filter", {
```
DavisVaughan, line 891:
```suggestion
test_that("if_any() on zero-column selection behaves like any() (#7059)", {
```

We often like to link to the issue when we add a test specifically for it

tests/testthat/test-across.R
```diff
@@ -888,6 +888,32 @@ test_that("if_any() and if_all() expansions deal with no inputs or single inputs
...
+    y = c(1, 4, 2, 4, 9),
+  )
+
+  expect_equal(
+    filter(tbl, if_any(c(), ~ is.na(.x))),
+    filter(tbl, FALSE)
+  )
+})
+
+test_that("if_all() on zero-column selection behaves as expected in filter", {
```
DavisVaughan, line 904:
```suggestion
test_that("if_all() on zero-column selection behaves like all() (#7059)", {
```

DavisVaughan (changes requested):
Looking good! Here are a few comments

R/across.R
```diff
@@ -4,14 +4,21 @@
 #' `across()` makes it easy to apply the same transformation to multiple
 #' columns, allowing you to use [select()] semantics inside in "data-masking"
 #' functions like [summarise()] and [mutate()]. See `vignette("colwise")` for
-#'  more details.
+#' more details.
 #'
 #' `if_any()` and `if_all()` apply the same
```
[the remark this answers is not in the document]
jrwinget: Sure thing!

R/across.R
```diff
@@ -4,14 +4,21 @@
...
 #' results into a single logical vector: `if_any()` is `TRUE` when
 #' the predicate is `TRUE` for *any* of the selected columns, `if_all()`
 #' is `TRUE` when the predicate is `TRUE` for *all* selected columns.
 #'
+#' **Important Behavior:** When there are no selected columns:
+#'
+#' - `if_any()` will return `FALSE`, consistent with the behavior of
+#'   `any()` when called without inputs.
+#' - `if_all()` will return `TRUE`, consistent with the behavior of
+#'   `all()` when called without inputs.
```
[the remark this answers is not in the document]
jrwinget: Sounds good!

R/across.R
```diff
@@ -4,14 +4,21 @@
...
+#'
 #' If you just need to select columns without applying a transformation to each
 #' of them, then you probably want to use [pick()] instead.
```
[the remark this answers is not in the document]
jrwinget: Yes, absolutely!

R/across.R
jrwinget: Got it! I assumed they would both go together since they were both mentioned within the same issue. Thanks for the feedback!

R/across.R
jrwinget: That makes sense. Thanks!

tests/testthat/test-across.R
jrwinget: No problem - will update that!

tests/testthat/test-across.R
jrwinget: Thanks for the feedback!

tests/testthat/test-across.R
jrwinget: @DavisVaughan I've made the requested changes and pushed the updates. Please take a look when you have a moment. Thanks!!

DavisVaughan:
Thanks @jrwinget!

jrwinget:
Of course - this was fun! Thanks again for your feedback!!

PR #7073, opened 2024-08-15
Add a ptype argument to between Fixes #6906

### LINKED ISSUE #6906: Add a `ptype` argument to `between()` since we can't make it cast to the type of `x` at this time
The result of using `between()` in filtering is not equivalent to evaluating a conditional expression when working with ordered factors.

```{r}
> (x <- data.frame(Visit = c("Discharge", "Month 3", "Month 6", "Month 12", "Month 20")))
      Visit
1 Discharge
2   Month 3
3   Month 6
4  Month 12
5  Month 20

> class(x$Visit)
[1] "character"
```

Now let's make it an ordered factor with assumed levels (used later for other operations):
```{r}
> (x$Visit <- factor(x$Visit, levels = c("Baseline", "Discharge", "Month 3", "Month 6", "Month 12", "Month 20", "Premature"), ordered = TRUE))
[1] Discharge Month 3   Month 6   Month 12  Month 20 
Levels: Baseline < Discharge < Month 3 < Month 6 < Month 12 < Month 20 < Premature
```

When filtering using the conjunction of conditions it works
```{r}
> x %>% filter(Visit >= "Discharge" & Visit <= "Month 20")
      Visit
1 Discharge
2   Month 3
3   Month 6
4  Month 12
5  Month 20
```
When using the convenient between(), it doesn't
```{r}
> x %>% filter(between(Visit, "Discharge", "Month 20"))
      Visit
1 Discharge
2  Month 12
3  Month 20
```
2 items are missing in the output.

-----
R version 4.2.0 (2022-04-22 ucrt)
Running under: Windows 10 x64 (build 19045)

> packageVersion("dplyr")
[1] ‘1.1.1’

JamesHWade:
Add `ptype` argument to `between()` function

This PR introduces a new `ptype` argument to the `between()` function, allowing users to specify the desired output type for comparisons. This enhancement is particularly useful when dealing with different data types or when specific type casting is required before comparison.

Key changes:
1. Added `ptype` parameter to `between()` function
2. Updated function logic to use `ptype` for casting when provided
3. Added new tests to cover `ptype` functionality
4. Updated documentation to explain the new argument


NEWS.md
```diff
@@ -1,5 +1,9 @@
 # dplyr (development version)
 
+* `between()` gains a new `ptype` argument, allowing users to specify the desired
+  output type. This is particularly useful for ordered factors and other complex
+  types where the default common type behavior might not be ideal (#6906).
```
DavisVaughan, line 5:
```suggestion
  types where the default common type behavior might not be ideal (#6906, @JamesHWade).
```

R/funs.R
```diff
@@ -27,16 +30,27 @@
...
-  # Common type of all inputs
-  args <- vec_cast_common(!!!args)
-  x <- args$x
-  args$x <- NULL
+#'
+#' # Using ptype argument with factors
+#' x <- factor(c("a", "b", "c", "d"), levels = c("a", "b", "c", "d"), ordered = TRUE)
+#' between(x, "b", "c", ptype = x)
+between <- function(x, left, right, ptype = NULL) {
+  if (is.null(ptype)) {
```
krlmlr, line 38: The special-casing here appears to work, but could we also just do:

```r
args <- vec_cast_common(!!!args, .to = ptype)
```

with most of the original code unchanged?

krlmlr (commented):
Nice, thanks!

R/funs.R
```diff
@@ -27,25 +30,36 @@
...
-  x <- args$x
-  args$x <- NULL
-
-  # But recycle to size of `x`
-  args <- vec_recycle_common(!!!args, .size = vec_size(x))
+#'
+#' # Using ptype argument with factors
+#' x <- factor(c("a", "b", "c", "d"), levels = c("a", "b", "c", "d"), ordered = TRUE)
+#' between(x, "b", "c", ptype = x)
+between <- function(x, left, right, ptype = NULL) {
```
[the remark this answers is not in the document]
DavisVaughan: I agree with @krlmlr here. If the user decides to use `ptype`, it is nice to force it to be a named argument, and this is a good way to do that. You should add

```
@inheritParams rlang::args_dots_empty
```

above

```
@param x A vector
```

and then redocument

R/funs.R
```diff
@@ -27,25 +30,36 @@
...
+    # Ensure x is also of type ptype
+    x <- vec_cast(x, ptype)
+  }
+  
+  # Recycle to size of `x`
+  args <- vec_recycle_common(left = left, right = right, .size = vec_size(x))
   left <- args$left
   right <- args$right
-
+
```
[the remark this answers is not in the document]
JamesHWade: Not intentionally. I'm using Positron, which is still new to me.

R/funs.R
DavisVaughan: Agree with @krlmlr that it could just be

```
  args <- list(x = x, left = left, right = right)

  # Common type of all inputs
  args <- vec_cast_common(!!!args, .to = ptype) # this is the only change from the original code
  x <- args$x
  args$x <- NULL
```

R/funs.R
```diff
@@ -27,16 +29,27 @@
...
+    left <- vec_cast(left, ptype)
+    right <- vec_cast(right, ptype)
+    # Ensure x is also of type ptype
+    x <- vec_cast(x, ptype)
+  }
 
-  # But recycle to size of `x`
-  args <- vec_recycle_common(!!!args, .size = vec_size(x))
+  # Recycle to size of `x`
+  args <- vec_recycle_common(left = left, right = right, .size = vec_size(x))
```
DavisVaughan, line 52: If you do what we suggest above, then you can revert this part

tests/testthat/test-funs.R
```diff
@@ -74,6 +74,77 @@ test_that("recycles `left` and `right` to the size of `x`", {
...
+test_that("ptype argument works as expected", {
+  x <- factor(c("a", "b", "c", "d"), levels = c("a", "b", "c", "d"), ordered = TRUE)
+  
+  # Test with ptype specified
+  expect_identical(
+    between(x, "b", "c", ptype = x),
+    c(FALSE, TRUE, TRUE, FALSE)
+  )
+  
+  # Test without ptype (actual behavior)
```
DavisVaughan, line 86:
```suggestion
  # Test without ptype (common type is character)
```

tests/testthat/test-funs.R
```diff
@@ -74,6 +74,77 @@ test_that("recycles `left` and `right` to the size of `x`", {
...
+
+test_that("ptype argument affects type casting", {
+  x <- 1:5
+  expect_identical(
+    between(x, 1.5, 3.5),
+    c(FALSE, TRUE, TRUE, FALSE, FALSE)
+  )
+  expect_error(
+    between(x, 1.5, 3.5, ptype = integer())
+  )
```
DavisVaughan, line 101: Please use `expect_snapshot(error = TRUE)` here, it helps us see what the actual error message is. When you switch to this, run the tests again and it will generate a new snapshot file with the results

tests/testthat/test-funs.R
```diff
@@ -74,6 +74,77 @@ test_that("recycles `left` and `right` to the size of `x`", {
...
+})
+
+test_that("ptype argument works with different types", {
+  # Test with numeric
+  x_num <- c(1, 2, 3, 4, 5)
+  expect_identical(
+    between(x_num, 2, 4, ptype = integer()),
+    c(FALSE, TRUE, TRUE, TRUE, FALSE)
+  )
+})
```
DavisVaughan, line 111: I think you can remove this test, I think the one above that shows that casting from double to integer fails proves that casting is occurring.

tests/testthat/test-funs.R
```diff
@@ -74,6 +74,77 @@ test_that("recycles `left` and `right` to the size of `x`", {
...
+  expect_error(
+    between(x, 2, 4, ptype = "not a valid ptype"),
+    "Can't convert"
+  )
+  
+  expect_error(
+    between(x, "2", "4", ptype = integer()),
+    "Can't convert"
+  )
+})
```
DavisVaughan, line 125: I think you can remove this test. Pretty much anything is a valid ptype. The error you are testing here is that the inputs can't be cast to the `ptype` type, which is already being tested in the double -> integer test from above

tests/testthat/test-funs.R
```diff
@@ -74,6 +74,77 @@ test_that("recycles `left` and `right` to the size of `x`", {
...
+test_that("ptype argument works with data frames", {
+  x <- tibble(a = 1:3, b = 4:6)
+  left <- tibble(a = c(0, 1, 2), b = c(3, 4, 5))
+  right <- tibble(a = c(2, 3, 4), b = c(6, 7, 8))
+  
+  expect_identical(
+    between(x, left, right, ptype = x),
+    c(TRUE, TRUE, TRUE)
+  )
+})
```
DavisVaughan, line 136: This is a nice idea but I think we are basically testing vctrs capabilities at this point, rather than `between()` capabilities, so I think we can remove this

tests/testthat/test-funs.R
```diff
@@ -74,6 +74,77 @@ test_that("recycles `left` and `right` to the size of `x`", {
   })
 })
 
+test_that("ptype argument works as expected", {
+  x <- factor(c("a", "b", "c", "d"), levels = c("a", "b", "c", "d"), ordered = TRUE)
+  
+  # Test with ptype specified
+  expect_identical(
+    between(x, "b", "c", ptype = x),
+    c(FALSE, TRUE, TRUE, FALSE)
```
DavisVaughan, line 83: Can you please find an example where casting to the ordered factor results in a value that is _different_ from not supplying a `ptype` at all? i.e. right now both results return `c(FALSE, TRUE, TRUE, FALSE)`, so it isn't super clear that the casting is useful

I think you could do that by messing with the order of the levels in `levels = c("a", "b", "c", "d")`, i.e. as long as they aren't in alphabetical order then you'll probably get a different result when you use the casting

DavisVaughan (commented):
Looking good! I first made a few tweaks of my own before leaving these comments, so make sure to `git pull` first before continuing your work to address these comments!

R/funs.R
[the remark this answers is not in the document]
DavisVaughan: I think you have to opt in to `"files.trimTrailingWhitespace": true,` in Positron settings, which we should probably considering automatically turning on in R files at least

I went ahead and fixed that up

R/funs.R
JamesHWade: I'm getting an error in my test when I try to revert.

JamesHWade: Error (test-funs.R:82:3): ptype argument works as expected with non-alphabetical ordered factors
Error in `vec_recycle_common(!!!args, .size = vec_size(x))`: Can't splice an object of type <closure> because it is not a vector.
Backtrace:
    ▆
 1. ├─testthat::expect_identical(...) at test-funs.R:82:3
 2. │ └─testthat::quasi_label(enquo(object), label, arg = "object")
 3. │   └─rlang::eval_bare(expr, quo_get_env(quo))
 4. ├─dplyr::between(x, "c", "a", ptype = x)
 5. │ └─vctrs::vec_recycle_common(!!!args, .size = vec_size(x)) at dplyr/R/funs.R:55:3
 6. └─rlang::abort(message = message)

PR #7080, opened 2024-09-02
Add group_split to the pkgdown reference page

MikeJohnPage:
The reference page for the pkgdown site is missing the experimental function `group_split()`. I presume this should be featured on the reference page as it exported by the package?


MikeJohnPage:
Closed as I have just observed somebody else has already submitted a fix.

PR #7676, opened 2025-04-14
Use {palmerpenguins} dataset in arrange() examples

KittJonathan:
I added examples using the penguins dataset in the arrange() documentation 
+ added {palmerpenguins} to suggests in the dplyr DESCRIPTION file

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -25,3 +25,4 @@ doc
 Meta
 compile_commands.json
 .cache
+*.Rproj
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -56,7 +56,8 @@ Suggests:
     stringi (>= 1.7.6),
     testthat (>= 3.1.5),
     tidyr (>= 1.3.0),
-    withr
+    withr,
+    palmerpenguins
 VignetteBuilder:
     knitr
 Config/Needs/website: tidyverse, shiny, pkgdown, tidyverse/tidytemplate
--- a/R/arrange.R
+++ b/R/arrange.R
@@ -59,14 +59,24 @@
 #'   help page under the `Default locale` section.
 #' @family single table verbs
 #' @examples
-#' arrange(mtcars, cyl, disp)
-#' arrange(mtcars, desc(disp))
+#' # Load the {palmerpenguins} package (R < 4.5.0)
+#' library(palmerpenguins)
 #'
-#' # grouped arrange ignores groups
-#' by_cyl <- mtcars %>% group_by(cyl)
-#' by_cyl %>% arrange(desc(wt))
+#' # Arrange by increasing order of bill depth
+#' arrange(penguins, bill_depth_mm)
+#'
+#' # Arrange by increasing order of bill depth and body mass
+#' arrange(penguins, bill_depth_mm, body_mass_g)
+#'
+#' # Arrange by increasing order of bill depth and decreasing order of body mass
+#' arrange(penguins, bill_depth_mm, desc(body_mass_g))
+#'
+#' # Grouped arrange ignores groups
+#' by_species <- penguins %>% group_by(species)
+#' by_species %>% arrange(desc(flipper_length_mm))
+
 #' # Unless you specifically ask:
-#' by_cyl %>% arrange(desc(wt), .by_group = TRUE)
+#' by_species %>% arrange(desc(flipper_length_mm), .by_group = TRUE)
 #'
 #' # use embracing when wrapping in a function;
 #' # see ?rlang::args_data_masking for more details
@@ -74,11 +84,12 @@
 #'   .data %>%
 #'     arrange({{ var }})
 #' }
-#' tidy_eval_arrange(mtcars, mpg)
+#' tidy_eval_arrange(penguins, bill_length_mm)
 #'
 #' # Use `across()` or `pick()` to select columns with tidy-select
-#' iris %>% arrange(pick(starts_with("Sepal")))
-#' iris %>% arrange(across(starts_with("Sepal"), desc))
+#' penguins %>% arrange(pick(contains("depth")))
+#' penguins %>% arrange(across(ends_with("mm"), desc))
+#'
 arrange <- function(.data, ..., .by_group = FALSE) {
   UseMethod("arrange")
 }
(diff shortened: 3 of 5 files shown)


DavisVaughan:
We really appreciate the PR! I do not currently see any reason to switch away from mtcars to penguins in this example. The story is a little complicated since R >=4.5 already has `penguins` built in, meaning we shadow it by loading this package and users would get a possibly confusing message

Screenshot 2025-04-14 at 8 46 36 AM

I don't think switching to penguins right now is worth the mental overhead of having to remember that there are now 2 datasets with the same name. Also, the column names of both are unfortunately different, making it even more confusing.

I do however like your additional inline comments, and how you added an extra example of arranging by just 1 column. I would accept a PR that kept mtcars but added those inline comments and the extra example, if you'd like to do that.

yutannihilation:
> I do not currently see any reason to switch away from mtcars to penguins in this example

A bit drifted topic: On the other hand, I believe there are reasons for dplyr (and other tidyverse packages) to switch from **`iris`** to `penguins` because I believe that's the very reason `penguins` was added.

ggplot2 removed `iris` example code [5 years ago](https://github.com/tidyverse/ggplot2/pull/4077). I was hoping other packages would follow, but it had not happen so far. I think it's a good time to consider migrating to `penguins`.

That said, I have no idea how the replacement should be done. [basepenguins](https://ellakaye.github.io/basepenguins/) might help, but I'm not fully sure if it won't make another confusion.

KittJonathan:
I get your point @DavisVaughan. I can indeed work on using mtcars. But I personally do think it would be nice having examples using palmerpenguins across as many tidyverse functions as possible, given the popularity of the dataset in tutorials and courses.
@hadley clarified by email the tidyverse version policy, supporting the current and previous four versions of R. 
Would it be worth while considering switching to palmerpenguins (or basepenguins) examples when four versions of R have been released?

hadley:
I think this change is worth it, but you need to (a) put palmerpenguins in imports, not suggests (since otherwise you need to protect all the examples) and (b) load the data using `data()` instead of by attaching the package. I'd also suggest keeping the existing pipe based examples primary, since I don't think we want to get into that discussion now.

I can't believe that R core changed the variable names, making the whole substitution more complicated.

R/arrange.R
```diff
@@ -59,26 +59,37 @@
...
-#' by_cyl <- mtcars %>% group_by(cyl)
-#' by_cyl %>% arrange(desc(wt))
+#' # Arrange by increasing order of bill depth
+#' arrange(penguins, bill_depth_mm)
+#'
+#' # Arrange by increasing order of bill depth and body mass
+#' arrange(penguins, bill_depth_mm, body_mass_g)
+#'
+#' # Arrange by increasing order of bill depth and decreasing order of body mass
+#' arrange(penguins, bill_depth_mm, desc(body_mass_g))
```
DavisVaughan, line 72:
```suggestion
#' penguins %>% arrange(bill_depth_mm)
#'
#' # Arrange by increasing order of bill depth and body mass
#' penguins %>% arrange(bill_depth_mm, body_mass_g)
#'
#' # Arrange by increasing order of bill depth and decreasing order of body mass
#' penguins %>% arrange(bill_depth_mm, desc(body_mass_g))
```

This is what Hadley means by

> I'd also suggest keeping the existing pipe based examples primary, since I don't think we want to get into that discussion now.

R/arrange.R
```diff
@@ -59,8 +59,8 @@
 #'   help page under the `Default locale` section.
 #' @family single table verbs
 #' @examples
-#' # Load the {palmerpenguins} package (R < 4.5.0)
-#' library(palmerpenguins)
+#' # Load the 'penguins' data set from {palmerpenguins} package (R < 4.5.0)
+#' data("penguins")
```
hadley, line 63:
```suggestion
#' data("penguins", package = "palmerpenguins")
```

No need for a comment if the is more clear 😄

DESCRIPTION
```diff
@@ -34,7 +34,8 @@ Imports:
     tibble (>= 3.2.0),
     tidyselect (>= 1.2.0),
     utils,
-    vctrs (>= 0.6.4)
+    vctrs (>= 0.6.4),
+    palmerpenguins
```
hadley, line 38: Can you please preserve the alphabetical ordering?

DESCRIPTION
```diff
@@ -56,7 +57,7 @@ Suggests:
     stringi (>= 1.7.6),
     testthat (>= 3.1.5),
     tidyr (>= 1.3.0),
-    withr
+    withr,
```
hadley, line 60:
```suggestion
    withr
```

R/arrange.R
```diff
@@ -59,26 +59,37 @@
...
 #' @examples
-#' arrange(mtcars, cyl, disp)
-#' arrange(mtcars, desc(disp))
+#' # Load the 'penguins' data set from {palmerpenguins} package (R < 4.5.0)
+#' data("penguins")
 #'
-#' # grouped arrange ignores groups
-#' by_cyl <- mtcars %>% group_by(cyl)
-#' by_cyl %>% arrange(desc(wt))
+#' # Arrange by increasing order of bill depth
```
hadley, line 65: I'm not sure we need one comment per line like this. I think it's fine in isolation, but it would be a really big change across all of the dplyr docs, and you'd need to introduce some new ways to break the docs up in to chunks.

.gitignore
```diff
@@ -25,3 +25,4 @@ doc
 Meta
 compile_commands.json
 .cache
+*.Rproj
```
hadley, line 28: It would be great if you could minimise unrelated changes in PRs. That makes it easier for us to review.

hadley (commented):
Thanks for continuing to work on this! It's really important to get the details 100% correct in this prototype, so I appreciate you bearing with us as we deliver very detailed feedback.

KittJonathan:
> Thanks for continuing to work on this! It's really important to get the details 100% correct in this prototype, so I appreciate you bearing with us as we deliver very detailed feedback.

Thanks for your help, I appreciate you taking the time to give very detailed feedback :-)

KittJonathan:
Hello there, I hadn't checked this issue since a while, and saw the faliling checks. If the idea of changing the dataset use cases to penguins is still useful, I'll try to sort it out.

PR #7688, opened 2025-06-05
remove defunct functions from prior to 1.0

olivroy:
I just had an issue with gt + a tidy evaluation of a column named id and that directed me to a defunct error. I saw that this has been deprecated for a while, so might be time to just remove those.

Just needs news + `devtools::document()`. I don't mind doing it if you agree with the PR!


olivroy:
Done in #7762

PR #7715, opened 2025-09-19
Add example for slice_sample

### LINKED ISSUE #7707: Feature request: `shuffle_rows()` to shuffle rows in a tibble
## Feature request

Add a function `shuffle_rows()` to randomly reorder rows in the current tibble/slice:

```R
dataset_tbl |>
  dplyr::shuffle_rows()
```

## Motivation

Currently, shuffling all rows in a slice/tibble requires 
`dataset_tbl |> dplyr::slice_sample(prop = 1)` or similar variations.

This approach has two main problems:

1. Unclear intent... because it is not obvious whether the goal is to shuffle rows or sample rows.
2. Fragile code... because small changes to the arguments passed can silently alter the result:
   - if `prop` is changed from `1`, the output is no longer a complete shuffle;
   - if `replace = TRUE` is used (accidentally or via copy-paste), rows may repeat or disappear.

Introducing `shuffle_rows()` would make the goal clear and make code more robust.

## Implementation sketch

```R
shuffle_rows <- function(.data, ..., .by = NULL) {
  dplyr::slice(
    .data,
    .by = {{ .by }},
    base::local({
      number_of_rows <- dplyr::n()
      sample.int(n = number_of_rows, size = number_of_rows, replace = FALSE)
    })
  )
}
```

Hzanib:
Fixes #7707

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,7 @@
 # dplyr (development version)
 
+* New `slice_sample()` example showing how to use it to shuffle rows (#7707, @Hzanib).
+
 * Updated `across()` examples to include an example using `everything()` (#7621, @JBrandenburg02).
 
 * Clarified how `slice_min()` and `slice_max()` work in the introduction vignette (#7717, @ccani007).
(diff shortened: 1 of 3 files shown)


DavisVaughan:
Thanks!

PR #7716, opened 2025-09-19
Updating the examples for `across()` to include an example using `everything()`

### LINKED ISSUE #7621: More examples in `across()` documentation
When I use `across()` I almost always want to do the thing to every single column. For whatever reason, that tidyselect action is hard for me to remember/find. Is it `all()`? Is it `everything()`? Etc. It would be great to have an example in the `across()` documentation that does the thing to every column. Additionally, it would be awesome to have an example that does the thing conditionally depending on the variable type. For example, rounding all the numeric variables.

JBrandenburg02:
Fixes #7621


DavisVaughan:
Thanks!

PR #7717, opened 2025-09-19
Clarified slice_max() and slice_min() in intro vignette # 7671

ccani007:
Closes # 7671
- Clarified wording for `slice_max()` and `slice_min()` in the introduction vignette
  (explicit default `n = 1`, clearer phrasing on NA handling).
- Added a space after the `slice_sample()` example code chunk to fix issues
  affecting subsequent chunks.

### DIFF
--- a/vignettes/dplyr.Rmd
+++ b/vignettes/dplyr.Rmd
@@ -124,14 +124,15 @@ starwars |> slice_head(n = 3)
 starwars |> slice_sample(n = 5)
 starwars |> slice_sample(prop = 0.1)
 ```
+
 Use `replace = TRUE` to perform a bootstrap sample. If needed, you can weight the sample with the `weight` argument.
 
-* `slice_min()` and `slice_max()` select rows with highest or lowest values of a variable. Note that we first must choose  only the values which are not NA.
+* `slice_max()` and `slice_min()` selects the rows with the largest or smallest values of the selected column, by default it returns the single maximum or minimum, but you can supply `n` to control how many rows remain. Rows with NA in the ordering column are dropped.
 
 ```{r}
 starwars |>
   filter(!is.na(height)) |>
-  slice_max(height, n = 3)
+  slice_max(height)
 ```
 
 ### Select columns with `select()`


vignettes/dplyr.Rmd
```diff
@@ -124,14 +124,15 @@ starwars |> slice_head(n = 3)
...
 Use `replace = TRUE` to perform a bootstrap sample. If needed, you can weight the sample with the `weight` argument.
 
-* `slice_min()` and `slice_max()` select rows with highest or lowest values of a variable. Note that we first must choose  only the values which are not NA.
+* `slice_max()` and `slice_min()` selects the rows with the largest or smallest values of the selected column, by default it returns the single maximum or minimum, but you can supply `n` to control how many rows remain. Rows with NA in the ordering column are dropped.
 
 ```{r}
 starwars |>
   filter(!is.na(height)) |>
-  slice_max(height, n = 3)
+  slice_max(height)
```
DavisVaughan: I actually do think showing `n = 3` is useful here to show how to control the number of rows returned

vignettes/dplyr.Rmd
```diff
@@ -124,14 +124,15 @@ starwars |> slice_head(n = 3)
 starwars |> slice_sample(n = 5)
 starwars |> slice_sample(prop = 0.1)
 ```
+
 Use `replace = TRUE` to perform a bootstrap sample. If needed, you can weight the sample with the `weight` argument.
 
-* `slice_min()` and `slice_max()` select rows with highest or lowest values of a variable. Note that we first must choose  only the values which are not NA.
+* `slice_max()` and `slice_min()` selects the rows with the largest or smallest values of the selected column, by default it returns the single maximum or minimum, but you can supply `n` to control how many rows remain. Rows with NA in the ordering column are dropped.
```
DavisVaughan:
```suggestion
* `slice_min()` and `slice_max()` select the rows with the smallest or largest values of the selected column. By default, they return a single minimum or maximum, but you can supply `n` to control how many rows remain.
```

I do think putting minimum first, then maximum second, makes a little more sense in my head (something about interval notation being `[minimum, maximum]`, i.e. the minimum is first).

For this simple overview example in an introduction vignette, I actually don't think we need to mention `NA` behavior at all

vignettes/dplyr.Rmd
DavisVaughan: In fact, I think you could just reduce the whole example down to this, since `NA` values don't actually matter here at all

```r
starwars |> slice_max(height, n = 3)
```

vignettes/dplyr.Rmd
DavisVaughan: (Try using the `Commit suggestion` button there in the GitHub UI. After committing my suggestion, don't forget to pull (i.e. `git pull`) the changes locally if you plan to make any more updates.)

vignettes/dplyr.Rmd
```diff
@@ -124,14 +124,15 @@ starwars |> slice_head(n = 3)
...
 Use `replace = TRUE` to perform a bootstrap sample. If needed, you can weight the sample with the `weight` argument.
 
-* `slice_min()` and `slice_max()` select rows with highest or lowest values of a variable. Note that we first must choose  only the values which are not NA.
+* `slice_min()` and `slice_max()` select the rows with the smallest or largest values of the selected column. By default, they return a single minimum or maximum, but you can supply `n` to control how many rows remain.
 
 ```{r}
 starwars |>
   filter(!is.na(height)) |>
-  slice_max(height, n = 3)
+  slice_max(height)
```
DavisVaughan: @ccani007 something still doesn't look right here. I'm expecting this example to be changed to

```r
starwars |> slice_max(height, n = 3)
```

DavisVaughan:
Thanks!

PR #7718, opened 2025-09-19
Add rename feature of dplyr select to docs

### LINKED ISSUE #7706: `select` renames columns if vector of column names is named
When I use a named vector to `select` columns, where the column name is specified by a named element in the vector, those columns are renamed. For example:
```
> df1 <- tibble::tibble(a = 1:4, b = letters[11:14])
> df1
# A tibble: 4 × 2
      a b    
  <int> <chr>
1     1 k    
2     2 l    
3     3 m    
4     4 n    
> col_sel <- c("a", b_but_renamed = "b")
> col_sel
              b_but_renamed 
          "a"           "b" 
> df2 <- df1 %>% dplyr::select(tidyselect::all_of(col_sel))
> df2
# A tibble: 4 × 2
      a b_but_renamed
  <int> <chr>        
1     1 k            
2     2 l            
3     3 m            
4     4 n            
```
Is this behavior expected? If so, I can't find it in the documentation (though this may be due to my failure to find it). If this behavior is expected, is there an argument to `dplyr::select` that would allow me to disable this feature, or do I need to always call `unname` on `col_sel` to ensure this doesn't happen? Thanks!

juliaapolonio:
Fixes #7706

### DIFF
--- a/man/rmd/select.Rmd
+++ b/man/rmd/select.Rmd
@@ -34,6 +34,18 @@ starwars |> select(homeworld, height, mass)
 iris |> select(Sepal.Length, Petal.Length)
 ```
 
+If you use a named vector to select columns, the output will have
+its columns renamed:
+
+```{r}
+selection <- c(
+  new_homeworld = "homeworld",
+  new_height = "height",
+  new_mass = "mass"
+)
+starwars |> select(all_of(selection))
+```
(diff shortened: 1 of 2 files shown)


DavisVaughan:
Thanks!

PR #7723, opened 2025-09-30
Switch `if_else()` to `vec_if_else()`

DavisVaughan:
Branched from #7722 

See https://github.com/r-lib/vctrs/pull/2030 for benchmarks!

One thing of note is that `vec_if_else()` disallows arrays for `condition`. I am not yet sure if we are going to have to provide a compat layer for 1D arrays in `if_else()` itself. We will have to run revdeps, and then add a test on the dplyr side if we need this. `"c6431dd7-a51a-4fe5-81a9-b889d0a1510a"`

### DIFF
--- a/tests/testthat/_snaps/if-else.md
+++ b/tests/testthat/_snaps/if-else.md
@@ -22,6 +22,14 @@
       Error in `if_else()`:
       ! `condition` must be a logical vector, not an integer vector.
 
+# `condition` can't be an array (#7723)
+
+    Code
+      if_else(array(TRUE), 1, 2)
+    Condition
+      Error in `if_else()`:
+      ! `condition` must be a logical vector, not a logical vector.
+
 # `true`, `false`, and `missing` must recycle to the size of `condition`
 
     Code
--- a/tests/testthat/_snaps/if-else.md
+++ b/tests/testthat/_snaps/if-else.md
@@ -12,7 +12,7 @@
       if_else(TRUE, 1, 2, missing = "x")
     Condition
       Error in `if_else()`:
-      ! Can't combine `true` <double> and `missing` <character>.
+      ! Can't combine `missing` <character> and <double>.
 
 # `condition` must be logical (and isn't cast to logical!)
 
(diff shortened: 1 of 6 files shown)


tests/testthat/_snaps/if-else.md
```diff
@@ -12,7 +12,7 @@
       if_else(TRUE, 1, 2, missing = "x")
     Condition
       Error in `if_else()`:
-      ! Can't combine `true` <double> and `missing` <character>.
+      ! Can't combine `missing` <character> and <double>.
```
hadley: Why do we lose the true label here?

hadley (approved):

tests/testthat/_snaps/if-else.md
DavisVaughan: Investigating this helped me uncover two vctrs bugs
https://github.com/r-lib/vctrs/pull/2048

---

TLDR is that I think I can fix this to get the old message back.

We do the ptype computation in two steps. First with `true` and `false` together to get their `ptype`, then we layer in the common type of that `ptype` and `default`.
https://github.com/r-lib/vctrs/blob/2e9e9109705aab0daa9c201c7d6059d5ec5a5e95/src/if-else.c#L549-L576

(we do it in two steps rather than using ptype-common so we can handle `true_arg`, `false_arg`, and `default_arg` lazily and not materialize them unless there is an error)

I wasn't sure what to use as the arg name that went along with `ptype` in that second ptype2 computation - does it go with `true_arg` or `false_arg`? But actually the `left` indicator there will tell me that, so I'll use that. 

---

I was also a bit worried about the case where the `ptype` is some _union of_ `true` and `false`, and doesn't exactly match either of them. But even `vec_ptype_common()` suffers from this subtle issue, so I'm not going to worry about that.

``` r
# Note how `x` has shape `[,2,1]` but in the error it is reported as `[,2,2]`
# because that is the common shape of `x` and `y` (neither has this shape on
# their own, so neither `x` nor `y` are technically correct to report here)
vec_ptype_common(
  x = array(character(), dim = c(0, 2, 1)),
  y = array(character(), dim = c(0, 1, 2)),
  z = 1
)
#> Error:
#> ! Can't combine `x` <character[,2,2]> and `z` <double>.
```

DavisVaughan: Fixed 🫡

DarwinAwardWinner:
For what it's worth, I just tripped over the 1D array case in my own code.

Issue #7082, opened 2024-09-05
`case_when` returns length-0 option if any option is length-0


antdurrant:
`case_when` always returns length-0 option if any option is length-0

I would expect an error as in `case_match`. This seems to only be an issue if the input to `case_when` is length 1 - it errors otherwise.

``` r
 x <- 5

# silently returns the wrong thing
dplyr::case_when(
    x == 5 ~ "hello",
    x > 10 ~ character(),
    .default = "other"
)
#> character(0)

# errors
dplyr::case_match(
    x,
    5 ~ "hello",
    10 ~ character(),
    .default = "other"
)
#> Error in `dplyr::case_match()`:
#> ! `..2 (right)` must have size 1, not size 0.

packageVersion("dplyr")
#> [1] '1.1.4'

y <- 5:7

# errors out as expected
dplyr::case_when(
    y == 5 ~ "hello",
    y > 10 ~ character(),
    .default = "other"
)
#> Error in `dplyr::case_when()`:
#> ! Can't recycle `..1 (left)` (size 3) to match `..2 (right)` (size 0).

# errors
dplyr::case_match(
    y,
    5 ~ "hello",
    10 ~ character(),
    .default = "other"
)
#> Error in `dplyr::case_match()`:
#> ! `..2 (right)` must have size 3, not size 0.
```

<sup>Created on 2024-09-05 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

dchiu911:
Yes I would have expected evaluation to stop when the first condition that is true has been met, but instead evaluation continues

DavisVaughan:
> Yes I would have expected evaluation to stop when the first condition that is true has been met

That's a separate request. You're asking for _lazy_ semantics here, which we have decided not to do in `case_when()` to ensure that is has size and type stability

DavisVaughan:
First off, just know that I do agree with you here. My ideal version of `case_when()` would follow two rules:
- The LHSs must all have the same `size`. No recycling is done. That `size` of the LHSs determines the output size.
- The RHSs must either have size 1 or size `size`. Size 1 RHSs get recycled to size `size`.

And in fact the lower level internal `dplyr:::vec_case_when()` is implemented this way, and we patch `case_when()` on top of it to be more liberal with the recycling.

[...]

PR #7724, opened 2025-09-30
Soft-deprecate `case_when()` with all size 1 LHS and any size >1 RHS

DavisVaughan:
*This description is outdated but I've left it for historical purposes, see below comments*

Closes https://github.com/tidyverse/dplyr/issues/7082

This is a breaking change.

Motivating example with CRAN dplyr:

``` r
x <- 5

# The user has made a typo here with `character()` on the RHS
#
# Note that all LHS inputs are size 1
#
# The common size from consulting BOTH the LHS and RHS inputs is 0, which
# results in a silent confusing result of `character()`
dplyr::case_when(
  x == 5 ~ "hello",
  x > 10 ~ character(),
  .default = "other"
)
#> character(0)
```

With this PR

``` r
x <- 5

# If you ONLY consult the size of the LHS inputs, then the common size is 1,
# and then all RHS inputs are recycled to this size, which RHS-2 `character()` can't
# do, and we correctly catch your error!
dplyr::case_when(
  x == 5 ~ "hello",
  x > 10 ~ character(),
  .default = "other"
)
#> Error in `dplyr::case_when()`:
#> ! `..2 (right)` must have size 1, not size 0.
```

In https://github.com/tidyverse/dplyr/issues/7082#issuecomment-2334173589 I explained how changing to this behavior increases the safety of `case_when()` (as seen above) and brings us one step closer to my ideal `case_when()` behavior. Unfortunately, having to support `TRUE ~` will keep us from reaching `case_when()` nirvana.

In https://github.com/tidyverse/dplyr/issues/7082#issuecomment-2334181190, I note that this should only affect an extremely small fraction of cases - a prerequisite for this to be a change in behavior is that all of your LHS inputs must have been length 1, which is quite rare. Regardless, I will run revdeps to assess. `"8fa7eaf9-c4e6-4fe9-b7e9-7855b11046a6"`

### DIFF
--- a/R/case-when.R
+++ b/R/case-when.R
@@ -16,10 +16,10 @@
 #'
 #'   The RHS inputs will be coerced to their common type.
 #'
-#'   All inputs will be recycled to their common size. That said, we encourage
-#'   all LHS inputs to be the same size. Recycling is mainly useful for RHS
-#'   inputs, where you might supply a size 1 input that will be recycled to the
-#'   size of the LHS inputs.
+#'   For historical reasons, all LHS inputs will be recycled to their common
+#'   size. That said, we encourage all LHS inputs to be the same size, which you
+#'   can optionally enforce with `.size`. All RHS inputs will be recycled to the
+#'   common size of the LHS inputs.
 #'
 #'   `NULL` inputs are ignored.
 #'
@@ -27,7 +27,7 @@
 #'   `FALSE` or `NA`.
 #'
 #'   `.default` must be size 1 or the same size as the common size computed
-#'   from `...`.
+#'   from the LHS inputs.
 #'
 #'   `.default` participates in the computation of the common type with the RHS
 #'   inputs.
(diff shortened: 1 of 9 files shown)


R/case-when.R
```diff
@@ -45,11 +45,12 @@
...
-#'   this overrides the common size computed from `...`.
+#'   this overrides the common size computed from the LHS inputs.
 #'
-#' @return A vector with the same size as the common size computed from the
-#'   inputs in `...` and the same type as the common type of the RHS inputs
-#'   in `...`.
+#' @return A vector
+#'
+#'   - The size of the vector is the common size of the LHS inputs, or `.size`.
+#'   - The type of the vector is the common type of the RHS inputs, or `.ptype`.
```
DavisVaughan, line 53:
```suggestion
#'   - The size of the vector is the size of the LHS inputs, where all LHS inputs must have the same size.
#'   - The type of the vector is the common type of the RHS inputs, or `.ptype`.
```

This would have been `case_when()` nirvana, but supporting `TRUE ~` prevents us from getting there. `vec_case_when()` has these semantics.

tests/testthat/test-case-when.R
```diff
@@ -269,6 +237,51 @@ test_that("throws chained errors when formula evaluation fails", {
...
+  # Used to return `integer()` (#3041).
+  # If you go back and look at #3041, it was an issue about the case where all
+  # LHS inputs have size 0 and all RHS inputs have size 1. That still works as
+  # expected. This extra test of LHS inputs having size 1 and RHS inputs having
+  # size 0 was not something the user wanted, and does not make sense to us.
+  expect_snapshot(error = TRUE, {
+    case_when(TRUE ~ integer(), FALSE ~ integer())
+  })
+  expect_identical(case_when(logical() ~ 1, logical() ~ 2), numeric())
+})
```
DavisVaughan: It's worth reading this test in full, you can ignore the others, which got absorbed into this one

PR #7726, opened 2025-10-02
Rework `vec_case_match()` on `vctrs::vec_case_when()`

DavisVaughan:
This effectively freezes `case_match()`'s implementation in preparation for it to be superseded by `recode_values()`.

This removes the final usage of `dplyr:::vec_case_when()` outside of `case_when()` itself, which will come next, allowing us to delete that dplyr helper entirely.

### DIFF
--- a/R/case-match.R
+++ b/R/case-match.R
@@ -168,6 +168,12 @@ vec_case_match <- function(
   obj_check_list(haystacks, arg = haystacks_arg, call = call)
   list_check_all_vectors(haystacks, arg = haystacks_arg, call = call)
 
+  if (length(haystacks) == 0L) {
+    # `case_match()` is like `case_when()` and doesn't allow empty `...`,
+    # even though `vec_case_when()` is well defined for this case.
+    abort("At least one condition must be supplied.", call = call)
+  }
+
   haystacks <- vec_cast_common(
     !!!haystacks,
     .to = needles,
@@ -175,19 +181,19 @@ vec_case_match <- function(
     .call = call
   )
 
-  conditions <- map(haystacks, vec_in, needles = needles)
+  cases <- map(haystacks, vec_in, needles = needles)
 
   size <- vec_size(needles)
 
-  vec_case_when(
-    conditions = conditions,
+  vctrs::vec_case_when(
+    cases = cases,
     values = values,
-    conditions_arg = "",
-    values_arg = values_arg,
     default = default,
-    default_arg = default_arg,
     ptype = ptype,
     size = size,
-    call = call
+    cases_arg = "",
+    values_arg = values_arg,
+    default_arg = default_arg,
+    error_call = call
   )
 }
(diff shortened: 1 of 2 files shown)


R/case-match.R
```diff
@@ -168,26 +168,32 @@ vec_case_match <- function(
   obj_check_list(haystacks, arg = haystacks_arg, call = call)
   list_check_all_vectors(haystacks, arg = haystacks_arg, call = call)
 
+  if (length(haystacks) == 0L) {
+    # `case_match()` is like `case_when()` and doesn't allow empty `...`,
+    # even though `vec_case_when()` is well defined for this case.
+    abort("At least one condition must be supplied.", call = call)
+  }
```
DavisVaughan, line 175: Unlike `dplyr:::vec_case_when()`, `vctrs::vec_case_when()` is well defined when you don't give it any inputs. It's useful for programmatic reasons.

But `case_when()` and `case_match()` don't support this since we don't want to expose the unspecified type to normal users.

So I've recreated the `dplyr:::vec_case_when()` error here.

``` r
library(vctrs)

vec_case_when(
  cases = list(),
  values = list()
)
#> <unspecified> [0]

vec_case_when(
  cases = list(),
  values = list(),
  size = 2
)
#> <unspecified> [2]

vec_case_when(
  cases = list(),
  values = list(),
  default = 1L
)
#> integer(0)

vec_case_when(
  cases = list(),
  values = list(),
  ptype = integer()
)
#> integer(0)
```

tests/testthat/_snaps/case-match.md
```diff
@@ -28,7 +28,7 @@
       case_match(1, 1 ~ 1L, .default = "x")
     Condition
       Error in `case_match()`:
-      ! Can't combine `..1 (right)` <integer> and `.default` <character>.
+      ! Can't combine <integer> and `.default` <character>.
```
DavisVaughan, line 31: Unlike with `if_else()`, we are going to have to accept this new error message where you don't have the label.

Essentially we need the common type of `list_of_rhs` and `default`, which we do at C level via

```r
ptype <- vec_ptype_common(!!!list_of_rhs)
ptype <- vec_ptype2(ptype, default, y_arg = default_arg)
```

https://github.com/r-lib/vctrs/blob/c46809139a63699bde2f1b4b58286553c2f2075e/src/list-combine.c#L1769-L1823

There is no easy way to smush `default` (and `default_arg`) into the `list_of_rhs` at the C level compute the common ptype all in one shot. This computation actually takes place in `list_combine()`, where the `list_of_rhs` could be a very large list of objects we want to combine, so allocating a list of `size + 1` just to insert `default` isn't worth it.

tests/testthat/_snaps/case-match.md
```diff
@@ -124,7 +124,7 @@
       vec_case_match(1, haystacks = list(1), values = list(1:2))
     Condition
       Error in `vec_case_match()`:
-      ! `values[[1]]` must have size 1, not size 2.
+      ! Can't recycle `values[[1]]` (size 2) to size 1.
```
DavisVaughan, line 127: I think this is actually more correct. We are recycling each `value` here to the size of `x`.

PR #7728, opened 2025-10-04
Implement `replace_when()`

DavisVaughan:
Part of https://github.com/tidyverse/tidyups/pull/29
Closes https://github.com/tidyverse/dplyr/issues/4050 (our second oldest issue!)


NEWS.md
```diff
@@ -1,5 +1,16 @@
...
+* `case_when()` is now part of a family of 4 related functions, 3 of which are new:
+
+  * Use `case_when()` to create a new vector based on logical conditions.
+  * Use `replace_when()` to update an existing vector based on logical conditions.
+  * Use `recode_values()` to create a new vector by mapping old values to new values.
+  * Use `replace_values()` to update an existing vector by mapping old values to new values.
+
+  We are particularly excited about `recode_values()`, which allows you to easily incorporate a lookup table, and serves as a more holistic replacement for both `case_match()` and `recode()`.
+
+  This work is a result of [Tidyup 7: Recoding and replacing values in the tidyverse](https://github.com/tidyverse/tidyups/blob/main/007-tidyverse-recoding-and-replacing.md), with a lot of great [feedback](https://github.com/tidyverse/tidyups/pull/29) from the community.
```
DavisVaughan: Preemptive NEWS bullet for the family of 4, the other 2 are coming up next

R/case-when.R
```diff
@@ -1,36 +1,83 @@
 #' A general vectorised if-else
 #'
 #' @description
```
DavisVaughan, line 3: I have fully refreshed the `case_when()` docs, so it's probably worth looking through them holistically

tests/testthat/test-case-when.R
```diff
@@ -374,3 +376,127 @@ test_that("Using scalar LHS with vector RHS is deprecated (#7082)", {
...
+})
+
+test_that("replace_when() does not recycle LHS values", {
+  # Unlike `case_when()` we get to do this right!
+  x <- c(1, 2, 3)
+
+  expect_snapshot(error = TRUE, {
+    replace_when(x, TRUE ~ 0)
+  })
+})
```
DavisVaughan, line 412: Notable test!

tests/testthat/test-case-when.R
```diff
@@ -374,3 +376,127 @@ test_that("Using scalar LHS with vector RHS is deprecated (#7082)", {
...
+    c(0, 2, 4)
+  )
+})
+
+test_that("replace_when() does not allow named `...`", {
+  # Purposefully stricter than `case_when()`
+  expect_snapshot(error = TRUE, {
+    replace_when(1, foo = TRUE ~ 2)
+  })
+})
```
DavisVaughan, line 467: Notable test!

tests/testthat/test-case-when.R
```diff
@@ -374,3 +376,127 @@ test_that("Using scalar LHS with vector RHS is deprecated (#7082)", {
...
+  )
+})
+
+test_that("replace_when() is a no-op with zero conditions", {
+  # Unlike `case_when()`, where when zero conditions are supplied
+  # we don't know what kind of vector to build (and we refuse to
+  # build an `unspecified` vector, unlike `vec_case_when()`)
+  expect_identical(replace_when(1), 1)
+  expect_identical(replace_when(1, NULL), 1)
+})
```
DavisVaughan, line 482: Notable test! `replace_when(1)` is well defined, while `case_when()` (that exact call, with no inputs) is not

(Well, `case_when()` could technically return `unspecified()`, or incorporate `.ptype`, `.size`, and `.default` if specified, all of which `vec_case_when()` does, but for the average user it's better to error, and historically that is what `case_when()` has always done here)

NEWS.md
```diff
@@ -1,5 +1,18 @@
 # dplyr (development version)
 
+* `case_when()` is now part of a family of 4 related functions, 3 of which are new:
+
+  * Use `case_when()` to create a new vector based on logical conditions.
+  * Use `replace_when()` to update an existing vector based on logical conditions.
+  * Use `recode_values()` to create a new vector by mapping old values to new values.
+  * Use `replace_values()` to update an existing vector by mapping old values to new values.
```
hadley:
```suggestion
  * Use `replace_values()` to update an existing vector by mapping some old values to new values.
```

?

OtherwIse I don't get the difference between "recode" and "replace"

R/case-match.R
```diff
@@ -123,15 +123,14 @@
 #'     .keep = "used"
 #'   )
 case_match <- function(.x, ..., .default = NULL, .ptype = NULL) {
-  args <- list2(...)
+  # Historically it did not allow this, even though in practice it could allow
+  # empty dots to be a no-op, like `replace_when()` and `replace_values()`.
+  allow_empty_dots <- FALSE
```
hadley, line 130: So why forbid it?

R/case-when.R
```diff
@@ -1,36 +1,83 @@
...
-#' the `.default` is used as a final "else" statment.
+#' `case_when()` and `replace_when()` are two forms of vectorized [if_else()].
+#' They work by evaluating each case sequentially and using the first match for
+#' each element to determine the corresponding value in the output vector.
+#'
+#' - Use `case_when()` when creating an entirely new vector.
+#'
+#' - Use `replace_when()` when updating an existing vector.
+#'
+#' If no cases match, then for `case_when()` a `.default` is used as a final
```
hadley:
```suggestion
#' They differ primarily in their behaviour when no cases match, then for `case_when()` a `.default` is used as a final
```

?

R/case-when.R
```diff
@@ -1,36 +1,83 @@
...
+#' "else" statement, and for `replace_when()` the original values from `x` are
+#' retained.
 #'
 #' `case_when()` is an R equivalent of the SQL "searched" `CASE WHEN` statement.
 #'
+#' @section Connection between `case_when()` and `replace_when()`:
```
hadley:
```suggestion
#' # Connection between `case_when()` and `replace_when()`:
```

I think this is preferred style now?

R/case-when.R
```diff
@@ -1,36 +1,83 @@
...
+#'
+#' The following two statements produce identical `result`s:
```
hadley: I'm not sure this section is currently that useful. Could you reduce to a sentence or two and add to the main description?

R/case-when.R
```diff
@@ -1,36 +1,83 @@
...
+#'
 #' @param ... <[`dynamic-dots`][rlang::dyn-dots]> A sequence of two-sided
 #'   formulas. The left hand side (LHS) determines which values match this case.
 #'   The right hand side (RHS) provides the replacement value.
 #'
-#'   The LHS inputs must evaluate to logical vectors.
+#'   For `case_when()`:
 #'
-#'   The RHS inputs will be coerced to their common type.
+#'   - The LHS inputs must be logical vectors.
```
hadley: For both `case_when()` and `replace_when()`, the LHS must be a logical vector. For backward compatibility, `case_when()` recycles scalar values to their common size, but we no longer recommend this. For `replace_when()`, the `LHS` must be the same length as `x`.

---

It feels like overall this documentation gets too into the weeds too quickly.  We should define the size/type guarantees somewhere, but I feel like it might be better in a details section, rather than in the argument docs.

R/case-when.R
```diff
@@ -39,45 +86,57 @@
...
 #'
 #' @param .size An optional size declaring the desired output size. If supplied,
 #'   this overrides the common size computed from the LHS inputs.
 #'
-#' @return A vector
+#' @returns
 #'
-#'   - The size of the vector is the common size of the LHS inputs, or `.size`.
-#'   - The type of the vector is the common type of the RHS inputs, or `.ptype`.
+#'   For `case_when()`, a new vector:
```
hadley: I think you can dedent this section

R/case-when.R
```diff
@@ -39,45 +86,57 @@
...
+#'
+#'   - The names are the same names as `x`, regardless of any replacements,
+#'     consistent with [base::replace()] and \code{base::`[<-`}.
+#'
+#' @name case-and-replace-when
 #'
-#' @export
 #' @examples
 #' x <- 1:70
 #' case_when(
```
hadley, line 81: Feels like you should change these to `replace_when()` to be truer to the spirit of fizzbuzz.

R/case-when.R
hadley: Maybe also worth mentioning factors explicitly?

tests/testthat/test-case-when.R
```diff
@@ -374,3 +376,127 @@ test_that("Using scalar LHS with vector RHS is deprecated (#7082)", {
...
+
+test_that("replace_when() retains names of `x`, consistent with `base::replace()`", {
+  x <- c(a = 1, b = 2, c = 3)
+
+  expect_identical(
+    x |>
+      replace_when(
+        x == 1 ~ 0,
+        x == 3 ~ c(z = 4)
+      ),
```
hadley:
```suggestion
      replace_when(
        x, 
        x == 1 ~ 0,
        x == 3 ~ c(z = 4)
      ),
```

Just because it looks a little nicer.

hadley (approved):

R/case-match.R
DavisVaughan: This was maybe a confusing comment on my part. It is now

```
  # Matching historical behavior of `case_match()`, which was to work like
  # `case_when()` and not allow empty `...`. Newer `replace_when()` and
  # `replace_values()` are a no-op for this case, but we superseded
  # `case_match()` at that time so it never moved to the new behavior.
  allow_empty_dots <- FALSE
```

R/case-when.R
DavisVaughan: Mostly removed, will put in vignette somehow

R/case-when.R
DavisVaughan: I am going to change `"unmatched"` back to `as.character(x)` to be truer to the fallthrough output of the fizzbuzz game

tests/testthat/test-case-when.R
DavisVaughan: I wasn't sure! Something about seeing _just_ the conditions in the function call was nice to me. But I also agree that having a pipe in the mutate() is also not great, I'll update

Issue #7696, opened 2025-06-23
Feature request `case_replace()`


rossholmberg:
I don't know about others, but my main use for `case_when` and `case_match` is to selectively replace values in a column. For example:

**Outline of the current situation (as I understand it)**
```
df <- tibble(cola = letters[1:5])
df |>
    mutate(cola = case_match(cola, c("b", "e") ~ "changed", .default = cola))
```

This works fine, but I find I regularly wish the `.default` value didn't need to be specified (its default is NULL). For example when using as part of a piped series of functions:

```
df |>
    mutate(cola = cola |>
        toupper() |>
        case_match(c("B", "E") ~ "changed")
    )
```
This will replace all unspecified values with NA, which is not what I want, but it's difficult to pass the piped vector as the `.default` value. It can be done by specifying a function, but it's a bit cumbersome:

```
df |>
    mutate(cola = cola |>
        toupper() |>
        (\(x) case_match(x, c("B", "E") ~ "changed", .default = x))()
    )
```

**Suggested solution**

My suggestion is a new function `case_replace`. The same as `case_match`, except without the `.default` parameter, always specifying that parameter as the input vector (the use case is similar to `base::replace`, but that function requires specifying by index instead of value):
```
case_replace <- \(x, ...) case_match(.x = x, ..., .default = x)
```

With that, we can pipe a vector nicely:
```
df |>
    mutate(cola = cola |>
        toupper() |>
        case_replace(c("B", "E") ~ "changed")
    )
```

I realise I'm nitpicking here, and `case_match` is really close to what I need. I've just found this to be a useful thing for me, and if others have similar needs, maybe it would be worth adding as a separate function in `dplyr::`. I'd be happy to have a go at contributing a PR if the interest is there.

Issue #7694, opened 2025-06-11
Canonical approach to lookup-based replacement?


MichaelChirico:
I am finding it a bit awkward to refactor code using `plyr::mapvalues()`, which looks like:

```r
mapvalues(x, from_vector, to_vector)
```

Items in `x` found in `from_vector` are replaced by the corresponding entry in `to_vector`.

The closest analogue is `case_match()`, but using it often requires pretty substantial refactoring, e.g. if `from_vector`/`to_vector` come from arguments to a function. Elsewhere it is often convenient to keep `from_vector`/`to_vector` instead of writing them out manually in different arguments.

Another approach I've tried would basically be to do a join-and-replace lookup with a `left_join()` to `tibble(from_vector, to_vector)`.

`plyr::revalue()` is similarly not trivial to migrate.

Am I missing something else?

DavisVaughan:
Not really. It's a hole in the API that I want to fix in the next minor vctrs release with some kind of new helper that probably lives slightly lower than dplyr itself (because it is typically useful for people writing functions, not people doing interactive analysis).

Out of curiosity, what is the data.table approach here?

(Also, assume that `from_vector` and `to_vector` are not character vectors, so assume you can't used a simple named vector to tie the two vectors together)

DavisVaughan:
> Another approach I've tried would basically be to do a join-and-replace lookup with a left_join() to tibble(from_vector, to_vector).

I think that would totally work, but it _feels_ like it's at the wrong level of abstraction (table level) since all of the required level is at the level of a single column (vector level)

MichaelChirico:
> Also, assume that `from_vector` and `to_vector` are not character vectors

I _do_ think that's the most common case (probably by far), but something more general makes sense. I will say that most often, this is about partially updating a column, where the data type would not be edited.

> not people doing interactive analysis

Maybe, maybe not. I would say it comes up interactively a fair amount, e.g. normalizing names:

```r
normalized_names <- data.frame(
  messy = c("Disney", "Disney, Inc.", "TWDCo"),
  clean = "The Walt Disney Company"
)
```

I guess your point is that interactively, it's easier+cleaner to just write it out, e.g. in `case_match()`. And the case I cited, where we take the look-up vectors as input, is more programmatic. But interactively, that could apply e.g. when we get the look-up as a data dictionary, e.g. in CSV. So it's a bit of a gray area, but lean towards yes, it's more programmatic.

> it feels like it's at the wrong level of abstraction

[...]

Issue #6962, opened 2023-11-09
Feature request: Pipe-supportive pass-through of unrecoded elements using `case_match()`?


jmobrien:
`case_match()` is great to have as a more pipe-able alternative to `case_when()` following modern tidyverse approaches.  However, it lacks an equivalent capabilities to one nice feature of its predecessor `recode()`: namely, where non-recoded elements could be allowed to pass through into the output.  While there is a workaround for this missing feature provided in the help examples, I'm realizing it may not generalize well across use cases.

An example similar to what I encountered:

``` r
# Some filenames:
filenames <- 
  c("juniors_2022_OU.csv", "roster_2023_juniors_notre_dame.csv",  "2022_SOU_juniors_roster.csv")

# Begin building tibble to (eventually) import & organize the content from these files:
tibble::tibble(

  # Import filenames (in practice, would be a call to list.files() ):
  file_names = filenames,
  
  # Get key info like year and school out of filenames: 
  
  # Year is straightforward:
  id_year = file_names |> stringr::str_detect("202[23]"),
  
  #School, however, needs initialisms expanded to minimize ambiguity:
  
  # recode() can do this inside a single set of piped instructions. It
  # fills in the original data when items are not recoded (& same 
  # type as recoded output):

  id_school_recode = 
    file_names |> 
    # Remove non-school-name content:
      stringr::str_remove(".csv$") |> 
      stringr::str_remove("_?202[23]_?") |> 
      stringr::str_remove("_?juniors_?") |> 
      stringr::str_remove("_?roster_?") |>
    # Recode initialisms:
      dplyr::recode(
        "OU" = "oklahoma",
        "SOU" = "southern_oregon"
      ),
  
  # Using case_match(), though, non-recoded elements become NA.
  id_school_casematch = 
    file_names |> 
      stringr::str_remove(".csv$") |> 
      stringr::str_remove("_?202[23]_?") |> 
      stringr::str_remove("_?juniors_?") |> 
      stringr::str_remove("_?roster_?") |> 
      dplyr::case_match(
        "OU" ~ "oklahoma",
        "SOU" ~ "southern_oregon"
      )
)
#> # A tibble: 3 × 4
#>   file_names                        id_year id_school_recode id_school_casematch
#>   <chr>                             <lgl>   <chr>            <chr>              
#> 1 juniors_2022_OU.csv               TRUE    oklahoma         oklahoma           
[...]

PR #7729, opened 2025-10-07
Implement `recode_values()` and `replace_values()`

DavisVaughan:
Branched from #7728 

Closes https://github.com/tidyverse/dplyr/issues/7696
Closes https://github.com/tidyverse/dplyr/issues/7694
Closes https://github.com/tidyverse/dplyr/issues/6962

Part of https://github.com/tidyverse/tidyups/pull/29

I have not marked `case_match()` as superseded yet, nor have I updated any docs mentioning `case_match()` to instead mention `recode_values()`, that will be in a follow up PR. This one is focused on just getting these new functions implemented. (In that follow up PR, I will also update `recode()`'s docs to mention this function, and `na_if()` to mention this function as an enhanced version of that).

### DIFF
--- a/R/recode-values.R
+++ b/R/recode-values.R
@@ -17,7 +17,7 @@
 #' A major difference between the two functions is what happens when no cases
 #' match:
 #'
-#' - `recode_values()` falls through to a `.default`.
+#' - `recode_values()` falls through to a `default`.
 #'
 #' - `replace_values()` retains the original values from `x`.
 #'
--- a/man/recode-and-replace-values.Rd
+++ b/man/recode-and-replace-values.Rd
@@ -104,7 +104,7 @@ better expresses intent.
 A major difference between the two functions is what happens when no cases
 match:
 \itemize{
-\item \code{recode_values()} falls through to a \code{.default}.
+\item \code{recode_values()} falls through to a \code{default}.
 \item \code{replace_values()} retains the original values from \code{x}.
 }
 


tests/testthat/test-recode-values.R
```diff
@@ -0,0 +1,369 @@
...
+    ~from, ~to,
+    "a", "A",
+    "b", "B",
+    "c", "C"
+  )
+
+  expect_snapshot(error = TRUE, {
+    recode_values(x, from = table$from, to = table$to, unmatched = "error")
+  })
+
```
DavisVaughan, line 125: Notable behavior

tests/testthat/test-recode-values.R
```diff
@@ -0,0 +1,369 @@
...
+    recode_values(x, from = table$from, to = table$to, unmatched = "error"),
+    c("A", "B", "A", NA, "C")
+  )
+})
+
+test_that("`NA` is matched exactly", {
+  # With logical `NA`
+  x <- c(1, NA)
+
+  expect_identical(recode_values(x, NA ~ 0), c(NA, 0))
```
DavisVaughan, line 138: As expected

R/recode-values.R
```diff
@@ -0,0 +1,506 @@
...
+#' x |> replace_values("NYC" ~ "NY")
+#'
+#' # `replace_values()` is particularly nice for replacing `NA`s with values...
+#' x |> replace_values(NA ~ "Unknown (NA)")
+#' # ...or values with `NA`s
+#' x |> replace_values("Unknown" ~ NA)
+#'
+#' # Multiple values can be grouped within a single left-hand side to normalize
+#' # all problematic values at once
+#' x |> replace_values(c(NA, "Unknown") ~ "<not recorded>")
```
DavisVaughan: I imagine this will be a popular way to use these

R/recode-values.R
```diff
@@ -0,0 +1,506 @@
...
+#' # into a standalone lookup table. You could even imagine reading this
+#' # `likert` lookup table in from a separate CSV file.
+#' likert <- tribble(
+#'   ~from, ~to,
+#'   1, "Strongly disagree",
+#'   2, "Disagree",
+#'   3, "Neutral",
+#'   4, "Agree",
+#'   5, "Strongly agree"
+#' )
```
DavisVaughan, line 190: Compelling example for using an external lookup table. The call out to the fact that it could come from an external CSV file is particularly compelling

R/recode-values.R
```diff
@@ -0,0 +1,506 @@
...
+#' # This `tribble()` is more complex than it may appear, it actually
+#' # creates a list column!
+#' standardized <- tribble(
+#'   ~from,                        ~to,
+#'   c("UNC", "Chapel Hill"),      "UNC",
+#'   c("Duke", "Duke University"), "Duke",
+#'   c("NC State"),                "NC State",
+#'   c("ECU", "East Carolina"),    "ECU",
+#'   NA,                           NA
+#' )
```
DavisVaughan, line 257: I think this is so cool

R/recode-values.R
```diff
@@ -0,0 +1,506 @@
+#' Recode and replace values
+#'
+#' @description
+#' `recode_values()` and `replace_values()` are two ways to map old values to
```
hadley:
```suggestion
#' `recode_values()` and `replace_values()` provide two ways to map old values to
```

R/recode-values.R
```diff
@@ -0,0 +1,506 @@
...
+#' - Use `replace_values()` when updating an existing vector.
+#'
+#' If no values match, then for `recode_values()` a `default` is used in
+#' unmatched locations, and for `replace_values()` the original values from `x`
+#' are retained.
+#'
+#' These functions have two mutually exclusive ways to use them:
+#'
+#' - A formula-based approach, i.e. `recode_values(x, from1 ~ to1, from2 ~
+#'   to2)`, similar to [case_when()], which is tailored for interactive usage.
```
hadley: Interactive vs programmatic doesn't seem quite the right distinction here. But I'm not sure I can offer anything better. The best I've come up with is "programmer-y" vs "database-y".

R/recode-values.R
```diff
@@ -0,0 +1,506 @@
...
+#'
+#' ```
+#' case_when(
+#'   x %in% c("a", "b") ~ 1,
+#'   x %in% "c" ~ 2,
+#'   x %in% c("d", "e") ~ 3
+#' )
+#'
+#' x |>
+#'   recode_values(
```
hadley: I don't love the way this style looks when the first argument is so short.

R/recode-values.R
```diff
@@ -0,0 +1,506 @@
...
+#' These functions have two mutually exclusive ways to use them:
+#'
+#' - A formula-based approach, i.e. `recode_values(x, from1 ~ to1, from2 ~
+#'   to2)`, similar to [case_when()], which is tailored for interactive usage.
+#'
+#' - A vector-based approach, i.e. `recode_values(x, from = from, to = to)`,
+#'   which is tailored for programmatic usage. This is particularly useful when
+#'   you have a pre-built lookup table.
+#'
+#' @section Connection between `recode_values()` and `case_when()`:
```
hadley: This section could go in the vignette when you write it.

R/recode-values.R
```diff
@@ -0,0 +1,506 @@
...
+#' `replace_values()` is always a better choice because it is type stable on
+#' `x`, pipes better, and better expresses intent.
+#'
+#' @param x A vector.
+#'
+#' @param ... <[`dynamic-dots`][rlang::dyn-dots]> A sequence of two-sided
+#'   formulas. The left hand side (LHS) determines which values match this case.
+#'   The right hand side (RHS) provides the replacement value.
+#'
+#'   - The LHS inputs will be [cast][vctrs::theory-faq-coercion] to the type of
```
hadley: Same feedback here as other PR.

R/recode-values.R
hadley: I love this sequence of examples 😄

R/recode-values.R
```diff
@@ -0,0 +1,506 @@
...
+#'   5, "Strongly agree"
+#' )
+#'
+#' data |>
+#'   mutate(
+#'     score = score |>
+#'       recode_values(
+#'         from = pull(likert, from),
+#'         to = pull(likert, to)
+#'       )
```
hadley:
```suggestion
#'       recode_values(from = likert$from, to = likert$to)
```

I think this is more succinct and clearer.

R/recode-values.R
```diff
@@ -0,0 +1,506 @@
...
+#'   unmatched = "error"
+#' )
+#'
+#' # ------------------------------------------------------------------------------
+#' # Lists of vectors
+#'
+#' # In some cases, your mapping may collapse multiple groups together into a
+#' # single value. For example, here we'd like to standardize the school names.
+#' schools <- tibble(
+#'   name = c(
```
hadley: Maybe just keep this as a vector? I think it's worth shrinking to fit on one line so it doesn't take up quite so much space.

R/recode-values.R
hadley: Yeah, it's super slick.

R/recode-values.R
```diff
@@ -0,0 +1,506 @@
...
+
+#' @rdname recode-and-replace-values
+#' @export
+replace_values <- function(
+  x,
+  ...,
+  from = NULL,
+  to = NULL
+) {
+  check_dots_unnamed()
```
hadley, line 334: Move into `determine_implementation()`? But I can all see the argument for keeping it here to make it very clear to the reader than `...` won't have names. That said, this function doesn't actually do anything with `...`.

R/recode-values.R
```diff
@@ -0,0 +1,506 @@
...
+  )
+}
+
+determine_implementation <- function(..., from, to, error_call) {
+  has_dots <- dots_n(...) != 0L
+  has_from <- !is.null(from)
+  has_to <- !is.null(to)
+
+  # Supplied `...`
+  if (has_dots) {
```
hadley, line 411: Would be nice if we could write something like `check_exclusive(c("from", "to"), "...")` but probably too specialised a case.

R/recode-values.R
```diff
@@ -0,0 +1,506 @@
...
+  list(
+    from = from,
+    to = to,
+    from_as_list_of_vectors = from_as_list_of_vectors,
+    to_as_list_of_vectors = to_as_list_of_vectors
+  )
+}
+
+determine_implementation <- function(..., from, to, error_call) {
+  has_dots <- dots_n(...) != 0L
```
hadley:
```suggestion
  has_dots <- !missing(...)
```

R/recode-values.R
```diff
@@ -0,0 +1,506 @@
...
+    cli::cli_abort(
+      "Must supply both {.arg from} and {.arg to}.",
+      call = error_call
+    )
+  }
+
+  # Empty call to `recode_values()` or `replace_values()` with no `...`
+  # or `from/to` specified. Let this use `...`, and then `recode_values()`
+  # will error and `replace_values()` will be a no-op
+  "dots"
```
hadley, line 443: I'd be tempted to move this earlier and do:

```R
if (!has_dots && !has_from && !has_to) {
  return("dot")
}
```

That way I think you could simplify the comment (since it's now more obviously what the condition is) and then you have all the successes before the failure.

Or maybe I'd prefer a big if statement like this:

```R
if (has_dots && (!has_from || !has_to)) {
  "dots"
} else if (!has_dots && !has_from && !has_to) {
  # fallback if no args supplied
  "dots"
} else if (!has_dots && has_from && has_to) {
  "from-to"
} else {
  cli::cli_abort("Either supply both from and to or ...")
}
```

That simplifies the error at the expense of making it less clear, but maybe that's ok?

(Not that this is really that important, but thought it would be fun to think it through since we've been talking about `if` statements)

tests/testthat/_snaps/recode-values.md
```diff
@@ -0,0 +1,199 @@
...
+      Error in `replace_values()`:
+      ! Case 2 (`2`) must be a two-sided formula, not the number 2.
+
+---
+
+    Code
+      recode_values(1, 1 ~ 1, ~2)
+    Condition
+      Error in `recode_values()`:
+      ! Case 2 (`~2`) must be a two-sided formula.
```
hadley: not the number 2?

hadley (approved):
People are going to love this 😄 

I focussed mostly on docs + code style since I assume the implementation is solid.

tests/testthat/_snaps/recode-values.md
DavisVaughan: It should really be "not a one-sided formula" 😄 , it's a special error from the case-when code about the case of getting a 1 sided formula

R/recode-values.R
DavisVaughan:
```
#' - A formula-based approach, i.e. `recode_values(x, from1 ~ to1, from2 ~
#'   to2)`, similar to [case_when()], which is useful when you have a small
#'   number of cases.
#'
#' - A vector-based approach, i.e. `recode_values(x, from = from, to = to)`,
#'   which is useful when you have a pre-built lookup table (which may come
#'   from an external source, like a CSV file).
```

R/recode-values.R
DavisVaughan: I think I prefer the current location for future readability. Tests ensure we don't mess this up

R/recode-values.R
DavisVaughan: I've gone for another alternative which is to better group the `from` and `to` cases together into a single if statement.

It now looks like:
- Supplied `...` early exit
- Supplied `from` and `to` early exit
- Supplied nothing fallthrough (which is typically where I'd put an error, but we allow this)

PR #7735, opened 2025-10-10
Supersede `case_match()`

DavisVaughan:
In favor of `recode_values()` and `replace_values()`

Bringing in dev lifecycle to ensure we don't release dplyr with degraded `case_match()` performance due to `lifecycle::signal_stage()`.

Updated all `case_match()` tests to show how to transition to `recode_values()` and `replace_values()`, which was very validating

### DIFF
--- a/_pkgdown.yml
+++ b/_pkgdown.yml
@@ -128,15 +128,15 @@ reference:
     to be superior, but we don't want to force you to change until you're
     ready, so the existing functions will stay around for several years.
   contents:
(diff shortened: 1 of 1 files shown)


_pkgdown.yml
```diff
@@ -137,6 +136,7 @@ reference:
   - vars
   - with_groups
   - recode
+  - case_match
```
hadley: In principle I think these should be in alphabetical order

PR #7741, opened 2025-10-22
C API compliance

DavisVaughan:
Here is what is left:

```
   File ‘dplyr/libs/dplyr.so’:
     Found non-API calls to R: ‘PRVALUE’, ‘Rf_allocSExp’, ‘SET_PRCODE’,
       ‘SET_PRENV’, ‘SET_PRVALUE’
```

These are related to promise creation. We ideally need `R_MakeDelayedBinding()` from our patch to Luke to fix this.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,7 @@
 # dplyr (development version)
 
+* Progress towards making dplyr conformant with the public C API of R (#7741).
+
 * `case_when()` now throws correctly indexed errors when `NULL`s are supplied in `...` (#7739).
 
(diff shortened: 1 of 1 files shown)


src/utils.h
```diff
@@ -0,0 +1,30 @@
...
+
+#define R_NO_REMAP
+#include <R.h>
+#include <Rinternals.h>
+#include <Rversion.h>
+
+static inline
+bool str_is_utf8(SEXP x) {
+#if (R_VERSION >= R_Version(4, 5, 0))
+  return Rf_charIsUTF8(x);
```
lionel-, line 12: oh nice

Issue #7651, opened 2025-02-04
Allow multiple arguments to `na_if`


nalimilan:
It's quite frequent to have to recode several distinct values to NA. So far the shortest solution I could find to do that is e.g. `na_if(na_if(vec, 88), 99)`. `if_else(vec %in% c(88, 99), NA_real_, vec)` or `case_match(vec, c(88, 99) ~ NA, .default=vec)` are also possible but relatively verbose. Could it be allowed to pass multiple arguments to `na_if ` so that one could write `na_if(vec, 88, 99)` instead?

Issue #6511, opened 2022-10-20
Version of `na_if` that takes conditionals.


orgadish:
I often have want to convert some of my data to NA, and keep the rest. If it is as simple as identifying a particular value, I can use `na_if`. If, instead, I want to use a conditional, I have to use `if_else`, which would not be a problem if I didn't then have to also keep track of what type of `NA` I have to use, since `if_else` is type strict.

(Apologies if this already exists...)

``` r
library(tidyverse)

ll <- LETTERS[1:5]

# Works
ll %>% na_if("A")
#> [1] NA  "B" "C" "D" "E"

# Does not work even with multiple values (which makes sense because it's designed to be parallelized along the vector). 
ll %>%  na_if(c("A","B"))
#> Error in `na_if()`:
#> ! `y` must be length 5 (same as `x`) or one, not 2.

# Fails due to type strictness.
ll %>% {if_else(. %in% c("A", "C"), NA, .)}
#> Error in `if_else()`:
#> ! `false` must be a logical vector, not a character vector.

# With type strictness.
ll %>% {if_else(. %in% c("A", "C"), NA_character_, .)}
#> [1] NA  "B" NA  "D" "E"

# `case_when` works but is much slower.
bench::mark(
  if_else = ll %>% {if_else(. %in% c("A", "C"), NA_character_, .)},
  case_when = ll %>% {case_when(!(. %in% c("A", "C")) ~ .)}
)
#> # A tibble: 2 × 6
#>   expression      min   median `itr/sec` mem_alloc `gc/sec`
#>   <bch:expr> <bch:tm> <bch:tm>     <dbl> <bch:byt>    <dbl>
#> 1 if_else      29.5µs   31.9µs    26601.        0B    10.6 
#> 2 case_when    94.5µs  102.6µs     8525.    91.9KB     6.13
```

<sup>Created on 2022-10-19 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

<details style="margin-bottom:10px;">
<summary>
Session info
</summary>

``` r
sessioninfo::session_info()
#> ─ Session info ───────────────────────────────────────────────────────────────
#>  setting  value
#>  version  R version 4.2.1 (2022-06-23)
#>  os       macOS Big Sur ... 10.16
#>  system   x86_64, darwin17.0
#>  ui       X11
#>  language (EN)
#>  collate  en_US.UTF-8
#>  ctype    en_US.UTF-8
#>  tz       America/Los_Angeles
[...]

Issue #1972, opened 2016-06-27
na_if: Support multiple values that are replaced with NA 


jlegewie:
The additions in 0.5 are great!! A lot of dataset contain multiple missing codes for a single variable though and I think it would be much more useful if the user could pass multiple values to `na_if`. For example, `1:10 %>% na_if(2, 4)` or even `1:10 %>% na_if(1:5, 10)` would both be useful. Maybe it's not desired because `na_if` is supposed to work like `NULL_IF` but I do think it's much more useful. So I propose modifying  `na_if` or adding a function `na_codes` that supports multiple values. Of course, `na_if(1:5, 5:1)` would lead to different results. If this is desired, I could submit a pull request.

Here are two example implementations:

#### Version 1 with support for vectors
```
na_codes <- function(x, ...) {
	x[x %in% c(...)] <- NA
	x
}
1:10 %>% na_codes(1:5, 10)
letters %>% na_codes("a", "b", c("z", "y"))
```
#### Version 2 without support for vectors
```
na_codes2 <- function(x, ...) {
	for(y in list(...))
		x[x == y] <- NA
	x
}
1:10 %>% na_codes(1, 2, 3, 4, 5, 10)
letters %>% na_codes("a", "b", "z", "y")
```

#### Performance comparison
Version 1 is slower but also supports vector arguments. Version 2 is as fast as `dplyr::na_if` with support for multiple arguments.

```
library("rbenchmark")
x <- rbinom(1e6, size = 1, prob = 0.5)
benchmark(dplyr::na_if(x, 1), na_codes(x, 1), na_codes2(x, 1))
#                test replications elapsed relative user.self sys.self user.child sys.child
#1 dplyr::na_if(x, 1)          100   2.870    1.000     2.138    0.468          0         0
#2     na_codes(x, 1)          100   4.895    1.706     4.150    0.518          0         0
#3    na_codes2(x, 1)          100   3.098    1.079     2.318    0.481          0         0
```

hadley:
The second argument is vectorised along x - I don't think it's possible to change this because it would break the symmetry with `coalesce()` and with SQL's `NULL_IF`

hadley:
You could have another function with the behaviour that you wanted though.

jlegewie:
That's what I thought. Would `na_codes` be a welcome addition to `dplyr` or too redundant with `na_if`?

hadley:
[...]

PR #7742, opened 2025-10-23
Refresh `na_if()` and `coalesce()` docs

DavisVaughan:
`na_if()` has always been super confusing to me because I want to reach for `na_if(x, c(-1, -99))`, but it doesn't work that way. It works off `vec_equal()` not `vec_match()`, so `y` has to be length 1 or the same length as `x`. This is due to the fact that we translate it from `NULLIF` in SQL.

I don't think I'm alone in this (https://github.com/tidyverse/dplyr/issues/7651, https://github.com/tidyverse/dplyr/issues/1972), and what you really want here is `replace_values(x, c(-1, -99) ~ NA)`. I've refreshed the `na_if()` docs to tell a story that points people towards this.

I've also noticed that people want `na_if(x, x < 0)` to work, which is a case for `replace_when(x, x < 0 ~ NA)` instead (https://github.com/tidyverse/dplyr/issues/5711, https://github.com/tidyverse/dplyr/issues/6511), so I'm mentioning that example now too.

---

Also refreshed the `coalesce()` docs, which goes the other way of replacing `NA` with a value. Here I tried to show:
- when `coalesce()` and `replace_values()` are equivalent
- when `coalesce()` shines (>2 vectors)
- when `replace_values()` is a better choice (i.e. you really wanted to replace a value with `NA`)

---

The real thing I realized a while back is that `replace_values()` is 1 tool that you can learn to do 95% of what `na_if()`, `coalesce()`, and `tidyr::replace_na()` were all trying to do, and I'm trying to surface that in the docs a bit. I'll also talk about this in the vignette.

### DIFF
--- a/R/na-if.R
+++ b/R/na-if.R
@@ -43,9 +43,12 @@
 #' )
 #'
 #' # If you have multiple problematic values that you'd like to replace with
-#' # `NA`, then `replace_values()` is a better choice than nested `na_if()`s
+#' # `NA`, then `replace_values()` is a better choice than `na_if()`
 #' x <- c(-99, 1, 4, 0, -99, 5, -1, 0, 5)
 #' replace_values(x, c(0, -1, -99) ~ NA)
+#'
+#' # You'd have to nest `na_if()`s to achieve this
+#' try(na_if(x, c(0, -1, -99)))
 #' na_if(na_if(na_if(x, 0), -1), -99)
 #'
 #' # If you'd like to replace values that match a logical condition with `NA`,
--- a/man/na_if.Rd
+++ b/man/na_if.Rd
@@ -44,9 +44,12 @@ na_if(
 )
 
 # If you have multiple problematic values that you'd like to replace with
-# `NA`, then `replace_values()` is a better choice than nested `na_if()`s
+# `NA`, then `replace_values()` is a better choice than `na_if()`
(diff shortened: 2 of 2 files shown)


R/coalesce.R
```diff
@@ -19,24 +19,48 @@
...
+#' coalesce(x, y)
+#'
+#' # `coalesce()` really shines when you have >2 vectors to coalesce with
+#' z <- c(NA, 2, 3, 4, 5, 6)
+#' coalesce(x, y, z)
+#'
+#' # If you're looking to replace values with `NA`, rather than replacing `NA`
+#' # with a value, then use `replace_values()`
+#' x <- c(0, -1, 5, -99, 8)
+#' replace_values(x, c(-1, -99) ~ NA)
```
hadley, line 58:
```suggestion
#' # If you're looking to replace multiple values with `NA`, rather than replacing `NA`
#' # with a value, then use `replace_values()`
#' x <- c(0, -1, 5, -99, 8)
#' replace_values(x, c(-1, -99) ~ NA)
#' compare with:
#' try(na_if(x, c(-1, -99)))
```

Maybe?

DavisVaughan: Hmm mentioning `na_if()` in the `coalesce()` docs? Seems like a bit much to mention _both_ `na_if()` and `replace_values()` here, but I do like the idea of showing `try(na_if` in the `na_if()` docs for the similar example there

Issue #7746, opened 2025-10-27
if_any inside mutate, unexpected return on single columns


mp8:
I am not sure if this will be use case for any, but trying to use if_any directly inside mutate changes behavior unexpectedly when only single column is in the .cols with nothing specified for .fns. Apologies if this was already documented:

```
library(dplyr)

# Sample data
zdata <- tibble(
    id = 1:5,
    col_a = c(10, 25, 5, 30, 15),
    col_b = c(20, 12, 8, 40, 22),
    col_c = c(5, 18, 10, 15, 7)
)

#Trying to use if_any directly inside mutate, expected logical output
#Case1
zdata %>% mutate(any_big = if_any(c(col_a,col_b,col_c), ~ .x>10))  #Returns logical

#Case2
zdata %>% mutate(any_big = if_any(c(col_a,col_b,col_c)))                #Returns logical

#Case3
zdata %>% mutate(any_big = if_any(c(col_a), ~ .x>10))                   #Returns logical

#Case4
zdata %>% mutate(any_big = if_any(c(col_a)))                       #Does not return logical output, returns column as it is..

# A tibble: 5 × 5
     id col_a col_b col_c any_big
  <int> <dbl> <dbl> <dbl>   <dbl>
1     1    10    20     5      10
2     2    25    12    18      25
3     3     5     8    10       5
4     4    30    40    15      30
5     5    15    22     7      15


#Adding sessionInfo:

sessionInfo()
#> R version 4.5.1 (2025-06-13 ucrt)
#> Platform: x86_64-w64-mingw32/x64
#> Running under: Windows 10 x64 (build 19045)
#> 
#> Matrix products: default
#>   LAPACK version 3.12.1
#> 
#> locale:
#> [1] LC_COLLATE=English_United States.utf8 
#> [2] LC_CTYPE=English_United States.utf8   
#> [3] LC_MONETARY=English_United States.utf8
#> [4] LC_NUMERIC=C                          
#> [5] LC_TIME=English_United States.utf8    
#> 
#> time zone: America/Los_Angeles
#> tzcode source: internal
#> 
#> attached base packages:
#> [1] stats     graphics  grDevices utils     datasets  methods   base     
#> 
#> loaded via a namespace (and not attached):
#>  [1] digest_0.6.37     fastmap_1.2.0     xfun_0.53         glue_1.8.0       
[...]

Issue #7077, opened 2024-08-27
`if_any()` does not work as expected inside `mutate` when no inputs are provided


ks8997:
This is related to #7074. `if_any()` is now supposed to return `FALSE` when no inputs are provided as per #7072. This works correctly inside `filter` but not `mutate`.

```r
iris %>% 
  filter(if_any(c(), \(x) x  > 4000)) # returns no rows as expected

iris %>%
    mutate(z = if_any(c(),  \(x) x  > 4000)) # incorrectly returns all TRUE
```

Apologies if I misunderstood something.

DavisVaughan:
Yea I think you're right, I didn't double check for `mutate()`. Thanks for following up!

ks8997:
@DavisVaughan So I poked around a bit and I think the fix seems straightforward. The problematic section is below:

https://github.com/tidyverse/dplyr/blob/0005f6768fa765b3bba5148711967d58b6013037/R/across.R#L345-L347

This unconditionally returns `TRUE` for empty input.

My proposed fix look like this:

```r
if_across <- function(op, df) {
  n <- nrow(df)
  across_if_fn <- context_peek_bare("across_if_fn") %||% "across"

  if (!length(df)) {
    if (across_if_fn == "if_any") {
      return(FALSE)
    } else {
      return(TRUE)
    }
  }
    
  combine <- function(x, y) {
    if (is_null(x)) {
      y
    } else {
      op(x, y)
    }
  }
  reduce(df, combine, .init = NULL)
}
```

The `across_if_fn` stores information about the type of across function we are calling. It will then return all `FALSE` for `if_any()` and all `TRUE` for `if_all()` (and `across()`). New to this but happy to submit a PR if this solution looks alright to you. I'll also add a test case for this.

PR #7747, opened 2025-10-28
Make `if_any()` and `if_all()` consistent in all contexts

DavisVaughan:
Closes https://github.com/tidyverse/dplyr/issues/7746
Closes https://github.com/tidyverse/dplyr/issues/7077

It turns out that our application of `if_any()` and `if_all()` was fairly inconsistent. This is due to the fact that they are tricky to get right, we have 2 different implementations of them. One for the expansion case and one for the evaluation case. I've now tried to unify these to use the same underlying implementation, `dplyr_list_pany()` or `dplyr_list_pall()` depending on the scenario (which just use `vec_pany()` and `vec_pall()` under the hood)

In addition to greater consistency across the board, you'll also note that in errors the `In argument:` label also now reports the _original_ expression pre-expansion in the `filter()` cases, which is a much better error

In the examples below, for `filter()`, note that adding `()` around the `if_any()` or `if_all()` calls triggers the evaluation case rather than the expansion case.

``` r
library(dplyr)

# With zero inputs, if_any

# Before
df <- tibble(x = 1:2)
filter(df, if_any(c(), identity))
#> # A tibble: 0 × 1
#> # ℹ 1 variable: x <int>
filter(df, (if_any(c(), identity)))
#> # A tibble: 2 × 1
#>       x
#>   <int>
#> 1     1
#> 2     2
filter(df, any())
#> # A tibble: 0 × 1
#> # ℹ 1 variable: x <int>

# After
df <- tibble(x = 1:2)
filter(df, if_any(c(), identity))
#> # A tibble: 0 × 1
#> # ℹ 1 variable: x <int>
filter(df, (if_any(c(), identity)))
#> # A tibble: 0 × 1
#> # ℹ 1 variable: x <int>
filter(df, any())
#> # A tibble: 0 × 1
#> # ℹ 1 variable: x <int>
```

```r
# With one non-logical input

# Before
df <- tibble(x = 1:2)
filter(df, if_any(x, identity))
#> Error in `filter()`:
#> ℹ In argument: `(function (x) ...`.
#> Caused by error:
#> ! `..1` must be a logical vector, not an integer vector.
filter(df, (if_any(x, identity)))
#> Error in `filter()`:
#> ℹ In argument: `(if_any(x, identity))`.
#> Caused by error:
#> ! `..1` must be a logical vector, not an integer vector.
mutate(df, a = if_any(x, identity))
#> # A tibble: 2 × 2
#>       x     a
#>   <int> <int>
#> 1     1     1
#> 2     2     2

# After
df <- tibble(x = 1:2)
filter(df, if_any(x, identity))
#> Error in `filter()`:
#> ℹ In argument: `if_any(x, identity)`.
#> Caused by error in `if_any()`:
#> ! `x` must be a logical vector, not an integer vector.
filter(df, (if_any(x, identity)))
#> Error in `filter()`:
#> ℹ In argument: `(if_any(x, identity))`.
#> Caused by error in `if_any()`:
#> ! `x` must be a logical vector, not an integer vector.
mutate(df, a = if_any(x, identity))
#> Error in `mutate()`:
#> ℹ In argument: `a = if_any(x, identity)`.
#> Caused by error in `if_any()`:
#> ! `x` must be a logical vector, not an integer vector.
```

```r
# In general, with non-logical types resulting from applying `.fns` we now error
# more appropriately

# Before
df <- tibble(x = c(TRUE, FALSE), y = c("a", "b"))
filter(df, if_any(c(x, y), identity))
#> Error in `filter()`:
#> ℹ In argument: `|...`.
#> Caused by error in `<function(x) x>(x) | <function(x) x>(y)`:
#> ! operations are possible only for numeric, logical or complex types
filter(df, (if_any(c(x, y), identity)))
#> Error in `filter()`:
#> ℹ In argument: `(if_any(c(x, y), identity))`.
#> Caused by error in `op()`:
#> ! operations are possible only for numeric, logical or complex types

# After
df <- tibble(x = c(TRUE, FALSE), y = c("a", "b"))
filter(df, if_any(c(x, y), identity))
#> Error in `filter()`:
#> ℹ In argument: `if_any(c(x, y), identity)`.
#> Caused by error in `if_any()`:
#> ! `y` must be a logical vector, not a character vector.
filter(df, (if_any(c(x, y), identity)))
#> Error in `filter()`:
#> ℹ In argument: `(if_any(c(x, y), identity))`.
#> Caused by error in `if_any()`:
#> ! `y` must be a logical vector, not a character vector.
```

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,13 @@
 # dplyr (development version)
 
+* `if_any()` and `if_all()` are now more consistent in all use cases (#7059, #7077, #7746, @jrwinget). In particular:
+
+  * When called with zero inputs, `if_any()` returns `FALSE` and `if_all()` returns `TRUE`.
+
+  * When called with one input, both now return logical vectors rather than the original column.
+
+  * The result of applying `.fns` now must be a logical vector.
+
 * `tally_n()` creates fully qualified funciton calls for duckplyr compatibility (#7046)
 
 * `storms` has been updated to include 2023 and 2024 data (#7111, @tomalrussell).
@@ -100,10 +108,6 @@
 
 * Fixed an issue where duckplyr's ALTREP data frames were being materialized
   early due to internal usage of `ncol()` (#7049).
-
-* `if_any()` and `if_all()` are now fully consistent with `any()` and `all()`.
-  In particular, when called with empty inputs `if_any()` returns `FALSE` and
-  `if_all()` returns `TRUE` (#7059, @jrwinget).
 
 ## Lifecycle changes
 
--- a/R/across.R
+++ b/R/across.R
@@ -349,31 +349,21 @@ across <- function(.cols, .fns, ..., .names = NULL, .unpack = FALSE) {
 if_any <- function(.cols, .fns, ..., .names = NULL) {
   context_local("across_if_fn", "if_any")
   context_local("across_frame", current_env())
-  if_across(`|`, across({{ .cols }}, .fns, ..., .names = .names))
+  df <- across({{ .cols }}, .fns, ..., .names = .names)
+  x <- dplyr_new_list(df)
+  size <- vec_size(df)
+  dplyr_list_pany(x, size = size)
 }
+
 #' @rdname across
 #' @export
 if_all <- function(.cols, .fns, ..., .names = NULL) {
   context_local("across_if_fn", "if_all")
   context_local("across_frame", current_env())
-  if_across(`&`, across({{ .cols }}, .fns, ..., .names = .names))
-}
-
-if_across <- function(op, df) {
-  n <- nrow(df)
-
-  if (!length(df)) {
-    return(TRUE)
-  }
-
-  combine <- function(x, y) {
-    if (is_null(x)) {
-      y
-    } else {
-      op(x, y)
-    }
-  }
-  reduce(df, combine, .init = NULL)
+  df <- across({{ .cols }}, .fns, ..., .names = .names)
+  x <- dplyr_new_list(df)
+  size <- vec_size(df)
+  dplyr_list_pall(x, size = size)
 }
 
 #' Combine values from multiple columns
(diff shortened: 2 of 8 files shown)


CLAUDE.md
```diff
@@ -0,0 +1,76 @@
+# CLAUDE.md
```
DavisVaughan: Plucked from ellmer, with modifications. I'm just trying it out.

NEWS.md
```diff
@@ -1,5 +1,13 @@
 # dplyr (development version)
 
+* `if_any()` and `if_all()` are now more consistent in all use cases (#7059, #7077, #7746, @jrwinget). In particular:
+
+  * When called with zero inputs, `if_any()` returns `FALSE` and `if_all()` returns `TRUE`.
+
+  * When called with one input, both now return logical vectors rather than the original column.
+
+  * The result of applying `.fns` now must be a logical vector.
```
DavisVaughan, line 9: This is a breaking change, but hopefully not a common issue? We really do want the result of `.fns` to be logical vectors, particularly for https://github.com/tidyverse/dplyr/issues/7746 to work right.

The vctrs functions for pany/pall will probably also be pretty strict about requiring logical inputs.

R/across.R
```diff
@@ -349,31 +349,87 @@ across <- function(.cols, .fns, ..., .names = NULL, .unpack = FALSE) {
...
+dplyr_list_pany <- function(
+  x,
+  ...,
+  size = NULL,
+  error_call = caller_env()
+) {
+  dplyr_list_pany_pall(x, "any", ..., size = size, error_call = error_call)
+}
+
+dplyr_list_pany_pall <- function(
```
DavisVaughan: I'm hoping to just remove this in favor of the vctrs versions soon, but I needed to get the semantics of it right enough to be able to add all the tests here

R/across.R
```diff
@@ -349,31 +349,87 @@ across <- function(.cols, .fns, ..., .names = NULL, .unpack = FALSE) {
...
+    all = {
+      op <- `&`
+      init <- TRUE
+    },
+    abort("Unreachable", .internal = TRUE)
+  )
+
+  init <- vec_rep(init, times = size)
+
+  reduce(x, op, .init = init)
```
DavisVaughan: Having an `init`ial value is an important reason that the 0 and 1 input cases now "just work" correctly. This was `NULL` before.

R/across.R
```diff
@@ -608,79 +664,68 @@ dplyr_quosures <- function(...) {
...
+
+    # In the evaluation path, `across()` automatically recycles to common size,
+    # so we must here as well for compatibility. `across()` also returns a 0
+    # col, 1 row data frame in the case of no inputs so that it will recycle to
+    # the group size, which we also do here.
+    size <- ns[["dplyr_list_size_common"]](x, absent = 1L, call = call(!!if_fn))
+    x <- ns[["dplyr_list_recycle_common"]](x, size = size, call = call(!!if_fn))
+
+    ns[[!!dplyr_fn]](x, size = size, error_call = call(!!if_fn))
+  })
```
DavisVaughan, line 674: This is the same kind of trick we use in `as_pick_expansion()` for the expansion path. I've carefully tried to match it to the evaluation path.

Basically we replace `if_any(c(x, y), fn)` with something like

```r
x <- list(x = x, y = y)
ns <- asNamespace("dplyr")
size <- ns[["dplyr_list_size_common"]](x, absent = 1L, call = call(if_any()))
x <- ns[["dplyr_list_recycle_common"]](x, size = size, call = call(if_any()))
ns[["dplyr_list_pany"]](x, size = size, error_call = call(if_any()))
```

R/vctrs.R
```diff
@@ -21,3 +21,28 @@ dplyr_vec_ptype_common <- function(chunks, name) {
     error = common_handler(name)
   )
 }
+
+# Version of `vec_size_common()` that takes a list.
+# Useful for delaying `!!!` when used within an `expr()` call.
+dplyr_list_size_common <- function(
```
DavisVaughan, line 27: I'm somewhat convinced that in vctrs we should have exported `list_size_common()` rather than `vec_size_common()`, i.e.:

```r
obj_check_vector(x)
list_check_all_vectors(xs)

vec_size(x)
list_size_common(xs)

vec_recycle(x)
list_recycle_common(xs)

vec_cast(x, to)
list_cast_common(xs, to)
```

tests/testthat/test-across.R
```diff
@@ -1256,6 +1272,431 @@ test_that("rowwise() preserves list-cols iff no `.fns` (#5951, #6264)", {
   )
 })
 
+test_that("`across()` recycle `.fns` results to common size", {
```
DavisVaughan, line 1283: Along the way I nearly changed `across()` to recycle inputs to the group size rather than recycling them to their common size. I think that would have been a mistake so I've added a test to prevent us from ever thinking of doing this.

tests/testthat/test-across.R
```diff
@@ -1256,6 +1272,431 @@ test_that("rowwise() preserves list-cols iff no `.fns` (#5951, #6264)", {
...
+  expect_snapshot(error = TRUE, {
+    # TODO: This error is bad
+    mutate(df, across(c(x, y), fn))
+  })
+  expect_snapshot(error = TRUE, {
+    mutate(df, (across(c(x, y), fn)))
+  })
+})
+
+test_that("`if_any()` and `if_all()` have consistent behavior across `filter()` and `mutate()`", {
```
DavisVaughan, line 1338: This is a mega test to make sure things are consistent everywhere.

It's obviously a lot of tests, but I think we really do need them all to be sure we aren't missing an edge case. These are all very hard to reason about since there are so many dimensions that intersect (filter vs mutate, expansion vs evaluation, groups vs no groups, etc)

Let's put it this way, I feel way more confident about this now that we have this test that hits every edge case

R/across.R
```diff
@@ -608,79 +664,67 @@ dplyr_quosures <- function(...) {
...
+    x <- list(!!!quos)
+    ns <- asNamespace("dplyr")
 
-  combine <- function(x, y) {
-    if (is_null(x)) {
-      y
-    } else {
-      call(op, x, y)
-    }
-  }
```
lionel-, line 678: This previous expansion was written with dbplyr and other backends in mind. We thought the dplyr expansions could be used as is in these other settings, removing some work for them as they would already know how to deal with the bare expansion.

Unfortunately, as the linked issues reveal, these bare expansions don't work that well for the dplyr backend because they are missing things like input validation. If we add these in the expansion, then the original purpose of generic translation is defeated.

We've never actually pushed towards using these expansions in other packages, so although this feels like a step backward, we don't currently lose anything by making the expansion untranslatable.

@DavisVaughan Maybe add a comment about why we still need the expansion at all (to avoid tidyselect getting evaluated on every group).

lionel- (approved):

PR #7750, opened 2025-10-29
Store `expr` not `quo` for `last_dplyr_warnings()`

### LINKED ISSUE #7649: Possible memory leak involving warnings
I recently updated my `dplyr` version (late to the party), and I'm hitting some increased memory usage. I've traced it back to how warnings are handled. Beginning in `dplyr 1.1.1`, I get the following output:
``` r
library(dplyr)
#> 
#> Attaching package: 'dplyr'
#> 
#>     filter, lag
#> 

identity <- function(x, warn) {
    if (warn) {
        warning("fake warning")
    }
    x
}

df <- tibble::tibble(e = rep(1, 1e8))

print(gc())
#>             used  (Mb) gc trigger   (Mb)  max used  (Mb)
#> Ncells    620941  33.2    1306337   69.8   1306337  69.8
#> Vcells 101049658 771.0  148096356 1129.9 101084366 771.3

df <- df %>% mutate(e = identity(e, warn = TRUE))
#> Warning: There was 1 warning in `mutate()`.
#> ℹ In argument: `e = identity(e, warn = TRUE)`.
#> Caused by warning in `identity()`:
#> ! fake warning

print(gc())
#>             used  (Mb) gc trigger   (Mb)  max used  (Mb)
#> Ncells    729780  39.0    1306337   69.8   1306337  69.8
#> Vcells 101287706 772.8  148096356 1129.9 102369359 781.1

rm(df)
print(gc())
#>             used  (Mb) gc trigger   (Mb)  max used  (Mb)
#> Ncells    729742  39.0    1306337   69.8   1306337  69.8
#> Vcells 101287654 772.8  148096356 1129.9 102369359 781.1
```

<sup>Created on 2025-01-31 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

If I restart R and rerun with `warn = FALSE`, the final memory usage is only 7.9 MB rather than 772.8 MB. Additionally, if I rewrite the mutate to avoid using a pipe via `df <- mutate(df, e = identity(e, warn = TRUE))`, the final memory usage is only 8.8 MB. Switching the pipe to `|>` also yields low memory usage. Under `dplyr 1.1.0`, the above reprex yields 18.8 MB.

I don't have a full appreciation for whether warnings would capture my environment, but I'm wondering if that's perhaps happening within either base R or dplyr's own record of warnings.

DavisVaughan:
Closes #7649 

Issue is described in https://github.com/tidyverse/dplyr/issues/7649#issuecomment-3464151946

### DIFF
--- a/R/conditions.R
+++ b/R/conditions.R
@@ -42,7 +42,7 @@ cnd_bullet_rowwise_unlist <- function() {
   if (peek_mask()$is_rowwise()) {
     glue_data(
       peek_error_context(),
-      "Did you mean: `{error_name} = list({quo_as_label(error_quo)})` ?"
+      "Did you mean: `{error_name} = list({expr_as_label(error_expr)})` ?"
     )
   }
 }
@@ -66,8 +66,7 @@ is_data_pronoun <- function(x) {
 }
 
 # Because as_label() strips off .data$<> and .data[[<>]]
-quo_as_label <- function(quo) {
-  expr <- quo_get_expr(quo)
+expr_as_label <- function(expr) {
   if (is_data_pronoun(expr)) {
     deparse(expr)[[1]]
   } else {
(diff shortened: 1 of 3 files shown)


R/conditions.R
```diff
@@ -87,15 +86,17 @@ new_error_context <- function(dots, i, mask) {
...
-    # Saving the quosure rather than the result of `quo_as_label()` to avoid
-    # slow label creation unless required
+    # Saving the expression rather than the result of `expr_as_label()` to avoid
+    # slow label creation unless required. Not saving the quosure itself because
+    # carrying around its environment past the scope of a dplyr verb's lifetime
+    # can be very expensive (#7649)!
     env(
       error_name = names(dots)[[i]],
-      error_quo = dots[[i]],
+      error_expr = quo_get_expr(dots[[i]]),
```
lionel-, line 99: I haven't looked where the quosure is created and if there might be injected quosures in there. If that's the case, the safer operation is `quo_squash()`. It recursively unwraps quosures (and so has some overhead).

lionel- (approved):

R/conditions.R
DavisVaughan: For our purposes, in the only place we previously used `error_quo` we were extracting the expression with `expr <- quo_get_expr(quo)`, so I think this is safe enough (at the very least it retains current behavior)

Issue #7760, opened 2025-11-08
Soft deprecate `dplyr.legacy_locale` global option usage


DavisVaughan:
It was always meant to be temporary, and was documented as such, so we should start the process of removal

PR #7766, opened 2025-11-12
Deprecate `options(dplyr.legacy_locale =)` usage

DavisVaughan:
Closes https://github.com/tidyverse/dplyr/issues/7760

This is what caused both `arrange()` and `group_by() |> summarise()` to fall back to the old system-locale ordering, rather than using the new C locale approach and requiring `arrange(.locale =)` instead.

It was always meant to be a temporary global option that people could slap at the top of their scripts if they just didn't want to deal with it right now.

### DIFF
--- a/R/group-by.R
+++ b/R/group-by.R
@@ -60,7 +60,10 @@
 #' Prior to dplyr 1.1.0, character vector grouping columns were ordered in the
 #' system locale. Setting the global option `dplyr.legacy_locale` to `TRUE`
 #' retains this legacy behavior, but this has been deprecated. Update existing
-#' code to explicitly call `arrange(.locale = )` instead.
+#' code to explicitly call `arrange(.locale = )` instead. Run
+#' `Sys.getlocale("LC_COLLATE")` to determine your system locale, and compare
+#' that against the list in [stringi::stri_locale_list()] to find an appropriate
+#' value for `.locale`, i.e. for American English, `"en_US"`.
 #'
 #' @export
 #' @examples
--- a/R/locale.R
+++ b/R/locale.R
@@ -36,7 +36,10 @@
 #' Prior to dplyr 1.1.0, character columns were ordered in the system locale.
 #' Setting the global option `dplyr.legacy_locale` to `TRUE` retains this legacy
 #' behavior, but this has been deprecated. Update existing code to explicitly
-#' call `arrange(.locale = )` instead.
+#' call `arrange(.locale = )` instead. Run `Sys.getlocale("LC_COLLATE")` to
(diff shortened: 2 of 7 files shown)


R/locale.R
```diff
@@ -64,30 +63,44 @@
...
+  # straight to `deprecate_warn()` since this is only an end user facing option.
+  lifecycle::deprecate_warn(
+    when = "1.2.0",
+    what = I("`options(dplyr.legacy_locale =)`"),
+    details = c(
+      i = "If needed for `arrange()`, use `arrange(.locale =)` instead.",
+      i = "If needed for `group_by() |> summarise()`, follow up with an additional `arrange(.locale =)` call."
+    ),
+    user_env = globalenv()
+  )
```
DavisVaughan, line 101: This is the most important paragraph to read, because this deprecation process is very special since it is a global option that is only used by end users:

- We force `user_env = globalenv()` to make usage always look "direct" to lifecycle (because an end user "directly" set the global option)
- We start with once per session `deprecate_warn()` rather than `deprecate_soft()` (that wouldn't really mean much anyways since `user_env = globalenv()`)

I started off with `deprecate_soft()` and tried to thread `user_env` up the stack, but it's quite a nightmare. It gets pushed up through `grouped_df()` which is used all over the place, way more than just `group_by()`.

Instead, this seems like a much saner option, especially since it is a global option that should only ever be set by end users anyway (a quick glance at the cran github org confirms that no R package is setting this for users).

R/group-by.R
```diff
@@ -55,13 +55,12 @@
...
 #' Prior to dplyr 1.1.0, character vector grouping columns were ordered in the
-#' system locale. If you need to temporarily revert to this behavior, you can
-#' set the global option `dplyr.legacy_locale` to `TRUE`, but this should be
-#' used sparingly and you should expect this option to be removed in a future
-#' version of dplyr. It is better to update existing code to explicitly call
-#' `arrange(.locale = )` instead. Note that setting `dplyr.legacy_locale` will
-#' also force calls to [arrange()] to use the system locale.
+#' system locale. Setting the global option `dplyr.legacy_locale` to `TRUE`
+#' retains this legacy behavior, but this has been deprecated. Update existing
+#' code to explicitly call `arrange(.locale = )` instead.
```
hadley: What value of `.locale` do you need to replicate `dplyr.legacy_locale`?

R/locale.R
hadley: Based on https://github.com/search?q=org%3Acran%20dplyr.legacy_locale&type=code, this is only going to affect user code, not CRAN packages, so this behaviour sounds reasonable to me.

PR #7770, opened 2025-11-19
Clarify `summarise()`'s `.groups` message

### LINKED ISSUE #6986: Update `.groups` message after `group_by() |> summarize()`
Currently this is the message dplyr emits for `summarize()` after `group_by()` with multiple variables.

``` r
library(dplyr)

mtcars |>
  group_by(vs, am) |>
  summarize(mean_mpg = mean(mpg))
#> `summarise()` has grouped output by 'vs'. You can override using the `.groups`
#> argument.
#> # A tibble: 4 × 3
#> # Groups:   vs [2]
#>      vs    am mean_mpg
#>   <dbl> <dbl>    <dbl>
#> 1     0     0     15.0
#> 2     0     1     19.8
#> 3     1     0     20.7
#> 4     1     1     28.4
```

<sup>Created on 2024-01-29 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

I think this message is still confusing and would be more clear if the grouping message was about the output and it explicitly stated `.groups` is an argument in `summarize()`, e.g.,

```
The output is grouped by `vs`. You can specify grouping structure of the output using the `.groups` argument in `summarize()`.
```

If going this route some things to keep in mind:

- Maybe "result" instead of "output" in two places in the message, or change the description of the `.groups` argument to say "Grouping structure of the output." Basically, we should match what we're calling the "thing" that the function spits out.
- It would be a nice-to-have if US/UK spelling of the function in the message matched what the spelling in the code that generates the message.

An alternative suggestion by @DavisVaughan was

```
summarize() has computed your expressions grouped by (foo, bar), and has regrouped the output by (foo).
```

I think this is an improvement over the current message too, but I'd suggest going with something simpler like the one above.

DavisVaughan:
Closes #6986 

There are two kinds of `.groups` related messages we emit. I will show the before/after for each below.

The precise goals are:
- Tell the user what the implicit behavior was (`"drop_last"` for grouped-df, `"keep"` plus type change for rowwise-df)
- Tell the user what groups were used for their summaries (so they don't get worried)
- Tell the user what the output is grouped by
- Tell the user how to silence the message

```r
df <- tibble(x = 1, y = 2)
```

### With a grouped_df and implicit `"drop_last"`

Before

``` r
df |>
  group_by(x, y) |>
  summarise(a = x + y)
#> `summarise()` has grouped output by 'x'. You can override using the `.groups`
#> argument.
```

After

```r
df |>
  group_by(x, y) |>
  summarise(a = x + y)
#> `summarise()` has regrouped the output by dropping the last grouping column.
#> ℹ Summaries were computed grouped by x and y.
#> ℹ Output is grouped by x.
#> ℹ Use `summarise(.groups = "drop_last")` to silence this message.
```

There is also color now:

Screenshot 2025-11-19 at 4 27 14 PM

### With a rowwise_df and implicit `"keep"` plus conversion to a grouped_df

Before

```r
df |>
  rowwise(x, y) |>
  summarise(a = x + y)
#> `summarise()` has grouped output by 'x', 'y'. You can override using the
#> `.groups` argument.
```

After

```r
df |>
  rowwise(x, y) |>
  summarise(a = x + y)
#> `summarise()` has converted the output from a rowwise data frame to a grouped
#> data frame.
#> ℹ Summaries were computed rowwise.
#> ℹ Output is grouped by x and y.
#> ℹ Use `summarise(.groups = "keep")` to silence this message.
```

There is also color now:

Screenshot 2025-11-19 at 4 27 24 PM

### DIFF
--- a/R/summarise.R
+++ b/R/summarise.R
@@ -477,7 +477,7 @@ inform_implicit_drop_last_for_grouped_df <- function(old, new) {
     i = "Summaries were computed grouped by {cli::col_blue(old)}.",
     i = "Output is grouped by {cli::col_blue(new)}.",
     i = "Use {.code summarise(.groups = \"drop_last\")} to silence this message.",
-    i = "Use {.code summarise(.by = {by})} for per-operation grouping instead."
+    i = "Use {.code summarise(.by = {by})} for {.topic [per-operation grouping](dplyr::dplyr_by)} instead."
   ))
 }
 
--- a/tests/testthat/_snaps/summarise.md
+++ b/tests/testthat/_snaps/summarise.md
@@ -105,7 +105,7 @@
       i Summaries were computed grouped by x and y.
       i Output is grouped by x.
       i Use `summarise(.groups = "drop_last")` to silence this message.
-      i Use `summarise(.by = c(x, y))` for per-operation grouping instead.
+      i Use `summarise(.by = c(x, y))` for per-operation grouping (`?dplyr::dplyr_by`) instead.
     Output
       # A tibble: 1 x 2
       # Groups:   x [1]


R/summarise.R
```diff
@@ -475,10 +467,20 @@ summarise_verbose <- function(.groups, .env) {
...
 }
 
-summarise_inform <- function(..., .env = parent.frame()) {
-  inform(paste0(
-    "`summarise()` ",
-    glue(..., .envir = .env),
-    '. You can override using the `.groups` argument.'
+inform_implicit_drop_last_for_grouped_df <- function(old, new) {
+  inform(cli_format_each_inline(
+    "{.fn summarise} has regrouped the output by dropping the last grouping column.",
```
mine-cetinkaya-rundel: How about

> summarise has dropped the last grouping column and regrouped the output

I feel like this reflects the order of operations better, no?

DavisVaughan: Or maybe its just 

> summarise has dropped the last grouping column

which is exactly what `"drop_last"` says

DavisVaughan: I ended up with

```
`summarise()` has dropped the last grouping column while regrouping the output.
```

Because I think saying something about the _output_ is critical

PR #7775, opened 2025-11-25
Implement `filter_out()`

### LINKED ISSUE #6560: `filter(.missing = )` option to optionally retain missing values
Currently, `filter()`:
- Retains `TRUE`
- Drops `FALSE` and `NA`
- This matches `subset()`

A number of requests have come up in the past desiring:
- Retains `TRUE` and `NA`
- Drops `FALSE`
- This matches `[`

Here are a few:
* https://stackoverflow.com/questions/46378437/how-to-filter-data-without-losing-na-rows-using-dplyr (17k views)
* https://stackoverflow.com/questions/32908589/why-does-dplyrs-filter-drop-na-values-from-a-factor-variable (8k views)
* https://github.com/tidyverse/dplyr/issues/6432 (august this year)
* https://github.com/tidyverse/dplyr/issues/6013
* https://github.com/tidyverse/dplyr/issues/4478 (requesting more docs about this)
* https://github.com/tidyverse/dplyr/issues/3196
* https://github.com/tidyverse/dplyr/issues/812
* https://github.com/tidyverse/dplyr/issues/1527 (more about drop(), but still related to NAs)
* My wife has brought this up to me at least 3 times

This is most apparently annoying when you have multiple columns to filter by

``` r
library(dplyr)

df <- tibble(
  x = c(TRUE, FALSE, NA, NA, NA),
  y = c(NA, TRUE, NA, NA, NA),
  z = c(TRUE, TRUE, TRUE, FALSE, NA)
)
df
#> # A tibble: 5 × 3
#>   x     y     z    
#>   <lgl> <lgl> <lgl>
#> 1 TRUE  NA    TRUE 
#> 2 FALSE TRUE  TRUE 
#> 3 NA    NA    TRUE 
#> 4 NA    NA    FALSE
#> 5 NA    NA    NA

filter(df, x, y, z)
#> # A tibble: 0 × 3
#> # … with 3 variables: x <lgl>, y <lgl>, z <lgl>

filter(df, x | is.na(x), y | is.na(y), z | is.na(z))
#> # A tibble: 3 × 3
#>   x     y     z    
#>   <lgl> <lgl> <lgl>
#> 1 TRUE  NA    TRUE 
#> 2 NA    NA    TRUE 
#> 3 NA    NA    NA
```

I propose a `.missing = c("drop", "keep", "error")` argument to `filter()` that would allow you to optionally keep rows with `NA`.

We'd have to carefully analyze the boolean algebra here to make sure we are being consistent. In particular I think we want to make sure these are the same if we do this, but I think they are:

```r
# these should be the same
filter(df, x, y, .missing = "drop")
filter(df, x & y, .missing = "drop")

# these should be the same
filter(df, x, y, .missing = "keep")
filter(df, x & y, .missing = "keep")
```

The `"drop"` case is probably already consistent because that is what we do today, and the `"keep"` case is probably like this, which seems consistent

``` r
na_to_true <- function(x) {
  x[is.na(x)] <- TRUE
  x
}

na_to_true(TRUE & NA)
#> [1] TRUE
na_to_true(TRUE) & na_to_true(NA)
#> [1] TRUE
```

---

When we do this, we should also think about whether `vec_pall()` or `vec_pany()` could be used in `filter()` in any way, since they are heavily optimized for performance.

### LINKED ISSUE #6891: The case for `drop_rows()` and `keep_rows()` to supersede `filter()`
There have been quite a few requests in the past for an "anti filter", i.e. I want to specify a set of conditions that determine which rows to _drop_. Additionally, it has traditionally been somewhat difficult to explain that `filter()` is about specifying rows to _keep_; that isn't really explained clearly in the verb name. Also, we've also seen in the past that it is mildly confusing that `select()` is about columns and `filter()` is about rows, again, there isn't anything in the verb names to describe the difference.

---

One thing we could consider doing is to add two new very explicit verbs:
- `keep_rows(data, ..., by = , missing = )`
- `drop_rows(data, ..., by = , missing = )`

Where `keep_rows()` is equivalent to `filter()`, and `drop_rows()` is the opposite.

The biggest annoyance when writing a "drop" style expression with `filter()` is that you first have to write a "keep" expression and then painfully invert it. i.e.:

"drop rows from `df` where `a` and `b` and `c` are `TRUE`"

``` r
library(dplyr)

df <- tibble(
  id = c(1, 2, 3),
  a = c(TRUE, NA, TRUE),
  b = c(FALSE, TRUE, TRUE),
  c = c(TRUE, TRUE, TRUE)
)
df
#> # A tibble: 3 × 4
#>      id a     b     c    
#>   <dbl> <lgl> <lgl> <lgl>
#> 1     1 TRUE  FALSE TRUE 
#> 2     2 NA    TRUE  TRUE 
#> 3     3 TRUE  TRUE  TRUE

# "keep rows where a and b and c are TRUE"
# this nicely drops NAs because they don't match our specified criteria
filter(df, a, b, c)
#> # A tibble: 1 × 4
#>      id a     b     c    
#>   <dbl> <lgl> <lgl> <lgl>
#> 1     3 TRUE  TRUE  TRUE

# "drop rows where a and b and c are TRUE"
# this AWFULLY drops NAs because of how filter() works!
# NA doesn't match our criteria so shouldnt be seen as something to drop
filter(df, !(a & b & c))
#> # A tibble: 1 × 4
#>      id a     b     c    
#>   <dbl> <lgl> <lgl> <lgl>
#> 1     1 TRUE  FALSE TRUE
```

Note that even the seemingly correct "drop" expression is actually wrong when it comes to handling missing values. It is fairly hard to get this right.

The `drop_rows()` version would be:

```r
df %>% drop_rows(a, b, c)
#> # A tibble: 2 × 4
#>      id a     b     c    
#>   <dbl> <lgl> <lgl> <lgl>
#> 1     1 TRUE  FALSE TRUE 
#> 2     2 NA    TRUE  TRUE 
```

Where `NA` isn't considered something you "drop" by default, but would be if `missing` was tweaked to whatever we decide means "treat a missing value like `TRUE`".

---

A few other notes:
- `missing` is from https://github.com/tidyverse/dplyr/issues/6560 and controls how missing values are treated. By default, both functions would treat an `NA` as `FALSE` (i.e. missing values are never kept or dropped), but could be made to treat them as `TRUE` or an error. Though I don't think `missing = c("keep", "drop", "error")` works uniformly for both verbs so we'd need to think of another parameterization.
- Both functions would support `if_all()` and `if_any()`, which I think form nice natural sentences. "drop rows if any are NA" sounds pretty good for `drop_rows(df, if_any(c(a, b), is.na))`. That is like `tidyr::drop_na()`.
- Neither would support `across()`, which we have been deprecating from `filter()` for a little while now.
- Both functions would combine multiple conditions using `&`, as that is typically the natural way to combine multiple conditions and you can always get `|` behavior with either an explicit `|` or by using multiple calls to the function. i.e. `df %>% drop_rows(x > 5 | y > 6)` is the same as `df %>% drop_rows(x > 5) %>% drop_rows(y > 6)` (and you can't do that split trick with `&`). `if_any()` can also work for `|` when you need to apply the same function to multiple columns.
- Both functions would support `by`

Some issues related to this:
- https://github.com/tidyverse/dplyr/issues/6888
- https://github.com/tidyverse/dplyr/issues/1527
- https://github.com/tidyverse/dplyr/issues/741
- https://github.com/tidyverse/dplyr/issues/1797

DavisVaughan:
Closes #6560 
Closes #6891

Part of https://github.com/tidyverse/tidyups/pull/30

### DIFF
--- a/tests/testthat/test-filter.R
+++ b/tests/testthat/test-filter.R
@@ -581,7 +581,7 @@ test_that("filter() gives useful error messages", {
   })
 })
 
-test_that("Using data frames in `filter()` and `filter_out()` is defunct (#7758)", {
+test_that("Using data frames in `filter()` is defunct (#7758)", {
   df <- data.frame(x = 1, y = 1)
   gdf <- group_by(df, x)
   rdf <- rowwise(df, x)
@@ -591,19 +591,10 @@ test_that("Using data frames in `filter()` and `filter_out()` is defunct (#7758)
     filter(df, across(everything(), ~ .x > 0))
   })
   expect_snapshot(error = TRUE, {
-    filter_out(df, across(everything(), ~ .x > 0))
-  })
-  expect_snapshot(error = TRUE, {
     filter(gdf, across(everything(), ~ .x > 0))
   })
   expect_snapshot(error = TRUE, {
-    filter_out(gdf, across(everything(), ~ .x > 0))
-  })
-  expect_snapshot(error = TRUE, {
     filter(rdf, across(everything(), ~ .x > 0))
-  })
-  expect_snapshot(error = TRUE, {
-    filter_out(rdf, across(everything(), ~ .x > 0))
   })
 
   # Can't filter with a data frame of logicals (same as the `across()` case)
@@ -611,19 +602,10 @@ test_that("Using data frames in `filter()` and `filter_out()` is defunct (#7758)
     filter(df, tibble(x > 0, y > 0))
   })
   expect_snapshot(error = TRUE, {
-    filter_out(df, tibble(x > 0, y > 0))
-  })
-  expect_snapshot(error = TRUE, {
     filter(gdf, tibble(x > 0, y > 0))
   })
   expect_snapshot(error = TRUE, {
-    filter_out(gdf, tibble(x > 0, y > 0))
-  })
-  expect_snapshot(error = TRUE, {
     filter(rdf, tibble(x > 0, y > 0))
-  })
-  expect_snapshot(error = TRUE, {
-    filter_out(rdf, tibble(x > 0, y > 0))
   })
 })
 
--- a/tests/testthat/_snaps/filter.md
+++ b/tests/testthat/_snaps/filter.md
@@ -186,7 +186,7 @@
       Caused by error:
       ! {
 
-# Using data frames in `filter()` and `filter_out()` is defunct (#7758)
+# Using data frames in `filter()` is defunct (#7758)
 
     Code
       filter(df, across(everything(), ~ .x > 0))
@@ -200,32 +200,9 @@
 ---
 
     Code
-      filter_out(df, across(everything(), ~ .x > 0))
-    Condition
-      Error in `filter_out()`:
-      i In argument: `across(everything(), ~.x > 0)`.
-      Caused by error:
-      ! `..1` must be a logical vector, not a <tbl_df/tbl/data.frame> object.
-      i If you used `across()` to generate this data frame, please use `if_any()` or `if_all()` instead.
-
----
-
-    Code
       filter(gdf, across(everything(), ~ .x > 0))
     Condition
       Error in `filter()`:
-      i In argument: `across(everything(), ~.x > 0)`.
-      i In group 1: `x = 1`.
-      Caused by error:
-      ! `..1` must be a logical vector, not a <tbl_df/tbl/data.frame> object.
-      i If you used `across()` to generate this data frame, please use `if_any()` or `if_all()` instead.
-
----
-
-    Code
-      filter_out(gdf, across(everything(), ~ .x > 0))
-    Condition
-      Error in `filter_out()`:
       i In argument: `across(everything(), ~.x > 0)`.
       i In group 1: `x = 1`.
       Caused by error:
@@ -247,32 +224,9 @@
 ---
 
     Code
-      filter_out(rdf, across(everything(), ~ .x > 0))
-    Condition
-      Error in `filter_out()`:
-      i In argument: `across(everything(), ~.x > 0)`.
-      i In row 1.
-      Caused by error:
-      ! `..1` must be a logical vector, not a <tbl_df/tbl/data.frame> object.
-      i If you used `across()` to generate this data frame, please use `if_any()` or `if_all()` instead.
-
----
-
-    Code
       filter(df, tibble(x > 0, y > 0))
     Condition
       Error in `filter()`:
-      i In argument: `tibble(x > 0, y > 0)`.
-      Caused by error:
-      ! `..1` must be a logical vector, not a <tbl_df/tbl/data.frame> object.
-      i If you used `across()` to generate this data frame, please use `if_any()` or `if_all()` instead.
-
----
-
-    Code
-      filter_out(df, tibble(x > 0, y > 0))
-    Condition
-      Error in `filter_out()`:
       i In argument: `tibble(x > 0, y > 0)`.
       Caused by error:
       ! `..1` must be a logical vector, not a <tbl_df/tbl/data.frame> object.
@@ -293,33 +247,9 @@
 ---
 
     Code
-      filter_out(gdf, tibble(x > 0, y > 0))
-    Condition
-      Error in `filter_out()`:
-      i In argument: `tibble(x > 0, y > 0)`.
-      i In group 1: `x = 1`.
-      Caused by error:
-      ! `..1` must be a logical vector, not a <tbl_df/tbl/data.frame> object.
-      i If you used `across()` to generate this data frame, please use `if_any()` or `if_all()` instead.
-
----
-
-    Code
       filter(rdf, tibble(x > 0, y > 0))
     Condition
       Error in `filter()`:
-      i In argument: `tibble(x > 0, y > 0)`.
-      i In row 1.
-      Caused by error:
-      ! `..1` must be a logical vector, not a <tbl_df/tbl/data.frame> object.
-      i If you used `across()` to generate this data frame, please use `if_any()` or `if_all()` instead.
-
----
-
-    Code
-      filter_out(rdf, tibble(x > 0, y > 0))
-    Condition
-      Error in `filter_out()`:
       i In argument: `tibble(x > 0, y > 0)`.
       i In row 1.
       Caused by error:


NEWS.md
```diff
@@ -1,5 +1,23 @@
...
+
+  ```r
+  df |> filter(count != 0 | is.na(count))
+
+  df |> filter_out(count == 0)
+  ```
+
+  With `filter()`, you must provide a "negative" condition of `!= 0` and must explicitly guard against accidentally dropping rows with `NA`. With `filter_out()`, you directly specify rows to drop and you don't have to guard against dropping rows with `NA`, which tends to result in much clearer code.
+
+  This work is a result of [Tidyup 8: Expanding the `filter()` family](https://github.com/tidyverse/tidyups/pull/30), with a lot of great feedback from the community (#6560, #6891).
```
DavisVaughan: Focus here

R/filter.R
```diff
@@ -107,52 +204,101 @@
...
   UseMethod("filter")
 }
 
+#' @rdname filter
+#' @export
+filter_out <- function(.data, ..., .by = NULL, .preserve = FALSE) {
+  check_by_typo(...)
+  check_not_both_by_and_preserve({{ .by }}, .preserve)
+  UseMethod("filter_out")
+}
```
DavisVaughan, line 222: The actual implementation is quite simple

- Extract out a common `filter_impl()`
- Provide `.verb = "filter"` or `.verb = "filter_out"`
    - If `filter_out`, provide `invert = TRUE` to the C implementation of filter, which inverts the final result on the way out

R/filter.R
```diff
@@ -1,17 +1,108 @@
-#' Keep rows that match a condition
```
DavisVaughan, line 1: Did a full doc overhaul. Things worth spending time on:

- The Description section
    - Note that I've declared `filter_out()` as experimental
- `@section Missing values:`
- Examples

src/filter.cpp
```diff
@@ -181,6 +182,12 @@ SEXP dplyr_mask_eval_all_filter(SEXP quos,
     DPLYR_MASK_ITERATION_FINALISE();
   }
 
+  if (LOGICAL_ELT(invert, 0)) {
+    for (R_xlen_t i = 0; i < n; ++i) {
+      p_keep[i] = !p_keep[i];
+    }
+  }
```
DavisVaughan, line 189: That's it!

tests/testthat/test-filter.R
DavisVaughan, line 1: I carefully looked at all `filter()` tests and added a corresponding `filter_out()` test if it felt like the test wasn't too niche and generally tested some kind of invariant (0 row behavior, no input behavior, etc)

vignettes/colwise.Rmd
```diff
@@ -211,24 +211,24 @@ For some verbs, like `group_by()`, `count()` and `distinct()`, you don't need to
 
 `across()` doesn't work with `select()` or `rename()` because they already use tidy select syntax; if you want to transform column names with a function, you can use  `rename_with()`.
 
-### filter()
```
DavisVaughan, line 214: Tweaked the vignettes only where it felt like using `filter_out()` was a noticeable improvement

NEWS.md
```diff
@@ -1,5 +1,23 @@
...
+* New experimental `filter_out()` companion to `filter()`.
+
+  * Use `filter()` when specifying rows to _keep_.
+
+  * Use `filter_out()` when specifying rows to _drop_.
+
+  `filter_out()` simplifies cases where you would have previously used a `filter()` to drop rows. It is particularly useful when missing values are involved. For example, to drop rows where the `count` is zero:
+
+  ```r
+  df |> filter(count != 0 | is.na(count))
```
hadley, line 12: Compelling

R/filter.R
```diff
@@ -1,17 +1,108 @@
...
 #'
-#' The `filter()` function is used to subset the rows of
-#' `.data`, applying the expressions in `...` to the column values to determine which
-#' rows should be retained. It can be applied to both grouped and ungrouped data (see [group_by()] and
-#' [ungroup()]). However, dplyr is not yet smart enough to optimise the filtering
-#' operation on grouped datasets that do not need grouped calculations. For this
-#' reason, filtering is often considerably faster on ungrouped data.
+#' - `filter()` _keeps_ rows where the conditions evaluate to `TRUE`.
+#'
+#' - `r lifecycle::badge("experimental")` `filter_out()` _drops_ rows where the
```
hadley: I think because of the way that pkgdown extract badges, in the reference index this might make it look like `filter()` is also experimental.

IMO it's sufficiently well thought through that you don't need to mark it as experimental.

R/filter.R
```diff
@@ -1,17 +1,108 @@
...
+#' Keep or drop rows that match a condition
 #'
-#' The `filter()` function is used to subset a data frame,
-#' retaining all rows that satisfy your conditions.
-#' To be retained, the row must produce a value of `TRUE` for all conditions.
-#' Note that when a condition evaluates to `NA`
-#' the row will be dropped, unlike base subsetting with `[`.
+#' @description
+#' These functions are used to subset a data frame, applying the expressions in
+#' `...` to determine which rows should be kept or dropped.
```
hadley:
```suggestion
#' These functions are used to subset a data frame, applying the expressions in
#' `...` to determine which rows should be kept (with `filter()`) or dropped (with `filter_out()`).
```

Then maybe you could drop the bulleted list?

R/filter.R
```diff
@@ -1,17 +1,108 @@
...
+#' behavior can impact how you write your conditions when missing values are
+#' involved. See the section on `Missing values` for important details and
+#' examples.
+#'
+#' @inheritParams arrange
+#' @inheritParams args_by
+#'
+#' @param ... <[`data-masking`][rlang::args_data_masking]> Expressions that
+#'   return a logical value, and are defined in terms of the variables in
+#'   `.data`. If multiple expressions are included, they are combined with the
```
hadley:
```suggestion
#' @param ... <[`data-masking`][rlang::args_data_masking]> Expressions that
#'   return a logical vector, defined in terms of the variables in
#'   `.data`. If multiple expressions are included, they are combined with the
```

?

R/filter.R
```diff
@@ -1,17 +1,108 @@
...
+#'   `&` operator. Only rows for which all conditions evaluate to `TRUE` are
```
hadley: Mention `when_any()` here?

R/filter.R
```diff
@@ -1,17 +1,108 @@
...
+#' @returns
+#' An object of the same type as `.data`. The output has the following
+#' properties:
+#'
+#' * Rows are a subset of the input, but appear in the same order.
+#' * Columns are not modified.
+#' * The number of groups may be reduced (if `.preserve` is not `TRUE`).
+#' * Data frame attributes are preserved.
+#'
+#' @section Missing values:
```
hadley, line 38:
```suggestion
#' ## Missing values:
```

And move up in to the description? I don't know if we have an official position on this style.

R/filter.R
```diff
@@ -1,17 +1,108 @@
...
+#'
+#' Both `filter()` and `filter_out()` treat `NA` like `FALSE`. This results in
+#' the following behavior:
+#'
+#' - `filter()` _drops_ both `NA` and `FALSE`.
+#'
+#' - `filter_out()` _keeps_ both `NA` and `FALSE`.
+#'
```
hadley, line 46: This means that `filter(df, whatever) + filter_out(df, whatever)` captures all rows in `df`.

R/filter.R
hadley: Given all the other sections below, probably not worth it.

R/filter.R
```diff
@@ -37,68 +128,74 @@
...
 #'
+#' # But this also drops rows with an `NA` hair color! To retain those:
+#' starwars |> filter(hair_color != "blond" | is.na(hair_color))
 #'
-#' # To refer to column names that are stored as strings, use the `.data` pronoun:
+#' # But explicit `NA` handling like this can quickly get unwieldy, especially
+#' # with multiple conditions. Since your intent was to specify rows to drop
+#' # rather than rows to keep, use `filter_out()`. This also removes the need
+#' # for any explicit `NA` handling.
+#' starwars |> filter_out(hair_color == "blond")
```
hadley, line 194: I don't know if you need to go in to the details again here (although maybe it's useful for the folks who only read examples?)

tests/testthat/test-filter.R
hadley: This feels a little heavy to me (given that you're basically testing the same five lines of C code again and again and again. But it's also not terrible to have the invariants clearly stated in the tests, independent of the implementation.

tests/testthat/test-filter.R
```diff
@@ -501,7 +581,7 @@ test_that("filter() gives useful error messages", {
   })
 })
 
-test_that("Using data frames in `filter()` is defunct (#7758)", {
+test_that("Using data frames in `filter()` and `filter_out()` is defunct (#7758)", {
```
hadley: I'd drop these tests — it's weird to test that it's not supported in a function that never supported it. (And it adds a bunch of noise in the snapshot tests).

vignettes/colwise.Rmd
```diff
@@ -211,24 +211,24 @@ For some verbs, like `group_by()`, `count()` and `distinct()`, you don't need to
...
+We cannot directly use `across()` in `filter()` or `filter_out()` because we need an extra step to combine
+the results into a single logical vector. To that end, `filter()` and `filter_out()` have two special purpose companion functions:
 
 * `if_any()` keeps the rows where the predicate is true for *at least one* selected
   column:
 
 ```{r}
 starwars |>
-  filter(if_any(everything(), ~ !is.na(.x)))
+  filter_out(if_any(everything(), is.na))
```
hadley, line 224: So much easier to read!

tests/testthat/test-filter.R
DavisVaughan: > invariants clearly stated in the tests, independent of the implementation

yea that's what I was going for, especially if we swap the C++ implementation for `vctrs::vec_pall()` in the future, I want it to feel well tested regardless of the implementation details

PR #7777, opened 2025-11-26
Implement `when_any()` and `when_all()`

DavisVaughan:
Branched from #7775 

Part of https://github.com/tidyverse/tidyups/pull/30

### DIFF
--- a/R/when.R
+++ b/R/when.R
@@ -102,14 +102,6 @@
 #' # ---------------------------------------------------------------------------
 #' # Programming with `when_any()` and `when_all()`
 #'
-#' # When no inputs are provided, these functions are consistent with `any()`
-#' # and `all()`
-#' any()
-#' when_any(size = 1)
-#'
-#' all()
-#' when_all(size = 1)
-#'
 #' # The `size` argument is useful for making these functions size stable when
 #' # you aren't sure how many inputs you're going to receive
 #' size <- length(x)
@@ -125,6 +117,14 @@
 #' # Zero inputs (without `size`, this would return `logical()`)
 #' inputs <- list()
 #' when_all(!!!inputs, size = size)
+#'
+#' # When no inputs are provided, these functions are consistent with `any()`
+#' # and `all()`
+#' any()
+#' when_any(size = 1)
+#'
+#' all()
+#' when_all(size = 1)
 NULL
 
 #' @rdname when-any-all
--- a/man/when-any-all.Rd
+++ b/man/when-any-all.Rd
@@ -108,14 +108,6 @@ countries |>
 # ---------------------------------------------------------------------------
 # Programming with `when_any()` and `when_all()`
 
-# When no inputs are provided, these functions are consistent with `any()`
-# and `all()`
-any()
-when_any(size = 1)
-
-all()
-when_all(size = 1)
-
 # The `size` argument is useful for making these functions size stable when
 # you aren't sure how many inputs you're going to receive
 size <- length(x)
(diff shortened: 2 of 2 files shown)


R/when.R
```diff
@@ -0,0 +1,162 @@
+#' Elementwise `any()` and `all()`
+#'
+#' @description
+#' These functions are variants of [any()] and [all()] that work elementwise
+#' across multiple inputs. You can also think of these functions as generalizing
+#' [`|`] and [`&`] to any number of inputs, rather than just two, for example:
+#'
+#' - `when_any(x, y, z)` is equivalent to `x | y | z`.
```
hadley, line 8: Are we sure `when_` is the correct prefix here? Like would you also expect `when_min()` and `when_max()`?

R/when.R
```diff
@@ -0,0 +1,162 @@
...
+#'   filter(
+#'     (name %in% c("US", "CA") & between(score, 200, 300)) |
+#'       (name %in% c("PR", "RU") & between(score, 100, 200))
+#'   )
+#'
+#' # With `when_any()`, you drop the explicit `|`, the extra `()`, and your
+#' # conditions are all indented to the same level
+#' countries |>
+#'   filter(when_any(
+#'     name %in% c("US", "CA") & between(score, 200, 300),
```
hadley, line 91: The indenting is so much nicer with `when_any()` 😄

R/when.R
```diff
@@ -0,0 +1,162 @@
...
+#'     name %in% c("PR", "RU") & between(score, 100, 200)
+#'   ))
+#'
+#' # ---------------------------------------------------------------------------
+#' # Programming with `when_any()` and `when_all()`
+#'
+#' # When no inputs are provided, these functions are consistent with `any()`
+#' # and `all()`
+#' any()
+#' when_any(size = 1)
```
hadley: Bit weird to use `size` before you describe what it does?

tests/testthat/test-when.R
```diff
@@ -0,0 +1,131 @@
...
+  expect_snapshot(error = TRUE, {
+    when_all(TRUE, c(TRUE, FALSE))
+  })
+
+  expect_snapshot(error = TRUE, {
+    when_any(TRUE, size = 2)
+  })
+  expect_snapshot(error = TRUE, {
+    when_all(TRUE, size = 2)
+  })
```
hadley, line 71:
```suggestion
  expect_snapshot(error = TRUE, {
    when_any(TRUE, c(TRUE, FALSE))
    when_all(TRUE, c(TRUE, FALSE))

    when_any(TRUE, size = 2)
    when_all(TRUE, size = 2)
  })
```

is my personal preference. Yes, it's slightly less safer because it doesn't check every expression errors, but that only matters on the first run, and it makes it way easier to see the whole at a glance.

R/filter.R
```diff
@@ -10,7 +10,8 @@
 #'   conditions evaluate to `TRUE`.
 #'
 #' Multiple conditions can be supplied separated by a comma. These will be
-#' combined with the `&` operator.
+#' combined with the `&` operator. To combine comma separated conditions using
+#' `|` instead, wrap them in [when_any()].
```
DavisVaughan, line 10: Mention `when_any()` in the `...` docs too https://github.com/tidyverse/dplyr/pull/7775#discussion_r2590959733

R/when.R
DavisVaughan: The full family is really something like

```r
pmin()
pmax()

pany()
pall()

psum()
pprod()
pmean()

pfirst() # first non missing
plast() # last non missing
```

See also https://cran.r-project.org/web/packages/kit/refman/kit.html#parallel-funs

And no, I don't think that `when_` would work as a prefix for all of these.

It's more like `reduce_*()` because it takes `K` vectors of size `N` and returns `1` vector of size `N`. But that's not so fun to read in `filter(reduce_any(...))`.

So I think of `when_any()` and `when_all()` as readable wrappers, and really `vec_pany()` and `vec_pall()` are the "true" additions to the `pmin()` and `pmax()` family.

PR #7780, opened 2025-12-03
Make `bind_rows()` a special S3 generic

### LINKED ISSUE #6905: `bind_rows()` for database tables
It would be nice if `bind_rows()` would work for database tables.

``` r
library(purrr)
con <- DBI::dbConnect(RSQLite::SQLite(), ":memory:")

# Create random slices of mtcars
tables <-
  1:5 |>
  map(function(n) {
    tbl_name <- paste0("mtcars_", n)
    copy_to(con, slice_sample(mtcars, n = 3), tbl_name)

    tbl(con, tbl_name)
  })
bind_rows(!!!tables)
#> Error in `bind_rows()`:
#> ! Argument 1 must be a data frame or a named atomic vector.
#> Backtrace:
#>     ▆
#>  1. └─dplyr::bind_rows(!!!tables)
#>  2.   └─rlang::abort(glue("Argument {i} must be a data frame or a named atomic vector."))
```

<sup>Created on 2023-08-07 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

I know there were issues with the double dispatch and what happens if local tables and database tables are mixed.
But I think it would be sufficient if either all inputs are local tables or database tables.
This was also requested again in dbplyr: https://github.com/tidyverse/dbplyr/issues/1342.

DavisVaughan:
Closes #6905 

I've tried this out "for real" over here and it seems to work https://github.com/DavisVaughan/testbindrows

This is a very special S3 generic because:

- It's only argument in the generic is `...`

- It splits the `...` into two categories by a new convention:
    - Named inputs where the name starts with `.` are treated as _optional arguments_
    - Everything else is treated as an input to bind with

This convention is used by a new `dplyr::bind_rows_dots()` utility that we could export and require that S3 method authors call at the beginning of their method.

All S3 method authors must use a `.` prefix for any additional optional arguments they want to include, so that `bind_rows_dots()` can correctly ignore them.

I don't think that an S3 method can use a tidyselect style optional argument, even with a `.` prefix, because we have to collect all arguments in the generic. This seems okay with me.

### DIFF
--- a/R/bind-rows.R
+++ b/R/bind-rows.R
@@ -24,38 +24,47 @@
 #' Methods available in currently loaded packages:
 #' \Sexpr[stage=render,results=rd]{dplyr:::methods_rd("bind_rows")}.
 #'
-#' `bind_rows()` is a special S3 generic due to the fact that its first
-#' argument is `...`. S3 methods for `bind_rows()` are written as:
+#' `bind_rows()` is a special S3 generic due to the fact that it dispatches on
+#' `...`. S3 methods for `bind_rows()` are written as:
 #'
 #' ```r
 #' #' @importFrom dplyr bind_rows
 #' #' @export
 #' bind_rows.my_subclass <- function(..., .id = NULL) {
+#'   dots <- dplyr::bind_rows_dots(...)
+#'   # Implementation
 #' }
 #' ```
 #'
 #' Dispatch is performed on the _first_ argument, so order matters when data
 #' frames of different types are being combined.
 #'
-#' For S3 methods authors:
+#' The very first thing you must do in your S3 method is call
+#' `dplyr::bind_rows_dots()`. This does the following:
 #'
-#' - `...` will always contain at least 1 input. The first input will have the
-#'   S3 class that dispatch was performed on, but no other inputs are validated.
+#' - Applies all legacy flattening behavior.
 #'
-#' - `...` must be captured by [rlang::list2()].
+#' - Removes all `NULL` values.
 #'
-#' - `bind_rows()`'s S3 generic will preprocess `...` by applying all legacy
-#'   flattening behavior and removing any `NULL` values. Dispatch occurs on the
-#'   first argument remaining after this preprocessing, and S3 methods should
-#'   expect that `...` contains a flat list of objects to bind.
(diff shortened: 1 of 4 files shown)


R/bind-rows.R
```diff
@@ -7,11 +7,56 @@
...
+#' - `...` must be captured by [rlang::list2()].
+#'
+#' - `bind_rows()`'s S3 generic will preprocess `...` by applying all legacy
+#'   flattening behavior and removing any `NULL` values. Dispatch occurs on the
+#'   first argument remaining after this preprocessing, and S3 methods should
+#'   expect that `...` contains a flat list of objects to bind.
+#'
+#' - `bind_rows()` S3 method signatures must exactly match the signature of the
+#'   `bind_rows()` generic. Due to how this S3 generic works, additional method
+#'   specific arguments are not allowed.
```
DavisVaughan: Update:

With the new approach of using `bind_rows_dots()`, we no longer have these restrictions.

---

**Everything below this no longer applies**

I'm a bit worried about how this might lock us into a specific `bind_rows()` signature

It basically means that we can't ever add any new optional arguments to `bind_rows()` in the future.

- We can't add extra arguments to the `data.frame` S3 method that aren't already in the S3 generic, because the S3 generic won't know how to pass that argument down

- Adding an extra argument `foo` to the S3 generic means that the S3 method `bind_rows.my_class` will immediately start to see `foo` as an input to `...` that it should bind, since it won't have `foo` as an argument. See https://github.com/DavisVaughan/testbindrows/commit/862670cad52ea0a53c7583efd129ccda24da1bcc where I tried this by creating an S3 method that lacks the current `.id` argument

I don't love that

R/bind-rows.R
```diff
@@ -7,11 +7,65 @@
...
+#' #' @export
+#' bind_rows.tbl_lazy <- function(..., .copy = FALSE) {
+#'   dots <- dplyr::bind_rows_dots(...)
+#'   # Implementation
+#' }
+#' ```
+#'
+#' Notice how this implementation both adds a new argument, `.copy`, and
+#' refuses to implement the `.id` argument from the data frame method. Both of
+#' these are allowed as long as `bind_rows_dots()` is called.
```
DavisVaughan, line 67: The all important docs for this

PR #7762, opened 2025-11-08
Advance all deprecations

DavisVaughan:
Deprecationpalooza!

In addition to these, see https://github.com/tidyverse/dplyr/issues/7761, https://github.com/tidyverse/dplyr/issues/7760, https://github.com/tidyverse/dplyr/issues/7758, and https://github.com/tidyverse/dplyr/issues/7757, all of which require a bit more individual care.

---

Revdeps in https://github.com/tidyverse/dplyr/issues/7763 show ~150 issues, but 100 of them are due to removal of the already defunct `dplyr::id()`. People were using `mutate(df, col = id)` and not listing `id` as a `globalVariable()` so now they will get a note about this. I can send all of these people a github issue / email, but I was not planning on doing individual PRs to them. It does seem worth pushing on this though.

```
Remotes:
    vctrs=url::https://github.com/DavisVaughan/vctrs/archive/test/remote-urls.tar.gz
```

Plus set to CRAN lifecycle

```
"97e2bc7a-8a36-4e8f-b963-036934944b6e"
```

PR #7782, opened 2025-12-16
Advance lazyeval and each deprecation

DavisVaughan:
- Advances all underscored (`mutate_()`) verbs from "always warn" to defunct. Deprecated since 0.7.0.

- Advances the each family (`mutate_each()`, `summarise_each()`) from "always warn" to defunct. Deprecated since 0.7.0.
    - These were tied up in the `mutate_()` lazyeval R file, I have split these into their own file.

- Removes `select_vars_()` and `rename_vars_()` outright, since we also removed `select_vars()` and `rename_vars()` in #7762.

- Added a NEWS bullet for the defunct functions that were fully removed in https://github.com/tidyverse/dplyr/pull/7762/changes/f8aca67e267cb7a133baa1d5fc4a7f71de985a74, I forgot to do that then, but needed a place to mention the `select_vars_()` and `rename_vars_()` removal.

As mentioned in https://github.com/tidyverse/dplyr/pull/7764, I'm expecting 50 or so breakages. But it's time to get rid of these.


R/defunct-each.R
```diff
@@ -0,0 +1,48 @@
...
+#'
+#' `mutate_each()` and `summarise_each()` are deprecated in favour of
+#' the new [across()] function that works within `summarise()` and `mutate()`.
+#'
+#' @name defunct-each
+#' @keywords internal
+NULL
+
+each_defunct <- function(fun) {
+  lifecycle::deprecate_stop(
```
hadley, line 14: You're making these defunct just because they're still used in packages right? (And you want slightly more informative errors during revdep checks)?

NEWS.md
```diff
@@ -145,6 +145,29 @@
 
 * The following were already deprecated, and are now defunct:
 
+  * All underscored standard evaluation versions of major dplyr verbs. Deprecated in 0.7.0, use the non-underscored version of the verb with unquoting instead, see `vignette("programming")`. This includes:
```
hadley, line 148: I'd put dates next to all the version numbers (claude seems reliable at doing that)

R/defunct-each.R
DavisVaughan: Right, and just for one more release, then we remove entirely.

NEWS.md
DavisVaughan: Will follow up with this on all news bullets

PR #7783, opened 2025-12-16
Fix links to blog posts about `dplyr` releases

etiennebacher:
Those links were giving 404.


etiennebacher:
And just in case it helps, I use [lychee](https://github.com/lycheeverse/lychee) in a few projects to check for dead links. It's a bit flaky sometimes, but it runs in ~5 seconds in CI.

Issue #7732, opened 2025-10-08
Write a replacing and recoding vignette


DavisVaughan:
Some fodder removed from the function docs that could be better served in the vignette

## The 2x2 table

| | Create a new column | Update an existing column |
|--------|--------|--------|
| **Logical conditions** | case_when | replace_when |
| **Values** | recode_values | replace_values | 

## Function connections

```r
#' @section Connection between `case_when()` and `replace_when()`:
#'
#' The following two statements produce identical `result`s:
#'
#' ```r
#' result <- replace_when(x, lhs ~ rhs)
#'
#' result <- case_when(lhs ~ rhs, .default = x, .ptype = x, .size = vec_size(x))
#' result <- vec_set_names(result, vec_names(x))
#' ```
#'
#' If you are replacing a few values within an existing vector, then
#' `replace_when()` is always a better choice because it is type stable and size
#' stable on `x`, pipes better, and better expresses intent.
```

```r
#' @section Connection between `recode_values()` and `case_when()`:
#'
#' While [case_when()] is a general vectorized [if_else()] that uses logical
#' expressions, `recode_values()` is a general vectorized [switch()] that uses
#' values. The following two statements are roughly equivalent:
#'
#' ```
#' case_when(
#'   x %in% c("a", "b") ~ 1,
#'   x %in% "c" ~ 2,
#'   x %in% c("d", "e") ~ 3
#' )
#'
#' recode_values(
#'   x,
#'   c("a", "b") ~ 1,
#'   "c" ~ 2,
#'   c("d", "e") ~ 3
#' )
#' ```
#'
#' For large inputs `recode_values()` is going to be much faster than
#' the equivalent `case_when()` statement.
```

```r
#' @section Connection between `recode_values()` and `replace_values()`:
#'
#' The following two statements produce identical `result`s:
#'
#' ```r
#' result <- replace_values(x, lhs ~ rhs)
#'
#' result <- recode_values(x, lhs ~ rhs, default = x, ptype = x)
#' result <- vec_set_names(result, vec_names(x))
#' ```
```

Replaceable with `replace_values()` or `replace_when()` (from https://github.com/tidyverse/dplyr/pull/7742)
- The idea of `na_if()` with >1 RHS values
- The idea of `na_if()` with a logical condition
- `coalesce(x, 0)` or `coalesce(x, y)` but not `coalesce(x, y, z)`
- `tidyr::replace_na()` 

## SQL connections

```r
[...]

PR #7793, opened 2026-01-22
Add recoding and replacing vignette

DavisVaughan:
Closes https://github.com/tidyverse/dplyr/issues/7732

### DIFF
--- a/vignettes/recoding-replacing.Rmd
+++ b/vignettes/recoding-replacing.Rmd
@@ -118,26 +118,23 @@ Note that missing values must be explicitly handled when setting `.unmatched = "
 
 ## `replace_when()`
 
-Let's assume that some of our racers had a false start, and need to be assigned a penalty of 20 seconds (1/3 of a minute).
-Additionally, some other racers used banned running shoes, and should incur a penalty of 1 minute.
+Let's assume that some of our racers used banned running shoes and are disqualified.
+Also, some racers had a false start and need to incur a 20 second (1/3 minute) penalty.
 
 ```{r}
-id_false_start <- sample(racers$id, size = 10)
-id_false_start
-
-id_banned_shoes <- sample(racers$id, size = 5)
-id_banned_shoes
+id_banned_shoes <- c(2, 10, 15, 32, 65)
+id_false_start <- c(1, 2, 5, 20, 55, 74, 91)
 ```
 
-We could add penalties a few ways.
+We could add this information in a few ways.
 With `case_when()`:
 
 ```{r}
 racers |>
   mutate(
     time = case_when(
+      id %in% id_banned_shoes ~ NA,
       id %in% id_false_start ~ time + 1 / 3,
-      id %in% id_banned_shoes ~ time + 1,
       .default = time
     )
   )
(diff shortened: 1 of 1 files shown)


vignettes/recoding-replacing.Rmd
```diff
@@ -0,0 +1,531 @@
...
+dplyr provides a family of functions for *recoding* columns and *replacing* values within a column.
+These are extremely common operations, so mastering these functions can be a big productivity boost!
+This family can be summarized by the following table:
+
+|                           | **Recoding**      | **Replacing**      |
+|---------------------------|-------------------|--------------------|
+| **Match with conditions** | `case_when()`     | `replace_when()`   |
+| **Match with values**     | `recode_values()` | `replace_values()` |
+
+It's also helpful to define exactly what we mean by recoding vs replacing:
```
hadley: Define the terms before the table?

vignettes/recoding-replacing.Rmd
```diff
@@ -0,0 +1,531 @@
...
+racers |>
+  mutate(
+    time = case_when(
+      id %in% id_false_start ~ time + 1 / 3,
+      id %in% id_banned_shoes ~ time + 1,
+      .default = time
+    )
+  )
+```
+
```
hadley, line 142: I think you probably want to ponder on what happens if they have both a false start and a banned shoe. I'd almost be tempted to tackle this problem like:

```R
racers |>
  mutate(
    banned_shoe = id %in% id_banned_shoes,
    false_start = id %in% id_false_starts,
    time = time + if_else(false_start, 1/3, 0) + if_else(banned_shoe, 1, 0)
  )
```

Maybe it'd  be worth coming up with an example where the clauses aren't so clearly additive? Maybe you could tweak this to a three strikes and you're out policy?

vignettes/recoding-replacing.Rmd
```diff
@@ -0,0 +1,531 @@
...
+      time < 33 ~ "D",
+      .default = "unknown"
+    ) |>
+      factor(levels = c("A", "B", "C", "D", "unknown"))
+  )
+
+tiers
+```
+
+Note how a `case_when()` solution results in `tier` becoming a *character* column, losing its factor class.
```
hadley:
```suggestion
Note that a `case_when()` solution results in `tier` becoming a *character* column, losing its factor class.
```

For some reason that "note how" made me look at the previous chunk, instead of the next chunk.

vignettes/recoding-replacing.Rmd
```diff
@@ -0,0 +1,531 @@
...
+Using a `tribble()`, we can extract out the lookup table into its own standalone data frame.
+
+```{r}
+lookup <- tribble(
+  ~from , ~to                 ,
+      1 , "Strongly disagree" ,
+      2 , "Disagree"          ,
+      3 , "Neutral"           ,
+      4 , "Agree"             ,
+      5 , "Strongly agree"
```
hadley, line 293: Random thought: if we knew that `tribble()` allowed trailing commas somehow (like a sort of type annotation) air could automatically add it here.

vignettes/recoding-replacing.Rmd
```diff
@@ -0,0 +1,531 @@
...
+
+### `if_else()`
+
+`if_else()` is a type of recoding function, as it creates an entirely new column.
+In fact, it's closely tied to `case_when()`:
+
+```{r, eval = FALSE}
+if_else(condition, true, false, missing)
+
+case_when(condition ~ true, !condition ~ false, .default = missing)
```
hadley:
```suggestion
case_when(condition ~ true, !condition ~ false, is.na(condition) ~ missing)
```

Maybe more clear?

vignettes/recoding-replacing.Rmd
```diff
@@ -0,0 +1,531 @@
...
+CASE x
+  WHEN 'E' THEN 'East'
+  WHEN 'W' THEN 'West'
+  WHEN 'N' THEN 'North'
+  WHEN 'S' THEN 'South'
+  ELSE 'Unknown'
+END
+```
+
+As of dbplyr 2.5.1, we don't currently have a translation for `recode_values()` since it is so new, but we expect to have one soon.
```
hadley, line 528: You (claude) should do this!

vignettes/recoding-replacing.Rmd
DavisVaughan: https://github.com/posit-dev/air/issues/462

PR #7807, opened 2026-02-09
Prototype of webr on the landing page

DavisVaughan:
- [ ] Could we add some kind of Plausible integration?
- [ ] Fonts are hardcoded to `font-family: 'Source Code Pro'`. Should it be `var(--bs-font-monospace)` in `webr.css`? Would that just work?
- [ ] Can this be templated enough to live as `usethis::use_pkgdown_webr()`?
    - Package name is easy to extract
    - Sample code would have to be provided somewhere for usethis to inject into the template. It would be package specific.
- [x] Can we get the javascript out of `_pkgdown.yml` and have `pkgdown::preview_site()` still work? Because it doesn't start a real local server and only serves from the file system, we get CORS security errors when trying to run a script on a `file:///` path. Is there another way?
- [x] I don't have a clue what its going to do for plots
- [x] Make sure the "webr can't load" case is decently handled so we don't confuse people if that happens
- [ ] Should `Try it` be a permanent section in the README, even on the GitHub README? It could link to the website and tell you to go there to try it live. Right now the section doesn't show up at all in the GitHub README, but that makes it float kind of awkwardly in your `README.Rmd`.
- [x] Is `in_header` the right place for this? Is that being pulled into _every_ subpage?

https://github.com/user-attachments/assets/f2011dfc-708a-4fdc-9071-ee6c6f4a64ae

### DIFF
--- a/README.md
+++ b/README.md
@@ -121,7 +121,7 @@ On GitHub, the JS never runs, so nothing is displayed.
   outline: 0;
   box-shadow: 0 0 0 0.25rem rgba(13, 110, 253, 0.25);
 }
-.webr-run-btn {
+.webr-run-button {
   margin-bottom: 0.75rem;
   padding: 0.5rem 1.5rem;
   font-size: 0.9rem;
@@ -133,10 +133,10 @@ On GitHub, the JS never runs, so nothing is displayed.
   cursor: pointer;
   transition: background-color 0.15s ease-in-out;
 }
-.webr-run-btn:hover:not(:disabled) {
+.webr-run-button:hover:not(:disabled) {
   background-color: #375a7a;
 }
-.webr-run-btn:disabled {
+.webr-run-button:disabled {
   opacity: 0.65;
   cursor: not-allowed;
 }
@@ -190,7 +190,7 @@ Try it
 
 </div>
 
-<button class="webr-run-btn" disabled>
+<button class="webr-run-button" disabled>
 
 Run Code
 </button>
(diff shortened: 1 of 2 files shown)


pkgdown/assets/webr-readme.html
```diff
@@ -0,0 +1,26 @@
...
+<div id="webr-container">
+
+<div class="webr-status"></div>
+
+<button class="webr-run-btn" disabled>Run Code</button>
+
+<textarea class="webr-editor">starwars |>
+  filter(species == "Human") |>
+  select(name, homeworld) |>
+  arrange(name)</textarea>
```
jennybc: It feels like we would want such a code snippet(s) to be isolated in a `.R` file somewhere, for a whole bunch of reasons. Perhaps that's part of what you mean by:

> Sample code would have to be provided somewhere for usethis to inject into the template. It would be package specific.

DavisVaughan: Yes! Likely some well specified location for this particular use case

Or maybe it can be in `_pkgdown.yml` somewhere, but that wouldn't solve the "you get nice things when this is an R file" problem. But it's also pretty static code, so that matters less. You just want it to be template-able really.

jennybc: Could it be an R chunk in `README.Rmd` but with a special class or some such? That renders in the normal way in a non-webR context and in the "try me! and edit me!" way when webR is available.

PR #7809, opened 2026-02-11
Make dplyr tests more robust to randomness

DavisVaughan:
There was an issue where if you run `devtools::test()` interactively twice in a row, then it would consistently fail on this test:

```r
Error ([test-count-tally.R:85:3](vscode-file://vscode-app/Applications/Positron.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html#)): works with dbplyr
Error in `db_copy_to(dest$con, name, df, overwrite = overwrite, types = types, temporary = temporary, unique_indexes = unique_indexes, indexes = indexes, analyze = analyze, in_transaction = in_transaction, ...)`: Can't copy data to table `dbplyr_4NRy23f5MY`.
Caused by error in `dplyr::db_write_table()`:
! Can't write table table `dbplyr_4NRy23f5MY`.
Caused by error:
! Table `dbplyr_4NRy23f5MY` exists in database, and both overwrite and append are FALSE
```

Turns out that we were fixing the seed on each test run via a test in `test-across.R`, which called `set.seed()`, so dbplyr's random table names weren't actually all that random!

- Remove `set.seed()` usage in `test-across.R`
- Ensure dbplyr related tests can `overwrite` preexisting temporary tables of the same name (Unfortunately `dbplyr::memdb_frame()` doesn't expose `overwrite` so we have to dip down a layer)

---

In the future we should switch to dbplyr's new `local_memdb_frame()` once it makes it to CRAN, because that would be very useful here.

### DIFF
--- a/tests/testthat/test-count-tally.R
+++ b/tests/testthat/test-count-tally.R
@@ -82,12 +82,14 @@ test_that("works with dbplyr", {
   skip_if_not_installed("dbplyr")
   skip_if_not_installed("RSQLite")
 
+  # Use `dbplyr::local_memdb_frame()` when available
   db <- copy_to(
     dbplyr::src_memdb(),
     tibble(x = c(1, 1, 1, 2, 2)),
     name = "dplyr_test_count",
     overwrite = TRUE
   )
+
   df1 <- db |> count(x) |> as_tibble()
   expect_equal(df1, tibble(x = c(1, 2), n = c(3, 2)))
 
@@ -99,6 +101,7 @@ test_that("dbplyr `count()` method has transient internal grouping (#6338, tidyv
   skip_if_not_installed("dbplyr")
   skip_if_not_installed("RSQLite")
 
+  # Use `dbplyr::local_memdb_frame()` when available
   db <- copy_to(
     dbplyr::src_memdb(),
     tibble(x = c(1, 1, 1, 2, 2), y = c("a", "a", "b", "c", "c")),


R/compat-dbplyr.R
```diff
@@ -58,16 +58,23 @@ utils::globalVariables("!<-")
...
 #' @export
+#' @examplesIf requireNamespace("dbplyr", quietly = TRUE)
+#' # sql() just adds a class
+#' sql("x + 1")
+#' dbplyr::is.sql(sql("x + 1"))
+#'
+#' # You can use it when you need to insert some literal SQL in a query
+#' db <- dbplyr::memdb_frame(x = 1:3)
+#' db |> mutate(y = sql("CAST(x as VARCHAR)"))
 sql <- function(...) {
```
DavisVaughan, line 61: We inherit all of dbplyr's sql docs, including examples, which it now has. The examples reference `is.sql()` without any `dbplyr::` because it assumes it is loaded. But that isn't the case for us, so it failed.

I'm just overriding those new examples with a copy-paste version of them adapted for dplyr

tests/testthat/test-across.R
```diff
@@ -1927,14 +1927,18 @@ test_that("across() evaluates ... with promise semantics (#5813)", {
...
+    mutate(across(var, function(x, y) y, {
+      counter <<- counter + 1L
+      counter
+    })) |>
     pull(var)
 
-  set.seed(0)
-  expect_equal(out, rnorm(3))
+  expect_identical(counter, 3L)
+  expect_identical(out, c(1L, 2L, 3L))
```
DavisVaughan, line 1941: Writing out the seed manipulation, keeping the spirit of the test

tests/testthat/test-count-tally.R
```diff
@@ -82,7 +82,11 @@ test_that("works with dbplyr", {
   skip_if_not_installed("dbplyr")
   skip_if_not_installed("RSQLite")
 
-  db <- dbplyr::memdb_frame(x = c(1, 1, 1, 2, 2))
+  db <- dbplyr::local_memdb_frame(
+    "dplyr_test_count",
+    x = c(1, 1, 1, 2, 2)
+  )
```
DavisVaughan, line 91: Cleans itself up after the test scope ✅ , much nicer

PR #7819, opened 2026-03-20
Remove all C level non-API calls

### LINKED ISSUE #7818: Remove reliance on rlang `env_unbind()` callable
```r
static inline
void r_env_unbind(r_obj* env, r_obj* sym) {
  R_removeVarFromFrame(sym, env);
}
```

As of R 4.0 it's just R_removeVarFromFrame

DavisVaughan:
Closes #7818 

Now that we finally have the new C level bindings API

Best reviewed one commit at a time.

https://github.com/tidyverse/dplyr/pull/7819/changes/69421d5684ae0dbb5e4a1a54020130eb4d5ca737 was particularly tricky. We were trying to reference a `caller` variable in the DataMask's private R6 environment that no longer existed as of https://github.com/tidyverse/dplyr/pull/6444. The new APIs we are switching to error on unbound bindings, which revealed this. By luck, this happened to be totally fine before. We would pass this `caller` env on to `rlang::eval_tidy()`, but we always provided it a quosure, so the `env` argument there was never looked at. I've encoded this assumption in new `check_quosure()` checks that we do before setting the `env` to `R_EmptyEnv`, since the environment doesn't matter.

### DIFF
--- a/src/dplyr.h
+++ b/src/dplyr.h
@@ -104,7 +104,8 @@ namespace rlang {
 SEXP eval_tidy(SEXP expr, SEXP data, SEXP env);
 SEXP as_data_pronoun(SEXP x);
 SEXP new_data_mask(SEXP bottom, SEXP top);
-SEXP str_as_symbol(SEXP);
+SEXP str_as_symbol(SEXP str);
+bool is_quosure(SEXP x);
 }
 
 namespace vctrs {
--- a/src/utils.h
+++ b/src/utils.h
@@ -1,10 +1,7 @@
 #ifndef DPLYR_UTILS_H
 #define DPLYR_UTILS_H
 
-#define R_NO_REMAP
-#include <R.h>
-#include <Rinternals.h>
-#include <Rversion.h>
+#include "dplyr.h"
 
 // String encoding normalization
 // From https://github.com/r-lib/vctrs/pull/2085
@@ -23,14 +20,8 @@ static inline SEXP string_as_utf8(SEXP x) {
   return Rf_mkCharCE(Rf_translateCharUTF8(x), CE_UTF8);
 }
 
-// rlang's C level `is_quosure()`
-// https://github.com/r-lib/rlang/blob/01c296faa170d874cb8a3971e9cf7e5cd384b27b/src/internal/quo.c#L17
-static inline bool is_quosure(SEXP x) {
-  return TYPEOF(x) == LANGSXP && Rf_inherits(x, "quosure");
-}
-
 static inline void check_quosure(SEXP x) {
-  if (!is_quosure(x)) {
+  if (!rlang::is_quosure(x)) {
     Rf_errorcall(R_NilValue, "Internal error: `x` must be a quosure.");
   }
 }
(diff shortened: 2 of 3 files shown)


src/dplyr.h
```diff
@@ -36,6 +36,23 @@ SEXP R_mapAttrib(SEXP x, SEXP (*FUN)(SEXP, SEXP, void *), void *data) {
...
 #endif
 
+// Poorman's version of rlang's `r_env_get()`, which more faithfully handles
+// edge cases like active bindings, promises, and missing args. Only use this
+// for extracting a simple variable that you KNOW exists in the env, like
+// anything in our `private` envs.
+static inline
+SEXP env_get(SEXP env, SEXP sym) {
+#if R_VERSION >= R_Version(4, 5, 0)
+  return R_getVar(sym, env, TRUE);
```
lionel-: Does this need to be `FALSE` to match the "find in frame" in the other branch?

src/utils.h
```diff
@@ -23,4 +23,27 @@ static inline SEXP string_as_utf8(SEXP x) {
   return Rf_mkCharCE(Rf_translateCharUTF8(x), CE_UTF8);
 }
 
+// rlang's C level `is_quosure()`
+// https://github.com/r-lib/rlang/blob/01c296faa170d874cb8a3971e9cf7e5cd384b27b/src/internal/quo.c#L17
+static inline bool is_quosure(SEXP x) {
+  return TYPEOF(x) == LANGSXP && Rf_inherits(x, "quosure");
+}
```
lionel-: Can you please use the C callable to keep the quosure representation opaque.

lionel-: https://github.com/r-lib/rlang/blob/5e54655d483704f0bdc81d60c5c4801aebe73450/src/internal/internal.c#L376

src/chop.cpp
```diff
@@ -121,23 +121,48 @@ SEXP dplyr_make_mask_bindings(SEXP env_chops, SEXP data) {
...
+//
+// We first require that the binding exists, otherwise that would
+// be an internal error on our part.
+//
+// Then we check to see if the binding is delayed. If it is, then it is
+// considered "unused".
+//
+// Used bindings include:
+// - Forced chop promises (i.e. if the user referenced an original column in an expression)
+// - Newly installed columns from a previous expression
```
lionel-, line 161: Likely not impacting you here, but worth noting that checking for used bindings with the new API produces different result because it's the innermost promise, rather than outermost, that determines whether the binding is used or not. I've seen an rlang revdep failure due to this.

src/dplyr.h
```diff
@@ -43,7 +43,7 @@ SEXP R_mapAttrib(SEXP x, SEXP (*FUN)(SEXP, SEXP, void *), void *data) {
 static inline
 SEXP env_get(SEXP env, SEXP sym) {
 #if R_VERSION >= R_Version(4, 5, 0)
-  return R_getVar(sym, env, TRUE);
+  return R_getVar(sym, env, FALSE);
```
lionel-, line 46: ah there it is :)

src/dplyr.h
DavisVaughan: Yea I think I want to apply https://github.com/r-lib/rlang/pull/1886 here as well

DavisVaughan: Oh I did it already and you caught that below

PR #7844, opened 2026-06-06
docs: clarify between comparison semantics

### LINKED ISSUE #7702: `between()` doesn't work if `x` is `Date` and `left` and `right` are character
The docs of `between()` say:

> This is a shortcut for `x >= left & x <= right`

However, `between()` doesn't work if `x` is `Date` and `left` and `right` are character, while `<=` and `>=` work:

``` r
packageVersion("dplyr")
#> [1] '1.1.4.9000'

x <- as.Date("2010-01-02")
x >= "2010-01-01" & x <= "2010-01-04"
#> [1] TRUE

between(x, "2010-01-01", "2010-01-04")
#> Error in `between()`:
#> ! Can't combine `x` <date> and `left` <character>.
```

The "Details" section says:

> `x`, `left`, and `right` are all cast to their common type before the comparison is made.

but maybe it's worth updating the first line or mention somewhere that characters and dates are not compatible?

LeonidasZhak:
Fixes #7702.

## Summary
- describe local `between()` comparisons as similar to `x >= left & x <= right`, rather than an exact shortcut
- mention that local comparisons use vctrs after common-type casting
- clarify that character strings are not parsed as dates or date-times

## Checks
- `Rscript -e 'tools::checkRd("man/between.Rd")'`\n- `git diff --check`\n- `R CMD build --no-build-vignettes .` was attempted, but this local machine has `rlang` 1.1.6 loaded while dplyr now requires >= 1.1.7

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,8 @@
 # dplyr (development version)
 
+* `between()` documentation now more precisely describes how local vectors are
+  cast to a common type and compared with vctrs (#7702, @LeonidasZhak).
+
 # dplyr 1.2.1
 
 * dplyr is now fully compliant with the R C API (#7819).
--- a/R/funs.R
+++ b/R/funs.R
@@ -1,11 +1,13 @@
 #' Detect where values fall in a specified range
 #'
-#' This is a shortcut for `x >= left & x <= right`, implemented for local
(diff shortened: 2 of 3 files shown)


DavisVaughan:
Assuming this is the same as https://github.com/tidymodels/broom/pull/1281#issuecomment-4642851835

